PR #1, opened 2015-10-15
MPT-32 run external-james test when JAMES_ADDRESS env is set

mbaechler:
Switch to Guice 4 because exception in java 8 lambda
	breaks Guice 3 error reporting.


aduprat:
:+1:

rouazana:
:+1:

PR #2, opened 2015-10-15
MPT-33 introduce namespace support feature and run related tests only…

mbaechler:
… when supported

	This uses junit Assume that allowes to disable a test when a
	condition if false.

	TestModules have to define a boolean to advertise support of namespace.


aduprat:
Great :+1:

chibenwa:
Really nice :+1:

rouazana:
:+1:

PR #3, opened 2015-10-19
MPT-33 introduce namespace support feature and run related tests only…

mbaechler:
… when supported

	This uses junit Assume that allowes to disable a test when a
	condition if false.

	ImapHostSystem is supposed to implements supports(Feature... feature)
	for this purpose


mbaechler:
It's a new try at implementing features based testing

rouazana:
:+1:

mpt/core/src/main/java/org/apache/james/mpt/api/ImapFeatures.java
```diff
@@ -0,0 +1,36 @@
+package org.apache.james.mpt.api;
```
aduprat: license missing

PR #12, opened 2015-10-30
Mailbox 211 : distributed event systems

chibenwa:
In this PR I propose 3 kind of event system : 
 - the default one : the previous one, revisited. Better failure handling, and package relocation
 - a broadcast based event system : Well if you have global mailbox listeners that will not perform well in a distributed environment.
 - a registered based event system : the scalability limit is the numbers of topics you can handle on your publish subscribe system. It does not support non distributed global Listeners.

Note that the on going discussion about Message Sequence Number handling via the event system is still an open debate ;-)


chibenwa:
I updated the PR after Matthieu comments.

It still needs to be unit test validated ( will take ~ an hour ).

PR #13, opened 2015-10-31
Mailbox 257

chibenwa:
This tool can be used if you change the technology you used as an indexer.

It an also be used to correct index inconsistency.

These inconsistencies can arrise from re ordering of non CRDT operations (example deletes), on server stops, on index information loss.

The idea with this tool is to drop all documents from the idex, read the whole mailbox, and reindex these documents.

Note that this tool is designed to be used against a running, client exposed James server. The idea is to listen on events and correct not yet indexed values (eg flags, or deletes). We do not care about additions as they are well handled by the normal indexer.

I took this to the next level : reindex the whole mailboxmanager. The idea is to get the snapshot of all mailboxes at time T. Then we reindex mailboxes one by one using the process described above. Of course the following process is event corrected (upon Mailbox deletion). We o not care about mailbox additions as they are well handled by the normal indexer.

### DIFF
--- a/mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/FakeReIndexer.java
+++ b/mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/FakeReIndexer.java
@@ -1,0 +1,36 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.mailbox.indexer;
+
+import org.apache.james.mailbox.exception.MailboxException;
+import org.apache.james.mailbox.model.MailboxPath;
+
+public class FakeReIndexer implements ReIndexer {
+
+    @Override
+    public void reIndex(MailboxPath path) throws MailboxException {
+        throw new MailboxException("Not implemented");
+    }
+
+    @Override
+    public void reIndex() throws MailboxException {
+        throw new MailboxException("Not implemented");
+    }
+}
--- a/mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
+++ b/mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
@@ -1,0 +1,138 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.mailbox.indexer;
+
+import org.apache.james.mailbox.MailboxManager;
+import org.apache.james.mailbox.MailboxSession;
+import org.apache.james.mailbox.exception.MailboxException;
+import org.apache.james.mailbox.indexer.events.FlagsMessageEvent;
+import org.apache.james.mailbox.indexer.events.ImpactingEventType;
+import org.apache.james.mailbox.indexer.events.ImpactingMessageEvent;
+import org.apache.james.mailbox.indexer.registrations.GlobalRegistration;
+import org.apache.james.mailbox.indexer.registrations.MailboxRegistration;
+import org.apache.james.mailbox.model.MailboxPath;
+import org.apache.james.mailbox.model.MessageRange;
+import org.apache.james.mailbox.store.MailboxSessionMapperFactory;
+import org.apache.james.mailbox.store.mail.MessageMapper;
+import org.apache.james.mailbox.store.mail.model.Mailbox;
+import org.apache.james.mailbox.store.mail.model.MailboxId;
+import org.apache.james.mailbox.store.mail.model.Message;
+import org.apache.james.mailbox.store.search.ListeningMessageSearchIndex;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Note about live re-indexation handling :
+ *
+ *  - Data races may arise... If you modify the stored value between the received event check and the index operation,
+ *  you have an inconsistent behavior.
+ *
+ *  This class is more about supporting changes in real time for future indexed values. If you change a flags / delete
+ *  mails for instance, you will see it in the indexed value !
+ *
+ *  Why only care about updates and deletions ? Additions are already handled by the indexer that behaves normaly. We
+ *  should just "adapt" our indexed value to the latest value, if any. The normal indexer will take care of new stuff.
+ */
+public class ReIndexerImpl<Id extends MailboxId> implements ReIndexer {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ReIndexerImpl.class);
+    public static final int LIMIT = 0;
+
+    private final MailboxManager mailboxManager;
+    private final ListeningMessageSearchIndex<Id> messageSearchIndex;
+    private final MailboxSessionMapperFactory<Id> mailboxSessionMapperFactory;
+
+    public ReIndexerImpl(MailboxManager mailboxManager,
+                         ListeningMessageSearchIndex<Id> messageSearchIndex,
+                         MailboxSessionMapperFactory<Id> mailboxSessionMapperFactory) {
+        this.mailboxManager = mailboxManager;
+        this.messageSearchIndex = messageSearchIndex;
+        this.mailboxSessionMapperFactory = mailboxSessionMapperFactory;
+    }
+
+    public void reIndex(MailboxPath path) throws MailboxException {
+        MailboxSession mailboxSession = mailboxManager.createSystemSession("re-indexing", LOGGER);
+        reIndex(path, mailboxSession);
+    }
+
+
+    public synchronized void reIndex() throws MailboxException {
+        MailboxSession mailboxSession = mailboxManager.createSystemSession("re-indexing", LOGGER);
+        List<MailboxPath> mailboxPaths = mailboxManager.list(mailboxSession);
+        GlobalRegistration globalRegistration = new GlobalRegistration();
+        mailboxManager.addGlobalListener(globalRegistration, mailboxSession);
+        for (MailboxPath mailboxPath : mailboxPaths) {
+            if (globalRegistration.indexThisPath(mailboxPath)) {
+                reIndex(mailboxPath, mailboxSession);
+            }
+        }
+        mailboxManager.removeGlobalListener(globalRegistration, mailboxSession);
+    }
+
+    private void reIndex(MailboxPath path, MailboxSession mailboxSession) throws MailboxException {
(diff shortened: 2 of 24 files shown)


mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/FakeReIndexer.java
```diff
@@ -0,0 +1,36 @@
...
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.mailbox.indexer;
+
+import org.apache.james.mailbox.exception.MailboxException;
+import org.apache.james.mailbox.model.MailboxPath;
+
+public class FakeReIndexer implements ReIndexer {
```
mbaechler: It's not fake if it throws.

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+ *  This class is more about supporting changes in real time for future indexed values. If you change a flags / delete
+ *  mails for instance, you will see it in the indexed value !
+ *
+ *  Why only care about updates and deletions ? Additions are already handled by the indexer that behaves normaly. We
+ *  should just "adapt" our indexed value to the latest value, if any. The normal indexer will take care of new stuff.
+ */
+public class ReIndexerImpl<Id extends MailboxId> implements ReIndexer {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ReIndexerImpl.class);
+    public static final int LIMIT = 0;
```
mbaechler: does this mean "no limit" ?

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+        this.mailboxSessionMapperFactory = mailboxSessionMapperFactory;
+    }
+
+    public void reIndex(MailboxPath path) throws MailboxException {
+        MailboxSession mailboxSession = mailboxManager.createSystemSession("re-indexing", LOGGER);
+        reIndex(path, mailboxSession);
+    }
+
+
+    public synchronized void reIndex() throws MailboxException {
```
mbaechler: what do you think you are protecting with this synchronized ? why no synchronized when you index a single mailbox ? what do you expect in distributed context ?

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+ *  - Data races may arise... If you modify the stored value between the received event check and the index operation,
+ *  you have an inconsistent behavior.
+ *
+ *  This class is more about supporting changes in real time for future indexed values. If you change a flags / delete
+ *  mails for instance, you will see it in the indexed value !
+ *
+ *  Why only care about updates and deletions ? Additions are already handled by the indexer that behaves normaly. We
+ *  should just "adapt" our indexed value to the latest value, if any. The normal indexer will take care of new stuff.
+ */
+public class ReIndexerImpl<Id extends MailboxId> implements ReIndexer {
```
mbaechler: I'm sure you could find a better class name

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+                reIndex(mailboxPath, mailboxSession);
+            }
+        }
+        mailboxManager.removeGlobalListener(globalRegistration, mailboxSession);
+    }
+
+    private void reIndex(MailboxPath path, MailboxSession mailboxSession) throws MailboxException {
+        MailboxRegistration mailboxRegistration = new MailboxRegistration(path);
+        mailboxManager.addListener(path, mailboxRegistration, mailboxSession);
+        LOGGER.info("Intend to reindex " + path);
```
mbaechler: slf4j has placeholders with `info("message {}", argument)` syntax

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+        messageSearchIndex.delete(mailboxSession, mailbox, MessageRange.all());
+        handleIterations(mailboxSession,
+            mailboxRegistration,
+            mailbox,
+            mailboxSessionMapperFactory.getMessageMapper(mailboxSession)
+                .findInMailbox(mailbox,
+                    MessageRange.all(),
+                    MessageMapper.FetchType.Full,
+                    LIMIT));
+        mailboxManager.removeListener(path, mailboxRegistration, mailboxSession);
```
mbaechler: what if you throw before reaching this line ?

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+        MailboxSession mailboxSession = mailboxManager.createSystemSession("re-indexing", LOGGER);
+        List<MailboxPath> mailboxPaths = mailboxManager.list(mailboxSession);
+        GlobalRegistration globalRegistration = new GlobalRegistration();
+        mailboxManager.addGlobalListener(globalRegistration, mailboxSession);
+        for (MailboxPath mailboxPath : mailboxPaths) {
+            if (globalRegistration.indexThisPath(mailboxPath)) {
+                reIndex(mailboxPath, mailboxSession);
+            }
+        }
+        mailboxManager.removeGlobalListener(globalRegistration, mailboxSession);
```
mbaechler: what if you throw before reaching this line ?

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+                messageSearchIndex.add(mailboxSession, mailbox, message);
+            } else if (impactingMessageEvent instanceof FlagsMessageEvent) {
+                message.setFlags(((FlagsMessageEvent) impactingMessageEvent).getFlags());
+                messageSearchIndex.add(mailboxSession, mailbox, message);
+            }
+        }
+    }
+
+    private ImpactingMessageEvent findMostRelevant(Collection<ImpactingMessageEvent> messageEvents) {
+        if (messageEvents == null) {
```
mbaechler: why would you get null instead of empty collection ?

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/registrations/MailboxRegistration.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private final Multimap<Long, ImpactingMessageEvent> impactingMessageEvents;
+    private final MailboxPath mailboxPath;
+
+    public MailboxRegistration(MailboxPath mailboxPath) {
+        this.impactingMessageEvents = Multimaps.synchronizedMultimap(ArrayListMultimap.<Long, ImpactingMessageEvent>create());
+        this.mailboxPath = mailboxPath;
+    }
+
+    public Collection<ImpactingMessageEvent> getImpactingEvents(long uid) {
```
mbaechler: why do you return Collection when you can return a list ?

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
```diff
@@ -0,0 +1,138 @@
...
+            return null;
+        }
+        ImpactingMessageEvent last = null;
+        for (ImpactingMessageEvent impactingMessageEvent : messageEvents) {
+            if (impactingMessageEvent.getType().equals(ImpactingEventType.Deletion)) {
+                return impactingMessageEvent;
+            }
+            last = impactingMessageEvent;
```
mbaechler: if you get a list of messageEvents, you can just split your "find deletion" loop from the getLast by using Iterables.getLast()

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/FakeReIndexer.java
chibenwa: Ok

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
chibenwa: Yes...

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/registrations/MailboxRegistration.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.Collection;
+
+public class MailboxRegistration implements MailboxListener {
+
+    private final Multimap<Long, ImpactingMessageEvent> impactingMessageEvents;
+    private final MailboxPath mailboxPath;
+
+    public MailboxRegistration(MailboxPath mailboxPath) {
+        this.impactingMessageEvents = Multimaps.synchronizedMultimap(ArrayListMultimap.<Long, ImpactingMessageEvent>create());
```
mbaechler: what do you expect from this synchronized ? Are event dispatched from multiple threads ?

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
chibenwa: I think I 'll remove it, it makes no sense in a distributed context.

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/registrations/GlobalRegistration.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public boolean indexThisPath(MailboxPath mailboxPath) {
+        return impactingEvents.get(mailboxPath) != null;
+    }
+
+    @Override
+    public void event(Event event) {
+        if (event instanceof MailboxDeletion) {
+            impactingEvents.putIfAbsent(event.getMailboxPath(), true);
```
mbaechler: why putIfAbsent and not put ?

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/registrations/GlobalRegistration.java
```diff
@@ -0,0 +1,47 @@
...
+
+public class GlobalRegistration implements MailboxListener {
+
+    private final ConcurrentHashMap<MailboxPath, Boolean> impactingEvents;
+
+    public GlobalRegistration() {
+        this.impactingEvents = new ConcurrentHashMap<MailboxPath, Boolean>();
+    }
+
+    public boolean indexThisPath(MailboxPath mailboxPath) {
```
mbaechler: the name suggest it does the indexing but the code is more "does this path needs indexing"

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
chibenwa: If you say so....

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
chibenwa: Ok

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
chibenwa: Thanks for this! If a single document can not be indexed, the process stops and we leak the listener. I guess the wanted behaviour is to not throw during a single docurement indexation, and ensure we remove the listener.

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
chibenwa: Same thing.

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
chibenwa: Ok

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/registrations/MailboxRegistration.java
chibenwa: Ok

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/ReIndexerImpl.java
chibenwa: Ok.

chibenwa:
Thanks for the review.

mailbox/tool/src/main/java/org/apache/james/mailbox/indexer/registrations/MailboxRegistration.java
chibenwa: Yes it is. . . The goal is to protect from local concurency

mbaechler: And you do you prevent deadlock from happening ?

chibenwa: From having lock contained in a guava object, obviously. This means only one lock can be held by a bunch of code. At most one lock all the time is equal, according to me to "no situations where somthing bad can arise".

PR #17, opened 2015-11-04
JAMES-1635 Increase embedded Cassandra startup timeout

chibenwa:
Increase embedded Cassandra startup timeout to the value configured in CassandraCluster used for unit tests.

This will avoid tests to fail on less effective computers.


backends-common/cassandra/src/test/java/org/apache/james/backends/cassandra/CassandraCluster.java
```diff
@@ -42,10 +42,12 @@
 
     private static final long SLEEP_BEFORE_RETRY = 200;
     private static final int MAX_RETRY = 200;
+    private static final long EMBEDDED_CASSANDRA_TIMEOUT = MAX_RETRY * SLEEP_BEFORE_RETRY;
```
aduprat: I would prefer an independent value, as the MAX_RETRY & SLEEP_BEFORE_RETRY don't make sense for this, IMHO.
You may also add the unit in the constant name.

chibenwa: Ok for the unit.

I do not agree with you for the timeout independent value. I do not want Cassandra unit to throw before the expected timeout value expressed by the retry loop. I'm OK with it throwing after that. For me it make perfect sense.

In fact SLEEP_BEFORE_RETRY should also indicate its value

aduprat:
:+1:

chibenwa:
Committed

PR #23, opened 2015-12-20
Jsieve 73 - Sieve body extension

chibenwa:
This PR proposes patchs so that Jsieve implements RFC 5173 : Sieve Body extension. It contains SIEVE test integrtion for this functionnality. 

Note that matching feature is achived threw the tools used by the IMAP search feature without indexing. (It was a little expended so that it can search the raw message and filter content by mime type).

**There is a new dependency between mailbox/store and sieve/mailet**. I don't think this is the better way to do it. I think we need to relocate MessageSearcher to some other projects. Don't hesitate to tell me your thoghts about that.

See https://github.com/apache/james-jsieve/pull/6 , a related PR.


chibenwa:
See https://github.com/apache/james-mime4j/pull/5

chibenwa:
I'm going to force push here  : 
 - Deletes MesageSearcher and depend on mime4j/james-utils

PR #26, opened 2016-01-05
JAMES-1650 fix an assertion in JMAP integration test (totalMessages=0…

fvn-linagora:
…), following previous


fvn-linagora:
merged on trunk

PR #32, opened 2016-01-26
Update ToRecipientFolder.java to use folder parameter

Ahmet-Kaplan:
If mailet's "folder" parameter different from "INBOX", then this parameter should be used.


aduprat:
I know that there's no unit test yet on this class.
But it may be the time to implement some.

chibenwa:
You could have a look to james-project/server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/delivery/LocalDeliveryTest.java

I think you will end up with tests quite similar...

Ahmet-Kaplan:
I will add a unit test for ToRecipientFolder when available and already created issue [JAMES-1670] (https://issues.apache.org/jira/browse/JAMES-1670).

Ahmet-Kaplan:
This is done at [34] (https://github.com/apache/james-project/pull/34)
So you can dismiss the request.

PR #33, opened 2016-01-26
Update ToRecipientFolder.java to use consume parameter

Ahmet-Kaplan:
If value of "consume" parameter is true then mail process should be discontinued.


Ahmet-Kaplan:
I'll documenting it in that file ASAP.

Ahmet-Kaplan:
This is done at [34] (https://github.com/apache/james-project/pull/34)
So you can dismiss the request.

PR #34, opened 2016-01-28
JAMES-1670 ToRecipientFolder mailet should use init parameters

Ahmet-Kaplan:
I fixed the mailet, created a unit test (taken mostly from LocalDeliveryTest) and added some info about ToRecipientFolder mailet to dev-provided-mailets docs.
Hopefully I don't make any mistake :)

### DIFF
--- a/server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/delivery/RecipientFolderTest.java
+++ b/server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/delivery/RecipientFolderTest.java
@@ -1,0 +1,197 @@
+/**
+ * **************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one * or more
+ * contributor license agreements. See the NOTICE file * distributed with this
+ * work for additional information * regarding copyright ownership. The ASF
+ * licenses this file * to you under the Apache License, Version 2.0 (the *
+ * "License"); you may not use this file except in compliance * with the
+ * License. You may obtain a copy of the License at * *
+ * http://www.apache.org/licenses/LICENSE-2.0 * * Unless required by applicable
+ * law or agreed to in writing, * software distributed under the License is
+ * distributed on an * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY *
+ * KIND, either express or implied. See the License for the * specific language
+ * governing permissions and limitations * under the License. *
+ * **************************************************************
+ */
+package org.apache.james.transport.mailets.delivery;
+
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import com.google.common.collect.Lists;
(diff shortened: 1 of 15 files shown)


server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/delivery/RecipientFolderTest.java
```diff
@@ -0,0 +1,197 @@
+/**
+ * **************************************************************
```
mbaechler: I think you have formating issue on the license header

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/delivery/RecipientFolderTest.java
```diff
@@ -0,0 +1,197 @@
...
+    private MailboxManager mailboxManager;
+    private SieveRepository sieveRepository;
+    private ToRecipientFolder recipientFolder;
+    private FakeMailetConfig mailetConfig;
+
+    private String folder = null;
+
+    private String consume = null;
+
+    private void setFolder(String folder) {
```
mbaechler: why do you need this setter ?

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/delivery/RecipientFolderTest.java
```diff
@@ -0,0 +1,197 @@
...
+        Mail mail = createMail();
+        recipientFolder.service(mail);
+
+        Assert.assertEquals(Boolean.valueOf(consume), mail.getState().equals(Mail.GHOST));
+    }
+
+    @Test
+    public void mailShouldBeWellReceivedByDefaultToUserWhenvirtualHostingIsTurnedOn() throws Exception {
+
+        when(usersRepository.supportVirtualHosting()).thenAnswer(new Answer<Boolean>() {
```
mbaechler: thenReturn is shorter

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/delivery/RecipientFolderTest.java
```diff
@@ -0,0 +1,197 @@
...
+            @Override
+            public Boolean answer(InvocationOnMock invocationOnMock) throws Throwable {
+                return true;
+            }
+        });
+        recipientTest();
+    }
+
+    private void recipientTest() throws Exception {
```
mbaechler: we usually inline the test code in the test method

mbaechler:
Thank you for your contribution, it looks good. There are only some minor style issues.

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/delivery/RecipientFolderTest.java
Ahmet-Kaplan: Those are not so necessary but I added if we may add another test which can use the parameters separately.

mbaechler: We don't need setter for properties of the current object, we can access it directly

PR #35, opened 2016-02-02
JAMES-1678 Don't throw unsupported exception when collapseThreads is …

Laura-Royet:
…false
rebased from : https://github.com/linagora/james-project/pull/118


Laura-Royet:
already merged

PR #36, opened 2016-02-11
Jsieve 103

chibenwa:
This PR :

-    Adds activation time for active sieve scipt (necessary for SIEVE vacation extension)
-    Refactor so that ResourceLocator is a class parameter and reinstanciated on each call
-    Add tests for vacation.


chibenwa:
See https://github.com/apache/james-jsieve/pull/9

mbaechler:
No joda-time ?

chibenwa:
rework - SieveRepository API should rely on yoda-time ... 
Seems like I forgot to push this one...

PR #37, opened 2016-03-04
JAMES-1700 Don't fail in account creation if user already exists

rouazana:
Using thread-weaver to really test racing condition.
Manual mock seems to be necessary as Mockito and thread-weaver seem to conflict.


rouazana:
Reopen here: https://github.com/linagora/james-project/pull/178

PR #42, opened 2016-06-03
V3.0 roadmap proposal

chibenwa:
In this pull request, I propose tagging different features with one of these three tags : 
- supported
- experimental
- unsupported

I give a proposal about what **supported** means.

From this, we deduce what tasks needs to be done.

Once reviewed, I will create the tickets on JIRA. These documents have been discutted with the rest of the team at Linagora.

If you want a pretty version to read the documents, I propose you can read https://github.com/chibenwa/james-project/blob/v3.0_roadmap/v3.0_roadmap.adoc and https://github.com/chibenwa/james-project/blob/v3.0_roadmap/v3.0_mailet_inventory.adoc

### DIFF
--- a/v3.0_mailet_inventory.adoc
+++ b/v3.0_mailet_inventory.adoc
@@ -1,0 +1,580 @@
+= Inventory for Mailets and matchers
+
+== Mailets
+
+=== /mailet/standard
+
+.Inventory for /mailet/standard
+|===
+|Name |Tested |Documented |Status |Remarks
+
+|AddFooter
+|Yes
+|Yes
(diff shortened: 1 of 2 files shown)


v3.0_mailet_inventory.adoc
```diff
@@ -0,0 +1,580 @@
...
+|Supported
+|
+
+|RemoveMimeHeader
+|Yes
+|Yes
+|Supported
+|
+
+|ReplaceSubject
```
aduprat: ReplaceContent

PR #47, opened 2016-08-09
Fix JAMES-1815: Copy potentially immutable list to allow modifications

tfg13:
Fixes https://issues.apache.org/jira/browse/JAMES-1815

### DIFF
--- a/server/data/data-library/src/main/java/org/apache/james/domainlist/lib/AbstractDomainList.java
+++ b/server/data/data-library/src/main/java/org/apache/james/domainlist/lib/AbstractDomainList.java
@@ -119,7 +119,7 @@ public abstract class AbstractDomainList implements DomainList, LogEnabled, Conf
         if (domains != null) {
 
             // create mutable copy, some subclasses return ImmutableList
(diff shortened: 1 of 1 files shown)


server/data/data-library/src/main/java/org/apache/james/domainlist/lib/AbstractDomainList.java
```diff
@@ -118,6 +118,9 @@ public String getDefaultDomain() throws DomainListException {
         List<String> domains = getDomainListInternal();
```
chibenwa: I tend to don't like variable re-affectation....

Here I would prefer a variable called "mutableDomains" to be created rather than seing domains reused.

By the way, nice spot.

tfg13:
Ok, I changed the fix to introduce a new variable.
I also noticed something else: When domains is null, ImmutableList.copyof(domains) will crash with a NPE.
Also fixed that.

chibenwa:
:+1:

rouazana:
:+1:

PR #49, opened 2016-08-13
James 1773 : Provide tests for RelayLimit

chibenwa:
RelayLimit is used in every configuration but was not tested yet.

I modified it so that the relay number is compelled to be superior to zero.


rouazana:
:+1:

chibenwa:
Merged

PR #52, opened 2016-08-16
Add docker image tags to README.txt instructions

maguro:
Adding separate image tags, java-6 and java-8, to the docker images
is a nice way to organize the docker images as well as minimize confusion
due to the shared image name, james/project.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -54,10 +54,10 @@ We need to check the compilation in both Java 6 & Java 8:
 
 * Java 6
 First step, you have to build the Docker image
-$ docker build -t james/project dockerfiles/compilation/java-6
+$ docker build -t james/project:java-6 dockerfiles/compilation/java-6
 
 In order to run the build, you have to launch the following command:
-$ docker run -v $PWD/.m2:/root/.m2 -v $PWD:/origin -v $PWD/dockerfiles/run/spring/destination:/destination -t james/project -s SHA1
+$ docker run -v $PWD/.m2:/root/.m2 -v $PWD:/origin -v $PWD/dockerfiles/run/spring/destination:/destination -t james/project:java-6 -s SHA1
 
 Where:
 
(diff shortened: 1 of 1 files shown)


mbaechler:
It looks nice, we'll do a CI build on that to check everything is ok.

ieugen:
This PR has been sitting for quite some time. It's mostly deprecated by now. Sorry @maguro for not looking at this sooner. If you are still interested in updating it I can take care of review and merge. 

If not I propose to close the issue so we can clean things here. I will keep it open for the next 7 days.

chibenwa:
Since the Java 11 adoption throughout the code base we rely on a single build image, based on Java 11.

Using a `java-11` tag would not really serve anymore a disambiguation purpose. But adding it would I believe not hurt.

ieugen:
Ok. Let's do that when we start experimenting with other JDK versions. We should test that james builds and starts with the LTS version and the latest non LTS version (now JDK 14).

I'm going to close this for now as the things have changed. Thanks @maguro .

chibenwa:
Let's add the tag for current `java11` version to make subsequent changes easier to keep the spirit of this work, no?

2 factor authentication is required for PMC members to have access to github administration. More details can be found on this page on how to do it https://cwiki.apache.org/confluence/display/OPENWHISK/Accessing+Apache+GitHub+as+a+Committer

(I must admit I did not do it as last time I turned 2 factor authentication on things did not really go as planned and I ended up doing quickly a recovery (oops)).

Another method is to fill an INFRA JIRA ticket requesting them to come and close the pull requests (in which case it is better to close them by batch)

I end up often asking politely to contributors to close them themselves when it makes sense.

Cheers,

Benoit

ieugen:
Sure. Let's do it. I'm still getting the hang of it. I expected to find published docker images on docker hub. I don't know what is the policy around that.

chibenwa:
It's complicated, as part of ApacheCon Europe during the licensing workshop, the question was raised.

Docker images relies on non apachev2 licensed code, having a project delivering them gives a false state of ownership, and accountability. Wich is slightely different from standalone compiled code.  Despite these doubts, no clear position was exposed.

So far, linagora do publishes these image.

We could of course discuss publishing them as apache.

mbaechler:
Let's close that, then

PR #53, opened 2016-08-17
James 1773 Rework SizeGreaterThan

chibenwa:
- Specified size limit should not be negative or null
 - Right more tests for hedge cases and all possible units
 - Some imports were not needed


chibenwa:
- I rewrotte tests for all standard matcher (that do not have tests based on inheritance) with assertj.
 - I provide also some tests for HasAttachment
 - HasAttachment now is conform to RFC-8153 and uses content disposition.
 - I clarified FetchFrom did not removed non matched headers

PR #62, opened 2016-09-01
Mailet 124 Composite matchers (And, Or, Not and Xor)

chibenwa:
- Improve code quality a litlle. Not having java-8 for this task forced me not to use reduce. Hence, simplification works not that well for some matchers (And).
 - Rewrite tests using  mocks


chibenwa:
Merged

PR #63, opened 2016-09-25
JAMES-1823 quotaUpdater is invoked more than one time for a message addition or deletion

leepaul21:
I'm not sure if my solution is appropriate and on the right direction from the perspective of the system and its architecture because I don't fully understand it. My purpose in the patch is not to register the same listener duplicately in an instance of DefaultDelegatingMailboxListener. For this, I've added some code and data structure for checking duplication. If my solution isn't suitable for the whole system, you don't have to use it.

### DIFF
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/event/DefaultDelegatingMailboxListener.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/event/DefaultDelegatingMailboxListener.java
@@ -20,6 +20,9 @@
 package org.apache.james.mailbox.store.event;
 
 import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.james.mailbox.MailboxListener;
 import org.apache.james.mailbox.MailboxSession;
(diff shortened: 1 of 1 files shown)


chibenwa:
In DefaultDelegatingMailboxListener, the intent was to add a business logic layer over listener lists, especially about updating stored listeners upon events.

In over worlds, there is no need to add logic for avoiding multiple storage of listeners as it is some tasks of underlying layers (not to mention that your implementation is not concurrent friendly, as applied at the wrong layer).

To obtain desired effect, you should modify type for **globalListeners** in **MailboxListenerRegistry** to **ConcurrentSkipListSet** or over concurrent friendly set datastructure.

chibenwa:
After investigations, **Collections.newSetFromMap(new ConcurrentHashMap<MailboxListener, Boolean>())** is better.

ieugen:
Thank you for your contribution @leepaul21 . 
Sorry for not merging it sooner. 
Is this something that we can merge? 

CC @chibenwa : Can you please give a hand?

If this is not an issue anymore, let's close it.

chibenwa:
It had been adressed in another pull request, it is no longer an issue. We can safely close this.

PR #65, opened 2016-11-27
JAMES-1865 JDBCGreylistHandler will lead to can't start james server

luo877280:
rename set method to init on WhiteListedNetworks to avoild dependencyCheck exception


luo877280:
So sorry for pull the test request . Please help remove the test commit.

chibenwa:
The first commit makes sens, we can remove all the followings.

I would go then one step further : remove the 's/setWhiteListedNetworks/initWhiteListedNetworks/g' directly, and initialize the field directly inside init.

gmcdonald:
Closed as per INFRA-13760

PR #67, opened 2017-03-31
JAMES-1975 Move mailets over from james-jdkim

shrumit:
https://issues.apache.org/jira/browse/JAMES-1975

Added the necessary files and changed the pom as needed. Need guidance on whether this was done correctly.

### DIFF
--- a/server/mailet/mailets/pom.xml
+++ b/server/mailet/mailets/pom.xml
@@ -140,6 +140,10 @@
             <groupId>org.slf4j</groupId>
             <artifactId>jcl-over-slf4j</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.james.jdkim</groupId>
+            <artifactId>apache-jdkim-library</artifactId>        
+        </dependency>
         <!--
                <dependency>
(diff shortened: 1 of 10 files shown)


chibenwa:
I will trigger a build, and wait over reviews before merging this.

shrumit:
I'm really not sure whether I did this correctly. I just moved over the mailet files and updated the POM. I've run a successful build without tests. When I run a build with tests I get errors unrelated to this, probably something wrong with my local setup.

ieugen:
I'm looking around and found this PR that seems to be waiting for a merge. 
@chibenwa, @shrumit: Should we continue the work and merge? It seems to me like we should.

chibenwa:
Hello @ieugen ,

As far as I can tell, this PR was not passing unit tests.

The move of the mailet from dkim into James Server code had been contributed here https://github.com/linagora/james-project/pull/1709

We should have been notifying it here.

This pull request likely can be closed.

ieugen:
Something has been merged there so I second the closing of this issue. Thank you for your contribution @shrumit . Please let us know if we missed something.

PR #73, opened 2017-10-01
JAMES-992 Added HasMimeType matcher and tests.

vVv-AA:
Added HasMimeType matcher that will check if a given MimeMessage's content type is present in a possibility set.

### DIFF
--- a/mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
+++ b/mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
@@ -1,0 +1,74 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+
+package org.apache.james.transport.matchers;
+
+import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import org.apache.james.core.MailAddress;
+import org.apache.mailet.Mail;
+import org.apache.mailet.base.GenericMatcher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.activation.MimeType;
+import javax.activation.MimeTypeParseException;
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
+import java.util.Collection;
+import java.util.Optional;
+import java.util.Set;
+
+
+/**
+ * <p>This matcher checks if the content type matches.</p>
+ *
+ * use: <pre><code><mailet match="HasMimeType=text/plain,text/html" class="..." /></code></pre>
+ */
+public class HasMimeType extends GenericMatcher {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(HasMimeType.class);
+
+    private Set<String> acceptedContentTypes;
+
+    public void init() throws javax.mail.MessagingException {
+        acceptedContentTypes = ImmutableSet.copyOf(Splitter.on(",").trimResults().split(getCondition()));
+    }
+
+    public Collection<MailAddress> match(Mail mail) throws javax.mail.MessagingException {
+        Optional<String> mimeTypes = getMimeTypeFromMessage(mail.getMessage());
+
+        return mimeTypes.filter(acceptedContentTypes::contains)
+                .map(any -> mail.getRecipients())
+                .orElse(ImmutableList.of());
+    }
+
+    private static Optional<String> getMimeTypeFromMessage(MimeMessage message) throws MessagingException {
+        try {
+            return Optional.of(new MimeType(message.getContentType()).getBaseType());
+        } catch (MimeTypeParseException e) {
+            LOGGER.warn(String.format("Error while parsing message's mimeType %s", message.getContentType()), e);
+            return Optional.empty();
+        }
+    }
+
+}
+
--- a/mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
+++ b/mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
@@ -1,0 +1,93 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.transport.matchers;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
+
+import org.apache.james.core.MailAddress;
+import org.apache.mailet.base.test.FakeMail;
+import org.apache.mailet.base.test.FakeMatcherConfig;
+import org.apache.mailet.base.test.MimeMessageBuilder;
+import org.junit.Before;
+import org.junit.Test;
+
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
+import java.util.Collections;
+
+public class HasMimeTypeTest {
+
+    private static final String RECIPIENT = "test@james.apache.org";
+    private static final String FROM = "test@james.apache.org";
+    private static final String MIME_TYPES = "multipart/mixed";
+    private static final String NON_MATCHING_MIME_TYPES = "text/plain, application/zip";
+
+    private HasMimeType matcher;
+    private FakeMail mail;
+
+    @Before
+    public void setUp() throws Exception {
+        matcher = new HasMimeType();
+        MimeMessage message = MimeMessageBuilder.mimeMessageBuilder()
+                .setMultipartWithBodyParts(
+                        MimeMessageBuilder.bodyPartBuilder()
+                                .data("simple text")
+                                .disposition("text")
+                                .build(),
+                        MimeMessageBuilder.bodyPartBuilder()
+                                .filename("text_file.txt")
+                                .disposition("attachment")
+                                .build(),
+                        MimeMessageBuilder.bodyPartBuilder()
+                                .type("application/zip")
+                                .filename("zip_file.zip")
+                                .disposition("attachment")
+                                .build())
+                .setSubject("test")
+                .build();
+
+        mail = FakeMail.builder()
+                .mimeMessage(message)
+                .sender(new MailAddress(FROM))
+                .recipient(new MailAddress(RECIPIENT))
+                .build();
+    }
+
+    @Test
+    public void hasMimeType() throws MessagingException {
+        matcher.init(FakeMatcherConfig.builder()
+                .matcherName("HasMimeType")
+                .condition(MIME_TYPES)
+                .build());
+
+        assertThat(matcher.match(mail)).containsAll(mail.getRecipients());
+    }
+
+    @Test
+    public void doesNotHaveMimeType() throws MessagingException{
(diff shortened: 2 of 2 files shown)


mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,61 @@
...
+
+import javax.activation.MimeType;
+import javax.activation.MimeTypeParseException;
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
+
+import java.util.*;
+
+/**
+ * <p>This matcher checks if the content type matches.</p>
```
chibenwa: Could you give in the javadoc a configuration examplefor this mailet?

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public class HasMimeType extends GenericMatcher {
+
+    private Set<String> content_types;
```
chibenwa: We enforce contentType :-)

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,61 @@
...
+
+    public void init() throws javax.mail.MessagingException {
+        content_types = ImmutableSet.copyOf(getCondition().split("\\s*,\\s*"));
```
chibenwa: You can have a look to the excellent guava's Splitter for this.

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    public Collection<MailAddress> match(Mail mail) throws javax.mail.MessagingException {
+        String mimeType;
+        try {
+            mimeType = getMimeTypeFromMessage(mail.getMessage());
+        } catch (MimeTypeParseException e) {
+            mimeType = null;
+        }
+        return (content_types.contains(mimeType)) ? mail.getRecipients() : null;
```
chibenwa: We should avoid returning null. An empty list is better :-)

chibenwa: A if here would be more readable...

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,61 @@
...
+    public void init() throws javax.mail.MessagingException {
+        content_types = ImmutableSet.copyOf(getCondition().split("\\s*,\\s*"));
+    }
+
+    public Collection<MailAddress> match(Mail mail) throws javax.mail.MessagingException {
+        String mimeType;
+        try {
+            mimeType = getMimeTypeFromMessage(mail.getMessage());
+        } catch (MimeTypeParseException e) {
+            mimeType = null;
```
chibenwa: We should avoid swallowing exceptions. We should add a warning log here...

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private Set<String> content_types;
+
+    public void init() throws javax.mail.MessagingException {
+        content_types = ImmutableSet.copyOf(getCondition().split("\\s*,\\s*"));
+    }
+
+    public Collection<MailAddress> match(Mail mail) throws javax.mail.MessagingException {
+        String mimeType;
+        try {
```
chibenwa: You can move the try/catch in getMimeTypeFrom message.

It avoid mimeType double allocation in parallel branches.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
```diff
@@ -0,0 +1,60 @@
+package org.apache.james.transport.matchers;
```
chibenwa: We tend to start with the Apache license on the top of our files

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Test;
+
+import javax.mail.internet.MimeMessage;
+
+import static org.junit.Assert.assertNotNull;
+
+public class HasMimeTypeTest {
+
+    private HasMimeType matcher;
+    private final String RECIPIENT = "test@james.apache.org";
```
chibenwa: You can add the static modifier here for these constants.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
```diff
@@ -0,0 +1,60 @@
...
+                .build();
+    }
+
+    @Test
+    public void hasMimeType() throws Exception {
+        matcher.init(FakeMatcherConfig.builder()
+                .matcherName("HasMimeType")
+                .condition(MIME_TYPES)
+                .build());
+        assertNotNull(matcher.match(mail));
```
chibenwa: Can we be more explicit than is not null? It sould contain the list of recipients.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
```
chibenwa: You miss one test on a non matching case.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
```diff
@@ -0,0 +1,60 @@
...
+import javax.mail.internet.MimeMessage;
+
+import static org.junit.Assert.assertNotNull;
+
+public class HasMimeTypeTest {
+
+    private HasMimeType matcher;
+    private final String RECIPIENT = "test@james.apache.org";
+    private final String FROM = "test@james.apache.org";
+    private final String MIME_TYPES = "multipart/mixed";
```
chibenwa: It's more MIME_TYPE

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.james.core.MailAddress;
+import org.apache.mailet.base.test.FakeMail;
+import org.apache.mailet.base.test.FakeMatcherConfig;
+import org.apache.mailet.base.test.MimeMessageBuilder;
+import org.junit.Before;
+import org.junit.Test;
+
+import javax.mail.internet.MimeMessage;
+
+import static org.junit.Assert.assertNotNull;
```
chibenwa: We tend to put static imports above normal imports

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
vVv-AA: You can specify multiple mime types in MIME_TYPES.

vVv-AA:
@chibenwa Updated.

I did a soft reset before pushing and its screwed up your review comments. Did not know that would happen. Sorry.

Also, I kept MIME_TYPES as the variable name because it can contain multiple mime types as is the case with NON_MATCHING_MIME_TYPES if that is fine with you.

chibenwa:
No problem, the PR is short. I can do another review pass on it?

Ok, your explanation convince me for MIME_TYPES

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,74 @@
...
+ * use: <pre><code><mailet match="HasMimeType=text/plain,text/html" class="..." /></code></pre>
+ */
+public class HasMimeType extends GenericMatcher {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(HasHabeasWarrantMark.class);
+
+    private Set<String> contentType;
+
+    public void init() throws javax.mail.MessagingException {
+        contentType = ImmutableSet.copyOf(Splitter.onPattern("\\s*,\\s*").split(getCondition()));
```
chibenwa: Actually, we do believe regex are counter intuitive.

Maybe:

```
Splitter.on(',').split(...)
```

Is more explicit.

chibenwa: Also what happens if I got:

```
text/plain, text/html
```

As a condition? Could we trimm the results? (Splitter have an option for that...)

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    public Collection<MailAddress> match(Mail mail) throws javax.mail.MessagingException {
+        String mimeType;
```
chibenwa: You can do : 

```
Strin mimeType = getMimeTypeFromMessage(...);
```

Directly....

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,74 @@
...
+        mimeType = getMimeTypeFromMessage(mail.getMessage());
+        if (contentType.contains(mimeType))
```
chibenwa: We use to always put blocks ^^ 

```
if {
    //...
} else {
    //...
}
```

Sorry for being picky...

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,74 @@
...
+        else
+            return Collections.emptyList();
+    }
+
+    private static String getMimeTypeFromMessage(MimeMessage message) throws MessagingException {
+        try {
+            return new MimeType(message.getContentType()).getBaseType();
+        } catch (MimeTypeParseException e) {
+            LOGGER.warn(e.toString());
+            return "";
```
chibenwa: If we return nothing, that has to be explicit on the type.

The way we handle this in a clean way is using java Optional.

You can then:
 - Modify the mimeType of this method to return Optional<String>
 - When returning something, return Optional.of(something)
 - When we encouter the error, we can then return Optional.empty()
 - Finally in **match** methods, we can handle the if resoluion in a functional manner using *filter* and *orElse* (that might replace the if block)

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasMimeTypeTest.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Test
+    public void doesNotHaveMimeType() throws MessagingException{
+        matcher.init(FakeMatcherConfig.builder()
+                .matcherName("HasMimeType")
+                .condition(NON_MATCHING_MIME_TYPES)
+                .build());
+
+        assertEquals(matcher.match(mail), Collections.emptyList());
```
chibenwa: You can do this one also with assertJ : 

```
assertThat(matcher.match(mail)).isEmpty();
```

Then you might simplify imports.

chibenwa:
By the way, you might add the Hacktaberfest label on this PR, no?

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
vVv-AA: The regex was to pre-trim to handle cases like text/plain, text/html. I'll replace regex.

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * <p>This matcher checks if the content type matches.</p>
+ *
+ * use: <pre><code><mailet match="HasMimeType=text/plain,text/html" class="..." /></code></pre>
+ */
+public class HasMimeType extends GenericMatcher {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(HasHabeasWarrantMark.class);
+
+    private Set<String> contentType;
```
chibenwa: Shouldn't the name of the variable be "acceptedContentTypes" ? This would be more descriptive.

vVv-AA:
@chibenwa Done.

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -0,0 +1,74 @@
...
+        return mimeTypes.filter(acceptedContentTypes::contains)
+                .map(any -> mail.getRecipients())
+                .orElse(ImmutableList.of());
+    }
+
+    private static Optional<String> getMimeTypeFromMessage(MimeMessage message) throws MessagingException {
+        try {
+            return Optional.of(new MimeType(message.getContentType()).getBaseType());
+        } catch (MimeTypeParseException e) {
+            LOGGER.warn(e.toString());
```
chibenwa: Ah, I missed this one.

My big concern here is that "we loose the stacktrace", which makes understanding and categorizing possible errors harder.

We should add an error message explaining what's happening ("Error while checking message mimeType") and pass the exception as a second argument.

chibenwa:
Thanks @vVv-AA  for your great work and  this added bonus matcher!

I scheduled a build on your PR in https://github.com/linagora/james-project/pull/1005 . We might have further review pass on it, but I will handle it.

If you are looking forward more Hacktoberfest contributions, I let you have look at some other Hacktoberfest related issues. 

Especially https://github.com/linagora/james-project/issues/993 will allow you to discover the James Command Line client, and would ease Sieve scripts deployment. Also in this issue, a good part of the wiring is already done. Hence the functionality should not be too hard to implement.

Appart the Hacktober fest, the James community is actively looking for new contributors (be it communication, articles, blogpost, documentation, code, packaging, or anything else.). I let you check the James website (http://james.apache.org/#second).

vVv-AA:
Thanks @chibenwa 
I will check out [linagora#993](https://github.com/linagora/james-project/issues/993)

vVv-AA:
Merged as https://github.com/linagora/james-project/pull/1005. Closing.

PR #74, opened 2017-10-18
 JAMES-2194 server/app/src/main/resources/recipientrewritetable.xml: fix package …

PascalSchumacher:
…of XMLRecipientRewriteTable class reference


strokyl:
Thank you a lot for your PR Pascal.
I have just created a Apache ticket linked to it: https://issues.apache.org/jira/browse/JAMES-2194.
I would need that you rename you commit so you could add JAMES-2194 before it is name. We ask you that so we can link your commit to the given ticket when we watch the git log.

Also if you can add JAMES-2194 as a prefix to your PR title that would be perfect!

PascalSchumacher:
Thanks for the feedback.  👍  

I have prefixed the commit message and the pulll request title with `JAMES-2194` as requested.

PR #76, opened 2017-10-20
JAMES-2199 SLF4J-related fixes

dtrebbien:
I used [SLF4J Helper for NetBeans IDE](http://plugins.netbeans.org/plugin/72557/) to find some fixes and improvements to logging code using SLF4J.

### DIFF
--- a/mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java
+++ b/mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java
@@ -146,12 +146,8 @@ public class ElasticSearchListeningMessageSearchIndex extends ListeningMessageSe
                 .map(uid ->  indexIdFor(mailbox, uid))
                 .collect(Collectors.toList()));
         } catch (Exception e) {
-            if (LOGGER.isErrorEnabled()) {
-                LOGGER.error("Error when deleting messages {} in mailbox {} from index",
-                    mailbox.getMailboxId().serialize(),
-                    ImmutableList.copyOf(expungedUids),
-                    e);
-            }
+            LOGGER.error("Error when deleting messages {} in mailbox {} from index",
+                    mailbox.getMailboxId().serialize(), expungedUids, e);
         }
     }
 
--- a/mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java
+++ b/mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java
@@ -121,19 +121,19 @@ public class ElasticSearchListeningMessageSearchIndex extends ListeningMessageSe
             indexer.indexMessage(indexIdFor(mailbox, message.getUid()), messageToElasticSearchJson.convertToJson(message, ImmutableList.of(session.getUser())));
         } catch (Exception e) {
             try {
-                LOGGER.warn(String.format("Indexing mailbox %s-%s of user %s on message %s without attachments ",
+                LOGGER.warn("Indexing mailbox {}-{} of user {} on message {} without attachments ",
                         mailbox.getName(),
                         mailbox.getMailboxId().serialize(),
                         session.getUser().getUserName(),
-                        message.getUid().toString()),
-                    e);
+                        message.getUid(),
+                        e);
                 indexer.indexMessage(indexIdFor(mailbox, message.getUid()), messageToElasticSearchJson.convertToJsonWithoutAttachment(message, ImmutableList.of(session.getUser())));
             } catch (JsonProcessingException e1) {
-                LOGGER.error(String.format("Error when indexing mailbox %s-%s of user %s on message %s without its attachment",
+                LOGGER.error("Error when indexing mailbox {}-{} of user {} on message {} without its attachment",
                         mailbox.getName(),
                         mailbox.getMailboxId().serialize(),
                         session.getUser().getUserName(),
-                        message.getUid().toString()),
+                        message.getUid(),
                         e1);
             }
         }
@@ -146,10 +146,12 @@ public class ElasticSearchListeningMessageSearchIndex extends ListeningMessageSe
                 .map(uid ->  indexIdFor(mailbox, uid))
                 .collect(Collectors.toList()));
         } catch (Exception e) {
-            LOGGER.error(String.format("Error when deleting messages %s in mailbox %s from index",
-                mailbox.getMailboxId().serialize(),
-                ImmutableList.copyOf(expungedUids).toString()),
-                e);
+            if (LOGGER.isErrorEnabled()) {
+                LOGGER.error("Error when deleting messages {} in mailbox {} from index",
+                    mailbox.getMailboxId().serialize(),
+                    ImmutableList.copyOf(expungedUids),
+                    e);
+            }
         }
     }
 
@@ -161,7 +163,7 @@ public class ElasticSearchListeningMessageSearchIndex extends ListeningMessageSe
                     JsonMessageConstants.MAILBOX_ID,
                     mailbox.getMailboxId().serialize()));
         } catch (Exception e) {
-            LOGGER.error(String.format("Error when deleting all messages in mailbox %s", mailbox.getMailboxId().serialize()), e);
+            LOGGER.error("Error when deleting all messages in mailbox {}", mailbox.getMailboxId().serialize(), e);
         }
     }
 
@@ -172,7 +174,7 @@ public class ElasticSearchListeningMessageSearchIndex extends ListeningMessageSe
                 .map(updatedFlags -> createUpdatedDocumentPartFromUpdatedFlags(mailbox, updatedFlags))
                 .collect(Collectors.toList()));
         } catch (Exception e) {
-            LOGGER.error(String.format("Error when updating index on mailbox %s", mailbox.getMailboxId().serialize()), e);
+            LOGGER.error("Error when updating index on mailbox {}", mailbox.getMailboxId().serialize(), e);
         }
     }
 
--- a/mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java
+++ b/mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java
@@ -146,8 +146,9 @@ public class ElasticSearchListeningMessageSearchIndex extends ListeningMessageSe
                 .map(uid ->  indexIdFor(mailbox, uid))
                 .collect(Collectors.toList()));
         } catch (Exception e) {
-            LOGGER.error("Error when deleting messages {} in mailbox {} from index",
-                    mailbox.getMailboxId().serialize(), expungedUids, e);
+            if (LOGGER.isErrorEnabled()) {
+                LOGGER.error("Error when deleting messages {} in mailbox {} from index", mailbox.getMailboxId().serialize(), expungedUids.toArray(), e);
+            }
         }
     }
 
--- a/mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
+++ b/mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
@@ -339,12 +339,12 @@ public class MaildirFolder {
                 String uidString = String.valueOf(uid.asLong());
                 String line = reader.readLine(); // the header
                 int lineNumber = 1;
-                while ((line = reader.readLine()) != null) {
+                for (; (line = reader.readLine()) != null; lineNumber++) {
                     if (!line.equals("")) {
                         int gap = line.indexOf(" ");
                         if (gap == -1) {
                             // there must be some issues in the file if no gap can be found
-                            LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber++);
+                            LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber);
                             continue;
                         }
 
@@ -582,12 +582,12 @@ public class MaildirFolder {
             if (line != null)
                 readUidListHeader(line);
             int lineNumber = 1;
-            while ((line = reader.readLine()) != null) {
+            for (; (line = reader.readLine()) != null; lineNumber++) {
                 if (!line.equals("")) {
                     int gap = line.indexOf(" ");
                     if (gap == -1) {
                         // there must be some issues in the file if no gap can be found
-                        throw new MailboxException("Corrupted entry in uid-file " + uidList + " line " + lineNumber++);
+                        throw new MailboxException("Corrupted entry in uid-file " + uidList + " line " + lineNumber);
                     }
                     MessageUid uid = MessageUid.of(Long.valueOf(line.substring(0, gap)));
                     String name = line.substring(gap + 1, line.length());
--- a/mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
+++ b/mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
@@ -344,7 +344,7 @@ public class MaildirFolder {
                         int gap = line.indexOf(" ");
                         if (gap == -1) {
                             // there must be some issues in the file if no gap can be found
-                            LOGGER.info("Corrupted entry in uid-file " + uidList + " line " + lineNumber++);
+                            LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber++);
                             continue;
                         }
 
@@ -510,7 +510,7 @@ public class MaildirFolder {
                         if (gap == -1) {
                             // there must be some issues in the file if no gap can be found
                             // there must be some issues in the file if no gap can be found
-                            LOGGER.info("Corrupted entry in uid-file " + uidList + " line " + lines.size());
+                            LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lines.size());
                             continue;
                         }
 
@@ -637,7 +637,7 @@ public class MaildirFolder {
 
                     if (gap == -1) {
                         // there must be some issues in the file if no gap can be found
-                    	LOGGER.info("Corrupted entry in uid-file " + uidList + " line " + lineNumber++);
+                    	LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber++);
                         continue;
                     }
                     
@@ -861,7 +861,7 @@ public class MaildirFolder {
                     int gap = line.indexOf(" ");
                     if (gap == -1) {
                         // there must be some issues in the file if no gap can be found
-                        LOGGER.info("Corrupted entry in uid-file " + uidList + " line " + lineNumber++);
+                        LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber++);
                         continue;
                     }
 
--- a/mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
+++ b/mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
@@ -338,8 +338,9 @@ public class MaildirFolder {
                 reader = new BufferedReader(fileReader);
                 String uidString = String.valueOf(uid.asLong());
                 String line = reader.readLine(); // the header
-                int lineNumber = 1;
-                for (; (line = reader.readLine()) != null; lineNumber++) {
+                int lineNumber = 1; // already read the first line
+                while ((line = reader.readLine()) != null) {
+                    lineNumber++;
                     if (!line.equals("")) {
                         int gap = line.indexOf(" ");
                         if (gap == -1) {
@@ -581,8 +582,9 @@ public class MaildirFolder {
             // the first line in the file contains the next uid and message count
             if (line != null)
                 readUidListHeader(line);
-            int lineNumber = 1;
-            for (; (line = reader.readLine()) != null; lineNumber++) {
+            int lineNumber = 1; // already read the first line
+            while ((line = reader.readLine()) != null) {
+                lineNumber++;
                 if (!line.equals("")) {
                     int gap = line.indexOf(" ");
                     if (gap == -1) {
@@ -630,8 +632,9 @@ public class MaildirFolder {
             // count
             if (line != null)
                 readUidListHeader(line);
-            int lineNumber = 1;
-            for (; (line = reader.readLine()) != null; lineNumber++) {
+            int lineNumber = 1; // already read the first line
+            while ((line = reader.readLine()) != null) {
+                lineNumber++;
                 if (!line.equals("")) {
                     int gap = line.indexOf(" ");
 
@@ -856,8 +859,9 @@ public class MaildirFolder {
                 // It may be possible that message count is 0 so we should better not try to calculate the size of the ArrayList
                 ArrayList<String> lines = new ArrayList<>();
                 String line;
-                int lineNumber = 1;
-                for (; (line = reader.readLine()) != null; lineNumber++) {
+                int lineNumber = 0;
+                while ((line = reader.readLine()) != null) {
+                    lineNumber++;
                     int gap = line.indexOf(" ");
                     if (gap == -1) {
                         // there must be some issues in the file if no gap can be found
--- a/mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
+++ b/mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
@@ -631,13 +631,13 @@ public class MaildirFolder {
             if (line != null)
                 readUidListHeader(line);
             int lineNumber = 1;
-            while ((line = reader.readLine()) != null) {
+            for (; (line = reader.readLine()) != null; lineNumber++) {
                 if (!line.equals("")) {
                     int gap = line.indexOf(" ");
 
                     if (gap == -1) {
                         // there must be some issues in the file if no gap can be found
-                    	LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber++);
+                    	LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber);
                         continue;
                     }
                     
@@ -857,11 +857,11 @@ public class MaildirFolder {
                 ArrayList<String> lines = new ArrayList<>();
                 String line;
                 int lineNumber = 1;
-                while ((line = reader.readLine()) != null) {
+                for (; (line = reader.readLine()) != null; lineNumber++) {
                     int gap = line.indexOf(" ");
                     if (gap == -1) {
                         // there must be some issues in the file if no gap can be found
-                        LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber++);
+                        LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber);
                         continue;
                     }
 
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
@@ -91,7 +91,7 @@ public abstract class ListeningMessageSearchIndex implements MessageSearchIndex,
                     try {
                         delete(session, expunged.getMailbox(), expunged.getUids());
                     } catch (MailboxException e) {
-                        LOGGER.error("Unable to deleted messages " + expunged.getUids() + " from index for mailbox " + expunged.getMailbox(), e);
+                        LOGGER.error("Unable to deleted messages {} from index for mailbox {}", expunged.getUids(), expunged.getMailbox(), e);
                     }
                 } else if (event instanceof EventFactory.FlagsUpdatedImpl) {
                     EventFactory.FlagsUpdatedImpl flagsUpdated = (EventFactory.FlagsUpdatedImpl) event;
@@ -100,7 +100,7 @@ public abstract class ListeningMessageSearchIndex implements MessageSearchIndex,
                     try {
                         update(session, mailbox, flagsUpdated.getUpdatedFlags());
                     } catch (MailboxException e) {
-                        LOGGER.error("Unable to update flags in index for mailbox " + mailbox, e);
+                        LOGGER.error("Unable to update flags in index for mailbox {}", mailbox, e);
                     }
                 }
             } else if (event instanceof EventFactory.MailboxDeletionImpl) {
@@ -132,7 +132,7 @@ public abstract class ListeningMessageSearchIndex implements MessageSearchIndex,
         try {
             add(session, mailbox, message);
         } catch (MailboxException e) {
-            LOGGER.error("Unable to index message " + message.getUid() + " for mailbox " + mailbox, e);
+            LOGGER.error("Unable to index message {} for mailbox {}", message.getUid(), mailbox, e);
         }
     }
 
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
@@ -121,8 +121,7 @@ public abstract class ListeningMessageSearchIndex implements MessageSearchIndex,
                     .findInMailbox(mailbox, MessageRange.one(next), FetchType.Full, UNLIMITED)
                     .next());
             } catch (Exception e) {
-                LOGGER.error("Could not retrieve message {} in mailbox {}",
-                    next, mailbox.getMailboxId().serialize(), e);
+                LOGGER.error("Could not retrieve message {} in mailbox {}", next, mailbox.getMailboxId().serialize(), e);
                 return Optional.empty();
             }
         }
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
@@ -121,8 +121,8 @@ public abstract class ListeningMessageSearchIndex implements MessageSearchIndex,
                     .findInMailbox(mailbox, MessageRange.one(next), FetchType.Full, UNLIMITED)
                     .next());
             } catch (Exception e) {
-                LOGGER.error(String.format("Could not retrieve message %d in mailbox %s",
-                    next, mailbox.getMailboxId().serialize()), e);
+                LOGGER.error("Could not retrieve message {} in mailbox {}",
+                    next, mailbox.getMailboxId().serialize(), e);
                 return Optional.empty();
             }
         }
--- a/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
+++ b/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
@@ -455,14 +455,7 @@ public abstract class AbstractSign extends GenericMailet {
             throw new MessagingException("Exception thrown", e);
         } finally {
             if (isDebug()) {
-                StringBuilder logBuffer =
-                new StringBuilder(1024)
-                .append("Other parameters:")
-                .append(", signerName=").append(getSignerName())
-                .append(", postmasterSigns=").append(postmasterSigns)
-                .append(", rebuildFrom=").append(rebuildFrom)
-                .append(" ");
-                LOGGER.debug(logBuffer.toString());
+                LOGGER.debug("Other parameters: signerName={}, postmasterSigns={}, rebuildFrom={}", getSignerName(), postmasterSigns, rebuildFrom);
             }
         }
         
--- a/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
+++ b/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
@@ -227,7 +227,7 @@ public abstract class AbstractSign extends GenericMailet {
      */
     protected void initExplanationText() {
         setExplanationText(getInitParameter("explanationText"));
-        if (isDebug()) {
+        if (isDebug() && LOGGER.isDebugEnabled()) {
             LOGGER.debug("Explanation text:\r\n" + getExplanationText());
         }
     }
--- a/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
+++ b/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
@@ -294,14 +294,11 @@ public abstract class AbstractSign extends GenericMailet {
         }
         
         if (isDebug()) {
-            StringBuilder logBuffer =
-            new StringBuilder(1024)
-            .append("KeyStore related parameters:")
-            .append("  keyStoreFileName=").append(keyStoreFileName)
-            .append(", keyStoreType=").append(keyStoreType)
-            .append(", keyAlias=").append(keyAlias)
-            .append(" ");
-            LOGGER.debug(logBuffer.toString());
+            LOGGER.debug("KeyStore related parameters:"
+                         + "  keyStoreFileName={}"
+                         + ", keyStoreType={}"
+                         + ", keyAlias={}",
+                    keyStoreFileName, keyStoreType, keyAlias);
         }
             
         // Certificate preparation
--- a/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
+++ b/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
@@ -294,11 +294,7 @@ public abstract class AbstractSign extends GenericMailet {
         }
         
         if (isDebug()) {
-            LOGGER.debug("KeyStore related parameters:"
-                         + "  keyStoreFileName={}"
-                         + ", keyStoreType={}"
-                         + ", keyAlias={}",
-                    keyStoreFileName, keyStoreType, keyAlias);
+            LOGGER.debug("KeyStore related parameters: keyStoreFileName={}, keyStoreType={}, keyAlias={}", keyStoreFileName, keyStoreType, keyAlias);
         }
             
         // Certificate preparation
--- a/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
+++ b/mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
@@ -202,7 +202,7 @@ public abstract class AbstractSign extends GenericMailet {
             throw new MessagingException("The specified <keyHolderClass> does not exist: " + keyHolderClassName);
         }
         if (isDebug()) {
-            LOGGER.debug("keyHolderClass: " + getKeyHolderClass());
+            LOGGER.debug("keyHolderClass: {}", getKeyHolderClass());
         }
     }
     
@@ -309,7 +309,7 @@ public abstract class AbstractSign extends GenericMailet {
         setKeyHolder((KeyHolder)keyHolderConstructor.newInstance(parameters));
         
         if (isDebug()) {
-            LOGGER.debug("Subject Distinguished Name: " + getKeyHolder().getSignerDistinguishedName());
+            LOGGER.debug("Subject Distinguished Name: {}", getKeyHolder().getSignerDistinguishedName());
         }
         
         if (getKeyHolder().getSignerAddress() == null) {
@@ -538,7 +538,7 @@ public abstract class AbstractSign extends GenericMailet {
             mail.setAttribute(SMIMEAttributeNames.SMIME_SIGNER_ADDRESS, getKeyHolder().getSignerAddress());
             
             if (isDebug()) {
-                LOGGER.debug("Message signed, reverse-path: " + mail.getSender() + ", Id: " + messageId);
+                LOGGER.debug("Message signed, reverse-path: {}, Id: {}", mail.getSender(), messageId);
             }
             
         } catch (MessagingException me) {
@@ -583,7 +583,7 @@ public abstract class AbstractSign extends GenericMailet {
         String authUser = (String) mail.getAttribute(Mail.SMTP_AUTH_USER_ATTRIBUTE_NAME);
         // was the sender user SMTP authorized?
         if (authUser == null) {
-            LOGGER.info("Can not sign mail for sender <" + mail.getSender() + "> as he is not a SMTP authenticated user");
+            LOGGER.info("Can not sign mail for sender <{}> as he is not a SMTP authenticated user", mail.getSender());
             return false;
         }
         
@@ -598,7 +598,7 @@ public abstract class AbstractSign extends GenericMailet {
             // is the reverse-path user different from the SMTP authorized user?
             String username = getUsername(reversePath);
             if (!username.equals(authUser)) {
-                LOGGER.info("SMTP logged in as <" + authUser + "> but pretend to be sender <" + username + ">");
+                LOGGER.info("SMTP logged in as <{}> but pretend to be sender <{}>", authUser, username);
                 return false;
             }
             // is there no "From:" address same as the reverse-path?
@@ -661,7 +661,7 @@ public abstract class AbstractSign extends GenericMailet {
                     try {
                         mailAddress = new MailAddress(aFromArray);
                     } catch (ParseException pe) {
-                        LOGGER.info("Unable to parse a \"FROM\" header address: " + aFromArray.toString() + "; ignoring.");
+                        LOGGER.info("Unable to parse a \"FROM\" header address: {}; ignoring.", aFromArray);
                         continue;
                     }
                     if (mailAddress.equals(reversePath)) {
--- a/mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
+++ b/mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
@@ -21,11 +21,13 @@ package org.apache.james.transport.mailets;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.Serializable;
+import java.nio.charset.Charset;
 import java.util.Map;
 import java.util.stream.Stream;
 
 import javax.mail.MessagingException;
 
+import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.lang3.tuple.Pair;
 import org.apache.mailet.Mail;
 import org.apache.mailet.base.GenericMailet;
@@ -134,10 +136,28 @@ public class ICalendarParser extends GenericMailet {
             ByteArrayInputStream inputStream = new ByteArrayInputStream(icsContent);
             return Stream.of(Pair.of(key, builder.build(inputStream)));
         } catch (IOException e) {
-            LOGGER.error("Error while reading input: " + icsContent, e);
+            if (LOGGER.isErrorEnabled()) {
+                String icsContentAsString;
+                try {
+                    icsContentAsString = new String(icsContent, Charset.forName("UTF-8"));
+                } catch (Throwable t) {
+                    LOGGER.error("Error while decoding ics content", t);
+                    icsContentAsString = new String(Hex.encodeHex(icsContent));
+                }
+                LOGGER.error("Error while reading input: " + icsContentAsString, e);
+            }
             return Stream.of();
         } catch (ParserException e) {
-            LOGGER.error("Error while parsing ICal object: " + icsContent, e);
+            if (LOGGER.isErrorEnabled()) {
+                String icsContentAsString;
+                try {
+                    icsContentAsString = new String(icsContent, Charset.forName("UTF-8"));
+                } catch (Throwable t) {
+                    LOGGER.error("Error while decoding ics content", t);
+                    icsContentAsString = new String(Hex.encodeHex(icsContent));
+                }
+                LOGGER.error("Error while parsing ICal object: " + icsContentAsString, e);
+            }
             return Stream.of();
         }
     }
--- a/mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
+++ b/mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
@@ -21,7 +21,7 @@ package org.apache.james.transport.mailets;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.Serializable;
-import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.stream.Stream;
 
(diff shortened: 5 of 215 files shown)


server/container/spring/src/main/java/org/apache/james/container/spring/bean/factory/mailrepositorystore/MailRepositoryStoreBeanFactory.java
```diff
@@ -160,19 +160,12 @@ public synchronized MailRepository select(String destination) throws MailReposit
...
-            if (LOGGER.isDebugEnabled()) {
-                logBuffer = new StringBuffer(128).append("obtained repository: ").append(repID).append(",").append(reply.getClass());
-                LOGGER.debug(logBuffer.toString());
-            }
+            LOGGER.debug("obtained repository: {},{}", repID, reply.getClass());
             return reply;
         } else {
             String repClass = classes.get(protocol);
-            if (LOGGER.isDebugEnabled()) {
-                logBuffer = new StringBuffer(128).append("obtained repository: ").append(repClass).append(" to handle: ").append(protocol).append(" with key ").append(protocol);
```
dtrebbien: I faithfully converted this to a parameterized logger call, but the log does not look right to me. Notice that `protocol` is inserted twice into the log message.

mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java
```diff
@@ -146,10 +146,12 @@ public void delete(MailboxSession session, Mailbox mailbox, List<MessageUid> exp
                 .map(uid ->  indexIdFor(mailbox, uid))
                 .collect(Collectors.toList()));
         } catch (Exception e) {
-            LOGGER.error(String.format("Error when deleting messages %s in mailbox %s from index",
-                mailbox.getMailboxId().serialize(),
-                ImmutableList.copyOf(expungedUids).toString()),
-                e);
+            if (LOGGER.isErrorEnabled()) {
```
chibenwa: Maybe we can remove this "isErrorEnabled"

mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
```diff
@@ -134,10 +136,28 @@ public String getMailetInfo() {
             ByteArrayInputStream inputStream = new ByteArrayInputStream(icsContent);
             return Stream.of(Pair.of(key, builder.build(inputStream)));
         } catch (IOException e) {
-            LOGGER.error("Error while reading input: " + icsContent, e);
+            if (LOGGER.isErrorEnabled()) {
+                String icsContentAsString;
```
chibenwa: Extract "icsContentAsString" in an other method to make this more readable, please ;-)

mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
```diff
@@ -134,10 +136,28 @@ public String getMailetInfo() {
...
+                try {
+                    icsContentAsString = new String(icsContent, Charset.forName("UTF-8"));
+                } catch (Throwable t) {
```
chibenwa: Catch Exception, not Throwable, please

mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
```diff
@@ -134,10 +136,28 @@ public String getMailetInfo() {
             ByteArrayInputStream inputStream = new ByteArrayInputStream(icsContent);
             return Stream.of(Pair.of(key, builder.build(inputStream)));
         } catch (IOException e) {
-            LOGGER.error("Error while reading input: " + icsContent, e);
+            if (LOGGER.isErrorEnabled()) {
+                String icsContentAsString;
+                try {
+                    icsContentAsString = new String(icsContent, Charset.forName("UTF-8"));
```
chibenwa: StandardCharsets.UTF-8

mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
```diff
@@ -134,10 +136,28 @@ public String getMailetInfo() {
...
+                    LOGGER.error("Error while decoding ics content", t);
+                    icsContentAsString = new String(Hex.encodeHex(icsContent));
+                }
+                LOGGER.error("Error while reading input: " + icsContentAsString, e);
+            }
             return Stream.of();
         } catch (ParserException e) {
-            LOGGER.error("Error while parsing ICal object: " + icsContent, e);
+            if (LOGGER.isErrorEnabled()) {
+                String icsContentAsString;
```
chibenwa: You have code duplication. Please prefer extracting this byte array -> string method.

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMessageRangeProcessor.java
```diff
@@ -113,7 +113,9 @@ protected void doProcess(M request, ImapSession session, String tag, ImapCommand
             }
             taggedBad(command, tag, responder, HumanReadableText.INVALID_MESSAGESET);
         } catch (MailboxException e) {
-            LOGGER.error(getOperationName() + " failed from mailbox " + currentMailbox.getPath() + " to " + targetMailbox + " for sequence-set " + ImmutableList.copyOf(idSet), e);
+            if (LOGGER.isErrorEnabled()) {
```
chibenwa: What is the point of logger.is\*Enabled() everywhere in this PR?

I know it is to avoid string formatting, but I feel like it might be sometime premature optimization... I don't think this log operation is expensive, and I would value not making the code look more complex... 

Also what forbids us to use "placeholders"?

server/container/spring/src/main/java/org/apache/james/container/spring/bean/factory/mailrepositorystore/MailRepositoryStoreBeanFactory.java
chibenwa: Having looked at the code we can remove the "with key..." part IMO.

server/data/data-file/src/main/java/org/apache/james/mailrepository/file/FileMailRepository.java
```diff
@@ -244,7 +244,7 @@ public Mail retrieve(String key) throws MessagingException {
 
             return mc;
         } catch (Exception me) {
-            LOGGER.error("Exception retrieving mail: " + me);
+            LOGGER.error("Exception retrieving mail: {}", me.toString());
```
chibenwa: LOGGER.error("Exception retrieving mail", me);

Overwize we swallow the stacktrace

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/delivery/SimpleMailStore.java
```diff
@@ -110,7 +110,9 @@ private String computeUsername(MailAddress recipient) throws MessagingException
         try {
             return usersRepository.getUser(recipient);
         } catch (UsersRepositoryException e) {
-            LOGGER.warn("Unable to retrieve username for " + recipient.asPrettyString(), e);
+            if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn("Unable to retrieve username for " + recipient.asPrettyString(), e);
```
chibenwa: Why no placeholder here?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/jsieve/delivery/SieveExecutor.java
```diff
@@ -124,10 +124,14 @@ protected boolean sieveMessage(MailAddress recipient, Mail aMail) throws Messagi
             sieveMessageEvaluate(recipient, aMail, userSieveInformation);
             return true;
         } catch (ScriptNotFoundException e) {
-            LOGGER.info("Can not locate SIEVE script for user " + recipient.asPrettyString());
+            if (LOGGER.isInfoEnabled()) {
+                LOGGER.info("Can not locate SIEVE script for user " + recipient.asPrettyString());
```
chibenwa: And here?

chibenwa:
Don't you mind opening an APache ticket for this?

https://issues.apache.org/jira/browse/JAMES

mailet/standard/src/main/java/org/apache/james/transport/mailets/MimeDecodingMailet.java
```diff
@@ -84,8 +84,8 @@ public void service(Mail mail) throws MessagingException {
         Serializable attributeContent = mail.getAttribute(attribute);
         if (! (attributeContent instanceof Map)) {
             LOGGER.debug("Invalid attribute found into attribute "
-                    + attribute + "class Map expected but "
-                    + attributeContent.getClass() + " found.");
+                    + "{} class Map expected but "
+                    + "{} found.", attribute, attributeContent.getClass());
             return ImmutableMap.of();
```
strokyl: Maybe you could inline the string now there is no concatenation anymore

mailet/standard/src/main/java/org/apache/james/transport/mailets/StripAttachment.java
```diff
@@ -411,7 +411,7 @@ private String decodeFilename(String fileName) throws UnsupportedEncodingExcepti
         boolean result = isMatchingPattern(name, regExPattern).orElse(false)
                 || !isMatchingPattern(name, notRegExPattern).orElse(true);
 
-        LOGGER.debug("attachment " + name + " " + ((result) ? "matches" : "does not match"));
+        LOGGER.debug("attachment {} {}", name, (result) ? "matches" : "does not match");
```
strokyl: () arround result are useless IMO

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/redirect/MailModifier.java
```diff
@@ -120,7 +120,7 @@ public void setTo(List<MailAddress> mailAddresses) throws MessagingException {
             InternetAddress[] internetAddresses = MailAddressUtils.toInternetAddressArray(mailAddresses);
             mail.getMessage().setRecipients(Message.RecipientType.TO, internetAddresses);
             if (mailet.getInitParameters().isDebug()) {
-                LOGGER.debug("apparentlyTo set to: " + internetAddresses);
+                LOGGER.debug("apparentlyTo set to: {}", (Object) internetAddresses);
```
strokyl: Why this cast ?

server/protocols/jmap/src/main/java/org/apache/james/jmap/send/PostDequeueDecorator.java
```diff
@@ -81,7 +81,7 @@ public void done(boolean success) throws MailQueueException {
                     moveFromOutboxToSent(messageId, mailboxSession);
                     getMail().setAttribute(IS_DELIVERED, IS_DELIVERED);
                 } catch (MailShouldBeInOutboxException e) {
-                    LOG.info("Message does not exist on Outbox anymore, it could have already been sent {}", e);
+                    LOG.info("Message does not exist on Outbox anymore, it could have already been sent {}", e.toString());
```
strokyl: Why do you add "toString()"

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMessageRangeProcessor.java
strokyl: :+1:

server/data/data-library/src/main/java/org/apache/james/repository/file/AbstractFileRepository.java
```diff
@@ -188,7 +188,7 @@ public Repository getChildRepository(String childName) {
             throw new RuntimeException("Cannot initialize child " + "repository " + childName + " : " + e);
         }
 
-        if (DEBUG) {
+        if (DEBUG && LOGGER.isDebugEnabled()) {
```
strokyl: why ``if (DEBUG)`` is not enough ?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/DSNBounce.java
```diff
@@ -248,7 +248,7 @@ private void trySendBounce(Mail originalMail) throws MessagingException {
             newMail.setRemoteAddr(getRemoteAddr());
             newMail.setRecipients(getSenderAsList(originalMail));
        
-            if (getInitParameters().isDebug()) {
+            if (getInitParameters().isDebug() && LOGGER.isDebugEnabled()) {
```
strokyl: Same question here

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/redirect/ProcessRedirectNotify.java
```diff
@@ -57,7 +57,7 @@ public void process(Mail originalMail) throws MessagingException {
             mailModifier.setRemoteAddr();
             mailModifier.setRemoteHost();
 
-            if (mailet.getInitParameters().isDebug()) {
+            if (mailet.getInitParameters().isDebug() && LOGGER.isDebugEnabled()) {
```
strokyl: Same question here

mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java
strokyl: Hum I think the copyOf can be a bit expansive, it is good to have a if here

mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
```diff
@@ -121,8 +121,8 @@ public void event(Event event) {
                     .findInMailbox(mailbox, MessageRange.one(next), FetchType.Full, UNLIMITED)
                     .next());
             } catch (Exception e) {
-                LOGGER.error(String.format("Could not retrieve message %d in mailbox %s",
-                    next, mailbox.getMailboxId().serialize()), e);
+                LOGGER.error("Could not retrieve message {} in mailbox {}",
```
strokyl: I think who can online those two lines

mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
```diff
@@ -294,14 +294,11 @@ protected void initKeyHolder() throws Exception {
...
         if (isDebug()) {
-            StringBuilder logBuffer =
-            new StringBuilder(1024)
-            .append("KeyStore related parameters:")
-            .append("  keyStoreFileName=").append(keyStoreFileName)
-            .append(", keyStoreType=").append(keyStoreType)
-            .append(", keyAlias=").append(keyAlias)
-            .append(" ");
-            LOGGER.debug(logBuffer.toString());
+            LOGGER.debug("KeyStore related parameters:"
```
strokyl: Same here you can remove this now useless string concatenation

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SendMailHandler.java
```diff
@@ -77,8 +77,7 @@ public HookResult onMessage(SMTPSession session, Mail mail) {
                 recipientString = theRecipients.toString();
             }
             if (LOGGER.isInfoEnabled()) {
-                String infoBuffer = "Successfully spooled mail " + mail.getName() + " from " + mail.getSender() + " on " + session.getRemoteAddress().getAddress().toString() + " for " + recipientString;
-                LOGGER.info(infoBuffer.toString());
+                LOGGER.info("Successfully spooled mail " + mail.getName() + " from " + mail.getSender() + " on " + session.getRemoteAddress().getAddress().toString() + " for " + recipientString);
```
strokyl: You can remove the if and use placeholder here I think

dtrebbien:
I created a new JIRA issue: https://issues.apache.org/jira/browse/JAMES-2199

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
```diff
@@ -344,7 +344,7 @@ public MaildirMessageName getMessageNameByUid(final MailboxSession session, fina
                         int gap = line.indexOf(" ");
                         if (gap == -1) {
                             // there must be some issues in the file if no gap can be found
-                            LOGGER.info("Corrupted entry in uid-file " + uidList + " line " + lineNumber++);
+                            LOGGER.info("Corrupted entry in uid-file {} line {}", uidList, lineNumber++);
```
mbaechler: are arguments evaluated _at all_ ? because evaluating lineNumber++ or not make a difference here.

mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
dtrebbien: Done

mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
dtrebbien: Done

mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
dtrebbien: Done

mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
dtrebbien: Done

server/data/data-file/src/main/java/org/apache/james/mailrepository/file/FileMailRepository.java
dtrebbien: Done

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
dtrebbien: Yes, arguments are evaluated. You are right about the side effects making a difference.

Perhaps it would be better to move the increment of `lineNumber` to immediately after the info() call?

dtrebbien: Actually, it appears that the intent is to increment `lineNumber` with each line read.

I went ahead and pushed a commit to increment `lineNumber` in the "increment" portion of a for loop, both here and a similar place further down in the file.

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMessageRangeProcessor.java
dtrebbien: In addition to string concatenations, the is*Enabled() checks help prevent unnecessary work from being performed if, ultimately, the work will be wasted. Here, for example, there is a call to `ImmutableList.copyOf(idSet)`.

If I wasn't sure if the amount of work could somehow be controlled by an external actor, possibly leading to a DoS vulnerability, I added the check for is*Enabled(). If this concern is not applicable, then parameterized logging is probably the way to go.

The Guava implementation of ImmutableList.copyOf(E[]) generally will create a copy of the array:

https://github.com/google/guava/blob/e3b8c0b6f07c1017e80d86758191d6898517a546/guava/src/com/google/common/collect/ImmutableList.java#L285..L294

Now that I look at the code again, I think we can use Arrays.asList() here because a copy doesn't actually need to be made. The reason is to convert the array to a `List`.

dtrebbien: I just pushed a commit making this change.

server/container/spring/src/main/java/org/apache/james/container/spring/bean/factory/mailrepositorystore/MailRepositoryStoreBeanFactory.java
dtrebbien: Done

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/delivery/SimpleMailStore.java
dtrebbien: Changed to use a placeholder

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/jsieve/delivery/SieveExecutor.java
dtrebbien: Changed to use placeholders

mailet/standard/src/main/java/org/apache/james/transport/mailets/MimeDecodingMailet.java
dtrebbien: Done

mailet/standard/src/main/java/org/apache/james/transport/mailets/StripAttachment.java
dtrebbien: Removed

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/redirect/MailModifier.java
dtrebbien: The cast was to avoid the varargs debug() overload from being called, because `internetAddresses` is an array.

I now think Arrays.asList() should be used here, and I just pushed a commit to make this change.

server/protocols/jmap/src/main/java/org/apache/james/jmap/send/PostDequeueDecorator.java
dtrebbien: This is to work around a regression introduced in SLF4J 1.7.15. James Project is using SLF4J 1.7.25, which is affected.

See http://mailman.qos.ch/pipermail/slf4j-user/2017-August/001639.html and the follow-up message http://mailman.qos.ch/pipermail/slf4j-user/2017-August/001640.html

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/DSNBounce.java
dtrebbien: Changed to parameterized logging

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/redirect/ProcessRedirectNotify.java
dtrebbien: Changed to parameterized logging

mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
dtrebbien: Done

mailet/crypto/src/main/java/org/apache/james/transport/mailets/AbstractSign.java
dtrebbien: Done

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SendMailHandler.java
dtrebbien: Done.  Additionally, I merged the`recipientString` calculation into the info() logger call.

dtrebbien:
Rebased to resolve merge conflict

server/protocols/jmap/src/main/java/org/apache/james/jmap/send/PostDequeueDecorator.java
strokyl: I guess here we want to have the error with the full stacktrace so I would go for:
```LOG.info("Message does not exist on Outbox anymore, it could have already been sent", e);""

File: mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java (hunk 21 above)
Comment (strokyl) [reply mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java]: Imo we should have the increment on a separate line before then the call for logging

File: server/data/data-library/src/main/java/org/apache/james/repository/file/AbstractFileRepository.java (hunk 15 above)
Comment (aduprat) [reply server/data/data-library/src/main/java/org/apache/james/repository/file/AbstractFileRepository.java]: I agree with you, the aim of the _DEBUG_ variable is not clear.
BTW, depending on what you want to print, it's a good habit to have a condition before calling `LOGGER.debug`
No in this case, but in some cases, formatting the string to be logged is time consuming (extracting things from stream...).

To summarize:
- in this case I agree that we don't need the condition
- in some cases, a condition on the logger level is needed

File: server/protocols/jmap/src/main/java/org/apache/james/jmap/send/PostDequeueDecorator.java (hunk 14 above)
Comment (aduprat) [reply server/protocols/jmap/src/main/java/org/apache/james/jmap/send/PostDequeueDecorator.java]: :+1:

Comment (dtrebbien) [reply server/protocols/jmap/src/main/java/org/apache/james/jmap/send/PostDequeueDecorator.java]: Done

File: protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMessageRangeProcessor.java (hunk 7 above)
Comment (dtrebbien) [reply protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMessageRangeProcessor.java]: I was mistaken in thinking that SLF4J supported logging Collections directly. Actually, it only appeared to work because most implementations of Collection.toString() come from `java.util.AbstractCollection`.

SLF4J "natively" supports printing arrays (see https://github.com/qos-ch/slf4j/blob/a81440ea36676f3b090f5a4ccba70d17c7b80f6e/slf4j-api/src/main/java/org/slf4j/helpers/MessageFormatter.java#L265 ), so I think that it is better to pass `idSet ` directly to the logger call as a formatting arg. I have just pushed a commit making this change.

**Comment by dtrebbien** [comment]
I have pushed some more commits that remove many of the is*Enabled() checks, using parameterized logging instead.

File: mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java (hunk 2 above)
Comment (dtrebbien) [reply mailbox/elasticsearch/src/main/java/org/apache/james/mailbox/elasticsearch/events/ElasticSearchListeningMessageSearchIndex.java]: **UPDATE:** I have pushed another commit changing the logger call to pass `expungedUids.toArray()` rather than `expungedUids`.

SLF4J supports logging arrays directly, but not Collections. For Collection instances, SLF4J relies on toString(), which will probably produce the same output; however, toString() is not a method of the `List` interface, so its behavior is unspecified and left to implementations.

**Comment by dtrebbien** [comment]
Rebased to resolve the merge conflicts.

File: mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java [hunk 22]
```diff
@@ -339,12 +339,12 @@ public MaildirMessageName getMessageNameByUid(final MailboxSession session, fina
                 String uidString = String.valueOf(uid.asLong());
                 String line = reader.readLine(); // the header
                 int lineNumber = 1;
-                while ((line = reader.readLine()) != null) {
+                for (; (line = reader.readLine()) != null; lineNumber++) {
```
chibenwa: I don't like that loop change. while seems more expressive that for...

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
```diff
@@ -582,12 +582,12 @@ public MaildirMessageName getMessageNameByUid(final MailboxSession session, fina
             if (line != null)
                 readUidListHeader(line);
             int lineNumber = 1;
-            while ((line = reader.readLine()) != null) {
+            for (; (line = reader.readLine()) != null; lineNumber++) {
```
chibenwa: idem

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
```diff
@@ -631,13 +631,13 @@ public MaildirMessageName getMessageNameByUid(final MailboxSession session, fina
...
```
chibenwa: idem

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
```diff
@@ -857,11 +857,11 @@ public MaildirMessageName delete(final MailboxSession session, final MessageUid
                 ArrayList<String> lines = new ArrayList<>();
                 String line;
                 int lineNumber = 1;
-                while ((line = reader.readLine()) != null) {
+                for (; (line = reader.readLine()) != null; lineNumber++) {
```
chibenwa: And idem

mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
```diff
@@ -738,16 +737,16 @@ protected void ping(InetAddress address) throws Exception {
...
             if (isDebug()) {
-                LOGGER.debug("Trial #" + ping + "/" + getMaxPings() + " - creating socket connected to " + address + " on port " + getPort());
+                LOGGER.debug("Trial #{}/{} - creating socket connected to {} on port {}", ping, getMaxPings(), address, getPort());
             }
             try {
                 socket = new Socket(address, getPort());
                 break;
             } catch (ConnectException ce) {
-                LOGGER.debug("Trial #" + ping + "/" + getMaxPings() + " - exception caught: " + ce.toString() + " while creating socket connected to " + address + " on port " + getPort());
+                LOGGER.debug("Trial #{}/{} - exception caught: {} while creating socket connected to {} on port {}", ping, getMaxPings(), ce, address, getPort());
```
chibenwa: This log line seems still poor.

Something like 

```
LOGGER.debug("Trial #{}/{} - exception caught while creating socket connected to {} on port {}", ping, getMaxPings(), address, getPort(), ce);
```

That would avoid loosing the stacktrace...

server/container/cli/src/main/java/org/apache/james/cli/ServerCmd.java
```diff
@@ -94,7 +94,7 @@ public static void main(String[] args) {
         } catch (IOException ioe) {
             failWithMessage("Error connecting to remote JMX agent : " + ioe.getMessage());
         } catch (Exception e) {
-            LOG.error("Error on command: {}", e);
+            LOG.error("Error on command: {}", e.toString());
```
chibenwa: Swallowing stacktrace ;-)

I would be in favour of:

```
LOG.error("Error while playing command", e);
```

server/container/util-java8/src/test/java/org/apache/james/util/streams/SwarmGenericContainer.java
```diff
@@ -63,7 +63,7 @@ public SwarmGenericContainer(ImageFromDockerfile imageFromDockerfile) {
         }
     }
     private void logAndCheckSkipTest(IllegalStateException e) {
-        LOGGER.error("Cannot initial a docker container because: " + e);
+        LOGGER.error("Cannot initial a docker container because: {}", String.valueOf(e));
```
chibenwa:
```
 LOGGER.error("Cannot initial a docker container", e);
````

Would be better ;-)

File: server/container/util/src/main/java/org/apache/james/util/retry/naming/LoggingRetryHandler.java [hunk 29]
```diff
@@ -50,9 +50,7 @@ public LoggingRetryHandler(Class<?>[] exceptionClasses, ExceptionRetryingProxy p
     @Override
     public void postFailure(NamingException ex, int retryCount) {
         super.postFailure(ex, retryCount);
-        LOGGER.info(
-                "Retry failure: " + ex.getLocalizedMessage() + "\n Retrying in " + getRetryInterval(retryCount) / 1000 + " seconds"
-                );
+        LOGGER.info("Retry failure: {}\n Retrying in {} seconds", ex.getLocalizedMessage(), getRetryInterval(retryCount) / 1000);
```
chibenwa: Again here the expection is swallowed.

Moreover \\n in log messages should be avoided.

What about:

```
LOGGER.info("Retry failure. Retrying in {} seconds", getRetryInterval(retryCount) / 1000, ex);
```

server/data/data-file/src/main/java/org/apache/james/mailrepository/file/FileMailRepository.java
```diff
@@ -145,13 +143,9 @@ public void init() throws Exception {
...
             }
-            if (LOGGER.isDebugEnabled()) {
-                String logBuffer = getClass().getName() + " created in " + destination;
-                LOGGER.debug(logBuffer);
-            }
+            LOGGER.debug("{} created in {}", getClass().getName(), destination);
         } catch (Exception e) {
-            final String message = "Failed to retrieve Store component:" + e.getMessage();
-            LOGGER.error(message, e);
+            LOGGER.error("Failed to retrieve Store component:{}", e.getMessage(), e);
```
chibenwa: IMO

```
LOGGER.error("Failed to retrieve Store component:", e);
```

Would be enough

server/data/data-file/src/main/java/org/apache/james/user/file/UsersFileRepository.java
```diff
@@ -98,14 +98,9 @@ public void init() throws Exception {
...
-            if (LOGGER.isDebugEnabled()) {
-                String logBuffer = this.getClass().getName() + " created in " + destination;
-                LOGGER.debug(logBuffer);
-            }
+            LOGGER.debug("{} created in {}", getClass().getName(), destination);
         } catch (Exception e) {
-            if (LOGGER.isErrorEnabled()) {
-                LOGGER.error("Failed to initialize repository:" + e.getMessage(), e);
-            }
+            LOGGER.error("Failed to initialize repository: {}", e.getMessage(), e);
```
chibenwa: We can get ride of getMessage

server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
```diff
@@ -255,15 +247,10 @@ public void init() throws Exception {
...
             }
 
-            if (LOGGER.isDebugEnabled()) {
-                String logBuf = this.getClass().getName() + " created according to " + destination;
-                LOGGER.debug(logBuf);
-            }
+            LOGGER.debug("{} created according to {}", getClass().getName(), destination);
         } catch (Exception e) {
             final String message = "Failed to retrieve Store component:" + e.getMessage();
             LOGGER.error(message, e);
```
chibenwa: Looks like we can:
 - get ride of the getMessage
 - inline the message in the log operation

server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
```diff
@@ -380,14 +361,12 @@ protected void checkJdbcAttributesSupport(DatabaseMetaData dbMetaData) throws SQ
...
-                logBuffer.append("JDBC Mail Attributes support ready.");
-                LOGGER.info(logBuffer.toString());
-            }
+            logBuffer.append("JDBC Mail Attributes support ready.");
+            LOGGER.info("{}", logBuffer);
         } else {
             jdbcMailAttributesReady = false;
             logBuffer.append("JDBC Mail Attributes support not activated. " + "Missing both 'updateMessageAttributesSQL' " + "and 'retrieveMessageAttributesSQL' " + "statements for table '").append(tableName).append("' in sqlResources.xml. ").append("Will not persist in the repository '").append(repositoryName).append("'.");
-            LOGGER.warn(logBuffer.toString());
+            LOGGER.warn("{}", logBuffer);
```
chibenwa: Oups?

server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
```diff
@@ -380,14 +361,12 @@ protected void checkJdbcAttributesSupport(DatabaseMetaData dbMetaData) throws SQ
         }
         if (hasUpdateMessageAttributesSQL && hasRetrieveMessageAttributesSQL) {
             jdbcMailAttributesReady = true;
-            if (LOGGER.isInfoEnabled()) {
-                logBuffer.append("JDBC Mail Attributes support ready.");
-                LOGGER.info(logBuffer.toString());
-            }
+            logBuffer.append("JDBC Mail Attributes support ready.");
+            LOGGER.info("{}", logBuffer);
```
chibenwa: Oups?

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailSpooler.java
```diff
@@ -165,17 +159,14 @@ public void run() {
...
+                        LOGGER.debug("==== Begin processing mail {} ====", mail.getName());
 
                         try {
                             mailProcessor.service(mail);
                             queueItem.done(true);
                         } catch (Exception e) {
-                            if (active.get() && LOGGER.isErrorEnabled()) {
-                                LOGGER.error("Exception processing mail while spooling " + e.getMessage(), e);
+                            if (active.get()) {
+                                LOGGER.error("Exception processing mail while spooling {}", e.getMessage(), e);
```
chibenwa: getMessage

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailSpooler.java
```diff
@@ -184,8 +175,8 @@ public void run() {
                             mail = null;
                         }
                     } catch (Throwable e) {
-                        if (active.get() && LOGGER.isErrorEnabled()) {
-                            LOGGER.error("Exception processing mail while spooling " + e.getMessage(), e);
+                        if (active.get()) {
+                            LOGGER.error("Exception processing mail while spooling {}", e.getMessage(), e);
```
chibenwa: getMessage

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
dtrebbien: Done

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
dtrebbien: Done

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
dtrebbien: Done

mailbox/maildir/src/main/java/org/apache/james/mailbox/maildir/MaildirFolder.java
dtrebbien: Done

server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
dtrebbien: This is intentional; its purpose is to avoid creating a String of the contents of `logBuffer` if logging at the info level is not enabled.

mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
dtrebbien: Done

server/container/cli/src/main/java/org/apache/james/cli/ServerCmd.java
dtrebbien: Done

server/container/util-java8/src/test/java/org/apache/james/util/streams/SwarmGenericContainer.java
dtrebbien: Done

server/container/util/src/main/java/org/apache/james/util/retry/naming/LoggingRetryHandler.java
dtrebbien: Done

server/data/data-file/src/main/java/org/apache/james/mailrepository/file/FileMailRepository.java
dtrebbien: Done

server/data/data-file/src/main/java/org/apache/james/user/file/UsersFileRepository.java
dtrebbien: Done

server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
dtrebbien: Done

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailSpooler.java
dtrebbien: Done

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailSpooler.java
dtrebbien: Done

dtrebbien:
I think that line 110 of `HasHabeasWarrantMark.java` needs a better log message, but I don't know what to suggest:

https://github.com/apache/james-project/blob/21188d00e37d425289b77da4b2ab650991c656bc/mailet/standard/src/main/java/org/apache/james/transport/matchers/HasHabeasWarrantMark.java#L110

chibenwa:
Caught an exception while reading message?

chibenwa:
```
[b7f0b0ab682766ae39ef8578ecdd9d246bdcf4e3] [INFO] Apache James :: Standard Mailets ................... FAILURE [ 43.264 s]

[b7f0b0ab682766ae39ef8578ecdd9d246bdcf4e3] [ERROR]   LogMessageTest.serviceShouldFailWhenMailHasNoStream:100 
[b7f0b0ab682766ae39ef8578ecdd9d246bdcf4e3] Argument(s) are different! Wanted:
[b7f0b0ab682766ae39ef8578ecdd9d246bdcf4e3] logger.info("Logging mail null");
[b7f0b0ab682766ae39ef8578ecdd9d246bdcf4e3] -> at org.apache.james.transport.mailets.LogMessageTest.serviceShouldFailWhenMailHasNoStream(LogMessageTest.java:100)
[b7f0b0ab682766ae39ef8578ecdd9d246bdcf4e3] Actual invocation has different arguments:
[b7f0b0ab682766ae39ef8578ecdd9d246bdcf4e3] logger.info("Logging mail {}", null);
[b7f0b0ab682766ae39ef8578ecdd9d246bdcf4e3] -> at org.apache.james.transport.mailets.LogMessage.service(LogMessage.java:90)
```

Some test failures....

```

**Comment by dtrebbien** [comment]
Unit test failures fixed and merge conflict resolved.

**Comment by chibenwa** [comment]
I started a new build against your branch in https://github.com/linagora/james-project/pull/1176

PS: We have the habit to prefix commit message with the ticket number:

```
JAMES-2199 My awesome fix about logging
```

I reworded your commits on the `slf4j-related-fixes` of my repository.

**Comment by chibenwa** [comment]
There is still some unit test failures:

```
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR] Failures: 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.exceptionLoggingInFatalModeShouldWorkWithQuiet:225 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error(
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "Message",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     java.lang.Exception
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] );
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.exceptionLoggingInFatalModeShouldWorkWithQuiet(CommonsLoggingAdapterTest.java:225)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error(
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "{}",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "Message",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     java.lang.Exception
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] );
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.fatal(CommonsLoggingAdapter.java:124)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.exceptionLoggingInVerboseModeShouldWorkInDebug:129 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.debug(
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "Message",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     java.lang.Exception
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] );
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.exceptionLoggingInVerboseModeShouldWorkInDebug(CommonsLoggingAdapterTest.java:129)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.debug(
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "{}",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "Message",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     java.lang.Exception
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] );
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.debug(CommonsLoggingAdapter.java:100)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.exceptionLoggingInWarnModeShouldWorkByDefault:175 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.warn(
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "Message",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     java.lang.Exception
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] );
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.exceptionLoggingInWarnModeShouldWorkByDefault(CommonsLoggingAdapterTest.java:175)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.warn(
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "{}",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "Message",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     java.lang.Exception
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] );
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.warn(CommonsLoggingAdapter.java:184)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.logShouldHandleNullException:257 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error("NULL", null);
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.logShouldHandleNullException(CommonsLoggingAdapterTest.java:257)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error("{}", "NULL", null);
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.fatal(CommonsLoggingAdapter.java:124)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.logShouldHandleNullValue:235 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error("NULL");
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.logShouldHandleNullValue(CommonsLoggingAdapterTest.java:235)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error("{}", "NULL");
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.fatal(CommonsLoggingAdapter.java:118)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.logShouldHandleNullValueWithException:246 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error(
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "NULL",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     java.lang.Exception
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] );
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.logShouldHandleNullValueWithException(CommonsLoggingAdapterTest.java:246)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error(
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "{}",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     "NULL",
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c]     java.lang.Exception
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] );
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.fatal(CommonsLoggingAdapter.java:124)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.simpleLoggingInFatalModeShouldWorkWithQuiet:212 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error("Message");
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.simpleLoggingInFatalModeShouldWorkWithQuiet(CommonsLoggingAdapterTest.java:212)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.error("{}", "Message");
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.fatal(CommonsLoggingAdapter.java:118)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.simpleLoggingInVerboseModeShouldWorkInDebug:116 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.debug("Message");
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.simpleLoggingInVerboseModeShouldWorkInDebug(CommonsLoggingAdapterTest.java:116)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.debug("{}", "Message");
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.debug(CommonsLoggingAdapter.java:94)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] [ERROR]   CommonsLoggingAdapterTest.simpleLoggingInWarnModeShouldWorkByDefault:163 
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Argument(s) are different! Wanted:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.warn("Message");
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapterTest.simpleLoggingInWarnModeShouldWorkByDefault(CommonsLoggingAdapterTest.java:163)
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] Actual invocation has different arguments:
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] logger.warn("{}", "Message");
[5c3cf5e5b153e22315bb2b37228e94b594e8a66c] -> at org.apache.james.transport.mailets.jsieve.CommonsLoggingAdapter.warn(CommonsLoggingAdapter.java:178)
```

dtrebbien:
Closing as this has been merged.

PR #79, opened 2017-10-26
Consume by copying to NULL_OUTPUT_STREAM

dtrebbien:
Consuming bytes individually can be inefficient if the input stream is not buffered. IOUtils.copy() buffers the input internally, so there is no need to use a BufferedInputStream.

### DIFF
--- a/protocols/imap/src/main/java/org/apache/james/imap/processor/AppendProcessor.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/processor/AppendProcessor.java
@@ -19,6 +19,8 @@
 
 package org.apache.james.imap.processor;
 
+import static org.apache.commons.io.output.NullOutputStream.NULL_OUTPUT_STREAM;
+
 import java.io.Closeable;
 import java.io.IOException;
 import java.io.InputStream;
@@ -26,6 +28,7 @@ import java.util.Date;
 
 import javax.mail.Flags;
 
+import org.apache.commons.io.IOUtils;
 import org.apache.james.imap.api.ImapCommand;
 import org.apache.james.imap.api.ImapSessionUtils;
 import org.apache.james.imap.api.display.HumanReadableText;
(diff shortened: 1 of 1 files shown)


protocols/imap/src/main/java/org/apache/james/imap/processor/AppendProcessor.java
```diff
@@ -26,6 +26,8 @@
 
 import javax.mail.Flags;
 
+import org.apache.commons.io.IOUtils;
+import static org.apache.commons.io.output.NullOutputStream.NULL_OUTPUT_STREAM;
```
chibenwa: We tend to position static imports above normal  imports

protocols/imap/src/main/java/org/apache/james/imap/processor/AppendProcessor.java
```diff
@@ -103,8 +105,9 @@ protected void doProcess(AppendRequest request, ImapSession session, String tag,
 
     private void consume(InputStream in) {
         try {
-            while (in.read() != -1)
-                ; // NOPMD false positive
+            // IOUtils.copy() buffers the input internally, so there is no need
```
chibenwa: IMO you can remove that comment

protocols/imap/src/main/java/org/apache/james/imap/processor/AppendProcessor.java
dtrebbien: Done

protocols/imap/src/main/java/org/apache/james/imap/processor/AppendProcessor.java
dtrebbien: Done

dtrebbien:
Can this PR be merged?

dtrebbien:
Opened https://issues.apache.org/jira/browse/JAMES-2252

chibenwa:
This issue had been merged. Wouldn't you mind closing it?

Thanks again for your contribution ;-)

Benoit

PR #80, opened 2017-10-30
Misc. improvements

dtrebbien:


server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/BayesianAnalysis.java
```diff
@@ -374,10 +374,10 @@ private String getAddressesString(Collection<MailAddress> addresses) {
         StringBuilder sb = new StringBuilder();
         sb.append('[');
         for (int i = 0; iter.hasNext(); i++) {
-            sb.append(iter.next());
-            if (i + 1 < addresses.size()) {
```
chibenwa:
```
i+1 < addresses.size()
i < addesses.size() - 1 
```

That seems very different from i > 0

IMO if you want to fix it, simply extract the size call in a variable out of the loop

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/delivery/MailDispatcher.java
```diff
@@ -18,11 +18,11 @@
  ****************************************************************/
```
chibenwa: This one is cool.

(Use ArrayList instead of obsolete Vector )

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailetContext.java
```diff
@@ -129,9 +128,10 @@ public void removeAttribute(String key) {
 
     @Override
     public Iterator<String> getAttributeNames() {
-        Vector<String> names = new Vector<>();
-        for (Enumeration<String> e = attributes.keys(); e.hasMoreElements(); ) {
-            names.add(e.nextElement());
+        final List<String> names;
+        synchronized (attributes) {
+            names = new ArrayList<>(attributes.size());
```
chibenwa: What about : 

```
 names = new ArrayList<>(attributes.keySet());
```

mailbox/cassandra/src/test/java/org/apache/james/mailbox/cassandra/mail/CassandraMapperProvider.java
```diff
@@ -111,7 +111,7 @@ public AnnotationMapper createAnnotationMapper() throws MailboxException {
 
     @Override
     public List<Capabilities> getSupportedCapabilities() {
-        return ImmutableList.copyOf(Capabilities.values());
```
chibenwa: copyOf does not necessarilly copy (I know name is misleading).

Also we value readability, and old version is way more redable than the new. Finnaly we prefer relying on guava's collection.

So I disagree with this commit.

server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
```diff
@@ -543,8 +543,8 @@ protected void internalStore(Mail mc) throws IOException, MessagingException {
                 PreparedStatement insertMessage = null;
                 try {
                     String insertMessageSQL = sqlQueries.getSqlString("insertMessageSQL", true);
-                    int number_of_parameters = getNumberOfParameters(insertMessageSQL);
                     insertMessage = conn.prepareStatement(insertMessageSQL);
+                    final int number_of_parameters = insertMessage.getParameterMetaData().getParameterCount();
```
chibenwa: remove the final qualifier?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/BayesianAnalysis.java
dtrebbien: It looks like the purpose of the code is to insert ", " between each `MailAddress` returned by the iterator. In the existing code, ", " is appended after a `MailAddress`, except for after the last `MailAddress`. In the modified code, ", " is appended before the next `MailAddress`, if not the first `MailAddress`. Unless I am mistaken, the existing and modified codes have the same behavior.

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailetContext.java
dtrebbien: Done

server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
dtrebbien: Done

mailbox/cassandra/src/test/java/org/apache/james/mailbox/cassandra/mail/CassandraMapperProvider.java
dtrebbien: You are right that copyOf() does not make a copy if the collection is already an `ImmutableList`, with [some exceptions](https://github.com/google/guava/wiki/ImmutableCollectionsExplained#copyof-is-smarter-than-you-think). However, in this case, the collection is an `ArrayList`, which means that toArray() is called:

https://github.com/google/guava/blob/e3b8c0b6f07c1017e80d86758191d6898517a546/guava/src/com/google/common/collect/ImmutableList.java#L252..L259

As an alternative, what about using Iterables.concat()? I have added a commit, "Avoid creating mutableDomains in AbstractDomainList.getDomains()" to show what I mean.

dtrebbien:
Rebased to resolve the merge conflict.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/BayesianAnalysis.java
chibenwa: Right.

I would prefer a more explicit:

```
sb.append("[")
    .append(Joiner.on(',')
       .join(addresses))
    .append("]");
```

(Guava is available, so let's use it ;-) )

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailetContext.java
```diff
@@ -132,9 +130,9 @@ public void removeAttribute(String key) {
 
     @Override
     public Iterator<String> getAttributeNames() {
-        Vector<String> names = new Vector<>();
-        for (Enumeration<String> e = attributes.keys(); e.hasMoreElements(); ) {
-            names.add(e.nextElement());
+        final List<String> names;
+        synchronized (attributes) {
```
chibenwa: I don't like that **synchronize** at all.

I had a quick look, and **attributes** can be a **ConcurrentHashMap**.

This method become:

```
    @Override
    public Iterator<String> getAttributeNames() {
        return attributes.keySet().iterator();
    }
```

Sources: http://javarevisited.blogspot.com/2011/04/difference-between-concurrenthashmap.html#axzz50RtNZKVw

server/data/data-library/src/main/java/org/apache/james/domainlist/lib/AbstractDomainList.java
```diff
@@ -151,32 +151,33 @@ public boolean containsDomain(String domain) throws DomainListException {
     }
 
     @Override
-    public List<String> getDomains() throws DomainListException {
+    public ImmutableList<String> getDomains() throws DomainListException {
```
chibenwa: How about:

```
    @Override
    public List<String> getDomains() throws DomainListException {
        ImmutableList<String> domainsWithoutIp = Stream.concat(
            getDomainListInternal().stream(),
            detectDomains().stream())
            .collect(Guavate.toImmutableList());
        List<String> detectedIps = detectIps(domainsWithoutIp);
        
        return Stream.concat(
            domainsWithoutIp.stream(),
            detectedIps.stream())
            .peek(domain ->  LOGGER.debug("Handling mail for {}" + domain))
            .collect(Guavate.toImmutableList());
    }
```

?

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailetContext.java
dtrebbien: I traced the introduction of `Hashtable attributes` all the way back to [revision 107158](https://svn.apache.org/viewvc?view=revision&revision=107158) of `james/server/trunk/src/org/apache/james/James.java` on September 13, 2000.

I'm not seeing why it needs to be a synchronized collection.

The only other implementation of `MailetContext` that I found is `FakeMailContext` that is used in test code. `FakeMailContext` uses a `HashMap`.

I think it's safe to use a `HashMap` in `JamesMailetContext` as well.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/BayesianAnalysis.java
dtrebbien: That's a lot cleaner.

I have pushed a commit to use `Joiner`.

dtrebbien:
Created a new issue for to track merging this PR: https://issues.apache.org/jira/browse/JAMES-2251

server/data/data-library/src/main/java/org/apache/james/domainlist/lib/AbstractDomainList.java
dtrebbien: I think that this has the same issue as the current code in that it creates temporary `ImmutableList` instances, which, in order to be concatenated, need to be copied to a larger `ImmutableList` that stores the result of concatenation.

Another source of temporary `ImmutableList` instances is getDomainsIP()/getDomainIP(), which currently returns a new `ImmutableList` for each domain. I have pushed a commit to refactor these private utility methods so that they work with `Stream`s instead. This should be more efficient, I think.

server/data/data-library/src/main/java/org/apache/james/domainlist/lib/AbstractDomainList.java
```diff
@@ -158,21 +159,23 @@ public boolean containsDomain(String domain) throws DomainListException {
         // https://stackoverflow.com/questions/37919648/concatenating-immutablelists
         // A work-around is to use Iterables.concat() until something like
         // https://github.com/google/guava/issues/1029 is implemented.
-        ImmutableList<String> detectedIps = detectIps(Iterables.concat(domains, detectedDomains));
-        Iterable<String> allDomains = Iterables.concat(domains, detectedDomains, detectedIps);
+        Iterable<String> domainsWithoutIp = (detectedDomains.isEmpty() ? domains : Iterables.concat(domains, detectedDomains));
+        ImmutableList<String> detectedIps = detectIps(domainsWithoutIp);
+        ImmutableList<String> allDomains = ImmutableList.copyOf(detectedIps.isEmpty() ? domainsWithoutIp : Iterables.concat(domainsWithoutIp, detectedIps));
```
chibenwa: I would be in favour of always doing the concatenation:
 - Having a condition is misleading
 - Concant with an empty collection is not expensive.

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailetContext.java
```diff
@@ -71,7 +70,7 @@
     /**
      * A hash table of server attributes These are the MailetContext attributes
      */
-    private final Hashtable<String, Object> attributes = new Hashtable<>();
+    private final Map<String, Object> attributes = new HashMap<>();
```
chibenwa: Should be a ConcurrentHashMap....

server/data/data-library/src/main/java/org/apache/james/domainlist/lib/AbstractDomainList.java
dtrebbien: Done

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/JamesMailetContext.java
dtrebbien: Done

chibenwa:
Integration tests started here https://github.com/linagora/james-project/pull/1181

aduprat:
@dtrebbien Thank you very much for your contribution, it has been merged.
:clap:

aduprat:
@dtrebbien Can you close this PR?

PR #81, opened 2017-11-06
Try cassandra 3 support

mbaechler:
follow up of https://github.com/linagora/james-project/pull/1072


mbaechler:
arf, bad target ... closing

PR #83, opened 2017-11-14
randymo-JAMES-2217

randymo:


chibenwa:
This issue have been merged some time ago after playing tests there https://github.com/linagora/james-project/pull/1127

Thanks for this cool contribution.

One last thing... Could you close this issue? I don't have the right to do it, and I would be glad not having to ask Apache about it...

Cheers,

Benoit

PR #90, opened 2017-12-18
Fix for JAMES-2265

apptaro:


apptaro:
Hi, thank you very much!

PR #96, opened 2018-01-09
JAMES-2275 Exception Error Handling Feature

apptaro:


apptaro:
See JAMES-2275 for recommended implementation method. I have implemented 1, 2, 4 and partly 3. I have not implemented the test for mailet error test for 3, because I have no way to check if the exception is attached in MailetProcessorListener.afterMailet in AbstractStateMailetProcessorTest.mailetProcessingShouldNotResultInAnExceptionWhenMailetThrows. I don't think it's good idea to change interface signature just for this. Any advice?

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,126 @@
...
+     * @param mail
+     */
+    public Collection<MailAddress> match(Mail mail) throws MessagingException
+    {
+        Object exceptionValue = mail.getAttribute(Mail.MAILET_ERROR_ATTRIBUTE_NAME);
+
+        if (exceptionValue != null && exceptionClass.isAssignableFrom(exceptionValue.getClass())) {
+            return mail.getRecipients();
+        } else {
+            return null;
```
chibenwa: We would prefer return ImmutableList.of(), which returns an empty list, rather than an error prone null value...

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,126 @@
...
+ * 
+ * <p>Sample configuration:</p>
+ * <pre><code>
+ * &lt;mailet match="HasException=org.apache.james.managesieve.api.ManageSieveException" class=&quot;&lt;any-class&gt;&quot;&gt;
+ * </code></pre>
+ *
+ * @version CVS $Revision$ $Date$
+ * @since 3.0.2
+ **/
+public class HasException extends GenericMatcher
```
chibenwa: From a style perspective, we tend to enforce:

```
public class HasException extends GenericMatcher {
    //...
}
```

We have a checkstyle pull request almost merged on master, maybe it can help with such concerns : https://github.com/linagora/james-project/pull/1206

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
chibenwa: We should return an empty immutable list rather than a null value...

```
return immutableList.of();
```

Would do the trick.

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,126 @@
...
+     * (non-Javadoc)
+     * @see org.apache.mailet.base.GenericMatcher#init()
+     */
+    @SuppressWarnings("unchecked")
+	public void init() throws MessagingException
+    {
+        String condition = getCondition().trim();
+        setExceptionClassName(condition);
+        
+    	try {
```
chibenwa: You have some indent issues. We enforce a 4 space indentation.

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,126 @@
...
+     * @param exceptionClassName The exceptionClassName to set
+     */
+    protected void setExceptionClassName(String exceptionClassName)
+    {
+    	this.exceptionClassName = exceptionClassName;
+    }
+
+    /*
+     * (non-Javadoc)
+     * @see org.apache.mailet.base.GenericMatcher#init()
```
chibenwa: This comment seems not really usefull...

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,126 @@
...
+     */
+    @SuppressWarnings("unchecked")
+	public void init() throws MessagingException
+    {
+        String condition = getCondition().trim();
+        setExceptionClassName(condition);
```
chibenwa: I would rather have the assignement directly here (and remove the setter)

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,112 @@
...
+        matcher.init(FakeMatcherConfig.builder()
+                .matcherName("HasException")
+                .condition("javax.mail.MessagingException")
+                .build());
+
+        assertThat(matcher.match(mockedMail)).isNull();
+    }
+
+    @Test
+    public void matchShouldReturnNullWhenSuperclassOfSpecifiedExceptionHasOccurred() throws MessagingException {
```
chibenwa: According to what I wrote above, it should rather return an empty list.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,112 @@
...
+    public void matchShouldReturnNullWhenNoExceptionHasOccurred() throws MessagingException {
+        matcher.init(FakeMatcherConfig.builder()
+                .matcherName("HasException")
+                .condition("java.lang.Exception")
+                .build());
+
+        assertThat(matcher.match(mockedMail)).isNull();
+    }
+
+}
```
chibenwa: The test you wrote are really cool. I still think we can get it a bit further. I would propose to test:

 - What happens when the Mail.MAILET_ERROR_ATTRIBUTE_NAME is not an exception?

(Expected behaviour: not matching without error)

 - What happens when this matcher is configured with an invalid class name?
 - What happens when this matcher is configured with a class that is not an exception?

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,126 @@
...
+        Object exceptionValue = mail.getAttribute(Mail.MAILET_ERROR_ATTRIBUTE_NAME);
+
+        if (exceptionValue != null && exceptionClass.isAssignableFrom(exceptionValue.getClass())) {
+            return mail.getRecipients();
+        } else {
+            return null;
+        }
+    }
+
+    /**
```
chibenwa: As explained in this PR, I would remove the getter and setter from this class.

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,99 @@
...
+    private Class<? extends Throwable> exceptionClass;
+
+    /**
+     * <p>Answers the recipients of the mail if the specified exception or
+     * the subclasses of it has occured.</p>
+     * 
+     * @param mail
+     */
+    public Collection<MailAddress> match(Mail mail) throws MessagingException
+    {
```
chibenwa: Idem, that `{` should be on it's single line.

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,99 @@
...
+        if (exceptionValue != null && exceptionClass.isAssignableFrom(exceptionValue.getClass())) {
+            return mail.getRecipients();
+        } else {
+            return ImmutableList.of();
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    public void init() throws MessagingException
+    {
```
chibenwa: Idem, that `{` should be on it's single line.

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,99 @@
...
+
+        try {
+            Class<?> exceptionClass = Class.forName(exceptionClassName);
+            if (Throwable.class.isAssignableFrom(exceptionClass)) {
+                this.exceptionClass = (Class<? extends Throwable>) exceptionClass;
+            } else {
+                throw new MessagingException("Specified class name is not a throwable.");
+            }
+        } catch (ClassNotFoundException e) {
+            throw new MessagingException("Specified exception class not found.");
```
chibenwa: I belive we should try to give as much indication as possible to the user...

Here we should rather do:

```
   throw new MessagingException("Specified exception class not found.", e);
```

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,99 @@
...
+     */
+    private Class<? extends Throwable> exceptionClass;
+
+    /**
+     * <p>Answers the recipients of the mail if the specified exception or
+     * the subclasses of it has occured.</p>
+     * 
+     * @param mail
+     */
+    public Collection<MailAddress> match(Mail mail) throws MessagingException
```
chibenwa: maybe we can add an `@Override` here?

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,99 @@
...
+                this.exceptionClass = (Class<? extends Throwable>) exceptionClass;
+            } else {
+                throw new MessagingException("Specified class name is not a throwable.");
+            }
+        } catch (ClassNotFoundException e) {
+            throw new MessagingException("Specified exception class not found.");
+        }
+    }
+    
+    /**
```
chibenwa: I would propose adding an `@Override` and remove this redoundant javadoc.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Test
+    public void initShouldRaiseMessagingExceptionWhenInvalidClassName() throws MessagingException {
+        FakeMatcherConfig mci = FakeMatcherConfig.builder()
+                .matcherName("HasException")
+                .condition("java.lang.InvalidClassName")
+                .build();
+
+        boolean exceptionRaised = false;
```
chibenwa: You can use : 

```
assertThatThrownBy(() -> matcher.init(mci))
    .isInstanceOf(MessagingException.class);
```

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    @Test
+    public void initShouldRaiseMessagingExceptionWhenClassNameIsNotException() throws MessagingException {
+        FakeMatcherConfig mci = FakeMatcherConfig.builder()
+                .matcherName("HasException")
+                .condition("java.lang.String")
+                .build();
+
+        boolean exceptionRaised = false;
```
chibenwa: idem here...

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,157 @@
...
+        MimeMessage mimeMessage = MailUtil.createMimeMessage();
+        mockedMail = MailUtil.createMockMail2Recipients(mimeMessage);
+        matcher = new HasException();
+    }
+
+    @Test
+    public void matchShouldReturnAddressesWhenSpecifiedExceptionHasOccurred() throws MessagingException {
+        mockedMail.setAttribute(Mail.MAILET_ERROR_ATTRIBUTE_NAME, new javax.mail.internet.AddressException());
+
+        FakeMatcherConfig mci = FakeMatcherConfig.builder()
```
chibenwa: I know other tests uses the acronym `mci` but I really do not find it explicit.

I would propose to replace `mci` by `matcherConfig`.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,157 @@
...
+                .matcherName("HasException")
+                .condition("javax.mail.internet.AddressException")
+                .build();
+
+        matcher.init(mci);
+
+        assertThat(matcher.match(mockedMail)).containsAll(mockedMail.getRecipients());
```
chibenwa: You should rather do a `containsExactlyElementsOf` whish ensures order and no extra entries.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,157 @@
...
+        mockedMail.setAttribute(Mail.MAILET_ERROR_ATTRIBUTE_NAME, new javax.mail.internet.AddressException());
+
+        FakeMatcherConfig mci = FakeMatcherConfig.builder()
+                .matcherName("HasException")
+                .condition("javax.mail.MessagingException")
+                .build();
+
+        matcher.init(mci);
+
+        assertThat(matcher.match(mockedMail)).containsAll(mockedMail.getRecipients());
```
chibenwa: idem

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,101 @@
...
+ * Sample configuration:
+ * </p>
+ * 
+ * <pre>
+ * <code>
+ * &lt;mailet match="HasException=org.apache.james.managesieve.api.ManageSieveException" class=&quot;&lt;any-class&gt;&quot;&gt;
+ * </code>
+ * </pre>
+ *
+ * @version CVS $Revision$ $Date$
```
aduprat: Please remove this line

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+import org.apache.mailet.base.test.FakeMail;
+import org.apache.mailet.base.test.FakeMatcherConfig;
+import org.apache.mailet.base.test.MailUtil;
+import org.junit.Before;
+import org.junit.Test;
+
+public class HasExceptionTest {
+
+    private FakeMail mockedMail;
+    private Matcher matcher;
```
aduprat: Please rename _matcher_ to _testee_ it increases the readability of the test methods.

mbaechler:
There's a PR pending about using checkstyle to avoid loosing time on coding convention during PRs, maybe you could import the file https://github.com/mbaechler/james-project/blob/checkstyle-bootstrap/checkstyle.xml in your IDE to check that everything is ok ?

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+        MimeMessage mimeMessage = MailUtil.createMimeMessage();
+        mockedMail = MailUtil.createMockMail2Recipients(mimeMessage);
+        matcher = new HasException();
+    }
+
+    @Test
+    public void matchShouldReturnAddressesWhenSpecifiedExceptionHasOccurred() throws MessagingException {
+        mockedMail.setAttribute(Mail.MAILET_ERROR_ATTRIBUTE_NAME, new javax.mail.internet.AddressException());
+
+        FakeMatcherConfig matcherConfig = FakeMatcherConfig.builder().matcherName("HasException")
```
aduprat: Add some line breaks when calling multiple methods (one call per line).
```
        FakeMatcherConfig matcherConfig = FakeMatcherConfig.builder()
                .matcherName("HasException")
                .condition("javax.mail.internet.AddressException")
                .build();
```

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.mailet.base.GenericMatcher;
+
+import com.google.common.collect.ImmutableList;
+
+/**
+ * <p>
+ * This Matcher determines if the exception specified in the condition or the
+ * subclasses of it has occured during the processing of the mail. If true, all
+ * recipients are returned, else null. This matcher presupposes that the
+ * exception occured has set at the attribute
```
mbaechler: maybe replace "that the exception occured has set at the attribute X in the process" by "that the exception has been captured as a Mail attribute X"

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
mbaechler: I think we don't need that kind of headers

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,101 @@
...
+public class HasException extends GenericMatcher {
+
+    /**
+     * The class of the specified exception class to match
+     */
+    private Class<? extends Throwable> exceptionClass;
+
+    /**
+     * <p>
+     * Answers the recipients of the mail if the specified exception or the
```
mbaechler: I would suggest using "returns" instead of "answers" in this sentence. WDYT ?

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public void init() throws MessagingException {
+        String exceptionClassName = getCondition().trim();
+
+        try {
+            Class<?> exceptionClass = Class.forName(exceptionClassName);
+            if (Throwable.class.isAssignableFrom(exceptionClass)) {
+                this.exceptionClass = (Class<? extends Throwable>) exceptionClass;
```
mbaechler: What about extract this cast into another method so that the @SuppressWarnings is more obvious ?

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasException.java
```diff
@@ -0,0 +1,101 @@
...
+            } else {
+                throw new MessagingException("Specified class name is not a throwable.");
```
mbaechler: I'd rather use Throwable with a capital T, don't you think ?

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+    @Before
+    public void setUp() throws Exception {
+        MimeMessage mimeMessage = MailUtil.createMimeMessage();
+        mockedMail = MailUtil.createMockMail2Recipients(mimeMessage);
+        matcher = new HasException();
+    }
+
+    @Test
+    public void matchShouldReturnAddressesWhenSpecifiedExceptionHasOccurred() throws MessagingException {
+        mockedMail.setAttribute(Mail.MAILET_ERROR_ATTRIBUTE_NAME, new javax.mail.internet.AddressException());
```
mbaechler: is there a reason to use a qualified class name for AddressException ?

server/mailet/mailetcontainer-camel/src/test/java/org/apache/james/mailetcontainer/lib/AbstractStateMailetProcessorTest.java
```diff
@@ -198,6 +200,7 @@ public void afterMailet(Mailet m, String mailName, String state, long processTim
 
         // the source mail should have state error as the exception was thrown
         assertEquals(Mail.ERROR, mail.getState());
+        assertEquals(MessagingException.class, mail.getAttribute(Mail.MAILET_ERROR_ATTRIBUTE_NAME));
```
mbaechler: this assertions does not belong to this test because as the test name says : "matcher Processing Should Not Result In An Exception When Matcher Throws" but not "matcher processing should capture exception as an email attribute".

Would you mind adding a new test for that ?

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+        assertThat(matcher.match(mockedMail)).containsExactlyElementsOf(mockedMail.getRecipients());
+    }
+
+    @Test
+    public void matchShouldReturnAddressesWhenSubclassOfSpecifiedExceptionHasOccurred() throws MessagingException {
+        mockedMail.setAttribute(Mail.MAILET_ERROR_ATTRIBUTE_NAME, new javax.mail.internet.AddressException());
+
+        FakeMatcherConfig matcherConfig = FakeMatcherConfig.builder().matcherName("HasException")
+                .condition("javax.mail.MessagingException").build();
```
aduprat: You can add a similar test with a short class name (i.e. "MessagingException" instead of "javax.mail.MessagingException").

chibenwa: I disagree with that one.

We have an interest keeping the tests as close as possible to real world existing class, so that we **can use tests as actual documentation**. That is the reason why I did no remark to these Fully Qualified class names.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
chibenwa: I did not made that remark, as I valued making it explicit *that the value configured in the matcher* is the *fully qualified* class name, note the simple one.

If you add an import, then it becomes more obscur in the test.

I would say it is one of the rare cases where having a fully qualified name makes sens.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
aduprat: Nothing prevent us from a user to define such configuration in _mailetcontainer.xml_
```
<mailet match="HasException=ManageSieveException" class="...>
```
so it may be a good idea to cover this by a test.

chibenwa: Sorry, I misunderstood you ;-)

Of course.

mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java
apptaro: This is just for a readability. It matches to the code below
```.condition("javax.mail.internet.AddressException")```
Is it okay to keep it? Or should I change it?

Comment (mbaechler) [reply mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java]: no that's ok, I was just asking.

File: mailet/standard/src/test/java/org/apache/james/transport/matchers/HasExceptionTest.java [hunk 29]
```diff
@@ -36,93 +36,126 @@
...
+        FakeMatcherConfig matcherConfig = FakeMatcherConfig.builder()
+                .matcherName("HasException")
+                .condition("java.lang.String")
+                .build();
+
+        assertThatThrownBy(() -> testee.init(matcherConfig)).isInstanceOf(MessagingException.class);
+    }
+    
+    @Test
+    public void initShouldRaiseMessagingExceptionWhenClassNameIsNotFullyQualified() throws MessagingException {
```
aduprat: It seems that this method title is wrong.
What I understand from the test is that a short class name is well recognized, then the _match_ method of the mailer is working.
I propose that you change this method name to _matchShouldReturnAddressesWhenExceptionHasOccurredAndMatchesShortClassName_

BTW thank you for taking care of my comments.

apptaro: Oops, the method name is correct, but the assertion was wrong, and needs to be fixed.

apptaro:
One more thing, I have implemented
```AbstractStateMailetProcessorTest.matcherProcessingShouldCaptureExceptionAsMailAttributeWhenMatcherThrows``` but not ```AbstractStateMailetProcessorTest.mailetProcessingShouldCaptureExceptionAsMailAttributeWhenMailetThrows``` because I would have to change the MailetProcessorListener interface. Do you think it's okay?

chibenwa:
Hi,

Like you, I have problems writing such unit tests. Maybe the solution is to provide some integration tests. A bit like https://github.com/apache/james-project/blob/master/server/mailet/integration-testing/src/test/java/org/apache/james/mailets/MailetErrorsTest.java

What we can do is add new tests where we add the HasException mailet points to a custom repository. We run the throwing mailet, and ensure  that the HasException mailet matched. We should also write similar tests for matchers.

We tend to prefer integration tests in that case because:

 - No asumption is made on components. The tests stay valid even if we don't use Camel anymore.
 - Testing matches the user experience, and documents the end feature.
 - No need of "spying" API like the **MailetProcessorListener**

Of course, I would be glad to help you write such tests. You just need to ask ;-)

chibenwa:
Hi,

Thank you again for your dedication toward the James project. If you need help, want to contribute documentation, new features, tests or bugfix, please don't hesitate to reach us.

Thanks again,

Cheers,

Benoit

apptaro:
@chibenwa Thank you! Closing the PR.

PR #98, opened 2018-01-16
JAMES-2310 Use LineBasedFrameDecoder instead of DelimiterBasedFrameDecoder

chikei:
The resulting decoder should be the same as before commit b4f1ed6 except starttls handling. This is because delimiter decoder will try to use line decoder *if* the delimiter is CRLF and LF AND the decoder is not subclass of delimiter decoder. Therefore, before b4f1ed6 the underlying decoder is actually line decoder, after that commit the decoder became generic delimiter decoder which cause huge performance drop (in my simple benchmark, using line decoder only take 60% time of delimiter decoder to complete the same SMTP transcations).

### DIFF
--- a/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/AllButStartTlsLineChannelHandlerFactory.java
+++ b/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/AllButStartTlsLineChannelHandlerFactory.java
@@ -16,23 +16,22 @@
  * specific language governing permissions and limitations      *
  * under the License.                                           *
  ****************************************************************/
-package org.apache.james.imapserver.netty;
+package org.apache.james.protocols.smtp;
 
 import org.apache.james.protocols.netty.ChannelHandlerFactory;
 import org.jboss.netty.channel.ChannelHandler;
 import org.jboss.netty.channel.ChannelPipeline;
-import org.jboss.netty.handler.codec.frame.Delimiters;
 
-public class SwitchableLineDelimiterBasedFrameDecoderFactory implements ChannelHandlerFactory {
+public class AllButStartTlsLineChannelHandlerFactory implements ChannelHandlerFactory {
 
-    private int maxLineLength;
+    private int maxFrameLength;
 
-    public SwitchableLineDelimiterBasedFrameDecoderFactory(int maxLineLength) {
-        this.maxLineLength = maxLineLength;
+    public AllButStartTlsLineChannelHandlerFactory(int maxFrameLength) {
+        this.maxFrameLength = maxFrameLength;
     }
 
     @Override
     public ChannelHandler create(ChannelPipeline pipeline) {
-        return new SwitchableDelimiterBasedFrameDecoder(maxLineLength, false, Delimiters.lineDelimiter());
+        return new AllButStartTlsLineBasedChannelHandler(pipeline, maxFrameLength, false);
     }
 }
--- a/protocols/netty/src/main/java/org/apache/james/protocols/netty/LineDelimiterBasedChannelHandlerFactory.java
+++ b/protocols/netty/src/main/java/org/apache/james/protocols/netty/LineDelimiterBasedChannelHandlerFactory.java
@@ -20,10 +20,10 @@ package org.apache.james.protocols.netty;
 
 import org.jboss.netty.channel.ChannelHandler;
 import org.jboss.netty.channel.ChannelPipeline;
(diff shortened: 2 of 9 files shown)


chibenwa:
Hi, thank you very much for this interesting contribution.

I agree this is a must-have, as from a functional point of view both are equivalent, but the Line based decoder has good performance improvement.

By looking at the code, we can find additional places we can optimize using that technique:
 - LineDelimiterBasedChannelHandlerFactory, used by LMTP, ManageSieve and POP3
 - SwitchableDelimiterBasedFrameDecoder, used by IMAP
 - And the one you spotted: AllButStartTlsDelimiterChannelHandler, for SMTP

Obviously, we would need this improvement throughout the code base...

Additionally, we use to create tickets for new issues. You can do that here: https://issues.apache.org/jira/browse/JAMES. That being done, we prefix our commit comments with the ticket number.

Might you need any assistance, please don't hesitate, we will help you do that important contribution.

Thanks for the report, it is highly helpful.

Cheers,

Benoit Tellier

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/AllButStartTlsDelimiterChannelHandler.java
```diff
@@ -27,19 +27,19 @@
...
-public class AllButStartTlsDelimiterChannelHandler extends DelimiterBasedFrameDecoder {
+public class AllButStartTlsDelimiterChannelHandler extends LineBasedFrameDecoder {
 
     private static final String STARTTLS = "starttls";
     private final ChannelPipeline pipeline;
 
-    public AllButStartTlsDelimiterChannelHandler(ChannelPipeline pipeline, int maxFrameLength, boolean stripDelimiter, ChannelBuffer[] delimiters) {
-        super(maxFrameLength, stripDelimiter, delimiters);
+    public AllButStartTlsDelimiterChannelHandler(ChannelPipeline pipeline, int maxFrameLength, boolean stripDelimiter) {
+        super(maxFrameLength, stripDelimiter, stripDelimiter);
```
chibenwa: The second boolean parameter correspond to "fail fast" which was false before. We should preserve that behaviour.

I propose we introduce a constant named `FAIL_FAST` with value being `true`.

Thus parent constructor call becomes:

```
super(maxFrameLength, stripDelimiter, !FAIL_FAST);
```

chikei:
As the DelimiterBasedFrameDecoder from netty (which LineDelimiterBasedChannelHandlerFactory directly construct) will automatically use line decoder if we use it directly and with CRLF/LF as delimiter, so only IMAP and SMTP will actually gain some performance from this changes.

Therefore I only make change to SwitchableDelimiterBasedFrameDecoder/AllButStartTlsDelimiterChannelHandler in new commit, if you prefer also make LineDelimiterBasedChannelHandlerFactory explicitly use line frame decoder I'll make additional changes.

chibenwa:
> if you prefer also make LineDelimiterBasedChannelHandlerFactory explicitly use line frame decoder I'll make additional changes.

Yes please do. Maybe it don't make a difference from a performance point of view, but it makes the code more straightforward and explicit. It still is a good move.

chibenwa:
I opened another review here https://github.com/linagora/james-project/pull/1239 in order to play the Continuous Integration against your pull request. 

Might we have any request, we will not hesitate to ask you.

And thanks again for contributing.

Cheers,
Benoit

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/AllButStartTlsLineChannelHandlerFactory.java
```diff
@@ -21,18 +21,17 @@
 import org.apache.james.protocols.netty.ChannelHandlerFactory;
 import org.jboss.netty.channel.ChannelHandler;
 import org.jboss.netty.channel.ChannelPipeline;
-import org.jboss.netty.handler.codec.frame.Delimiters;
 
-public class AllButStartTlsLineDelimiterChannelHandlerFactory implements ChannelHandlerFactory {
+public class AllButStartTlsLineChannelHandlerFactory implements ChannelHandlerFactory {
```
rouazana: LineChannel -> LineBasedChannel?

chibenwa:
Hi,

Thank you again for your dedication toward the James project. If you need help, want to contribute documentation, new features, tests or bugfix, please don't hesitate to reach us.

Thanks again,

Cheers,

Benoit

PR #99, opened 2018-01-29
Mailbox 315 : port two patches from https://github.com/linagora/james-project/pull/1081

mbaechler:


chibenwa:
**clearAllTables** is yet still use. Given that our test strategy relies on creating new keyspace, we might try to remove it fully...

 - It is called in CassandraMapperProvider and I suspect can be safely removed.
 - The Predestroy seems like the mark of previous use in injection, and a quick inspection confirms there is no more such uses.

Can you check?

chibenwa:
We might also want to close the session in an async manner in **CassandraCluster::close** right?

chibenwa:
In CassandraMapperProvider we may want to call close instead.

mbaechler:
I would first merge this one then do what you suggest. The more you put in it, the later you merge.
It's the wrong target repository anyway, we have to close this one and create a new one on linagora.

chibenwa:
I opened https://github.com/linagora/james-project/pull/1273, please close this pull request.

PR #103, opened 2018-02-03
randymo-JAMES-2330

randymo:


server/protocols/protocols-library/src/main/java/org/apache/james/protocols/lib/netty/AbstractConfigurableAsyncServer.java
```diff
@@ -104,7 +104,7 @@
 
     private String secret;
 
-    private Encryption encryption;
+    protected Encryption encryption;
```
rouazana: hi, thank you for your contribution.
Just one question for me: if you ensure to call getEncryption instead of direct access to encryption (what you seem to have done), you don't need to change the visibility of this field, no?

randymo: true, I made it protected, mostly because that is what I did in my testing, but we could make it private, but would have to add a setter in that case as there is not one currently

rouazana: I dont't think a setter is needed. Just the overloading implementation can overload the getEncryption method().

randymo: the buildSSLContext() needs to be able to set the Encryption, without a way to set that field (protected or with setter), the overriding class would need to hold that variable to return it, and you would have to override the getEncryption(). I was thinking that getEncryption() would just need to be overridden in rare cases and the overridden buildSSLContext() would be the usual way of doing it.

chibenwa: Hi @randymo.

I understand your needs, and I understand current code do not allow to satisfy it.

In this discussion, people are trying to bring additional flexibility to classes using inheritance, which sounds like a bad idea. I know this area of the code is a bit "old school". But maybe we can give a try to do this well by composition.

I would propose to add a new **interface** : 

```
public interface EncryptionProvider {
    enum EncryptionMode {
        None,
        StartTls,
        Ssl
    }

    class KeystoreInformation { // Name could be improved?
        private final String keystoreLocation;
        private final char[] secret;
        private final String algorithm;

        // ... Boiler plate
    }

    Encryption get(EncryptionMode encryptionMode, KeystoreInformation keystoreInformation);
}
```
We can then have the 'standard' implementation of it being a lazy provider around the current version of `buildSSLContext`.

Then, **AbstractConfigurableAsyncServer** could use this interface in its `init` method. We would have removed the burdon of managing SSL from this already huge class while enabling you to pass **your** SSL logic.

It also brings in the possibility to store the keystore in a database. Or to generate a fake one for testing/ demonstration/ quick-start purposes.

Would that sound like a good idea?

randymo: Yes, this should work.

chiominto78: Can someone please contact me at Chiomintojames1978@gmail.com

chiominto78: Can someone please contact me at Chiomintojames1978@gmail.com

ieugen:
I moved this code to another PR since the soruce is no longer available and it is crashing the repo scanning https://builds.apache.org/job/ApacheJames/indexing/console . 

Thank you for your contribution @randymo . 
We are working to provide a way to start James with proper SSL.
See https://issues.apache.org/jira/browse/JAMES-3209 and https://github.com/ieugen/james-self-hosting-sandbox/issues/4 .
I hope we get you back contributing on the project.

PR #105, opened 2018-02-24
JAMES-2343 Use buffered file output stream for MimeMessageInputStreamSource

chikei:
Couldn't figure a way to reuse existing DeferredFileOutputStream, so I just copy it and add buffer.

### DIFF
--- a/server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
+++ b/server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
@@ -30,6 +30,18 @@ import org.apache.commons.io.output.ThresholdingOutputStream;
 
 /**
  * An almost copy of {@link DeferredFileOutputStream} with buffered file stream.
+ * <p>
+ * This copy is done because 1. the original class is not written in the way that
+ * we can decorate file output stream. 2. when switched to FileOutputStream the
+ * original class does not buffer file output stream which could cause up to 50%
+ * performance loss compare to buffered stream (see JIRA JAMES-2343 for simple
+ * benchmark).
+ * <p>
+ * The only difference is in the {@link #thresholdReached()} where when updating
+ * {@link #currentOutputStream}, instead of directly assign FileOutputStream to
+ * it, here we wrap the file output stream with BufferedOutputStream.
+ *
+ * @link https://issues.apache.org/jira/browse/JAMES-2343
  */
 public class BufferedDeferredFileOutputStream extends ThresholdingOutputStream {
 
--- a/server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
+++ b/server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
@@ -16,23 +16,22 @@
  */
 package org.apache.james.server.core;
 
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.io.output.ByteArrayOutputStream;
 import org.apache.commons.io.output.DeferredFileOutputStream;
 import org.apache.commons.io.output.ThresholdingOutputStream;
 
-import java.io.*;
-
-
 /**
  * An almost copy of {@link DeferredFileOutputStream} with buffered file stream.
  */
-public class BufferedDeferredFileOutputStream
-    extends ThresholdingOutputStream
-{
-
-    // ----------------------------------------------------------- Data members
-
+public class BufferedDeferredFileOutputStream extends ThresholdingOutputStream {
 
     /**
      * The output stream to which data will be written prior to the theshold
@@ -85,8 +84,7 @@ public class BufferedDeferredFileOutputStream
      * @param threshold  The number of bytes at which to trigger an event.
      * @param outputFile The file to which data is saved beyond the threshold.
      */
-    public BufferedDeferredFileOutputStream(final int threshold, final File outputFile)
-    {
+    public BufferedDeferredFileOutputStream(final int threshold, final File outputFile) {
         this(threshold,  outputFile, null, null, null);
     }
 
@@ -102,8 +100,7 @@ public class BufferedDeferredFileOutputStream
      *
      * @since 1.4
      */
-    public BufferedDeferredFileOutputStream(final int threshold, final String prefix, final String suffix, final File directory)
-    {
+    public BufferedDeferredFileOutputStream(final int threshold, final String prefix, final String suffix, final File directory) {
         this(threshold, null, prefix, suffix, directory);
         if (prefix == null) {
             throw new IllegalArgumentException("Temporary file prefix is missing");
@@ -132,10 +129,6 @@ public class BufferedDeferredFileOutputStream
         this.directory = directory;
     }
 
-
-    // --------------------------------------- ThresholdingOutputStream methods
-
-
     /**
      * Returns the current output stream. This may be memory based or disk
      * based, depending on the current state with respect to the threshold.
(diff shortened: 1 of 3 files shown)


server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,278 @@
...
+ * limitations under the License.
+ */
+package org.apache.james.server.core;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.output.ByteArrayOutputStream;
+import org.apache.commons.io.output.DeferredFileOutputStream;
+import org.apache.commons.io.output.ThresholdingOutputStream;
+
+import java.io.*;
```
chibenwa: We tend to not rely on wildcard imports.

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,278 @@
...
+
+
+/**
+ * An almost copy of {@link DeferredFileOutputStream} with buffered file stream.
```
chibenwa: This is IMO a bit sad...

Couldn't we avoid this copy-paste by wrapping by Bufferred streams where needed?

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,278 @@
...
+ */
+public class BufferedDeferredFileOutputStream
+    extends ThresholdingOutputStream
+{
```
chibenwa: This do not pass the project checkstyle.

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,278 @@
...
+
+    // ----------------------------------------------------------- Data members
```
chibenwa: We should limit the comments to the ones adding valuable information.

chibenwa:
Hi @chikei .

Thanks again for your report.

Do you have any news and input about it?

Best regards,

Benoit Tellier

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
mbaechler: maybe taking a OutputStreamSupplier as a parameter would allow to decouple the concerns ?

chikei:
Oops, I was waiting for other feedback then literally forgot this pr.

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
chikei: My observation was that MimeMessageInputStreamSource uses some method from DeferredFileOutputStream (`isInMemory/getData/getFile`) directly and wrapping with buffered streams would use an extra buffer even the input is so small that it fits buffer. So consider this extra buffer and extra logic to flush buffer, I was thinking not worth it.

Now I think it should just only need buffering during MimeMessageInputStreamSource's ctor and flush it after `IOUtils.copy` should do the work.

chibenwa:
No problem. On my side, I did completely forget that PR too!

I just have one little more question: "Do you still measure the impressive performance improvment you reports with your latest changes?"

Best regards,

Benoit Tellier

chibenwa:
For your information, I started a build for integration testing there: https://github.com/linagora/james-project/pull/1388

chibenwa:
For your information I started a build for integration testing there: https://github.com/linagora/james-project/pull/1388

chikei:
I just find out I made mistake yesterday that a simple flush during `MimeMessageInputStreamSource(String key, InputStream in)` is not enough because the code path originally I concerned was `MimeMessageInputStreamSource(String key)` followed by `MimeMessageInputStreamSource.getWritableOutputStream`.

I'll bring original patch back and add the tests from common-io then run the benchmark again to ensure the the improvement.

chikei:
As why I didn't choose wrapping DeferredFileOutputStream with BufferedOutputStream:
1. the buffer is completely useless for message smaller than threshold (which I believe is not a small percentage ?) yet still using memory.
2. there are more than one place in MimeMessageInputStreamSource using method of DeferredFileOutputStream. If we use BufferedOutputStream, we have to flush buffer every time when the method from DeferredFileOutputStream is used, I feel it's error-prone and messy

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * Constructs an instance of this class which will trigger an event at the
+     * specified threshold, and save data to a temporary file beyond that point.
+     *
+     * @param threshold  The number of bytes at which to trigger an event.
+     * @param prefix Prefix to use for the temporary file.
+     * @param suffix Suffix to use for the temporary file.
+     * @param directory Temporary file directory.
+     *
+     * @since 1.4
```
chibenwa: This `@since` and all followings makes no sens in a James context. They should be removed.

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,251 @@
...
+import java.io.IOException;
+import java.io.OutputStream;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.output.ByteArrayOutputStream;
+import org.apache.commons.io.output.DeferredFileOutputStream;
+import org.apache.commons.io.output.ThresholdingOutputStream;
+
+/**
+ * An almost copy of {@link DeferredFileOutputStream} with buffered file stream.
```
chibenwa: Could we get more information about why this was introduced?

Performance gain? Link to the JIRA issue with your bench as a comment?

server/container/core/src/test/java/org/apache/james/server/core/BufferedDeferredFileOutputStreamTest.java
```diff
@@ -0,0 +1,378 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.util.Arrays;
+
+import org.junit.Test;
+
+public class BufferedDeferredFileOutputStreamTest {
```
chibenwa: In my opinion either we have to clearly state this class is copied from a commons-io one (in a comment?), or we do have to make it's coding style match James standards...

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
chibenwa: Also can you clearly highlight what the differences are? (Which method, etc...) This will make updating that class easier.

chikei:
I use mail queue factory like this [NoopQueueFactory](https://gist.github.com/chikei/f35a5000a41a000a3e34fc1692f7aab7) (I actually use scala and manual wiring when doing benchmarks so I don't have guice code).

Running and timing each test with `smtp-source` from Postfix: `time smtp-source -l <size> -m 10000 -s 100 <james server ip>`

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
chiominto78: I'll tell you how I figured it out if you ask

ieugen:
Thank you for your contribution. 

This seems interesting. Can we revive and maybe merge this ?

chibenwa:
> This seems interesting. Can we revive and maybe merge this ?

So far we missed performance rationals about the benefits of this code change.

I will start a load testing campain, and see what I can come up with.

Cheers,

Benoit

ieugen:
Thanks for your contribution @chikei . 
We are working to merge this in #211 . 
Closing this PR.

PR #106, opened 2018-02-27
correct something about oracle

mostafavtp:
all varchar types converted to varchar2 in oracle tags and a commented section moved to uncommented region.

### DIFF
--- a/server/app/src/main/resources/sqlResources.xml
+++ b/server/app/src/main/resources/sqlResources.xml
@@ -641,10 +641,10 @@
     </sql>
     <sql name="createWhiteListTable" db="oracle">
         CREATE TABLE whitelist (
-            localUser varchar (64) NOT NULL,
-            localHost varchar (255) NOT NULL,
-            remoteUser varchar (64) NOT NULL,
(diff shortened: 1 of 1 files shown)


phansson:
This is correct. On Oracle you would almost always use `VARCHAR2` data type, not `VARCHAR`.  In fact Oracle discourages the use of `VARCHAR` as they threaten to change its definition in the future.

It is all down to history.  When the ANSI standard began to formalize the `VARCHAR` data type, Oracle already had one and unfortunately for Oracle it was slightly different then the ANSI definition.

chibenwa:
We need to have a JIRA ticket about the issue you describe here ( in https://issues.apache.org/jira/browse/JAMES ).

You will also need to prefix the commit message with the ticket number...

Here `JAMES-XXXX correct something about oracle `

Furthermore, maybe we can get a more meaningful commit message. I would propose `JAMES-XXXX Enforce varchar2 usage for WhiteLists on top of Oracle database`

mostafavtp:
Thanks to @phansson  @chibenwa.
I will close the pull request.

PR #112, opened 2018-04-23
JAMES-2384 When there is no body, getting a NullPointerException

provirus:
Fixing the issue by sending an empty array when the body is null

### DIFF
--- a/mailbox/jpa/src/test/java/org/apache/james/mailbox/jpa/mail/model/openjpa/JPAMailboxMessageTest.java
+++ b/mailbox/jpa/src/test/java/org/apache/james/mailbox/jpa/mail/model/openjpa/JPAMailboxMessageTest.java
@@ -1,0 +1,45 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+package org.apache.james.mailbox.jpa.mail.model.openjpa;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.nio.charset.StandardCharsets;
+
+import org.apache.commons.io.IOUtils;
+import org.junit.Test;
+
+public class JPAMailboxMessageTest {
+
+    /**
+     * Even though there should never be a null body, it does happen. See JAMES-2384
+     */
+    @Test
+    public void getFullContentShouldReturnOriginalContentWhenBodyFieldIsNull() throws Exception {
+
+        // Prepare the message
+        byte[] content = "Subject: the null message".getBytes(StandardCharsets.UTF_8);
+        JPAMailboxMessage message = new JPAMailboxMessage(content, null);
+
+        // Get and check
+        assertThat(IOUtils.toByteArray(message.getFullContent())).containsExactly(content);
+
+    }
+
+}
--- a/mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/model/openjpa/JPAMailboxMessage.java
+++ b/mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/model/openjpa/JPAMailboxMessage.java
@@ -39,9 +39,13 @@ import org.apache.james.mailbox.jpa.mail.model.JPAMailbox;
 import org.apache.james.mailbox.store.mail.model.MailboxMessage;
 import org.apache.james.mailbox.store.mail.model.impl.PropertyBuilder;
 
+import com.google.common.annotations.VisibleForTesting;
+
 @Entity(name = "MailboxMessage")
 @Table(name = "JAMES_MAIL")
 public class JPAMailboxMessage extends AbstractJPAMailboxMessage {
+
+    private static final byte[] EMPTY_BODY = new byte[] {};
 
     /** The value for the body field. Lazy loaded */
     /** We use a max length to represent 1gb data. Thats prolly overkill, but who knows */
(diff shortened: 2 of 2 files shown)


provirus:
Hi,

That was a strange bug do to my initial script to send email with msmtp which was like that:
```
msmtp --read-envelope-from $TO << _EOF
from: $FROM
to: $TO
subject: $SUBJECT
$BODY
_EOF
```

It was missing a line return between the subject line and the body and it was considering the `$BODY` to be another header; thus an empty body. 

No I don't have unit tests since those are failing. I wrote a message to "server-dev" 14 hours ago to know if that was normal. Considering that it is HBASE tests that are failing and that I could run the tests in `JpaMessageMapperTest` , I will add a test.

provirus:
I will try tomorrow since I don't know exactly how that `body` gets a null.
I tried creating a `SimpleMailboxMessage` with the body start when there is nothing else and that gives me an empty string; not a null.

I also checked the DB and nothing can be null...

I will reproduce and take a DB dump to see what is actually stored in the DB to reproduce in unit tests.

chibenwa:
I answered you concerning the issue you have with HBase on the mailing list.

Concerning your issue, we need to have a test demonstrating that buggy behaviour. At the very least for non-regression.

I had a quick look at the unit-test level in *MessageMapperTest*. I could not reproduce.

Could you find exactly which `IMAP FETCH` command is executed? What I suspect is more a *FetchType conversion issue*, where James is not using a "big enough" FetchType to retrieve data from the mailbox.

provirus:
FYI, here is the faulty message in the DB. The MAIL_BYTES is not null; just empty and I cannot reproduce in unit tests, but it always reproduces when using for real.

```
INSERT INTO `JAMES_MAIL` (`MAILBOX_ID`, `MAIL_UID`, `MAIL_IS_ANSWERED`, `MAIL_BODY_START_OCTET`, `MAIL_CONTENT_OCTETS_COUNT`, `MAIL_IS_DELETED`, `MAIL_IS_DRAFT`, `MAIL_IS_FLAGGED`, `MAIL_DATE`, `MAIL_MIME_TYPE`, `MAIL_MODSEQ`, `MAIL_IS_RECENT`, `MAIL_IS_SEEN`, `MAIL_MIME_SUBTYPE`, `MAIL_TEXTUAL_LINE_COUNT`, `MAIL_BYTES`, `HEADER_BYTES`) VALUES (1,1,'\0',496,496,'\0','\0','\0','2018-04-25 20:46:50','text',3,'\0','','plain',0,'','Return-Path: <a2@foilen-lab.com>\r\nMessage-ID: <1465402259.0.1524689210620@localhost>\r\nMIME-Version: 1.0\r\nX-UserIsAuth: true\r\nDelivered-To: a1@foilen-lab.com\r\nReceived: from 192.168.0.114 (EHLO localhost) ([192.168.0.114])\r\n          by james_server (JAMES SMTP Server ) with ESMTPA ID -164520987\r\n          for <a1@foilen-lab.com>;\r\n          Wed, 25 Apr 2018 20:46:50 +0000 (UTC)\r\nDate: Wed, 25 Apr 2018 16:46:50 -0400\r\nfrom: a2@foilen-lab.com\r\nto: a1@foilen-lab.com\r\nsubject: test null body\r\n\r\n');
```


> Could you find exactly which IMAP FETCH command is executed?

Is there a way to find it from the James log since I cannot find a way to know the commands sent by Thunderbird and Geary?

chibenwa:
>> Could you find exactly which IMAP FETCH command is executed?

> Is there a way to find it from the James log since I cannot find a way to know the commands sent by Thunderbird and Geary?

No that is OK, I just wanted to know a way to reproduce.

For the record, James logs the context of IMAP requests unising MDC context to log command details (MDC is not logged on the log file, but can be retrieved for instance with tools like ElasticSearch + Kibana).

provirus:
I want to test that out in a unit test:

- saving that message in the DB (e.g H2 database)
- loading it back

I searched for a unit test using `JPAMessageManager` to see how you set that up, but couldn't find one. 
I checked also for a `JPAMessageManagerTest` or `StoreMessageManagerTest`, but they don't exist.

How are you testing the JPA implementation?

thanks

chibenwa:
We are testing it at the mapper level. Have a look at `JPAMessageMapperTest`

Alternatively, you can write higher level tests through the `JPAMailboxManagerTest`

provirus:
I looked around, I tried adding a test in JpaMessageMapperTest:

- messageMapper.add
- tried to clear the cache with `JPA_TEST_CLUSTER.getEntityManagerFactory().getCache().evictAll()` to make sure it loads the object from the DB
- messageMapper.findInMailbox

and the body is never null.

In summary:
- I fixed the issue in 10 minutes
- I am trying to reproduce in unit tests for 4-5 hours without success

Maybe it is due to the MariaDB database ; not H2?

What's the next step since that is a very weird issue?

provirus:
I could always create a unit test that just creates the JPAMailboxMessage and sets the values on the fields directly (setting the `body` as null) and then call `getBodyContent()`. Since we do not know how it gets loaded in that state, but it does and at least, it will be clear that we need to check for null there even thought it makes no sense since it is non-nullable.

chibenwa:
That would be a good idea.

I gave several goes to try to reproduce this, with mitigated success. Artificially injecting the null value would be better than nothing IMO.

Could you contribute such a test?

provirus:
I updated this pull request with that test case.
Enjoy :)

mailbox/jpa/src/test/java/org/apache/james/mailbox/jpa/mail/model/openjpa/JPAMailboxMessageTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.nio.charset.StandardCharsets;
+
+import org.apache.commons.io.IOUtils;
+import org.junit.Test;
+
+public class JPAMailboxMessageTest {
+
+    /**
+     * Even thought there should never be a null body, it does happen. See JAMES-2384
```
rouazana: s/thought/though

provirus:
done

chibenwa:
Hi,

Thank you again for your dedication toward the James project. If you need help, want to [contribute](https://issues.apache.org/jira/issues/?jql=project%20%3D%20JAMES%20AND%20resolution%20%3D%20Unresolved%20AND%20labels%20%3D%20newbie%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC) [documentation](https://issues.apache.org/jira/issues/?jql=project%20%3D%20JAMES%20AND%20resolution%20%3D%20Unresolved%20AND%20labels%20%3D%20documentation%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC), new [features](https://issues.apache.org/jira/issues/?jql=project%20%3D%20JAMES%20AND%20resolution%20%3D%20Unresolved%20AND%20labels%20%3D%20feature%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC), tests or [bugfix](https://issues.apache.org/jira/issues/?jql=project%20%3D%20JAMES%20AND%20resolution%20%3D%20Unresolved%20AND%20labels%20%3D%20easyfix%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC), please don't hesitate to reach us.

Thanks again,

Cheers,

Benoit

PR #114, opened 2018-05-13
JAMES-2389 fix bug: when sending email with different case, making new inbox for destination user.

mostafavtp:
[JAMES-2389](https://issues.apache.org/jira/browse/JAMES-2389)
the bug:
When sending an email to a user with no lower case user name, a new Inbox is created for that user.
For example assume:
I have a user with "test1@t.t1" name for example. When I send mail to "Test1@t.t1", James accepts it but creates a new mail box with "Test1@t.t1" name in database(JAMES_MAILBOX table in USER_NAME field) and "test1@t.t1" user can't see the email.
It seems user existence check is case insensitive but mailbox creation is case sensitive.
James should not create a new inbox in this situation.

fix: In all implementation of `getUser ` method of `org.apache.james.user.api.UsersRepository` interface, I converted the result to lower case to avoid generating new inbox.

### DIFF
--- a/server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyUsersLDAPRepository.java
+++ b/server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyUsersLDAPRepository.java
@@ -770,7 +770,7 @@ public class ReadOnlyUsersLDAPRepository implements UsersRepository, Configurabl
         if (supportVirtualHosting()) {
             return mailAddress.asString();
         } else {
-            return mailAddress.getLocalPart();
+            return mailAddress.getLocalPart().toLowerCase();
         }
     }
 
--- a/server/data/data-library/src/main/java/org/apache/james/user/lib/AbstractUsersRepository.java
+++ b/server/data/data-library/src/main/java/org/apache/james/user/lib/AbstractUsersRepository.java
@@ -124,7 +124,7 @@ public abstract class AbstractUsersRepository implements UsersRepository, Config
         if (supportVirtualHosting()) {
             return mailAddress.asString();
         } else {
-            return mailAddress.getLocalPart();
+            return mailAddress.getLocalPart().toLowerCase();
         }
     }
 


server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyUsersLDAPRepository.java
```diff
@@ -770,7 +770,7 @@ public String getUser(MailAddress mailAddress) throws UsersRepositoryException {
         if (supportVirtualHosting()) {
             return mailAddress.asString();
         } else {
-            return mailAddress.getLocalPart();
+            return mailAddress.getLocalPart().toLowerCase();
```
chibenwa: `toLowerCase()` needs to take a Locale as parameter (as the locale might impact the operation.

Here you might want: `toLowerCase(Locale.US)`

server/data/data-library/src/main/java/org/apache/james/user/lib/AbstractUsersRepository.java
```diff
@@ -124,7 +124,7 @@ public String getUser(MailAddress mailAddress) throws UsersRepositoryException {
         if (supportVirtualHosting()) {
             return mailAddress.asString();
         } else {
-            return mailAddress.getLocalPart();
+            return mailAddress.getLocalPart().toLowerCase();
```
chibenwa: We also needa locale here...

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyUsersLDAPRepository.java
chibenwa: Additionally, we need tests about these changes.

Have a look at `AbstractUsersRepositoryTest`

chibenwa:
Hi @mostafavtp .

One alternative solution is to position the `RecipientToLowerCase` mailet just before the `LocalDelivery` mailet.

It will solve your use case while still keeping flexibility to users for which "Test@domain" might be a different user than "test@domain".

The sequence in `mailetcontainer.xml` becomes:

```
<mailet matcher="RecipientIsLocal" class="RecipientToLowerCase"/>
<mailet matcher="RecipientIsLocal" class="LocalDelivery"/>
```

Would that solve your problem?

If yes, we might consider adding this to default configurations:

```
... /server/app/src/main/resources/mailetcontainer.xml
... /server/container/guice/jpa-smtp/sample-configuration/mailetcontainer.xml
... /dockerfiles/run/guice/cassandra/destination/conf/mailetcontainer.xml
... /dockerfiles/run/guice/cassandra-ldap/destination/conf/mailetcontainer.xml
```

mostafavtp:
Thank you @chibenwa .
I added
`<mailet match="RecipientIsLocal" class="RecipientToLowerCase"/>`  (match not matcher)
and it works correctly. Thank you. 👍 

But one more thing:
I think this config must be hard coded because we can not create user Test@domain and test@domain. When we add user Test@domail it creates test@domail in database and adding test@domail user generates error. And also user existence check is case insensitive.
Finally I think using lower case must be hard coded or these behaviors must be changed also.

chibenwa:
Not necessarily...

While being more simple, the restriction (to avoid case variation) still stands without address normalization (to lower case).

If I want to use `Benoit.Tellier@domain.com`, forbidding the use of `Benoit.TELLIER@domain.com` avoids confusions, but I can not see reasons to restrict it to `benoit.tellier@domain.com`.

(Though I do agree case normalisation works badly... We would need a mailet allowing to normalize to the right case variation in the example I gave above.)

But that is just my opinion here. Maybe more feedback would be constructive...

chibenwa:
https://issues.apache.org/jira/browse/JAMES-2949 have been solving this.

We did strongly type usernames and then enforced a lower case usage within UsersRepository, eventually resolving this issue.

Thank you very much for this contribution.

ieugen:
Thank you for your contribution .

It seems the attached JIRA issues have been fixed. 
I'll close this issue since the PR is not relevant.
Please reopen if you think otherwise.

PR #115, opened 2018-05-14
PROTOCOLS-118 Continuation Response can't handle Login

Flyffies:
It seems James does not handle the repeated continuation request correctly.

This is the first part of the fix.
NGINX for example is expecting some text after the initial continuation Tag ("+").
Changed it so an "+ Ok" is send.
It seems to be specified by the RFC 3501, though it is a very vague description.

### DIFF
--- a/protocols/imap/src/main/java/org/apache/james/imap/encode/AuthenticateResponseEncoder.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/encode/AuthenticateResponseEncoder.java
@@ -38,7 +38,8 @@ public class AuthenticateResponseEncoder  extends AbstractChainedImapEncoder {
 
     @Override
     protected void doEncode(ImapMessage acceptableMessage, ImapResponseComposer composer, ImapSession session) throws IOException {
-        composer.continuationResponse("");
+        // Some clients dont recognise the continuation, unless it is followed by text
+        composer.continuationResponse("Ok");
     }
 
 }
--- a/protocols/imap/src/main/java/org/apache/james/imap/encode/AuthenticateResponseEncoder.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/encode/AuthenticateResponseEncoder.java
@@ -38,7 +38,7 @@ public class AuthenticateResponseEncoder  extends AbstractChainedImapEncoder {
 
     @Override
     protected void doEncode(ImapMessage acceptableMessage, ImapResponseComposer composer, ImapSession session) throws IOException {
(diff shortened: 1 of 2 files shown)


Flyffies:
Ohh, for got the link, here it is:[PROTOCOLS-118](https://issues.apache.org/jira/projects/PROTOCOLS/issues/PROTOCOLS-118?filter=allopenissues)

chibenwa:
Hi @Flyffies ,

Thank you for the contribution, it is an interesting one!

I had a quick look, and this seems like a good move. I will be launching some integration tests on this.

I just have two little comments:

 - We tend to prefix commit messages with the JIRA tickets. Here:

```
PROTOCOLS-118 Fixed continuation request not getting recognised by some clients
```

 - We tend to avoid comments in the code. These comments can be moved as part of the commit message if needed.

Cheers,

Benoit Tellier

Flyffies:
Hi, thanks I didn't saw the comment policy anywhere, so its noted for the future.

As to the second part of my ticket, I Tcpdumped the actual NGINX to James communication with my change applied, here it is:

N -> J: 1 login {8}
J -> N: + Ok
N -> J: testUser {8}

Then James doen not reqest more content. 
While attaching to it in Debug I saw the field `retry` in `AbstractNettyImapRequestLineReader` chenges to true as `AbstractNettyImapRequestLineReader` is recreated.

Thats prevents any other continuation request from being send.

But... When I changed the value to false during debug, nothing got written to nginx and James tried to send multiple continueation request that time... Its really strange...

Flyffies:
Closed it, because I switched away from james. To many things just dont work

rouazana:
Sorry about that. Can you share a little your experience so that others can benefit? Don't hesitate to open tickets on JIRA, some contributors could be interested fixing them.

Flyffies:
Well, the nginx reverse proxy for mail doesn't work, the Mail Clients complain, because james apperently doesn't do some things like its specified in the RFCs.

It caused me a lot of headaches. I found a all in one docker, easy to setup, with postfix and dovecot.
Way less pain.

I opened a Ticket on Jira, but nothing happend there, so I figuered I might as well close it.

rouazana:
Thank you for the feedback. Even if we have not the time to work on it, we are very concerned by the IMAP support, and for the moment pretty happy with it (at least with some MUA). But I agree there are some improvement to do on the authentication and continuation code.
Would you mind to reopen the https://issues.apache.org/jira/browse/PROTOCOLS-118 issue please? Even if you'll not fix it, we'll probably do it one day.

PR #119, opened 2018-06-09
JAMES-2149 Create domain mappings via webadmin

nstdio:


server/protocols/webadmin/webadmin-data/pom.xml
```diff
@@ -97,6 +100,13 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
+            <version>${junit-jupiter-engine.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>${junit-jupiter-engine.version}</version>
```
chibenwa: Include this dependency in james-project's pom.xml (we tend to avoid versions in children's POM to make librairy upgrades easier.)

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -0,0 +1,155 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.webadmin.routes;
+
+import com.github.fge.lambdas.consumers.ThrowingBiConsumer;
+import io.swagger.annotations.*;
```
chibenwa: We do avoid wildcard imports. Prefer having all the headers required here.

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -0,0 +1,155 @@
...
+                        e -> e.getValue()
+                                .select(Mapping.Type.Domain)
+                                .asStream()
+                                .map(Mapping::asString)
+                                .map(Mapping.Type.Domain::withoutPrefix)
+                                .collect(toImmutableList())
+                ));
+    }
+
+    private void doMapping(final Request request, final ThrowingBiConsumer<MappingSource, Domain> op) {
```
chibenwa: We avoid method parameters being final. Could you remove these final qualifiers?

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -0,0 +1,155 @@
...
+    @ApiResponses(value = {
+            @ApiResponse(code = HttpStatus.OK_200, message = "Domain mappings.", responseContainer = "Map"),
+            @ApiResponse(code = HttpStatus.INTERNAL_SERVER_ERROR_500,
+                    message = "Internal server error - Something went bad on the server side.")
+    })
+    public Map<String, List<String>> get(Request request, Response response) throws RecipientRewriteTableException {
+        return recipientRewriteTable.getAllMappings()
+                .entrySet()
+                .stream()
+                .collect(toImmutableMap(e -> e.getKey().getFixedDomain(),
```
chibenwa: `e` is a bit too short as a name. I'd propose to name it `mappingsEntry`

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -0,0 +1,155 @@
...
+import spark.Service;
+
+import javax.inject.Inject;
+import javax.ws.rs.*;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+
+import static com.github.steveash.guavate.Guavate.toImmutableList;
+import static com.github.steveash.guavate.Guavate.toImmutableMap;
```
chibenwa: We generally put static imports first.

Also, we tend to not import statically Guavate methods.

Wouldn't you mind doing this change?

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
chibenwa: What if the mappings contains no `DomainMapping`? Wouldn't it be included in the response with an empty list?

Wouldn't it make more sens to filter them out? (map the value with the select(Mapping.Type.Domain) then filter out empty mappings?)

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @GET
+    @Path(DOMAIN_MAPPINGS)
+    @ApiOperation(value = "Lists all domain mappings.")
+    @ApiResponses(value = {
+            @ApiResponse(code = HttpStatus.OK_200, message = "Domain mappings.", responseContainer = "Map"),
+            @ApiResponse(code = HttpStatus.INTERNAL_SERVER_ERROR_500,
+                    message = "Internal server error - Something went bad on the server side.")
+    })
+    public Map<String, List<String>> get(Request request, Response response) throws RecipientRewriteTableException {
```
chibenwa: It smells like a MultiList. Guavate has some collectors for this. You may need to load Guava's Jackson module in JsonTransformer also...

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    private void doMapping(final Request request, final ThrowingBiConsumer<MappingSource, Domain> op) {
+        MappingSource fromDomain = createDomainOrThrow()
+                .andThen(MappingSource::fromDomain)
+                .apply(request.params(FROM_DOMAIN));
+
+        Domain toDomain = createDomainOrThrow().apply(request.body());
+
+        op.accept(fromDomain, toDomain);
```
chibenwa: `op` is a bit of a short name. I'd propose `mappingOperation`

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,217 @@
...
+    private RecipientRewriteTable recipientRewriteTable;
+    private WebAdminServer webAdminServer;
+
+    @SuppressWarnings("unused")
+    static Stream<Arguments> invalidInputs() {
+        return Stream.of(
+                Arguments.of("domain.com", ""),
+                // Why this params should pass the test ?
+                // "domain.com, '    '",
+                // "domain.com, '    \n\t\r'",
```
chibenwa: That is a good question ;-)

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,217 @@
...
+    void tearDown() {
+        webAdminServer.destroy();
+    }
+
+    @Nested
+    class NormalBehaviour {
+
+        @Test
+        void addDomainMappingShouldRespondWithNoContent() {
+            // @formatter:off
```
chibenwa: Here the syntax we do apply for RestAssured is : 

```
given()
   .body(...)
.when()
   .put()
.then()
   .statusCode(...);
```

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        @Test
+        void getDomainMappings() throws RecipientRewriteTableException {
+            Domain expectedDomain = Domain.of("abc.com");
+            MappingSource mappingSource = MappingSource.fromDomain(expectedDomain);
+            ImmutableList<String> expectedAliases = ImmutableList.of("to_1.com", "to_2.com", "to_3.com");
+
+            for (String alias : expectedAliases) {
+                recipientRewriteTable.addAliasDomainMapping(mappingSource, Domain.of(alias));
```
chibenwa: We tend to avoid logic (even the simplest one) in our tests, in order to make them easier to read.

Here we would prefer:

```
String alias1 = "to_1.com";
String alias2 = "to_2.com";
String alias3 = "to_3.com";
recipientRewriteTable.addAliasDomainMapping(mappingSource, Domain.of(alias1));
recipientRewriteTable.addAliasDomainMapping(mappingSource, Domain.of(alias2));
recipientRewriteTable.addAliasDomainMapping(mappingSource, Domain.of(alias3));

//...

assertThat(map)
              .containsOnly(entry(expectedDomain.name(), ImmutableList.of(alias1, alias2, alias3)));
```

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,217 @@
...
+                        .jsonPath()
+                        .getMap(".");
+            // @formatter:on
+
+            assertThat(map)
+                    .containsOnly(entry(expectedDomain.name(), expectedAliases));
+        }
+
+        @Test
+        void getDomainMappingsShouldBeEmpty() {
```
chibenwa: Maybe `getDomainMappingsShouldBeEmptyByDefault` is a better name?

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,217 @@
...
+        @Test
+        void deleteDomainMappingShouldRespondWithNoContent() {
+            // @formatter:off
+            given(with().body("to.com"))
+                        .delete("from.com")
+                    .then()
+                        .statusCode(HttpStatus.NO_CONTENT_204)
+                        .body(isEmptyString());
+            // @formatter:on
+        }
```
chibenwa: Could you add these test cases?

Delete should remove domain mappings:
  - Add a domain mapping
  - Remove it
  - Ensure the collection is empty

get domain mappings should group mappings by source domain
 - The same than `getDomainMappings` but with different source domains.

getDomains should filter out mappings without domain mappings
 - Add for instance an address mapping
 - Ensure that the answer is still `{}`

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,217 @@
...
+        void getDomainMappings() throws RecipientRewriteTableException {
+            Domain expectedDomain = Domain.of("abc.com");
+            MappingSource mappingSource = MappingSource.fromDomain(expectedDomain);
+            ImmutableList<String> expectedAliases = ImmutableList.of("to_1.com", "to_2.com", "to_3.com");
+
+            for (String alias : expectedAliases) {
+                recipientRewriteTable.addAliasDomainMapping(mappingSource, Domain.of(alias));
+            }
+
+            // @formatter:off
```
chibenwa: These comments are a bit annoying. Wouldn't you mind removing them?

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,217 @@
...
+            when()
+                .delete("")
+            .then()
+                .statusCode(HttpStatus.NOT_FOUND_404);
+            // @formatter:on
+        }
+
+        @ParameterizedTest
+        @MethodSource("org.apache.james.webadmin.routes.DomainMappingsRoutesTest#invalidInputs")
+        void addDomainMappingWithInvalidDomainInPath(String fromDomain, String toDomain) {
```
chibenwa: I think that parametrized test is a great idea and thank you for sharing it with us, that is really instructive.

However, you should know that we recently encountered issues with junit 5.2.0, namely:
 - tests not played, missing dependencies, etc...

As such I'd like to avoid using it until it get more mature and that the maven integration improves.

Furthermore, I'd like to avoid diverging junit version between projects. 

Here, we got 2 inputs... I'd propose you to flatten them and write one test for each invalid case, without parametrized tests (for the reasons I listed above).

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @Inject
+    public DomainMappingsRoutes(final RecipientRewriteTable recipientRewriteTable, final JsonTransformer jsonTransformer) {
+        this.recipientRewriteTable = recipientRewriteTable;
+        this.jsonTransformer = jsonTransformer;
+    }
+
+    @Override
+    public void define(final Service service) {
+        service.get(DOMAIN_MAPPINGS, this::get, jsonTransformer);
```
chibenwa: I just thought about it, but `GET /domainMappins` can be potentially an expensive computation as it iterates on the all content of RecipientRewrite tables to find domain mappings.

Maybe we should allow to get domain mappings for a single domain. Something like: 

```
GET /domainMappings/fromdomain.com

["domainalias1.com", "domainalias2.org"]
```

Wouldn't you mind adding it?

nstdio:
Hi @chibenwa, @mbaechler ,
Thanks for comments (👍) , will address all of them soon.
Regarding to `fixup commits`, currently I not familiar with that but I promise I will :)

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
nstdio: I'll add this endpoint. Good idea.

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
nstdio: If you mean `MultiMap` then, yes, of course it is!

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
nstdio: Recently I've created the [ticket](https://issues.apache.org/jira/browse/JAMES-2419) describing the potential issue.

@chibenwa @mbaechler Any thoughts on this?

chibenwa:
Sure:

The idea is to present the fix you do in separate commits to ease the work of reviewers.

You can generate them with this command:

```
git commit --fixup=SHA1
```

Where SHA-1 is the id (obtained via git log for instance) of the commit you want to fix, here 18af08e .

That do allow to perform the review in an incremental way.

Cheers,

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -114,28 +126,56 @@ public HaltException removeDomainMapping(Request request, Response response) {
...
-                .collect(toImmutableMap(e -> e.getKey().getFixedDomain(),
-                        e -> e.getValue()
-                                .select(Mapping.Type.Domain)
-                                .asStream()
-                                .map(Mapping::asString)
-                                .map(Mapping.Type.Domain::withoutPrefix)
-                                .collect(toImmutableList())
+                .filter(mappingsEntry -> !mappingsEntry.getValue().isEmpty())
+                .filter(mappingsEntry -> mappingsEntry.getValue().contains(Mapping.Type.Domain))
+                .collect(Guavate.toImmutableMap(
```
chibenwa: You can use Guavate:: toImmutableMultimap, make the return type of this method a multimap.

If you rebase on master, you will have Guava objects support (we added it very recently).

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -114,28 +126,56 @@ public HaltException removeDomainMapping(Request request, Response response) {
...
                 .stream()
-                .collect(toImmutableMap(e -> e.getKey().getFixedDomain(),
-                        e -> e.getValue()
-                                .select(Mapping.Type.Domain)
-                                .asStream()
-                                .map(Mapping::asString)
-                                .map(Mapping.Type.Domain::withoutPrefix)
-                                .collect(toImmutableList())
+                .filter(mappingsEntry -> !mappingsEntry.getValue().isEmpty())
+                .filter(mappingsEntry -> mappingsEntry.getValue().contains(Mapping.Type.Domain))
```
chibenwa: The problem here is that we do not remove entries of the mappings that are not a domainMapping.

What I would suggest you to do it:

```
.map(mappingEntry -> Pair.of(mappingEntry.getKey(), mappingEntry.getValue().select(Mapping.Type.Domain))
.filter(mappingsEntry -> !mappingsEntry.getValue().isEmpty())
```

(We ensure to have only domain mappings, then we check it's not empty...).

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -114,28 +126,56 @@ public HaltException removeDomainMapping(Request request, Response response) {
...
+            @ApiImplicitParam(required = true, dataType = "string", name = FROM_DOMAIN, paramType = "path")
+    })
+    @ApiResponses(value = {
+            @ApiResponse(code = HttpStatus.OK_200, message = "Domain mappings.", responseContainer = "List"),
+            @ApiResponse(code = HttpStatus.NOT_FOUND_404, message = "Not existing mappings."),
+            @ApiResponse(code = HttpStatus.INTERNAL_SERVER_ERROR_500,
+                    message = "Internal server error - Something went bad on the server side.")
+    })
+    public List<String> getMapping(Request request, Response response) throws RecipientRewriteTableException {
+        final MappingSource mappingSource = mappingSourceFrom(request);
```
chibenwa: We do not use final variable :-)

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
```diff
@@ -114,28 +126,56 @@ public HaltException removeDomainMapping(Request request, Response response) {
...
+
+        return Optional.ofNullable(recipientRewriteTable.getAllMappings().get(mappingSource))
```
chibenwa: You can rely  on `RecipientRewriteTable::getUserDomainMappings`, litterally:

```
recipientRewriteTable.getUserDomainMappings(mappingSource)
    .select(Mapping.Type.Domain)
```

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,387 @@
...
+                .body("to.com")
+            .when()
+                .put("from.com")
+            .then()
+                .statusCode(HttpStatus.NO_CONTENT_204)
+                .body(isEmptyString());
+        }
+
+        @Test
+        void getDomainMappings() throws RecipientRewriteTableException {
```
chibenwa: We tend to use a standard scheme for test names, to make them more readable.

The scheme is `[methodName]Should[Action]When[Condition]` The when part being optional.

Here that would be `getDomainMappingsShouldReturnAllDomainMappings`

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,387 @@
...
+                        .body()
+                        .jsonPath()
+                        .getMap(".");
+
+            assertThat(map)
+                    .containsOnly(entry(expectedDomain.name(), ImmutableList.of(alias1, alias2, alias3)));
+        }
+
+        @Test
+        void getDomainMappingsEmptyMappingsAreFilteredOut() throws RecipientRewriteTableException {
```
chibenwa: Here I would suggest `getDomainMappingShouldFilterOutEmptyMappings`

BTW semantic  change from passive form into active form is interesting, as the test name implies we know where the action is taken ;-).

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,387 @@
...
+            .then()
+                .contentType(ContentType.JSON)
+                .statusCode(HttpStatus.NOT_FOUND_404)
+                .body("type", is(ErrorResponder.ErrorType.NOT_FOUND.getType()))
+                .body("statusCode", is(HttpStatus.NOT_FOUND_404))
+                .body("message", is("Cannot find mappings for " + domain));
+        }
+
+        @Test
+        void getSpecificDomainMappingNotDomainsMappingsShouldBeFilteredOut() throws RecipientRewriteTableException {
```
chibenwa: Idem here `getSpecificDomainMappingShouldFilterOutNonDomainMappings`

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,387 @@
...
+                        .statusCode(HttpStatus.OK_200)
+                    .extract()
+                        .jsonPath()
+                        .getList(".");
+
+            assertThat(body).containsOnly(aliasDomain);
+        }
+
+        @Test
+        void getSpecificDomainMappingShouldResponseWithOK() throws RecipientRewriteTableException {
```
chibenwa: I'd say here you test more than 'OK' response. `getSpecificDomainMappingShouldReturnDomainMappings`

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,387 @@
...
+            String domain = "abc.com";
+            String aliasDomain = "a.com";
+            MappingSource mapping = MappingSource.fromDomain(Domain.of(domain));
+            Map<MappingSource, Mappings> mappings = ImmutableMap.of(mapping, MappingsImpl.fromRawString("domain:" + aliasDomain));
```
chibenwa: Can't you use MappingsImpl.fromMappings(Stream.of(Mapping.domain(aliasDomain))) ?

We might want to add in MappingsImpl a convenience `MappingsImpl fromMappings(Mapping... mappings) ` if you find it easier...

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
```diff
@@ -0,0 +1,387 @@
...
+
+            when(recipientRewriteTable.getAllMappings()).thenReturn(mappings);
```
chibenwa: Why do we need to spy here?

Can't we just  add the domainMapping using `recipientRewriteTable.addMapping(MappingSource.fromDomain(domain), Mapping.domain(aliasDomain))` ?

(Or even better: rely on the PUT operation of this endoint ^^)

chibenwa:
Thanks for the great enhancement for this pull request. 

I still have some minor comment that I believe you would be able to fix quickly.

Thanks again for your involvement on that issue.

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
mbaechler: It probably deserves some more rationals : 
* we don't allow variable reallocation in our code because it makes reading harder
* in java 8, you don't really need variable to be final to use them in lambda (what previously required an anonymous class + final variable)
* finally, final becomes useless because of the two previous point

nstdio: I should change the IDE preferences on that 👍

chibenwa:
Hi @nstdio 

Here is the build output for your PR:

```
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [INFO] There are 26 errors reported by Checkstyle 8.5 with checkstyle.xml ruleset.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[26] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'com.github.fge.lambdas.consumers.ThrowingBiConsumer'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[27] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'com.github.steveash.guavate.Guavate'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[28] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'io.swagger.annotations.Api'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[29] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'io.swagger.annotations.ApiImplicitParam'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[30] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'io.swagger.annotations.ApiImplicitParams'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[31] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'io.swagger.annotations.ApiOperation'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[32] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'io.swagger.annotations.ApiResponse'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[33] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'io.swagger.annotations.ApiResponses'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[34] (imports) CustomImportOrder: Import statement for 'org.apache.james.core.Domain' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[35] (imports) CustomImportOrder: Import statement for 'org.apache.james.rrt.api.RecipientRewriteTable' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[36] (imports) CustomImportOrder: Import statement for 'org.apache.james.rrt.api.RecipientRewriteTableException' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[37] (imports) CustomImportOrder: Import statement for 'org.apache.james.rrt.lib.Mapping' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[38] (imports) CustomImportOrder: Import statement for 'org.apache.james.rrt.lib.MappingSource' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[39] (imports) CustomImportOrder: Import statement for 'org.apache.james.rrt.lib.Mappings' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[40] (imports) CustomImportOrder: Import statement for 'org.apache.james.webadmin.Routes' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[41] (imports) CustomImportOrder: Import statement for 'org.apache.james.webadmin.utils.ErrorResponder' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[42] (imports) CustomImportOrder: Import statement for 'org.apache.james.webadmin.utils.JsonTransformer' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[43] (imports) CustomImportOrder: Import statement for 'org.eclipse.jetty.http.HttpStatus' is in the wrong order. Should be in the 'SPECIAL_IMPORTS' group, expecting group 'STANDARD_JAVA_PACKAGE' on this line.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[44] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'spark.HaltException'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[45] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'spark.Request'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[46] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'spark.Response'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[47] (imports) CustomImportOrder: Imports without groups should be placed at the end of the import list: 'spark.Service'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[55] (imports) CustomImportOrder: Wrong lexicographical order for 'java.util.List' import. Should be before 'javax.ws.rs.Produces'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[56] (imports) CustomImportOrder: Wrong lexicographical order for 'java.util.Map' import. Should be before 'javax.ws.rs.Produces'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[57] (imports) CustomImportOrder: Wrong lexicographical order for 'java.util.Optional' import. Should be before 'javax.ws.rs.Produces'.
[0f7cca769367be82fcf05803a81a1a2949cad8b1] [ERROR] src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java:[58] (imports) CustomImportOrder: Wrong lexicographical order for 'java.util.function.Function' import. Should be before 'javax.ws.rs.Produces'.
```

As you can see, there is some checkstyle violation due to impport ordering.

In order to help you solving it, here is my IntelliJ configuration for import ordering:

```
import static all other imports
<blank line>
import java.*
<blank line>
import javax.*
<blank line>
import org.*
<blank line>
import com.*
<blank line>
import all other imports
```

(intelliJ -> File -> Settings -> Editor -> Code style -> Java -> Imports -> Import layout)

BTW you can check the checkstyle by running `mvn clean install -DskipTests` then `mvn checkstyle:check`.

Cheers,

Benoit

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/DomainMappingsRoutes.java
nstdio: The branch is on top of master and there is no method named `Guavate#toImmutableMultimap`.

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/DomainMappingsRoutesTest.java
nstdio: Disagree with using `PUT`, by using it we couple this two different methods together. When `PUT` action fails this test will fail too which is not cool thing.

chibenwa:
Hi.

Once again, there is a little checkstyle issue due to unused import in MappingsImpl class.

I contributed a fix here https://github.com/linagora/james-project/pull/1491/commits/1bad15b732eaba9936b86ed3bf66bf69406eb62c that you can cherry-pick.

Cheers,

nstdio:
Hi, @chibenwa, I have one question should I squash this `fixup`s ?

chibenwa:
For me your history is still very readable, thus I would say you would not need to squash fixups.

mbaechler:
merged, thank you for your contribution @nstdio , keep contributing !

chibenwa:
Hi,

Thanks again,

Cheers,

Benoit

nstdio:
Thanks for great comments.

PR #123, opened 2018-07-05
JAMES-2301 JMS mail queue do not preserve per-recipient headers.

nstdio:
Now we serialize per-recipient headers as a message property and
deserialize it when dequeue.

### DIFF
--- a/server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
+++ b/server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
@@ -32,6 +32,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Optional;
 import java.util.StringTokenizer;
 import java.util.concurrent.TimeUnit;
 
@@ -50,7 +51,9 @@ import javax.mail.MessagingException;
 import javax.mail.internet.AddressException;
 import javax.mail.internet.MimeMessage;
 
+import org.apache.commons.codec.binary.Hex;
 import org.apache.commons.collections.iterators.EnumerationIterator;
+import org.apache.commons.lang3.SerializationUtils;
 import org.apache.james.core.MailAddress;
 import org.apache.james.lifecycle.api.Disposable;
 import org.apache.james.metrics.api.Metric;
@@ -63,10 +66,13 @@ import org.apache.james.queue.api.ManageableMailQueue;
 import org.apache.james.server.core.MailImpl;
 import org.apache.james.server.core.MimeMessageCopyOnWriteProxy;
 import org.apache.mailet.Mail;
+import org.apache.mailet.PerRecipientHeaders;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.threeten.extra.Temporals;
 
+import com.github.fge.lambdas.Throwing;
+import com.google.common.base.Joiner;
 import com.google.common.collect.Iterators;
 
 /**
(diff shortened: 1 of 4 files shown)


server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
```diff
@@ -297,27 +303,20 @@ protected void produceMail(Map<String, Object> props, int msgPrio, Mail mail) th
...
-            }
+        // won't serialize the empty headers so it is mandatory
+        // to handle nulls when reconstructing mail from message
+        if (!mail.getPerRecipientSpecificHeaders().getHeadersByRecipient().isEmpty()) {
+            byte[] serialize = SerializationUtils.serialize(mail.getPerRecipientSpecificHeaders());
+            props.put(JAMES_MAIL_PER_RECIPIENT_HEADERS, Hex.encodeHexString(serialize));
         }
-        props.put(JAMES_MAIL_RECIPIENTS, recipientsBuilder.toString());
+
+        String recipientsBuilder = Joiner.on(JAMES_MAIL_SEPARATOR).skipNulls().join(mail.getRecipients());
```
chibenwa: It is no more a `recipientsBuilder` but a `recipientsAsString`

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
```diff
@@ -297,27 +303,20 @@ protected void produceMail(Map<String, Object> props, int msgPrio, Mail mail) th
...
-        StringBuilder recipientsBuilder = new StringBuilder();
-
-        Iterator<MailAddress> recipients = mail.getRecipients().iterator();
-        while (recipients.hasNext()) {
-            String recipient = recipients.next().toString();
-            recipientsBuilder.append(recipient.trim());
-            if (recipients.hasNext()) {
-                recipientsBuilder.append(JAMES_MAIL_SEPARATOR);
-            }
+        // won't serialize the empty headers so it is mandatory
```
chibenwa: Why can't we serialize it when it's empty?

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
```diff
@@ -297,27 +303,20 @@ protected void produceMail(Map<String, Object> props, int msgPrio, Mail mail) th
...
         props.put(JAMES_MAIL_REMOTEHOST, mail.getRemoteHost());
 
-        String sender;
-        MailAddress s = mail.getSender();
-        if (s == null) {
-            sender = "";
-        } else {
-            sender = mail.getSender().toString();
-        }
+        String sender = Optional.ofNullable(mail.getSender()).map(MailAddress::toString).orElse("");
```
chibenwa: Use `asString` instead of `toString`.

toString generally serves for debugging, and it's value can change...

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
nstdio: The empty `PerRecipientHeaders` will cost us a `303` bytes and once we hexing it'll double it's size, so I think it'll way much better to not sent anything instead of pay additional bytes for just empty object. Generally speaking this little optimization intended to decrease number of bytes sent over the wire.

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
nstdio: Sure, I just used old code, it seem to me bit difficult to read. I spent some time understand 5 lines and fill bit confused.

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
chibenwa: Seems fine to me

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
chibenwa: The new "Optional" version looks great and is way easier to read, yes, thank you.

nstdio:
@chibenwa 
> Wouldn't you mind applying your Hexadecimal trick also to mail attributes?
I'd love to. Could you open the separate issue, and assign it to me ?

chibenwa:
> @chibenwa

>>  Wouldn't you mind applying your Hexadecimal trick also to mail attributes?

> I'd love to. Could you open the separate issue, and assign it to me ?

https://issues.apache.org/jira/projects/JAMES/issues/JAMES-2167 seems to be already opened. I remembered I had a non working fix for this in some PRs, but the Hex trick is better IMO.

chibenwa:
For your information, I launched CI builds for all you pull requests. I will keep you informed about the results ;-)

nstdio:
Just realized that we could even more optimize this without complexity gain.
What do you think about replace Hex with Base64? With Base64 we could reduce bytes ~30% for free.

chibenwa:
> What do you think about replace Hex with Base64? With Base64 we could reduce bytes ~30% for free.

Seems to be a good idea, yes.

chibenwa:
Hi,

Thank you again for your dedication toward the James project. 

Thanks again,

Cheers,

Benoit

PR #124, opened 2018-07-05
JAMES-2296 JMS remove by recipient is not remove message from queue.

nstdio:
Using proper separator and LIKE operator instead of equals operator to match single recipient
from serialized string.


chibenwa:
Hi,

Thank you again for your dedication toward the James project.

Thanks again,

Cheers,

Benoit

PR #126, opened 2018-07-06
JAMES-2167 Serializable attributes are not preserved by enqueue/dequeue on a JMS queue.

nstdio:


chibenwa:
Thanks again for the great work.

I did a very rapid review, but will be afk until Wednesday. 

Cheers

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSSerializationUtils.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * The convenient method that serialize provided {@code value} if it is not supported by JMS. If JMS does support
+     * {@code value} type the object itself will be returned.
+     *
+     * @param value The serialization candidate.
+     *
+     * @return The serialized base64 string or {@code value} itself.
+     */
+    static Object trySerialize(Serializable value) {
+        return hasJMSNativeSupport(value) ? value : serialize(value).orElse(null);
```
chibenwa: How a String is diferenciated from a serialized value?

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
```diff
@@ -442,16 +426,39 @@ protected void populateMail(Message message, MailImpl mail) throws JMSException
...
      *
-     * @param value
-     * @return convertedValue
+     * @param attrValue The input string.
+     * @return The deserialized object or itself.
      */
-    protected Object convertAttributeValue(Object value) {
-        if (value == null || value instanceof String || value instanceof Byte || value instanceof Long || value instanceof Double || value instanceof Boolean || value instanceof Integer || value instanceof Short || value instanceof Float) {
-            return value;
+    private Serializable tryDeserialize(String attrValue) {
```
chibenwa: This should be moved to `JMSSerializationUtils`.

Also, you should not rely on an exception of deserializing to handle "native" values. If doing so, a user providing a `base64 serialized integer string` as an input will get an `Integer` as an output.

We miss the `bijection` behaviour that we would expect from such a feature. That may lead to subtle bugs.

My opinion on this is to promote correctness over performances. Hence I'd advocate**use base64 serialization even for native values**.

This would also have some cool benefits:
 - We would stop assuming ActiveMQ behaviour in James code as it is done in `JMSSerializationUtils::hasJMSNativeSupport`
 - We will simplify the code by removing one conditional branch.
 - `JMSSerializationUtils` would not be JMS specific anymore. We will be able to change its name, & in a near future use it in another context.
 - This would also remove the `if (X instanceOf String)` in `this::setMailAttribute`.

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSSerializationUtils.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * This class is similar to {@link SerializationUtils}. Unlike {@link SerializationUtils} this class operates with
+ * {@code String}s and not byte arrays.
+ * <p>
+ * The main advantage of this utility is that it introduces an additional operation after serialization and before
+ * deserialization. The operation consists in encoding/decoding the serialized/deserialized data in Base64, so that
+ * data can be safely transmitted over the wire.
+ */
+final class JMSSerializationUtils {
```
chibenwa: Couldn't we make this class public?

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSSerializationUtils.java
```diff
@@ -0,0 +1,109 @@
...
+    private JMSSerializationUtils() {
```
chibenwa: That private constructor seems useless to me.

We don't have the habit to add such constructor in class with only static methods.

Might you remove it?

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.io.Serializable;
+import java.util.Objects;
+
+import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang3.SerializationException;
+import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.Test;
+
+class JMSSerializationUtilsTest {
```
chibenwa: What happens if I try to serialize, then deserialize this string? 

`rO0ABXNyABFqYXZhLmxhbmcuSW50ZWdlchLioKT3gYc4AgABSQAFdmFsdWV4cgAQamF2YS5sYW5nLk51bWJlcoaslR0LlOCLAgAAeHAAAAAp`

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSSerializationUtils.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Serializes and deserializes the provided object.
+     *
+     * @param obj The object that needs to be serialized.
+     * @param <T> The type of the provided object.
+     *
+     * @return The provided object.
+     */
+    static <T extends Serializable> Optional<T> roundtrip(T obj) {
```
chibenwa: This test method  should be moved to test.

chibenwa: Maybe we should make it return `T` instead of `Optional<T>` and throw if empty... `roundtrip` expresses the idea of identity to me, and expressing it threw the types seems a good idea to me.

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
```diff
@@ -0,0 +1,119 @@
...
+        SerializableStringHolder value = new SerializableStringHolder("value");
+
+        Object actual = trySerialize(value);
+
+        assertThat(actual)
+                .isInstanceOf(String.class);
+    }
+
+    @Test
+    void roundTripShouldReturnEqualObject() {
```
chibenwa: We need roundtrip testing also on String, and other native values....

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Test
+    void trySerializeShouldReturnString() {
+        SerializableStringHolder value = new SerializableStringHolder("value");
+
+        Object actual = trySerialize(value);
+
+        assertThat(actual)
+                .isInstanceOf(String.class);
```
chibenwa: Can we get an assert on the string content?

chibenwa:
We also have a merge conflict. Could you solve it?

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
nstdio: Basically, it'll behave like in case of any other string, will return same content.

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSSerializationUtils.java
nstdio: If you did not mind I'm perfectly OK with that.

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
chibenwa: Are you sure?

Because for me it returns `41` instead of `"rO0ABXNyABFqYXZhLmxhbmcuSW50ZWdlchLioKT3gYc4AgABSQAFdmFsdWV4cgAQamF2YS5sYW5nLk51bWJlcoaslR0LlOCLAgAAeHAAAAAp"`

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
nstdio: Wouldn't that just test for test?

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
nstdio: Thanks for pointing this out. Definitely I'll double check this.

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
chibenwa: It would allow while reading the test class to get a glance at what goes on the wire in JMS/ActiveMQ. While not necessary, I consider it allows seeing the length of data, and the structure.

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
chibenwa: This comment is strongly connected to https://github.com/apache/james-project/pull/126#discussion_r201560161. If you apply what I recommend, you will naturally solve the above issue, I guess.

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/JMSMailQueue.java
nstdio: > This would also remove the if (X instanceOf String) in this::setMailAttribute

I think there would remove type checking on `Serializable` rather then on `String`. Because `JMSSerializationUtils` is spinning around `String`.

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
nstdio: In respect that we remove `JMSSerializationUtils::hasJMSNativeSupport` Is this still relevant for us?

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSSerializationUtilsTest.java
```diff
@@ -0,0 +1,106 @@
...
+                .map(JMSSerializationUtils::serialize)
+                .<T>map(JMSSerializationUtils::deserialize)
+                .orElseThrow(() -> new IllegalArgumentException("Cannot serialize/deserialize: " + obj));
+    }
+
+    @Test
+    void trySerializeShouldReturnString() {
+        SerializableStringHolder value = new SerializableStringHolder("value");
+
+        String expected = "rO0ABXNyAE1vcmcuYXBhY2hlLmphbWVzLnF1ZXVlLmptcy5KTVNTZXJpYWxp" +
```
chibenwa: We should explain what this string is. I propose to rename it 'serializedInteger' to explain how it was generated

chibenwa:
Hi!

I open a pull request for playing CI tests: https://github.com/linagora/james-project/pull/1540

I also contributed 2 little fixups on top of your work.

Cheers!

chibenwa:
Hi,

Thanks again,

Cheers,

Benoit

PR #127, opened 2018-07-08
JAMES-2295 JMS mailqueue deadlock.

nstdio:
This issue is related to ActiveMQ prefetch mechanics. Now new give ability to configure prefetch size for all types of MailQueues.


server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSMailQueueTest.java
```diff
@@ -49,7 +49,8 @@ public void setUp(BrokerService broker) throws Exception {
         RawMailQueueItemDecoratorFactory mailQueueItemDecoratorFactory = new RawMailQueueItemDecoratorFactory();
         NoopMetricFactory metricFactory = new NoopMetricFactory();
         String queueName = BrokerExtension.generateRandomQueueName(broker);
-        mailQueue = new JMSMailQueue(connectionFactory, mailQueueItemDecoratorFactory, queueName, metricFactory);
+        MailQueue.ConsumerOptions consumerOptions = name -> name + "?consumer.prefetchSize=0";
```
chibenwa: I should have an explicit API:

```
ConsumerOptions.builder()
   .noPrefetchSize()
   .build();
```

And not rely on strings at the API level.

server/queue/queue-api/src/main/java/org/apache/james/queue/api/EmptyConsumerOptions.java
```diff
@@ -0,0 +1,39 @@
...
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.queue.api;
+
+/**
+ * The stub implementation for {@code ConsumerOptions}. Simply returns given argument.
+ */
+final class EmptyConsumerOptions implements MailQueue.ConsumerOptions {
```
chibenwa: Can't this be implemented as a lambda in the ConsumerOptions class directly?

server/queue/queue-api/src/main/java/org/apache/james/queue/api/MailQueue.java
```diff
@@ -128,4 +128,24 @@ public MailQueueException(String msg) {
...
+            return EmptyConsumerOptions.instance();
+        }
+
+        /**
+         * The queue name with options for performing {@link MailQueue#deQueue()} operation. The implementation may
+         * return {@code queueName} itself if it does not support any options.
+         *
+         * @return The queue name maybe with additional options.
+         */
+        String applyForDequeue(String queueName);
```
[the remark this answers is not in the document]
chibenwa: More generally, maybe JMS encodes consumer options as URL parameters, but we should not leak that in our APIs...

server/queue/queue-activemq/src/main/java/org/apache/james/queue/activemq/ActiveMQConsumerOptions.java
```diff
@@ -0,0 +1,110 @@
...
+        public ActiveMQConsumerOptionsBuilder selector(String selector) {
+            optionsMap.put(ActiveMQSupport.CONSUMER_SELECTOR, selector);
+            return this;
+        }
+
+        public ActiveMQConsumerOptions build() {
+            ImmutableMap<String, String> options = optionsMap.build();
+
+            if (options.isEmpty()) {
+                return new ActiveMQConsumerOptions(null);
```
chibenwa: Shouldn't it be `Optional.empty()`?

server/queue/queue-activemq/src/main/java/org/apache/james/queue/activemq/ActiveMQConsumerOptions.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public String applyForDequeue(String name) {
+        return dequeueParams
+                .map(params -> name + '?' + params)
+                .orElse(name);
+    }
+
+    public static class ActiveMQConsumerOptionsBuilder {
```
chibenwa: Couldn't this class just be named `Builder`?

chibenwa: We should also rely on a `Builder builder()` helper static method

server/queue/queue-api/src/main/java/org/apache/james/queue/api/EmptyConsumerOptions.java
nstdio: Yes, good idea, definitely it can.

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSMailQueueTest.java
nstdio: Does every JMS implementation support prefetch ? `ConsumerOptions` is a general propose interface...so we should use only features that supported by the JMS spec.

server/queue/queue-activemq/src/main/java/org/apache/james/queue/activemq/ActiveMQConsumerOptions.java
```diff
@@ -0,0 +1,117 @@
...
+                .orElse(name);
+    }
+
+    public static class Builder {
+        private ImmutableMap.Builder<String, String> optionsMap = ImmutableMap.builder();
+
+        private Builder() {
+        }
+
+        public static Builder builder() {
```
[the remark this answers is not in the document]
nstdio: If you look at [here](https://github.com/apache/james-project/pull/127/files/42f8e023198b79b280a1aac1fe9ca52482cc4a3f#diff-0a3d75f68f122c84b4fa2e1f9b51b5f0R43) you can find that method.

server/queue/queue-activemq/src/test/java/org/apache/james/queue/activemq/ActiveMQMailQueueBlobTest.java
```diff
@@ -102,6 +103,20 @@ public void dequeueCouldBeInterleavingWithOutOfOrderAck() {
 
     }
 
+    @Test
+    @Override
+    @Disabled("JAMES-2301 Per recipients headers are not attached to the message.")
+    public void queueShouldPreservePerRecipientHeaders() {
```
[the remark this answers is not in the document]
chibenwa: You need to remove these @Disabled...

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSMailQueueTest.java
```diff
@@ -49,7 +49,8 @@ public void setUp(BrokerService broker) throws Exception {
         RawMailQueueItemDecoratorFactory mailQueueItemDecoratorFactory = new RawMailQueueItemDecoratorFactory();
         NoopMetricFactory metricFactory = new NoopMetricFactory();
         String queueName = BrokerExtension.generateRandomQueueName(broker);
-        mailQueue = new JMSMailQueue(connectionFactory, mailQueueItemDecoratorFactory, queueName, metricFactory);
+        ConsumerOptions consumerOptions = name -> name + "?consumer.prefetchSize=0";
```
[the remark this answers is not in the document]
chibenwa: > From my point of view our configuration/options should not affect entire connection to the message broker. We should be able to configure queue consumer creation (ConsumerOptions#applyForDequeue).

I'm not arguing for changing the implementation much, just proposing to make the **ConsumerOptions** a dumb interface:

```
interface ConsumerOptions {
    ImmutableMap<String, String> parameters();
}
```

(Not that this interface is just holding a configuration and enforce no knowledge about the way the queue works.)

Then, We could get a JMSConsumerOptions, and an ActiveMQConsumerOptions with the options I can use in these various case.

Then, JMSMailQueue will add the right parameters for the queue name when appropriate (it has the knowledge of how to encode the parameters).

Thus:
 - We get strong typing for JMSConsumerOptions
 - The configuration is a dumb POJO, which is easier to reason about (and classic in James).

Would it be OK for you?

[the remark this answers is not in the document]
nstdio: Yes, in fact its a good idea, solving multiple problems/inconsistencies with a single shot.

Please, clarify how do you see the `JMSConsumerOptions` object creation? It would be through `new` or through builder class?

What do you think about both options combined together (public constructor + builder) ?

In case of builder, I suppose we should expose single `put(String, String)` and provide users context free consumer options builder.

server/queue/queue-activemq/src/test/java/org/apache/james/queue/activemq/ActiveMQMailQueueBlobTest.java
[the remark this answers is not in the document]
nstdio: Yes, sure. This is a result of incorrect rebase.

server/queue/queue-jms/src/test/java/org/apache/james/queue/jms/JMSMailQueueTest.java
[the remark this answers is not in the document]
chibenwa: > Please, clarify how do you see the JMSConsumerOptions object creation? It would be through new or through builder class?

Through a builder, no? 

> What do you think about both options combined together (public constructor + builder) ?

Why using a constructor when I can use a builder? Isn't it better to enforce builder use?

> In case of builder, I suppose we should expose single put(String, String) and provide users context free consumer options builder.

The problem with that kind of API is that I got not idea about values I should pass. We can still expose the `put` you suggest, but we need a self-discoverable API for options we know.

WDYT?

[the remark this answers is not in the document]
nstdio: > The problem with that kind of API is that I got not idea about values I should pass. We can still expose the put you suggest, but we need a self-discoverable API for options we know.

Agree with that, but as far as I know JMS spec does not support things like that. Although we can provide pretty well documentation on that. In ActiveMQ case the builder API is pretty straightforward and easy do understand (from API design point of view).

> Why using a constructor when I can use a builder? Isn't it better to enforce builder use?

I thought about when user already has ready . Probably we should expose also the `putAll(Map<String, String>)` method in builder to not force user iterate over and add options one by one.

nstdio:
Hi @chibenwa 

I was trying to implement the new version of `ConsumerOptions` that we discussed.

```java
interface ConsumerOptions {
    ImmutableMap<String, String> parameters();
}
```

Here is sort of contradiction:

[here](https://github.com/apache/james-project/pull/127/files#diff-57f321ad4650460e153b0e32c690cf79R214) we should reference the `prefetchSize` size, but we don't know whether underlying JMS implementation does support that option or not. Is there anything that I get wrong ?

nstdio:
> In the tests we can enforce the prefetchSize (as ActiveMQ is used under the hood) via a lambda.

I really can't figure out how...

server/queue/queue-activemq/src/main/java/org/apache/james/queue/activemq/ActiveMQConsumerOptions.java
```diff
@@ -0,0 +1,117 @@
...
+import org.apache.james.queue.jms.ConsumerOptions;
+
+/**
+ * The ActiveMQ specific consumer options.
+ * <p>
+ * See <a href="http://activemq.apache.org/destination-options.html">http://activemq.apache.org/destination-options.html</>
+ * for more details.
+ */
+public final class ActiveMQConsumerOptions implements ConsumerOptions {
+    private final Optional<String> dequeueParams;
```
mbaechler: Sun Java guidelines (that we mostly implement with checkstyle) states that static members should come before any non-static member (true for innerclass, methods, etc)

server/queue/queue-activemq/src/test/java/org/apache/james/queue/activemq/ActiveMQConsumerOptionsTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        assertThat(new URL(options.applyForDequeue(queueName)))
+                .hasQuery(ActiveMQSupport.CONSUMER_PRIORITY + "=-1");
+    }
+
+    @Test
+    void consumerOptionsShouldReturnArgWhenNoOptionsAreGiven() {
+        ActiveMQConsumerOptions options = ActiveMQConsumerOptions.builder().build();
+
+        assertThat(options.applyForDequeue(queueName)).isSameAs(queueName);
```
mbaechler: I think equality is enough, don't you think ?

server/queue/queue-activemq/src/test/java/org/apache/james/queue/activemq/ActiveMQMailQueueBlobTest.java
```diff
@@ -69,7 +69,8 @@ public void setUp(BrokerService broker) {
         RawMailQueueItemDecoratorFactory mailQueueItemDecoratorFactory = new RawMailQueueItemDecoratorFactory();
         NoopMetricFactory metricFactory = new NoopMetricFactory();
         String queueName = BrokerExtension.generateRandomQueueName(broker);
-        mailQueue = new ActiveMQMailQueue(connectionFactory, mailQueueItemDecoratorFactory, queueName, USE_BLOB, metricFactory);
+        ActiveMQConsumerOptions consumerOptions = ActiveMQConsumerOptions.builder().prefetchSize(0).build();
```
mbaechler: what about offering a `noPrefetch()` builder method ?

mbaechler: and BTW it's never clear if 0 is "no" or "infinity" for example, so having a dedicated method is always better IMO

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/ConsumerOptions.java
```diff
@@ -0,0 +1,42 @@
...
+import org.apache.james.queue.api.MailQueue;
+
+/**
+ * Provides additional options when creating consumers.
+ */
+public interface ConsumerOptions {
+    /**
+     * The empty consumer options.
+     */
+    static ConsumerOptions empty() {
```
mbaechler: as it's immutable, you can very well write : 
```
ConsumerOptions EMPTY = Function.identity();
```

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/ConsumerOptions.java
```diff
@@ -0,0 +1,42 @@
...
+        return queueName -> queueName;
+    }
+
+    /**
+     * The queue name with options for performing {@link MailQueue#deQueue()} operation. The implementation may
+     * return {@code queueName} itself if it does not support any options.
+     *
+     * @return The queue name maybe with additional options.
+     */
+    String applyForDequeue(String queueName);
```
mbaechler: maybe you already discussed that with @chibenwa but this API makes the assumption that the queue name is actually an url where you can encode options through queryparams.
I should clearly hide that fact because it's more than likely no useful for some kinds of MailQueue.
What about not putting this method only in implementations that make sense ?

server/queue/queue-activemq/src/test/java/org/apache/james/queue/activemq/ActiveMQMailQueueBlobTest.java
nstdio: I thought about that and had a feeling that introducing `noPrefetch()` method will be highly coupled with our issue.

Perfectly OK with that.

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/ConsumerOptions.java
nstdio: Do you suggest to use field instead of method?

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/ConsumerOptions.java
nstdio: Our issue basically is related to `JMSMailQueue` from with we have reference on `ConsumerOptions` and not `ActiveMQConsumerOptions`. Although we can do `instanceof ActiveMQConsumerOptions` and then call `applyForDequeue`. That `instanceof` check will be a bit ugly in my opinion.

nstdio: Unfortunately `queue-jms` does not depend on `queue-activemq` which is logical. Therefore previous comment cannot be implemented...

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/ConsumerOptions.java
nstdio: I would prefer to use method (especially having a successful examples of Optional.empty(), Function.identity(), etc).

What reasoning you have on that?

server/queue/queue-activemq/src/test/java/org/apache/james/queue/activemq/ActiveMQConsumerOptionsTest.java
nstdio: This is not relevant code anymore :)

server/queue/queue-jms/src/main/java/org/apache/james/queue/jms/ConsumerOptions.java
mbaechler: The rational is : it never changes so you can easily avoid useless object creation.
The reason why Optional and Function uses methods for that is generic handling.
see for example Optional : 
```
public static<T> Optional<T> empty() {
        @SuppressWarnings("unchecked")
        Optional<T> t = (Optional<T>) EMPTY;
        return t;
    }
```
It allows to return the right Optional<T> type for the caller. They however use a singleton anyway for the same reason I asked it.
In our case, the type is not generic so we don't need this extra complexity but if you want to return a singleton with a method, I don't have anything against it.

nstdio: Sounds very logical. Surely I would like to use method approach and return singleton instance through, but in our case we must choose between field and method. I think for this particular case field is better approach.

Thanks for details.

chiominto78:
Ok so why are y'all talking about me like I AM stupid

chibenwa:

nstdio:
@chibenwa sorry, I've lost context for this PR long ago and actually don't have a time to proceed with it.

chibenwa:
From what I can see the initial issue won't happen anymore because the dequeue semantic changed (ie reactor migration).

Having prefetch for ActiveMQ might make sense on a performance standpoint but I think that it is not required for correctness.

As such I will close this PR and ticket.

Thanks again for your time on diagnostic and proposing a fix.

Best regards,

Benoit

PR #129, opened 2018-07-24
JAMES-2298 FileMailQueue do not support remove.

nstdio:
Now FileMailQueue uses Lucene indexing feature to super fast searching
and remove needed items from queue.

### DIFF
--- a/server/queue/queue-file/pom.xml
+++ b/server/queue/queue-file/pom.xml
@@ -92,6 +92,11 @@
             <artifactId>javax.inject</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-core</artifactId>
+            <version>7.4.0</version>
+        </dependency>
+        <dependency>
             <groupId>org.assertj</groupId>
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
--- a/server/queue/queue-file/pom.xml
+++ b/server/queue/queue-file/pom.xml
@@ -94,7 +94,7 @@
         <dependency>
             <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-core</artifactId>
-            <version>7.4.0</version>
+            <version>${lucene.version}</version>
         </dependency>
         <dependency>
             <groupId>org.assertj</groupId>
--- a/server/queue/queue-file/src/main/java/org/apache/james/queue/file/FileMailQueue.java
+++ b/server/queue/queue-file/src/main/java/org/apache/james/queue/file/FileMailQueue.java
@@ -36,23 +36,40 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Stream;
 
 import javax.mail.MessagingException;
 import javax.mail.util.SharedFileInputStream;
 
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.mutable.MutableLong;
 import org.apache.james.lifecycle.api.Disposable;
 import org.apache.james.lifecycle.api.LifecycleUtil;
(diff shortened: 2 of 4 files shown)


server/queue/queue-file/pom.xml
```diff
@@ -86,6 +86,11 @@
             <groupId>javax.inject</groupId>
             <artifactId>javax.inject</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-core</artifactId>
+            <version>7.4.0</version>
```
nstdio: Still confused with versions...

server/queue/queue-file/src/main/java/org/apache/james/queue/file/FileMailQueue.java
```diff
@@ -197,20 +192,14 @@ public void enQueue(Mail mail, long delay, TimeUnit unit) throws MailQueueExcept
                 }
             }
 
+            indexWriter.addDocument(toIndexDocument(mail, key));
+            indexWriter.commit();
```
mbaechler: does it flush the index to disk ?

server/queue/queue-file/src/main/java/org/apache/james/queue/file/FileMailQueue.java
```diff
@@ -62,34 +79,40 @@
...
+    private final IndexWriter indexWriter;
 
     public FileMailQueue(MailQueueItemDecoratorFactory mailQueueItemDecoratorFactory, File parentDir, String queuename, boolean sync) throws IOException {
         this.mailQueueItemDecoratorFactory = mailQueueItemDecoratorFactory;
         this.sync = sync;
         this.queueName = queuename;
         this.queueDir = new File(parentDir, queueName);
         this.queueDirName = queueDir.getAbsolutePath();
+        StandardAnalyzer analyzer = new StandardAnalyzer();
+        this.indexWriter = new IndexWriter(FSDirectory.open(new File(queueDir, "index").toPath()),
```
mbaechler: maybe we could name the field so that we now what is actually indexed ?

server/queue/queue-file/src/main/java/org/apache/james/queue/file/FileMailQueue.java
```diff
@@ -222,6 +211,31 @@ public void enQueue(Mail mail, long delay, TimeUnit unit) throws MailQueueExcept
...
+    private void enqueueInMemory(String key) {
+        try {
+            inmemoryQueue.put(key);
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("Unable to init", e);
+        }
+    }
+
+    private Document toIndexDocument(Mail mail, String key) {
```
mbaechler: could we find a better name to describe what's in the document ?

server/queue/queue-file/src/main/java/org/apache/james/queue/file/FileMailQueue.java
```diff
@@ -395,21 +411,45 @@ public long clear() throws MailQueueException {
...
-        switch (type) {
-            case Name:
-                FileItem item = keyMappings.remove(value);
-                if (item != null) {
-                    item.delete();
-                    return 1;
-                } else {
-                    return 0;
-                }
+        try (IndexReader indexReader = DirectoryReader.open(indexWriter)) {
```
mbaechler: could we split this method, I find it too long

server/queue/queue-file/src/main/java/org/apache/james/queue/file/FileMailQueue.java
```diff
@@ -395,21 +411,45 @@ public long clear() throws MailQueueException {
      */
     @Override
     public long remove(Type type, String value) throws MailQueueException {
```
mbaechler: are we thread-safe ?

nstdio: I **guess** yes we are. Here I see two sources of potential danger for concurrency.

* `keyMappings::remove` - is a bulletproof.
* `indexWriter::deleteDocuments` - Again quote from javadoc

> All given deletes are applied and flushed atomically at the same time.

Maybe I should write tests on it to not **guess**.

nstdio: Oh yes, this one is for sure. The `indexReader` will be not _fsync_-ed if concurrent thread add mail to the queue.

chibenwa:
I am closing this as the FileMailQueue had been removed.

PR #131, opened 2018-07-24
JAMES-2303 JDBC Mail repository do not persist Per recipient headers.

nstdio:
Using JDK serialization to persist Per recipient headers.

### DIFF
--- a/server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
+++ b/server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
@@ -33,6 +33,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Types;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
```diff
@@ -542,12 +544,18 @@ protected void internalStore(Mail mc) throws IOException, MessagingException {
                     insertMessage.setString(6, recipients.toString());
                     insertMessage.setString(7, mc.getRemoteHost());
                     insertMessage.setString(8, mc.getRemoteAddr());
-                    insertMessage.setTimestamp(9, new java.sql.Timestamp(mc.getLastUpdated().getTime()));
+                    if (mc.getPerRecipientSpecificHeaders().getHeadersByRecipient().isEmpty()) {
```
trantienduchn: is `mc.getPerRecipientSpecificHeaders()` could be null?

nstdio: As I recall, no it can't. Field is initialized in the ctor.

chibenwa:
Hi,

Thanks again,

Cheers,

Benoit

PR #132, opened 2018-07-24
JAMES-2498 JDBCMailRepository should not materialize entities for counting them.

nstdio:
Executing SQL query for size.

### DIFF
--- a/server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
+++ b/server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
@@ -722,6 +722,27 @@ public class JDBCMailRepository extends AbstractMailRepository {
     }
 
     @Override
+    public long size() throws MessagingException {
+        Connection conn = null;
+        PreparedStatement count = null;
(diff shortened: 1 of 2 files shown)


server/data/data-jdbc/src/main/java/org/apache/james/mailrepository/jdbc/JDBCMailRepository.java
```diff
@@ -721,6 +721,27 @@ protected void internalRemove(MailKey key) throws MessagingException {
         }
     }
 
+    @Override
+    public long size() throws MessagingException {
+        Connection conn = null;
```
mbaechler: AFAIK all jdbc objects can by handled with a try-with-resource

nstdio: Sure! I was trying to follow common pattern used in the class.

mbaechler: In these cases, we often do a first commit to migrate code to the right code-style, then a second one with the change we want.

chibenwa:
Hi,

Thanks again,

Cheers,

Benoit

PR #145, opened 2018-12-11
JAMES-2171: refactor LogMessage class

ReturnHttp418:
Add LogLevel defined by enum and use it by switch case;
Concatenate log messages by using StringBuffer for its thread safety;
Use builder pattern to build LogMessage and set configuration value like specificHeaders, specificAttribute, Logger, header, body;
Add a specific exception in the initialization phase and throw exception when maxBody is negative;
Fix old uts and add new uts for new feature;

### DIFF
--- a/mailet/api/src/main/java/org/apache/mailet/MailetConfig.java
+++ b/mailet/api/src/main/java/org/apache/mailet/MailetConfig.java
@@ -29,9 +29,19 @@ import java.util.Iterator;
  * The configuration information consists of mailet-specific initialization
  * parameters in a set of name/value pairs, and a MailetContext object
  * which allows the mailet to interact with the mailet container.
- *
  */
 public interface MailetConfig {
+
+    /**
+     * Returns the value of the named initialization parameter,
+     * or null if the parameter does not exist.
+     *
+     * @param name  the name of the initialization parameter
+     * @param clazz the type of the returned value
+     * @return the value of the initialization parameter, or null
+     * @throws ClassCastException
+     */
+    <T> T getInitParameter(String name, Class<T> clazz) throws ClassCastException;
 
     /**
      * Returns the value of the named initialization parameter,
@@ -48,7 +58,7 @@ public interface MailetConfig {
      * no initialization parameters.
      *
      * @return an Iterator of String objects containing the names of the
-     *      mailet's initialization parameters
+     * mailet's initialization parameters
      */
     Iterator<String> getInitParameterNames();
 
@@ -57,7 +67,7 @@ public interface MailetConfig {
      * executing.
      *
      * @return a MailetContext object which can be used by the mailet
-     *      to interact with the mailet container
+     * to interact with the mailet container
      */
     MailetContext getMailetContext();
 
@@ -70,4 +80,5 @@ public interface MailetConfig {
      * @return the name of the mailet instance
      */
     String getMailetName();
+
 }
--- a/mailet/base/src/main/java/org/apache/mailet/base/GenericMailet.java
+++ b/mailet/base/src/main/java/org/apache/mailet/base/GenericMailet.java
@@ -88,6 +88,7 @@ public abstract class GenericMailet implements Mailet, MailetConfig {
         return MailetUtil.getInitParameter(config, name).orElse(defaultValue);
     }
 
+
     public Optional<String> getInitParameterAsOptional(String name) {
         String value = getInitParameter(name);
         if (Strings.isNullOrEmpty(value)) {
@@ -122,11 +123,27 @@ public abstract class GenericMailet implements Mailet, MailetConfig {
      * named parameter from the mailet's MailetConfig object.
      *
      * @param name - a String specifying the name of the initialization parameter
-     * @return a String containing the value of the initalization parameter
+     * @return a String containing the value of the initialization parameter
      */
     @Override
     public String getInitParameter(String name) {
         return config.getInitParameter(name);
+    }
+
+    /**
+     * Returns a String containing the value of the named initialization
+     * parameter, or null if the parameter does not exist.
+     * <p>
+     * This method is supplied for convenience. It gets the value of the
+     * named parameter from the mailet's MailetConfig object.
+     *
+     * @param name - a String specifying the name of the initialization parameter
+     * @param clazz - a class specifying the type of the returned value type
+     * @return an object containing the value of the initialization parameter
+     */
+    @Override
+    public <T> T getInitParameter(String name, Class<T> clazz) {
+        return config.getInitParameter(name, clazz);
     }
 
     /**
@@ -137,17 +154,29 @@ public abstract class GenericMailet implements Mailet, MailetConfig {
      * named parameter from the mailet's MailetConfig object.
      *
      * @param name - a String specifying the name of the initialization parameter
-     * @param defValue - a String specifying the default value when the parameter
+     * @param defaultValue - a String specifying the default value when the parameter
      *                    is not present
-     * @return a String containing the value of the initalization parameter
+     * @return a String containing the value of the initialization parameter
      */
-    public String getInitParameter(String name, String defValue) {
-        String res = config.getInitParameter(name);
-        if (res == null) {
-            return defValue;
-        } else {
-            return res;
-        }
+    public String getInitParameter(String name, String defaultValue) {
+        return Optional.ofNullable(config.getInitParameter(name)).orElse(defaultValue);
+    }
+
+    /**
+     * Returns a String containing the value of the named initialization
+     * parameter, or defValue if the parameter does not exist.
+     * <p>
+     * This method is supplied for convenience. It gets the value of the
+     * named parameter from the mailet's MailetConfig object.
+     *
+     * @param name - a String specifying the name of the initialization parameter
+     * @param clazz - a class type specifying the type of the returned value type
+     * @param defaultValue - a generic type specifying the default value when the parameter
+     *                    is not present
+     * @return a generic type containing the value of the initialization parameter
+     */
+    public <T> T getInitParameter(String name, Class<T> clazz, T defaultValue) {
+        return Optional.ofNullable(config.getInitParameter(name, clazz)).orElse(defaultValue);
     }
 
     /**
--- a/mailet/base/src/main/java/org/apache/mailet/base/MailetUtil.java
+++ b/mailet/base/src/main/java/org/apache/mailet/base/MailetUtil.java
@@ -96,19 +96,13 @@ public class MailetUtil {
      * <p>Gets a boolean valued init parameter.</p>
      * @param config not null
      * @param name name of the init parameter to be queried
-     * @return true when the init parameter is <code>true</code> (ignoring case);
-     * false when the init parameter is <code>false</code> (ignoring case);
-     * otherwise the default value
+     * @return true when the init parameter is <code>true</code>;
+     * false when the init parameter is <code>false</code>;
+     * throw an exception when the value is not of Boolean type;
+     * otherwise returns optional with empty value
      */
     public static Optional<Boolean> getInitParameter(MailetConfig config, String name) {
-        String value = config.getInitParameter(name);
-        if ("true".equalsIgnoreCase(value)) {
-            return Optional.of(true);
-        }
-        if ("false".equalsIgnoreCase(value)) {
-            return Optional.of(false);
-        }
-        return Optional.empty();
+        return Optional.ofNullable(config.getInitParameter(name, Boolean.class));
     }
 
     public static int getInitParameterAsStrictlyPositiveInteger(String condition, int defaultValue) throws MessagingException {
--- a/mailet/standard/src/main/java/org/apache/james/transport/mailets/AddFooter.java
+++ b/mailet/standard/src/main/java/org/apache/james/transport/mailets/AddFooter.java
@@ -114,7 +114,7 @@ public class AddFooter extends GenericMailet {
     /**
      * Prepends the content of the MimePart as text to the existing footer
      *
-     * @param part the MimePart to attach
+     * @param content the MimePart to attach
      */
     private String attachFooterToText(String content) throws MessagingException,
             IOException {
@@ -148,7 +148,7 @@ public class AddFooter extends GenericMailet {
         return new StringBuilder()
                 .append(content.substring(0, insertionIndex))
                 .append(getFooterHTML())
-                .append(content.substring(insertionIndex, content.length()))
+                .append(content.substring(insertionIndex))
                 .toString();
     }
 
--- a/mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
+++ b/mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
@@ -34,6 +34,7 @@ import java.net.Socket;
 import java.net.UnknownHostException;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.mail.MessagingException;
@@ -266,8 +267,7 @@ public class ClamAVScan extends GenericMailet {
      * Initializer for property debug.
      */
     protected void initDebug() {
-        String debugParam = getInitParameter("debug");
-        setDebug((debugParam == null) ? false : Boolean.valueOf(debugParam));
+        setDebug(Optional.ofNullable(getInitParameter("debug", Boolean.class)).orElse(false));
     }
 
     /**
@@ -332,8 +332,7 @@ public class ClamAVScan extends GenericMailet {
      * Initializer for property port.
      */
     protected void initPort() {
-        String portParam = getInitParameter("port");
-        setPort((portParam == null) ? DEFAULT_PORT : Integer.parseInt(portParam));
+        setPort(Optional.ofNullable(getInitParameter("port", Integer.class)).orElse(DEFAULT_PORT));
         if (isDebug()) {
             LOGGER.debug("port: {}", getPort());
         }
@@ -363,8 +362,8 @@ public class ClamAVScan extends GenericMailet {
      * Initializer for property maxPings.
      */
     protected void initMaxPings() {
-        String maxPingsParam = getInitParameter("maxPings");
-        setMaxPings((maxPingsParam == null) ? DEFAULT_MAX_PINGS : Integer.parseInt(maxPingsParam));
+        Integer maxPingsParam = getInitParameter("maxPings", Integer.class);
+        setMaxPings(Optional.ofNullable(getInitParameter("maxPings", Integer.class)).orElse(DEFAULT_MAX_PINGS));
         if (isDebug()) {
             LOGGER.debug("maxPings: {}", getMaxPings());
         }
@@ -394,8 +393,7 @@ public class ClamAVScan extends GenericMailet {
      * Initializer for property pingIntervalMilli.
      */
     protected void initPingIntervalMilli() {
-        String pingIntervalMilliParam = getInitParameter("pingIntervalMilli");
-        setPingIntervalMilli((pingIntervalMilliParam == null) ? DEFAULT_PING_INTERVAL_MILLI : Integer.parseInt(pingIntervalMilliParam));
+        setPingIntervalMilli(Optional.ofNullable(getInitParameter("pingIntervalMilli", Integer.class)).orElse(DEFAULT_PING_INTERVAL_MILLI));
         if (isDebug()) {
             LOGGER.debug("pingIntervalMilli: {}", getPingIntervalMilli());
         }
@@ -425,8 +423,7 @@ public class ClamAVScan extends GenericMailet {
      * Initializer for property streamBufferSize.
      */
     protected void initStreamBufferSize() {
-        String streamBufferSizeParam = getInitParameter("streamBufferSize");
-        setStreamBufferSize((streamBufferSizeParam == null) ? DEFAULT_STREAM_BUFFER_SIZE : Integer.parseInt(streamBufferSizeParam));
+        setStreamBufferSize(Optional.ofNullable(getInitParameter("streamBufferSize", Integer.class)).orElse(DEFAULT_STREAM_BUFFER_SIZE));
         if (isDebug()) {
             LOGGER.debug("streamBufferSize: {}", getStreamBufferSize());
         }
--- a/mailet/standard/src/main/java/org/apache/james/transport/mailets/LogLevel.java
+++ b/mailet/standard/src/main/java/org/apache/james/transport/mailets/LogLevel.java
@@ -1,0 +1,52 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+package org.apache.james.transport.mailets;
+
+
+import org.apache.mailet.base.LogWriter;
+
+public enum LogLevel {
+    DEBUG{
+        @Override
+        LogWriter log() {
+            return (logger, message) -> logger.debug(message);
+        }
+    },
+    INFO{
+        @Override
+        LogWriter log() {
+            return (logger, message) -> logger.info(message);
+        }
+    },
+    WARN{
+        @Override
+        LogWriter log() {
+            return (logger, message) -> logger.warn(message);
+        }
+    },
+    ERROR{
+        @Override
+        LogWriter log() {
+            return (logger, message) -> logger.error(message);
+        }
+    };
+
+    abstract LogWriter log();
+
+}
--- a/mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
+++ b/mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
@@ -18,20 +18,28 @@
  ****************************************************************/
 
 
-
 package org.apache.james.transport.mailets;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
+import java.util.Optional;
 import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
 
 import javax.mail.MessagingException;
 import javax.mail.internet.MimeMessage;
 
 import org.apache.commons.io.IOUtils;
-import org.apache.mailet.Mail;
+import org.apache.james.core.MailAddress;
 import org.apache.mailet.base.GenericMailet;
+import org.apache.mailet.Attribute;
+import org.apache.mailet.AttributeValue;
+import org.apache.mailet.AttributeName;
+import org.apache.mailet.Mail;
+import org.apache.mailet.PerRecipientHeaders;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -52,11 +60,15 @@ public class LogMessage extends GenericMailet {
      * or mark it as finished.
      */
     private final Logger logger;
+    private LogLevel logLevel;
     private boolean passThrough = true;
-    private boolean headers = true;
-    private boolean body = true;
-    private int bodyMax = 0;
-    private String comment = null;
+    private boolean headers;
+    private boolean body;
+    private int bodyMax;
+    private String comment;
+    private List<String> specificHeaderNames;
+    private List<String> specificAttributeNames;
+
 
     public LogMessage(Logger logger) {
         this.logger = logger;
@@ -66,16 +78,20 @@ public class LogMessage extends GenericMailet {
         this(LOGGER);
     }
 
+
     @Override
     public void init() {
         try {
+            logLevel = getInitParameter("level", LogLevel.class, LogLevel.INFO);
+            specificHeaderNames = getInitParameter("specificHeaders", List.class, Collections.emptyList());
+            specificAttributeNames = getInitParameter("specificAttributes", List.class, Collections.emptyList());
+            headers = getInitParameter("headers", false);
+            body = getInitParameter("body", false);
             passThrough = getInitParameter("passThrough", true);
-            headers = getInitParameter("headers", true);
-            body = getInitParameter("body", true);
-            bodyMax = (getInitParameter("maxBody") == null) ? 0 : Integer.parseInt(getInitParameter("maxBody"));
+            bodyMax = Optional.ofNullable(getInitParameter("maxBody", Integer.class)).orElse(0);
             comment = getInitParameter("comment");
-        } catch (Exception e) {
-            logger.error("Caught exception while initializing LogMessage", e);
+        } catch (ClassCastException e) {
+            logger.error("Caught exception while parsing initial parameter values", e);
         }
     }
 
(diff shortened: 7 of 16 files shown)


mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -52,18 +52,81 @@
...
-    private int bodyMax = 0;
-    private String comment = null;
+    private boolean headers;
+    private boolean body;
+    private int bodyMax;
+    private String comment;
+    private List<String> specificHeaders;
+    private List<String> specificAttributes;
+
+    private StringBuffer logBuffer = new StringBuffer();
```
chibenwa: Please consider that mailet are concurrently called. Your code needs to be safe in a multi-threaded context.

Hence sharing a buffer seems 'unsafe'

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -52,18 +52,81 @@
...
+            return new LogMessage(logger,logLevel,specificHeaders,specificAttributes,headers,body);
+        }
 
-    public LogMessage(Logger logger) {
-        this.logger = logger;
     }
 
-    public LogMessage() {
-        this(LOGGER);
+    private LogMessage(Logger logger, LogLevel logLevel, List<String> specificHeaders, List<String> specificAttributes, boolean headers, boolean body){
```
chibenwa: While having a builder is clearly valuable, James needs to be able to 'inject' fields in that mailet.

So, we need to read the LogLevel, specificHeaders, specificAttributes headers & body parameters from the configuration, in the `init` method.

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -73,9 +136,15 @@ public void init() {
...
             bodyMax = (getInitParameter("maxBody") == null) ? 0 : Integer.parseInt(getInitParameter("maxBody"));
+            if(bodyMax < 0){
+                logger.error("Parsed a negative maxBody value and caused initialization failure");
+                throw new IllegalArgumentException("The value of maxBody cannot be negative.");
+            }
             comment = getInitParameter("comment");
-        } catch (Exception e) {
-            logger.error("Caught exception while initializing LogMessage", e);
+        } catch (NumberFormatException e){
+            logger.error("Caught exception while parsing maxBody value", e);
```
chibenwa: I would argue that we don't want to start with an invalid configuration. Maybe catching these exceptions is not needed?

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -86,12 +155,30 @@ public String getMailetInfo() {
 
     @Override
     public void service(Mail mail) {
-        logger.info("Logging mail {}", mail.getName());
+//        logger.info("Logging mail {}", mail.getName());
```
chibenwa: Can this comment be removed?

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -100,25 +187,39 @@ public void service(Mail mail) {
         }
     }
 
+    private void logAttribute(List<String> specificAttributes) {
+        if(!specificAttributes.isEmpty()){
+            logBuffer.append('\n');
+            for(String attribute : specificAttributes){
```
chibenwa: Don't we need to read these specific attributes from the email rather than just log the configured value?

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -100,25 +187,39 @@ public void service(Mail mail) {
...
     private void logHeaders(MimeMessage message) throws MessagingException {
         if (headers && logger.isInfoEnabled()) {
-            logger.info("\n");
+            logBuffer.append('\n');
             for (String header : Collections.list(message.getAllHeaderLines())) {
-                logger.info(header + "\n");
+                logBuffer.append(header + '\n');
+            }
+            if(!specificHeaders.isEmpty()){
+                for(String header : specificHeaders){
```
chibenwa: Don't we need to read mail.perRecipientHeaders here rather than only listing configured header names?

ReturnHttp402:
@chibenwa Hi I made a few changes and I'm not sure if this is a best practice to use type casting, please have a check, thanks!

mailet/api/src/main/java/org/apache/mailet/MailetConfig.java
```diff
@@ -40,7 +41,7 @@
      * @param name the name of the initialization parameter
      * @return the value of the initialization parameter, or null
      */
-    String getInitParameter(String name);
+    Object getInitParameter(String name);
```
chibenwa: I tend to strongly believe strong typing is a nice property to have, as it enables compile time checks and improves safety + efficiency. It also improves API discovery, which is actually quite critical in an API that is the base of James extension model.

We should refrain from changing any API related to mailets.

That being said, maybe we can add a : 

```
<T> T getInitParameter(String name, Class<T> clazz) throws ClassCastException;
```

Might be a good idea, as it might allow the caller to avoid some job parsing strings.

Note that this method returns always the right type, which is now decided by the caller (and not by the underlying datatype).

Also, I would be in favor to keep the 'old' code using the previous 'String' version....

ReturnHttp402: Is the `Class<T> clazz` in the `getInitParameter` used to indicate the data type for the returned value to improve API discovery? If yes, what supposed to fill in the `config.getInitParameter(name, )` used in the method `getInitParameter` in class `GenericMailet`?

ReturnHttp402: The reason that I use the enum version is that I think using String might cause problems when fill string with different cases by mistake, but anyway, I'll roll the related code back.

ReturnHttp402: In `GenericMailet` it calls the` getInitParameter` in the `MailetConfig`, however, when added the parameter `Class<T> clazz`, there's something needed to be passed in or construct when calling the method, I don't know what to pass here.

chibenwa:
Well, this method needs to be implemented in MailetConfig & MailetConfigImpl, then GenericMailet only needs to be calling that:

```
    @Override
    public <T> T getInitParameter(String name, Class<T> clazz) throws ClassCastException {
        return getMailetConfig().getInitParameter(name, clazz);
    }
```

For the record, in MailetConfigImpl, this could look like this: 

```
    @Override
    public <T> T getInitParameter(String name, Class<T> clazz) throws ClassCastException {
        Object o = configuration.getProperty(name);
        if (clazz.isInstance(o)) {
            return (T) o;
        }
        throw new ClassCastException(name + " property is not of type " + clazz.getCanonicalName());
    }
```

Would this seems Ok to you?

mailet/api/src/main/java/org/apache/mailet/MailetConfig.java
chibenwa: Well, this method needs to be implemented in MailetConfig & MailetConfigImpl, then GenericMailet only needs to be calling that:

```
    @Override
    public <T> T getInitParameter(String name, Class<T> clazz) throws ClassCastException {
        return getMailetConfig().getInitParameter(name, clazz);
    }
```

For the record, in MailetConfigImpl, this could look like this: 

```
    @Override
    public <T> T getInitParameter(String name, Class<T> clazz) throws ClassCastException {
        Object o = configuration.getProperty(name);
        if (clazz.isInstance(o)) {
            return (T) o;
        }
        throw new ClassCastException(name + " property is not of type " + clazz.getCanonicalName());
    }
```

Would this seems Ok to you?

ReturnHttp402:
@chibenwa Thanks for your comments, it's very helpful, I've made some changes, can you review it again?

mailet/standard/src/main/java/org/apache/james/transport/mailets/AddFooter.java
```diff
@@ -53,7 +53,7 @@
     
     @Override
     public void init() throws MessagingException {
-        plainTextFooter = (String) getInitParameter(MailetConfiguration.TEXT.name());
+        plainTextFooter = getInitParameter("text", String.class);
```
chibenwa: Why don't we 'stick' with the previous String version, that is directly the type that we want?

mailet/standard/src/main/java/org/apache/james/transport/mailets/AddSubjectPrefix.java
```diff
@@ -46,7 +46,7 @@
 
     @Override
     public void init() throws MessagingException {
-        subjectPrefix = (String) getInitParameter(MailetConfiguration.SUBJECT_PREFIX.name());
+        subjectPrefix = getInitParameter("subjectPrefix", String.class);
```
chibenwa: Why don't we 'stick' with the previous String version, that is directly the type that we want?

mailet/standard/src/test/java/org/apache/james/transport/mailets/LogMessageTest.java
```diff
@@ -352,4 +354,11 @@ private FakeMail createMail() throws MessagingException {
                 .build();
     }
 
+    @Test
+    public void test() {
```
chibenwa: This "debug" tests is valuable. Can we do a real assertion for it?

mailet/standard/src/main/java/org/apache/james/transport/mailets/AddFooter.java
ReturnHttp402: You mean the explicit type casting version? I think it's redundant here because using generics can help us to do the auto casting job.

ReturnHttp402:
@chibenwa Hi, Sorry for the long-delayed reply, I've been fixing my PC recently, I have one question about your last comment on 'String' version, do you mean using the previous explicit type casting?

mailet/api/src/main/java/org/apache/mailet/MailetConfig.java
```diff
@@ -38,17 +38,19 @@
      * or null if the parameter does not exist.
      *
      * @param name the name of the initialization parameter
+     * @param clazz the type of the returned value
      * @return the value of the initialization parameter, or null
+     * @throws ClassCastException
      */
-    String getInitParameter(String name);
```
chibenwa: IMO we should keep the `String getInitParameter(String name);` method as it is more convenient most of the time...

chibenwa:
Hi @ReturnHttp402 

I mean we could have both:

```
<T> T getInitParameter(String name, Class<T> clazz) throws ClassCastException;

String getInitParameter(String name);
```

That way nothing changes if I want a String. If I want *something else*, I can use the new method....

chibenwa:
Also can you rebase your PR on master? I see a lot of non related commits in github...

ReturnHttp402:
@chibenwa Hi, I add back the former string version api and pickup only my commits and do the rebase, I suspect that other unrelated commits may come from the CI process.

mailet/base/src/main/java/org/apache/mailet/base/GenericMailet.java
```diff
@@ -88,8 +88,9 @@ public boolean getInitParameter(String name, boolean defaultValue) {
         return MailetUtil.getInitParameter(config, name).orElse(defaultValue);
     }
 
+
     public Optional<String> getInitParameterAsOptional(String name) {
-        String value = getInitParameter(name);
+        String value = getInitParameter(name, String.class);
```
chibenwa: `String value = getInitParameter(name);`

Should be enough, no?

mailet/base/src/main/java/org/apache/mailet/base/MailetUtil.java
```diff
@@ -96,19 +96,19 @@ public static String normalizeSubject(String subj, String prefix) {
...
+        if(value != null){
+            if(Boolean.class.isInstance(value)){
+                return Optional.of((Boolean)value);
+            } else throw new ClassCastException(name + " property is not of type " + Boolean.class);
         }
-        if ("false".equalsIgnoreCase(value)) {
-            return Optional.of(false);
-        }
-        return Optional.empty();
+        else return Optional.empty();
```
chibenwa: This will probably not pass the checkstyle rules. You need to explicitly define blocks of code using `{}`

mailet/base/src/main/java/org/apache/mailet/base/MailetUtil.java
```diff
@@ -96,19 +96,19 @@ public static String normalizeSubject(String subj, String prefix) {
...
+     * @return true when the init parameter is <code>true</code>;
+     * false when the init parameter is <code>false</code>;
+     * throw an exception when the value is not of Boolean type;
      * otherwise the default value
      */
     public static Optional<Boolean> getInitParameter(MailetConfig config, String name) {
-        String value = config.getInitParameter(name);
-        if ("true".equalsIgnoreCase(value)) {
-            return Optional.of(true);
+        Object value = config.getInitParameter(name, Boolean.class);
```
chibenwa: If we ask for a Boolean, then we got a Boolean, right? No need to type it as an object then cast it?

chibenwa: Moreover the entire point of using `Optional` is to not deal with null checks.

I bet this code snipset could actually be as simple as: 

```
return Optional.ofNullable(config.getInitParameter(name, Boolean.class));
```

mailet/standard/src/main/java/org/apache/james/transport/mailets/AddFooter.java
```diff
@@ -146,9 +146,9 @@ private String attachFooterToHTML(String content) throws MessagingException,
         }
         int insertionIndex = matcher.start(matcher.groupCount() - 1);
         return new StringBuilder()
-                .append(content.substring(0, insertionIndex))
+                .append(content, 0, insertionIndex)
```
chibenwa: `.append(content.substring(0, insertionIndex))` was IMO more explicit

mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
```diff
@@ -266,8 +266,8 @@ public String getMailetInfo() {
      * Initializer for property debug.
      */
     protected void initDebug() {
-        String debugParam = getInitParameter("debug");
-        setDebug((debugParam == null) ? false : Boolean.valueOf(debugParam));
+        Boolean debugParam = getInitParameter("debug", Boolean.class);
+        setDebug((debugParam == null) ? false : debugParam);
```
chibenwa: This can be `Optional.ofNullable(debugParam).orElse(false)` which would be more functional friendly and more readable.

mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
```diff
@@ -332,8 +332,8 @@ public void setHost(String host) throws UnknownHostException {
      * Initializer for property port.
      */
     protected void initPort() {
-        String portParam = getInitParameter("port");
-        setPort((portParam == null) ? DEFAULT_PORT : Integer.parseInt(portParam));
+        Integer portParam = getInitParameter("port", Integer.class);
+        setPort((portParam == null) ? DEFAULT_PORT : portParam);
```
chibenwa: Idem this could be handled with an Optional...

mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
```diff
@@ -363,8 +363,8 @@ public void setPort(int port) {
      * Initializer for property maxPings.
      */
     protected void initMaxPings() {
-        String maxPingsParam = getInitParameter("maxPings");
-        setMaxPings((maxPingsParam == null) ? DEFAULT_MAX_PINGS : Integer.parseInt(maxPingsParam));
+        Integer maxPingsParam = getInitParameter("maxPings", Integer.class);
+        setMaxPings((maxPingsParam == null) ? DEFAULT_MAX_PINGS : maxPingsParam);
```
chibenwa: If we are changing this code, then relying on Optional could be a good move.

mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
```diff
@@ -394,8 +394,8 @@ public void setMaxPings(int maxPings) {
      * Initializer for property pingIntervalMilli.
      */
     protected void initPingIntervalMilli() {
-        String pingIntervalMilliParam = getInitParameter("pingIntervalMilli");
-        setPingIntervalMilli((pingIntervalMilliParam == null) ? DEFAULT_PING_INTERVAL_MILLI : Integer.parseInt(pingIntervalMilliParam));
+        Integer pingIntervalMilliParam = getInitParameter("pingIntervalMilli", Integer.class);
+        setPingIntervalMilli((pingIntervalMilliParam == null) ? DEFAULT_PING_INTERVAL_MILLI : pingIntervalMilliParam);
```
chibenwa: If we are changing this code, then relying on Optional could be a good move.

mailet/standard/src/main/java/org/apache/james/transport/mailets/ClamAVScan.java
```diff
@@ -425,8 +425,8 @@ public void setPingIntervalMilli(int pingIntervalMilli) {
      * Initializer for property streamBufferSize.
      */
     protected void initStreamBufferSize() {
-        String streamBufferSizeParam = getInitParameter("streamBufferSize");
-        setStreamBufferSize((streamBufferSizeParam == null) ? DEFAULT_STREAM_BUFFER_SIZE : Integer.parseInt(streamBufferSizeParam));
+        Integer streamBufferSizeParam = getInitParameter("streamBufferSize", Integer.class);
+        setStreamBufferSize((streamBufferSizeParam == null) ? DEFAULT_STREAM_BUFFER_SIZE : streamBufferSizeParam);
```
chibenwa: If we are changing this code, then relying on Optional could be a good move.

mailet/standard/src/main/java/org/apache/james/transport/mailets/ClassifyBounce.java
```diff
@@ -54,7 +54,7 @@
 
     @Override
     public void init() throws MessagingException {
-        headerName = getInitParameter("headerName");
+        headerName = getInitParameter("headerName", String.class);
```
chibenwa: Isn't `headerName = getInitParameter("headerName");` enough?

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -18,19 +18,23 @@
...
 
 
-
 package org.apache.james.transport.mailets;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
-import java.util.Collections;
+import java.util.*;
```
chibenwa: This wilcard import will be rejected by our checkrule

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -66,16 +76,21 @@ public LogMessage() {
...
+            logLevel = getInitParameter("level", LogLevel.class, LogLevel.INFO);
+            specificHeaderNames = getInitParameter("specificHeaders", List.class, Collections.emptyList());
+            specificAttributeNames = getInitParameter("specificAttributes", List.class, Collections.emptyList());
+            headers = getInitParameter("headers", false);
+            body = getInitParameter("body", false);
             passThrough = getInitParameter("passThrough", true);
-            headers = getInitParameter("headers", true);
-            body = getInitParameter("body", true);
-            bodyMax = (getInitParameter("maxBody") == null) ? 0 : Integer.parseInt(getInitParameter("maxBody"));
+            bodyMax = (getInitParameter("maxBody", Integer.class) == null) ? 0 : getInitParameter("maxBody", Integer.class);
```
chibenwa: Can we use Optional ?

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -52,11 +56,17 @@
...
-    private String comment = null;
+    private boolean headers;
+    private boolean body;
+    private int bodyMax;
+    private String comment;
+    private List<String> specificHeaderNames;
+    private List<String> specificAttributeNames;
+
+    //    private StringBuffer logBuffer = new StringBuffer();
+    private ThreadLocal<StringBuilder> logBuffer = new ThreadLocal<>();
```
chibenwa: Or.... I am really not confortable with statefull mailet.

Maybe a better way of seeing things is that each `log*` method return a String (representing logParts) that gets then concatenated together & logged.

Mutability makes reasoning about programming hard. We tend to prefer predictable pure function...

mailet/standard/src/main/java/org/apache/james/transport/mailets/ReplaceContent.java
```diff
@@ -114,7 +114,7 @@ public void init() throws MailetException {
     }
 
     private boolean isDebug() {
-        return Integer.valueOf(getInitParameter("debug", "0")) == 1;
+        return getInitParameter("debug", false);
```
chibenwa: That, is an API change. You will break peoples ReplaceContent debug configuration as the `1` value (which I agree is not explicit at all) will be broken.

mailet/standard/src/main/java/org/apache/james/transport/mailets/UseHeaderRecipients.java
```diff
@@ -84,7 +84,7 @@
      */
     @Override
     public void init() {
-        isDebug = (getInitParameter("debug") == null) ? false : Boolean.valueOf(getInitParameter("debug"));
+        isDebug = (getInitParameter("debug", Boolean.class) == null) ? false : getInitParameter("debug", Boolean.class);
```
chibenwa: If we modify this changeset, then we could give a try to  using Optional here

mailet/standard/src/main/java/org/apache/james/transport/mailets/WrapText.java
```diff
@@ -48,7 +48,7 @@ public String getMailetInfo() {
     @Override
     public void init() throws MailetException {
         optionFlowedDelsp = getBooleanParameter(getInitParameter(PARAMETER_NAME_FLOWED_DELSP), optionFlowedDelsp);
-        optionWidth = Integer.parseInt(getInitParameter(PARAMETER_NAME_WIDTH, "" + optionWidth));
+        optionWidth = getInitParameter(PARAMETER_NAME_WIDTH,Integer.class, optionWidth);
```
chibenwa: You miss a space before `,Integer.class`

mailet/standard/src/test/java/org/apache/james/transport/mailets/LogMessageTest.java
```diff
@@ -31,16 +32,22 @@
 import javax.mail.MessagingException;
 import javax.mail.internet.MimeMessage;
 
+import org.apache.james.core.MailAddress;
 import org.apache.james.core.builder.MimeMessageBuilder;
 import org.apache.james.util.MimeMessageUtil;
-import org.apache.mailet.Mail;
+import org.apache.mailet.*;
```
chibenwa: This WILCARD will brake MVN compilation

mailet/test/src/main/java/org/apache/mailet/base/test/FakeMailetConfig.java
```diff
@@ -76,22 +74,33 @@ public FakeMailetConfig build() {
...
     }
 
+    @Override
+    public <T> T getInitParameter(String name, Class<T> clazz) {
+        Object o = properties.get(name);
+        if (o != null) {
+            if (clazz.isInstance(o)) {
+                return (T) properties.get(name);
+            } else throw new ClassCastException(name + " property is not of type " + clazz.getCanonicalName());
+        } else return null;
```
chibenwa: You need an explicit `{}` block

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -100,25 +131,55 @@ public void service(Mail mail) {
         }
     }
 
+    private void logAttribute(Mail mail) {
+        if (!specificAttributeNames.isEmpty()) {
+            logBuffer.get().append('\n');
+            for (String name : specificAttributeNames) {
+                AttributeName attributeName = AttributeName.of(name);
+                Optional<Attribute> attribute = mail.getAttribute(attributeName);
+                if (attribute.isPresent()) {
```
chibenwa: This condition defeat the use of Optional, which should be used as a monad...

```
logBuffer.get().append(name + ": " + attribute.map(Attribute::getValue).map(AttributeValue::value) + "\n");
```

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -100,25 +131,55 @@ public void service(Mail mail) {
...
+            logBuffer.get().append('\n');
+            MimeMessage message = mail.getMessage();
             for (String header : Collections.list(message.getAllHeaderLines())) {
-                logger.info(header + "\n");
+                logBuffer.get().append(header + '\n');
+            }
+
+            if (!specificHeaderNames.isEmpty()) {
+                for (MailAddress recipient : mail.getRecipients()) {
+                    List<PerRecipientHeaders.Header> headers = mail.getPerRecipientSpecificHeaders()
```
chibenwa: Can we extect line 165 to 172 in a dedicated `logSpecificHeadersFor(Mail mail, MailAddress recipient)` method?

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -100,25 +131,55 @@ public void service(Mail mail) {
...
+                            .getHeadersForRecipient(recipient).stream()
+                            .filter(header -> specificHeaderNames.contains(header.getName()))
+                            .collect(Collectors.toList());
+                    logBuffer.get().append("Recipient " + recipient.asPrettyString() + "'s headers are: \n");
+                    for (PerRecipientHeaders.Header header : headers) {
+                        logBuffer.get().append(header.getName() + ": " + header.getValue() + '\n');
```
chibenwa: This, as all the for loops you introduced in this class could easily be replaced by a foreach.

chibenwa:
Hi @ReturnHttp402 

I did an in depth review of the changeset you proposes.

ReturnHttp402:
Hi @chibenwa,
Thank you for your reviews and instructions, I followed your comments and made some changes.

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -23,18 +23,17 @@
...
 import javax.mail.MessagingException;
 import javax.mail.internet.MimeMessage;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.james.core.MailAddress;
-import org.apache.mailet.Attribute;
-import org.apache.mailet.AttributeName;
-import org.apache.mailet.Mail;
-import org.apache.mailet.PerRecipientHeaders;
+import org.apache.mailet.*;
```
chibenwa: You introduced another wilcard here. Maybe you can change your IDE settings?

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -101,26 +97,25 @@ public String getMailetInfo() {
 
     @Override
     public void service(Mail mail) {
-        logComment();
+        String logs = "";
```
chibenwa: Or we can use a stringBuilder here that we build at the end of the request to avoid variable re-allocation?

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -131,57 +126,45 @@ public void service(Mail mail) {
...
-        if (comment != null) {
-            logBuffer.get().append(comment);
-        }
+    private String logComment() {
+        return Optional.ofNullable(comment).orElse("");
     }
 
-    private void logHeaders(Mail mail) throws MessagingException {
+    private String logHeaders(Mail mail) throws MessagingException {
+        String headersLog = "";
```
chibenwa: Idem, we need to avoid variable re-allocation.

mailet/standard/src/test/java/org/apache/james/transport/mailets/LogMessageTest.java
```diff
@@ -239,7 +243,7 @@ void serviceShouldNotLogHeadersWhenFalse() throws Exception {
         mailet.service(createMail());
 
         verify(logger).isInfoEnabled();
-        verify(logger).info("This is a fake mail");
+        verify(logger).info("\nThis is a fake mail");
```
chibenwa: This `\\n` looks bad, no? Maybe we should avoid it?

chibenwa:
Many thanks for your work ^^

ReturnHttp402:
Hi @chibenwa I made some changes and squash my commits. Many thanks for your patient instructions😀

ReturnHttp402:
@chibenwa Your comments are always very helpful and informative, I've learned a lot😊.Thanks!

ReturnHttp402:
@chibenwa Hi, I've made some changes 6 days ago, can you help to check? Maybe we can introduce the automatic CI email alert when new changes are pushed. 😁

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -86,12 +103,27 @@ public String getMailetInfo() {
...
-        logger.info("Logging mail {}", mail.getName());
-        logComment();
+        StringBuilder logs = new StringBuilder();
+        logs.append(logComment());
         try {
-            MimeMessage message = mail.getMessage();
-            logHeaders(message);
-            logBody(message);
+            logs.append(logHeaders(mail) + logBody(mail) + logAttribute(mail));
+            switch (logLevel) {
```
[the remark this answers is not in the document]
ReturnHttp402: 👍

chibenwa:
@ReturnHttp402 sure I will. Sorry it seems that I missed them.

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -129,9 +165,33 @@ private int lengthToLog(MimeMessage message) throws MessagingException {
...
+        return headersLogs.toString();
+    }
+
+    private LogWriter selectLogWriter(LogLevel logLevel) {
+        switch (logLevel) {
+            case INFO:
+                return (logger, message) -> logger.info(message);
+            case WARN:
+                return (logger, message) -> logger.warn(message);
+            case DEBUG:
```
chibenwa: Maybe we can put DEBUG before INFO to keep the implicit order

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogLevel.java
```diff
@@ -0,0 +1,27 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+package org.apache.james.transport.mailets;
+
+
+public enum LogLevel {
+    DEBUG,
```
chibenwa: Thinking about this we can enhance this rnum a lot by adding the LogWritter as a field:

```
public enum LogLevel {
    Debug((logger, message) -> logger.debug(message)),
    ...;

    private final LogWritter logWritter;

    // Constructor & getter
}
```

That way we can no more need LogMessage::selectLogWriter

mailet/standard/src/main/java/org/apache/james/transport/mailets/LogMessage.java
```diff
@@ -66,16 +79,20 @@ public LogMessage() {
         this(LOGGER);
     }
 
+
     @Override
     public void init() {
         try {
+            logLevel = getInitParameter("level", LogLevel.class, LogLevel.INFO);
```
chibenwa: I am amazed and very glad at the same time that this works!

chibenwa:
For your information, I opened this PR for unit-testing & integration testing of this PR https://github.com/linagora/james-project/pull/2092

chibenwa:
Hi.

The CI reported some checkstyle violation: could you have a look?

`[72cb939ae6d1174914f555ec461788ec2859135e] [ERROR] src/main/java/org/apache/mailet/MailetConfig.java:[24] (imports) CustomImportOrder: 'java.util.Iterator' should be separated from previous import group by one line.`

ReturnHttp402:
> Hi.
> 
> The CI reported some checkstyle violation: could you have a look?
> 
> `[72cb939ae6d1174914f555ec461788ec2859135e] [ERROR] src/main/java/org/apache/mailet/MailetConfig.java:[24] (imports) CustomImportOrder: 'java.util.Iterator' should be separated from previous import group by one line.`

I fixed this checkstyle error in the newly pushed code.

ReturnHttp402:
@chibenwa Hi, I use an abstract method `log()` in the enum `LogLevel.class` and not introduce a new constructor and private field `LogWriter` to replace the `LogMessage::selectLogWriter` method, please have a check to see if it's okay to do in this way. Thanks!

chibenwa:

chibenwa:
No answer, closing.

Reopen if relevant.

PR #148, opened 2019-02-03
JAMES-2657: SSLEngine and SslHandler unaware of remote ip and port

randymo:
add remoteaddress ip and port, if available, to SSLEngine and SslHandler

if the channel is connected, ie has a remote connection, I add it to the SSLEngine before passing it to the SslHandler.

We added this logic to our jamse SMTP server to be able to provide different certs based on ip.

### DIFF
--- a/protocols/netty/src/main/java/org/apache/james/protocols/netty/SslEngineUtil.java
+++ b/protocols/netty/src/main/java/org/apache/james/protocols/netty/SslEngineUtil.java
@@ -1,0 +1,103 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.protocols.netty;
+
+import org.jboss.netty.channel.Channel;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+import java.net.InetSocketAddress;
+
+public enum SslEngineUtil {
+    INSTANCE;
+
+    public enum SslEngineUtilMode {
+        NONE,
+        REMOTE_ONLY,
+        LOCAL_ONLY,
+        REMOTE_LOCAL,
+        LOCAL_REMOTE
+    }
+
+    //default is NONE, so behavior is same as it was if users do nothing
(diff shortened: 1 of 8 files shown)


mbaechler:
Hi @randymo  and thank you for this contribution.
I see that you also opened a JIRA about that and that's great.
To be honest, I don't understand the problem you described, could you elaborate on behavior you need ?
Talking about this PR, I have two concerns about it:
* it's quite a lot of code duplication, it could probably be solved by abstracting a bit the SSL API.
* as it doesn't come with tests, we can't guarantee it won't break in the future.

I know that tests are hard, even more when we speak about SSL but maybe writing a couple of integration tests would help for both the code maintenance and the understanding of your problem. WDYT ?

randymo:
@mbaechler I can do that, What module and package should I create the common ssl engine logic wrapping class in?

The problem is, we serve more domains than could be added to a single cert so we have multi certs for each domain we provide mail for. We have added a keymanager, but in order for that to work the sslEngine it is passed needs have knowledge of where it came from otherwise we would always use the default. Hope this clears up the problem.

mbaechler:
Ok, I understand your problem (and it's an interesting one, does that mean you have a lot of users and domains on a single James instance ?).
You can probably put the common code into protocols/netty, WDYT ?

randymo:
will do, and yes we have between 7-10 different domains handled by a single(technically 2, 1 for starttls and 1 for always ssl) instance of james depending on the geolocation of the data center it is in. Each geolocation also has a different map of ip to cert because of this.

protocols/netty/src/main/java/org/apache/james/protocols/netty/SslEngineUtil.java
```diff
@@ -0,0 +1,103 @@
...
+package org.apache.james.protocols.netty;
+
+import org.jboss.netty.channel.Channel;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+import java.net.InetSocketAddress;
+
+public enum SslEngineUtil {
+    INSTANCE;
```
mbaechler: We usually use Singleton only for immutable things (and it looks like the class is mutable here).
What I suggest is to declare 5 static factories like:
```java
public static SslEngineUtil remoteOnly();
```
First, it allows to have a single instance if we think it worth limiting the number of instance.
Second, all instances will share the same type. We will be able to pass it to our code as a dependency and choose the implementation in the right place.
Third, there's no global shared mutable anymore (and we would be happy about that).

WDYT ?

If you need a hand, I can try to write the solution I have in mind tomorrow, just let me know.

protocols/netty/src/test/java/org/apache/james/protocols/netty/SslEngineUtilTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+import org.jboss.netty.channel.Channel;
+import org.junit.Assert;
+import org.junit.Test;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+import java.net.InetSocketAddress;
+
+public class SslEngineUtilTest {
```
mbaechler: I would really like a function test about that.

protocols/netty/src/main/java/org/apache/james/protocols/netty/SslEngineUtil.java
randymo: Yeah I'd like to see what you have in mind so I can better understand the coding practices you guys use.

ieugen:
The source is gone and this blocks our CI from scanning our repo.
I've added the patch here and to the issue https://issues.apache.org/jira/browse/JAMES-2657 
[148.patch.txt](https://github.com/apache/james-project/files/4868834/148.patch.txt) 

Thank you for your contribution @randymo . 
Sorry for not merging this sooner. 
I hope I'll have some time to take a look at this in the future.
I'm going to close it for now, please re-open if you like.

PR #152, opened 2019-03-22
James 2694 - [Export] Restore API

remk:
Just the API for validation and advices.

### DIFF
--- a/mailbox/backup/src/main/java/org/apache/james/mailbox/backup/MailAccountZipIterator.java
+++ b/mailbox/backup/src/main/java/org/apache/james/mailbox/backup/MailAccountZipIterator.java
@@ -1,0 +1,47 @@
+package org.apache.james.mailbox.backup;
+
+import java.io.IOException;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import org.apache.commons.lang3.NotImplementedException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class MailAccountZipIterator implements MailArchiveIterator {
+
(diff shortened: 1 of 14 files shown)


remk:
Before starting the implementation, i would like to define the API.
I made a proposition in this PR.
My goal is to  untangle the restore api to it's zip implementation.
What do you think of it ? Advices and suggestions are more than welcomed !

mailbox/backup/src/main/java/org/apache/james/mailbox/backup/MailArchiveEntry.java
```diff
@@ -0,0 +1,7 @@
+package org.apache.james.mailbox.backup;
+
+/**
+ * Represents an element in an archive (Mailbox, MailboxAnnotation, Message)
```
chibenwa: **instanceOf** expected here to know the entity?

Or is there a plan to have a generic MailArchiveEntry that knows how to restore itself?

mailbox/backup/src/main/java/org/apache/james/mailbox/backup/MailAccountZipIterator.java
```diff
@@ -0,0 +1,47 @@
...
+        MailArchiveEntry mailArchiveEntry = from(next);
+        try {
+            next = zipInputStream.getNextEntry();
+        } catch (IOException e) {
+            LOGGER.error("Error during iterating on mail account zip", e);
+        }
+        return mailArchiveEntry;
+    }
+
+    private MailArchiveEntry from(ZipEntry zipEntry) {
```
chibenwa: The mailbox entry hold mbx + annotation that from below comments I under are separated MailArchiveEntry. If so we might need a Stream<MailArchiveEntry> here.

PR #154, opened 2019-04-05
JAMES-2725 Remove length from Blob API

aduprat:
This means that the S3 implementation should implement its own way to put blobs without knowing the length


server/blob/blob-objectstorage/src/main/java/org/apache/james/blob/objectstorage/ObjectStorageBlobsDAO.java
```diff
@@ -53,13 +54,17 @@
 
     private final ContainerName containerName;
     private final org.jclouds.blobstore.BlobStore blobStore;
+    private final Function<Blob, String> putBlob;
     private final PayloadCodec payloadCodec;
 
     ObjectStorageBlobsDAO(ContainerName containerName, BlobId.Factory blobIdFactory,
-                          org.jclouds.blobstore.BlobStore blobStore, PayloadCodec payloadCodec) {
+                          org.jclouds.blobstore.BlobStore blobStore,
+                          Function<Blob, String> putBlob,
```
blackheaven: Use `@FunctionInterface`

aduprat:
https://github.com/linagora/james-project/pull/2298

PR #161, opened 2019-07-15
JAMES-2830 update changelog to include effective DKIM support

remk:


remk:
Already included in other PR since then.

PR #166, opened 2019-10-02
JAMES-2909 Use Architecture Decision Record

blackheaven:


src/adr/references/documenting-architecture-decisions.md
```diff
@@ -0,0 +1,160 @@
+# [Documenting Architecture Decisions](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions)
+
+## Context
+
+Architecture for agile projects has to be described and defined
```
chibenwa: Maybe we can say that its also true for community projects...

src/adr/references/documenting-architecture-decisions.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Context** This section describes the forces at play, including
+technological, political, social, and project local. These forces are
+probably in tension, and should be called out as such. The language in
+this section is value-neutral. It is simply describing facts.
+
+**Decision** This section describes our response to these forces. It is
+stated in full sentences, with active voice. "We will ..."
+
+**Status** A decision may be "proposed" if the project stakeholders
```
chibenwa: Who os the project stackholder?

src/adr/references/documenting-architecture-decisions.md
blackheaven: that's a copy and paste of the original article

chibenwa: Dont we need to adapt it?

blackheaven: if we want to keep it as a reference, no.

src/adr/references/documenting-architecture-decisions.md
```diff
@@ -0,0 +1,160 @@
+# [Documenting Architecture Decisions](http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions)
```
rouazana: Licence? Or maybe we can simply link to archive.org if we are afraid of this article to be lost? https://web.archive.org/web/20190824074401/http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions

blackheaven:
> I still miss something explaining me how to write an ADR

it's in the associated link.

I can add the formalism if you want, but a complete how to would be overkill IMO

src/adr/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,23 @@
...
+
+In order to be more community-oriented, we should adopt a process to have a structured way to have open architectural decisions.
+
+Using an Architectural Decision Records-based process as a support of discussion on the developers mailing-lists.
+
+## Decision
+
+We will use Architecture Decision Records, as [described by Michael Nygard](https://web.archive.org/web/20190824074401/http://thinkrelevance.com/blog/2011/11/15/documenting-architecture-decisions).
+
+Each ADR will be discussed on the Apache James' developers mailing-list before being accepted.
```
chibenwa: Can we state:

```
Following https://community.apache.org/committers/decisionMaking.html , we provide the following possible status, with their associated meaning:

 - `Proposed`: The decision is being discussed on the mailing list.
 - `Accepted (lazy consensus)` : the architecture decision was proposed on the mailing list, and a consensus emerged from people involved in the discussion on the mailing list.
 - `Accepted (voted)` : the architecture undergo a voting process.
 - `Rejected` : Consensus builded up against that proposal.
```

Also:

```
We should provide in a mutable `Reference` section links to related JIRA meta-ticket (not necessarily to all related sub  tickets) as well as a link to the mail archive discussion thread.

JIRA tickets implementing that architecture decision should also link the related Architecture Decision Record.
```

blackheaven: We lack also `Conditionally accepted`, `Accepted and withdrawed`, etc.

I'm in favor of being a bit more fluid about that.

rouazana: I add the possible status

rouazana:
see  #169.

PR #167, opened 2019-10-08
JamesMessageHook is not working

mwzero:
I'm trying to use only SMTP features directly from Java application. 

I've defined:

SMTPProtocolHandlerChain chain = new SMTPProtocolHandlerChain(new NoopMetricFactory());
chain.addAll(0, handlers);
chain.wireExtensibleHandlers();

Protocol protocol = new SMTPProtocol(chain, smtpConfiguration);
server = new NettyServer.Factory(new HashedWheelTimer())
       .protocol(protocol)
       .build();
server.setListenAddresses(new InetSocketAddress(properties.getPort()));
server.setTimeout(properties.getTimeout());
        server.bind();

one of the handler is JamesMessageQueue because i need the MimeMessage. 

It seems the hook is never intercepted.


chibenwa:
Hi @mwzero ,

Thanks for the report!

This problem had already been reported in https://issues.apache.org/jira/projects/JAMES/issues/JAMES-2902

We provided quickly some way of writing integration tests for SMTP hooks, and are looking for contributors to:
 - Write a similar integration tests for JamesMessageHook
 - Identify and fix the underlying issue.

Would you be interested to do that?

Also in your pull request, I don't see any code written by you. In the future when you encounter an issue with Apache James, please report it on the JIRA.

Best regards,

Benoit Tellier

chibenwa:
@ieugen should we consider closing this PR, given that we opened a JIRA for it and it contains no code contribution relevant for the issue?

PR #170, opened 2019-10-17
Various ADR regarding JMAP and blobStore performance enhancements

chibenwa:
The goal of these architecture decisions is to ensure a GetMessageList call does not end up calling the blob store.

### DIFF
--- a/src/adr/0012-jmap-partial-reads.md
+++ b/src/adr/0012-jmap-partial-reads.md
@@ -10,7 +10,7 @@ Accepted (lazy consensus)
 
 JMAP core RFC8620 requires that the server responds only properties requested by the client.
 
-James currently returns all of the properties regardless of their cost, and if it had been asked by the client.
+James currently computes all of the properties regardless of their cost, and if it had been asked by the client.
 
 Clearly we can save some latencies and resources by avoiding reading/computing expensive properties that had not been explicitly requested by the client.
 
@@ -18,11 +18,12 @@ This is furthermore an opportunity to conform to /get behavior of released RFC-8
 
 ## Decision
 
-Introduce two projections for JMAP messages:
+Introduce two new datastructures representing JMAP messages:
  - One with only metadata
  - One with metadata + headers
 
-Given the properties requested by the client, the most appropriate message projection will be returned.
+Given the properties requested by the client, the most appropriate message datastructure will be returned, on top of 
+existing message storage APIs that should remain unchanged.
 
 Some performance tests will be run in order to evaluate the improvements.
 
--- a/src/adr/0012-jmap-partial-reads.md
+++ b/src/adr/0012-jmap-partial-reads.md
@@ -1,0 +1,38 @@
+# 12. Projections for JMAP Messages
+
+Date: 2019-10-09
+
+## Status
+
+Accepted (lazy consensus)
+
+## Context
+
+JMAP core RFC8620 requires that the server responds only properties requested by the client.
+
+James currently returns all of the properties regardless of their cost, and if it had been asked by the client.
+
+Clearly we can save some latencies and resources by avoiding reading/computing expensive properties that had not been explicitly requested by the client.
+
+This is furthermore an opportunity to conform to /get behavior of released RFC-8620
+
+## Decision
+
+Introduce two projections for JMAP messages:
+ - One with only metadata
+ - One with metadata + headers
+
+Given the properties requested by the client, the most appropriate message projection will be returned.
+
+Some performance tests will be run in order to evaluate the improvements.
+
+## Consequences
+
+Fields that were previously returned in Messages might not be returned anymore if not explicitly asked by the client.
+
+In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'accepted but inconclusive'.
+
+## References
+
+ - /get method: https://tools.ietf.org/html/rfc8620#section-5.1
+ - [JIRA](https://issues.apache.org/jira/browse/JAMES-2919)
--- a/src/adr/0012-jmap-partial-reads.md
+++ b/src/adr/0012-jmap-partial-reads.md
@@ -4,7 +4,9 @@ Date: 2019-10-09
 
 ## Status
 
-Accepted (lazy consensus)
+Proposed
+
+Adoption needs to be backed by some performance tests.
 
 ## Context
 
@@ -14,22 +16,21 @@ James currently computes all of the properties regardless of their cost, and if 
 
 Clearly we can save some latencies and resources by avoiding reading/computing expensive properties that had not been explicitly requested by the client.
 
-This is furthermore an opportunity to conform to /get behavior of released RFC-8620
-
 ## Decision
 
 Introduce two new datastructures representing JMAP messages:
  - One with only metadata
  - One with metadata + headers
 
-Given the properties requested by the client, the most appropriate message datastructure will be returned, on top of 
+Given the properties requested by the client, the most appropriate message datastructure will be computed, on top of 
 existing message storage APIs that should remain unchanged.
 
 Some performance tests will be run in order to evaluate the improvements.
 
 ## Consequences
 
-Fields that were previously returned in Messages might not be returned anymore if not explicitly asked by the client.
+GetMessages with a limited set of requested properties will no longer result necessarily in full database message read. We
+thus expect a significant improvement, for instance when only metadata are requested.
 
 In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'rejected'.
 
--- a/src/adr/0012-jmap-partial-reads.md
+++ b/src/adr/0012-jmap-partial-reads.md
@@ -31,7 +31,7 @@ Some performance tests will be run in order to evaluate the improvements.
 
 Fields that were previously returned in Messages might not be returned anymore if not explicitly asked by the client.
 
-In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'accepted but inconclusive'.
+In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'rejected'.
 
 ## References
 
--- a/src/adr/0013-precompute-jmap-preview.md
+++ b/src/adr/0013-precompute-jmap-preview.md
@@ -4,7 +4,9 @@ Date: 2019-10-09
 
 ## Status
 
-Accepted (lazy consensus)
+Proposed
+
+Adoption needs to be backed by some performance tests.
 
 ## Context
 
@@ -26,15 +28,14 @@ When the preview is precomputed then for these messages we can consider the "pre
 
 When the preview is not precomputed then we should compute the preview for these messages, and save the result for later.
 
-We should provide a webAdmin task allowing to rebuild the projection.
+We should provide a webAdmin task allowing to rebuild the projection. The computing and storing in MessagePreviewStore 
+is idempotent and that the task can be run in live without any concurrency problem.
 
 Some performance tests will be run in order to evaluate the improvements.
 
 ## Consequences
 
 We expect a huge performance enhancement for JMAP clients relying on preview for listing mails.
-
-Messages whose preview is not yet computed will still require live preview computation as a fallback mechanism.
 
 In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'rejected'.
 
--- a/src/adr/0013-precompute-jmap-preview.md
+++ b/src/adr/0013-precompute-jmap-preview.md
@@ -36,7 +36,7 @@ We expect a huge performance enhancement for JMAP clients relying on preview for
 
 Messages whose preview is not yet computed will still require live preview computation as a fallback mechanism.
 
-In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'accepted but inconclusive'.
+In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'rejected'.
 
 ## References
 
--- a/src/adr/0013-precompute-jmap-preview.md
+++ b/src/adr/0013-precompute-jmap-preview.md
@@ -12,7 +12,7 @@ JMAP messages have a handy preview property displaying the firsts 256 characters
 
 This property is often displayed for message listing in JMAP clients, thus it is queried a lot.
 
-Currently, to get the preview, James retrieves the full message body, parse it using MIME parsers, removes HTML and keep meaningful text. This process is expensive, especially for clients relying on polling.
+Currently, to get the preview, James retrieves the full message body, parse it using MIME parsers, removes HTML and keep meaningful text.
 
 ## Decision
 
@@ -23,6 +23,8 @@ A MailboxListener will compute the preview and store it in a MessagePreviewStore
 We should have a Cassandra and memory implementation.
 
 When the preview is precomputed then for these messages we can consider the "preview" property as a metadata.
+
+When the preview is not precomputed then we should compute the preview for these messages, and save the result for later.
 
 We should provide a webAdmin task allowing to rebuild the projection.
 
--- a/src/adr/0013-precompute-jmap-preview.md
+++ b/src/adr/0013-precompute-jmap-preview.md
@@ -29,7 +29,7 @@ When the preview is precomputed then for these messages we can consider the "pre
 When the preview is not precomputed then we should compute the preview for these messages, and save the result for later.
 
 We should provide a webAdmin task allowing to rebuild the projection. The computing and storing in MessagePreviewStore 
-is idempotent and that the task can be run in live without any concurrency problem.
+is idempotent and the task can be run in live without any concurrency problem.
 
 Some performance tests will be run in order to evaluate the improvements.
 
--- a/src/adr/0013-precompute-jmap-preview.md
+++ b/src/adr/0013-precompute-jmap-preview.md
@@ -1,0 +1,45 @@
+# 13. Precompute JMAP Email preview
+
+Date: 2019-10-09
+
+## Status
+
+Accepted (lazy consensus)
+
+## Context
+
+JMAP messages have a handy preview property displaying the firsts 256 characters of meaningful test of a message.
+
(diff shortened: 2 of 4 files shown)


src/adr/0013-precompute-jmap-preview.md
```diff
@@ -0,0 +1,45 @@
...
+
+Accepted (lazy consensus)
+
+## Context
+
+JMAP messages have a handy preview property displaying the firsts 256 characters of meaningful test of a message.
+
+This property is often displayed for message listing in JMAP clients, thus it is queried a lot.
+
+Currently, to get the preview, James retrieves the full message body, parse it using MIME persers, removes HTML and keep meaningful text. This process is expensive, especially for clients relying on polling.
```
remk: s/persers/parsers

src/adr/0012-jmap-partial-reads.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Status
+
+Accepted (lazy consensus)
+
+## Context
+
+JMAP core RFC8620 requires that the server responds only properties requested by the client.
+
+James currently returns all of the properties regardless of their cost, and if it had been asked by the client.
```
mbaechler: this statement is probably false

src/adr/0012-jmap-partial-reads.md
```diff
@@ -0,0 +1,38 @@
...
+
+Clearly we can save some latencies and resources by avoiding reading/computing expensive properties that had not been explicitly requested by the client.
+
+This is furthermore an opportunity to conform to /get behavior of released RFC-8620
+
+## Decision
+
+Introduce two projections for JMAP messages:
```
mbaechler: projection refers to a deduplication of data in a store. After some discussion, you actually meant "datastructures".

src/adr/0012-jmap-partial-reads.md
```diff
@@ -0,0 +1,38 @@
...
+ - One with only metadata
+ - One with metadata + headers
+
+Given the properties requested by the client, the most appropriate message projection will be returned.
+
+Some performance tests will be run in order to evaluate the improvements.
+
+## Consequences
```
mbaechler: what are the consequences in term of storage? How are the actual projections updated? There's a lot of missing pieces IMO

src/adr/0013-precompute-jmap-preview.md
mbaechler: polling should not load the preview, at least with the ADR #12 implemented

src/adr/0013-precompute-jmap-preview.md
```diff
@@ -0,0 +1,45 @@
...
+
+We should provide a webAdmin task allowing to rebuild the projection.
+
+Some performance tests will be run in order to evaluate the improvements.
+
+## Consequences
+
+We expect a huge performance enhancement for JMAP clients relying on preview for listing mails.
+
+Messages whose preview is not yet computed will still require live preview computation as a fallback mechanism.
```
mbaechler: do you intend to store the result of this live preview?

src/adr/0012-jmap-partial-reads.md
```diff
@@ -0,0 +1,38 @@
...
+
+Fields that were previously returned in Messages might not be returned anymore if not explicitly asked by the client.
+
+In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'accepted but inconclusive'.
```
rouazana: did you add this status?

src/adr/0012-jmap-partial-reads.md
chibenwa: s/returns/computes/

src/adr/0014-blobstore-storage-policies.md
```diff
@@ -0,0 +1,52 @@
...
+
+As an example:
+
+ - Mailbox message headers are expected to be small and frequently accessed
+ - Mailbox message body are expected to have sizes ranging from small to big but are unfrequently accessed
+ - DeletedMessageVault message headers are expected to be small and unfrequently accessed
+
+Also, the capabilities of the various implementations of BlobStore have different strengths:
+
+ - CassandraBlobStore is efficient for small blobs and offer low latency. However it is known to be expensive for big blobs. Cassandra storage is expensive.
```
rouazana: offers

src/adr/0013-precompute-jmap-preview.md
chibenwa: IMO yes we should

src/adr/0014-blobstore-storage-policies.md
```diff
@@ -0,0 +1,52 @@
...
+
+Thus, significant performance could be unlocked by using the right blob store for the right blob.
+
+## Decision
+
+Introduce StoragePolicies at the level of the BlobStore API.
+
+The proposed policies includes:
+
+ - SizeBasedStoragePolicy: The blob underlying storage medium will be chosen depending of its size.
```
rouazana: depending on

src/adr/0014-blobstore-storage-policies.md
```diff
@@ -0,0 +1,52 @@
...
+ - LowCostStoragePolicy: The blob is expected to be saved in low cost storage. Access is expected to be unfrequent.
+ - PerformantStoragePolicy: The blob is expected to be saved in performant storage. Access is expected to be unfrequent.
```
rouazana: frequent?

src/adr/0014-blobstore-storage-policies.md
```diff
@@ -0,0 +1,52 @@
...
+
+An HybridBlobStore will be created to choose between Cassandra and ObjectStorage implementations depending on the policies.
+
+DeletedMessageVault, BlobExport & MailRepository will rely on LowCostStoragePolicy. Other BlobStore users will rely on SizeBasedStoragePolicy.
```
rouazana: do we really need to use the policy based blob store for all our blob store uses? I'm not sure LowCostStoragePolicy and PerformantStoragePolicy are really worth it.

src/adr/0015-objectstorage-blobid-list.md
```diff
@@ -0,0 +1,55 @@
...
+
+This usage pattern was also adopted for BlobStore on top of ObjectStorage.
+
+However writing in Object storage:
+ - Takes time
+ - Is billed by most cloud providers
+
+Thus choosing a right strategy to avoid writing blob twice is desirable.
+
+However, ObjectStorage (OpenStack Swift) exist method was not efficient enough to be a real cost and performance saver.
```
rouazana: `exist`

src/adr/0015-objectstorage-blobid-list.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Decision
+
+Rely on a StoredBlobIdsList API to know which blob is persisted or not in object storage. Provide a Cassandra implementation of it. Located in blob-api for convenience, this it not a top level API. It is intended to be used by some blobStore implementations (here only ObjectStorage).
+
+ - When saving a blob with precomputed blobId, we can check the existance of the blob in storage, avoiding possibly the expensive "save".
```
rouazana: existence

src/adr/0015-objectstorage-blobid-list.md
```diff
@@ -0,0 +1,55 @@
...
+ - When saving a blob too big to precompute its blobId, once the blob had been stream using a temporary random blobId, copy operation can be avoided and the temporary blob could be directly removed.
```
rouazana: streamed

src/adr/0014-blobstore-storage-policies.md
chibenwa: LowCostStoragePolicy -> For DeletedMessageVault (mail headers)

chibenwa: PerformantStoragePolicy might not be worth it though...

rouazana: I mean why configuring DeletedMessageVault with a particular blob store policy while we probably ever want to store its data in swift?

src/adr/0012-jmap-partial-reads.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Context
+
+JMAP core RFC8620 requires that the server responds only properties requested by the client.
+
+James currently computes all of the properties regardless of their cost, and if it had been asked by the client.
+
+Clearly we can save some latencies and resources by avoiding reading/computing expensive properties that had not been explicitly requested by the client.
+
+This is furthermore an opportunity to conform to /get behavior of released RFC-8620
```
mbaechler: how?

src/adr/0012-jmap-partial-reads.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Decision
+
+Introduce two new datastructures representing JMAP messages:
+ - One with only metadata
+ - One with metadata + headers
+
+Given the properties requested by the client, the most appropriate message datastructure will be returned, on top of
```
mbaechler: s/returned/computed/ ?

src/adr/0012-jmap-partial-reads.md
```diff
@@ -0,0 +1,39 @@
...
+ - One with metadata + headers
+
+Given the properties requested by the client, the most appropriate message datastructure will be returned, on top of 
+existing message storage APIs that should remain unchanged.
+
+Some performance tests will be run in order to evaluate the improvements.
+
+## Consequences
+
+Fields that were previously returned in Messages might not be returned anymore if not explicitly asked by the client.
```
mbaechler: Is it true? Didn't we have property filtering?

src/adr/0012-jmap-partial-reads.md
```diff
@@ -0,0 +1,39 @@
+# 12. Projections for JMAP Messages
+
+Date: 2019-10-09
+
+## Status
+
+Accepted (lazy consensus)
```
mbaechler: Is the status really Accepted? I mean, unless we have performance data, we can't really decide if it's good or not, right?

src/adr/0013-precompute-jmap-preview.md
```diff
@@ -0,0 +1,47 @@
...
+
+A MailboxListener will compute the preview and store it in a MessagePreviewStore.
+
+We should have a Cassandra and memory implementation.
+
+When the preview is precomputed then for these messages we can consider the "preview" property as a metadata.
+
+When the preview is not precomputed then we should compute the preview for these messages, and save the result for later.
+
+We should provide a webAdmin task allowing to rebuild the projection.
```
mbaechler: I think we can explain that the computing and storing in MessagePreviewStore is idempotent and that the task can be run in live without any concurrency problem.

src/adr/0013-precompute-jmap-preview.md
```diff
@@ -0,0 +1,47 @@
...
+
+Some performance tests will be run in order to evaluate the improvements.
+
+## Consequences
+
+We expect a huge performance enhancement for JMAP clients relying on preview for listing mails.
+
+Messages whose preview is not yet computed will still require live preview computation as a fallback mechanism.
```
mbaechler: looks a diverging repetition of:
> When the preview is not precomputed then we should compute the preview for these messages, and save the result for later.

src/adr/0014-blobstore-storage-policies.md
```diff
@@ -0,0 +1,52 @@
+# 14. Add storage policies for BlobStore
+
+Date: 2019-10-09
+
+## Status
+
+Proposed
+
+Adoption needs to be backed by some performance tests.
```
mbaechler: This wording is much better

src/adr/0014-blobstore-storage-policies.md
mbaechler: because having to declare what we expect in term of storage policy is better than knowing that store X has this specific behavior

src/adr/0014-blobstore-storage-policies.md
```diff
@@ -0,0 +1,52 @@
...
+ - LowCostStoragePolicy: The blob is expected to be saved in low cost storage. Access is expected to be unfrequent.
+ - PerformantStoragePolicy: The blob is expected to be saved in performant storage. Access is expected to be frequent.
+
+An HybridBlobStore will be created to choose between Cassandra and ObjectStorage implementations depending on the policies.
+
+DeletedMessageVault, BlobExport & MailRepository will rely on LowCostStoragePolicy. Other BlobStore users will rely on SizeBasedStoragePolicy.
+
+Some performance tests will be run in order to evaluate the improvements.
+
+## Consequences
```
mbaechler: what about migration?
what about expected space usage in each store?

src/adr/0015-objectstorage-blobid-list.md
```diff
@@ -0,0 +1,55 @@
...
+ - Takes time
+ - Is billed by most cloud providers
+
+Thus choosing a right strategy to avoid writing blob twice is desirable.
+
+However, ObjectStorage (OpenStack Swift) `exist` method was not efficient enough to be a real cost and performance saver.
+
+## Decision
+
+Rely on a StoredBlobIdsList API to know which blob is persisted or not in object storage. Provide a Cassandra implementation of it. Located in blob-api for convenience, this it not a top level API. It is intended to be used by some blobStore implementations (here only ObjectStorage).
```
mbaechler: the dependency graph is not obvious. Could you explain a bit more?

src/adr/0015-objectstorage-blobid-list.md
```diff
@@ -0,0 +1,55 @@
...
+
+ - When saving a blob with precomputed blobId, we can check the existence of the blob in storage, avoiding possibly the expensive "save".
+ - When saving a blob too big to precompute its blobId, once the blob had been streamed using a temporary random blobId, copy operation can be avoided and the temporary blob could be directly removed.
+
+Cassandra is faster doing "write every time" rather than "read before write" so we should not use the stored blob projection for it
```
mbaechler: s/is faster/is probably faster/

src/adr/0015-objectstorage-blobid-list.md
```diff
@@ -0,0 +1,55 @@
...
+Some performance tests will be run in order to evaluate the improvements.
+
+## Consequences
+
+We expect to reduce the amount of writes to the object storage. This is expected to improve:
+ - operational costs on cloud providers
+ - performance improvement
+ - latency reduction under load
+
+Inconsistencies in StoredBlobIdsList will lead to duplicated saved blobs, which is the current behaviour.
```
mbaechler: be careful to put the id into StoredBlobIdsList after storing it to the blobstore otherwise the inconsistency will be a problem.

src/adr/0015-objectstorage-blobid-list.md
```diff
@@ -0,0 +1,55 @@
...
+Rely on a StoredBlobIdsList API to know which blob is persisted or not in object storage. Provide a Cassandra implementation of it. Located in blob-api for convenience, this it not a top level API. It is intended to be used by some blobStore implementations (here only ObjectStorage).
+
+ - When saving a blob with precomputed blobId, we can check the existence of the blob in storage, avoiding possibly the expensive "save".
+ - When saving a blob too big to precompute its blobId, once the blob had been streamed using a temporary random blobId, copy operation can be avoided and the temporary blob could be directly removed.
+
+Cassandra is faster doing "write every time" rather than "read before write" so we should not use the stored blob projection for it
+
+Some performance tests will be run in order to evaluate the improvements.
+
+## Consequences
```
mbaechler: I wonder how it will interact with the blob garbage collection we plan to implement

src/adr/0012-jmap-partial-reads.md
chibenwa: This comment is irrelevant as we already conform to it.

src/adr/0014-blobstore-storage-policies.md
chibenwa: > what about expected space usage in each store?

What do you mean @mbaechler ?

src/adr/0015-objectstorage-blobid-list.md
chibenwa: Good question!

And I have an answer for it: as you plan to add a generation component in the blob id, blobs will be different, and this idList will thus be specific for a  generation. We can thus safely remove entries from it as we know there will be no new blobs generated with the old generation.

However ADR for blob deletion are not presented yet, thus detailing interactions is hard.

Another argument I have: blob deletion acts at the BlobStore level, while this proposed implementation will be implemented at the ObjectStorageBlobDAO level, which should limit conflicts.

src/adr/0014-blobstore-storage-policies.md
mbaechler: I mean we should describe the space usage: is it 20% more cassandra usage? etc
As an admin, it's the information I need

src/adr/0014-blobstore-storage-policies.md
```diff
@@ -0,0 +1,57 @@
...
+
+We expect small frequently accessed blobs to be located in Cassandra, allowing ObjectStorage to be used mainly for large costly blobs.
+
+In case of a less than 5% improvement, the code will not be added to the codebase and the proposal will get the status 'rejected'.
+
+We expect more data to be stored in Cassandra.
```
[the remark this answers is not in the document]
chibenwa: That depends on the data, and proportion of small mails. Here only "production" like data would be relevant IMO

src/adr/0014-blobstore-storage-policies.md
```diff
@@ -35,7 +35,7 @@ The proposed policies include:
  - LowCostStoragePolicy: The blob is expected to be saved in low cost storage. Access is expected to be unfrequent.
  - PerformantStoragePolicy: The blob is expected to be saved in performant storage. Access is expected to be frequent.
 
-An HybridBlobStore will be created to choose between Cassandra and ObjectStorage implementations depending on the policies.
+An HybridBlobStore will be replace current UnionBlobStore and will allow to choose between Cassandra and ObjectStorage implementations depending on the policies.
```
rouazana: s/be replace/replace

chibenwa:
Merged

PR #171, opened 2019-10-17
 Various Architecture Decision Records regarding our ElasticSearch usage

chibenwa:


src/adr/0011-remove-elasticsearch-document-source.md
```diff
@@ -0,0 +1,33 @@
...
+
+Though very handy to have around, the source field does incur storage overhead within the index. 
+
+## Decision
+
+Disable _source for ELasticSearch indexed documents.
+
+## Consequences
+
+We notice a 20% space reduction of data stored on ElasticSearch.
```
rouazana: Can you precise the context? This number seems strange, because it means that we are storing 5x more data on ElasticSearch than the raw data. It's huge!
I guess your emails are full of headers and text only, which is perhaps not representative of a modern usage? What about emails with attachments and HTML?

src/adr/0010-enable-elasticsearch-routing.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Status
+
+Proposed
+
+Additional performance testing is required for adoption.
+
+## Context
+
+Our queries are mostly bounded to a mailbox or an user. We can easily
```
rouazana: a user

src/adr/0010-enable-elasticsearch-routing.md
```diff
@@ -0,0 +1,41 @@
...
+limit the number of ElasticSearch nodes involved in a given query by
+grouping the underlying documents on the same node using a routingKey.
```
rouazana: routing key

src/adr/0011-remove-elasticsearch-document-source.md
```diff
@@ -0,0 +1,33 @@
...
+
+The benefits do not outweigh the costs.
+
+## Context
+
+Though very handy to have around, the source field does incur storage overhead within the index. 
+
+## Decision
+
+Disable _source for ELasticSearch indexed documents.
```
rouazana: did you break the MD formatting with this `_`?

src/adr/0011-remove-elasticsearch-document-source.md
chibenwa: > because it means that we are storing 5x more data on ElasticSearch than the raw data

We discussed that, we store (in space) 5 time the raw document, which is not that surprising given than some fields are indexed twice (for search & for sort), that we rely on nested objects.

There was a confusion between the original MIME document and the indexed document.

chibenwa: `Given a dataset composed of small text/plain messages, we notice a 20% space reduction of data stored on ElasticSearch.`

Details added ;-)

src/adr/0009-disable-elasticsearch-dynamic-mapping.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Consequences
+
+The index needs to be re-created. Document reIndexation is needed.
+
+This solves the aforementionned bugs [1].
+
+Regarding performance:
+ - Default message list performance is unimpacted
+ - We noticed a 4% performance improvment upon indexing throughput
```
Arsnael: s/improvment/improvement

src/adr/0010-enable-elasticsearch-routing.md
```diff
@@ -0,0 +1,41 @@
...
+
+Additional performance testing is required for adoption.
+
+## Context
+
+Our queries are mostly bounded to a mailbox or a user. We can easily
+limit the number of ElasticSearch nodes involved in a given query by
+grouping the underlying documents on the same node using a routing key.
+
+Without routing key, each shard needs to execute the query. The coordinator
```
Arsnael: Without `a` routing key

src/adr/0011-remove-elasticsearch-document-source.md
```diff
@@ -0,0 +1,33 @@
...
+
+The benefits do not outweigh the costs.
+
+## Context
+
+Though very handy to have around, the source field does incur storage overhead within the index. 
+
+## Decision
+
+Disable \_source for ELasticSearch indexed documents.
```
Arsnael: s/ELasticSearch/ElasticSearch

src/adr/0010-enable-elasticsearch-routing.md
```diff
@@ -0,0 +1,41 @@
...
+needs also to be waiting for the slowest shard.
+
+Using the routing key unlocks significant throughput enhancement (proportional
```
mbaechler: I understand the latency thing but I don't get why the routing would enhance scalability

src/adr/0010-enable-elasticsearch-routing.md
```diff
@@ -0,0 +1,41 @@
...
+to the number of shard) and also a possible high percentile latencies enhancement.
+This allows to be more linearly scalable.
+
+## Decision
+
+Enable ElasticSearch routing.
+
+Messages should be indexed by mailbox.
```
mbaechler: that means a Message that is present in several mailboxes is indexed several time?

src/adr/0011-remove-elasticsearch-document-source.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Consequences
+
+Given a dataset composed of small text/plain messages, we notice a 20% space reduction of data stored on ElasticSearch.
+
+However, patch updates can no longer be performed upon flags updates. Upon flag update we need to fully read the mail
```
mbaechler: Unless you choose another indexing structure

What about indexing flags separately from messages?
I could also prevent the multiple indexation of a given message when it is present in several mailboxes, don't you think?
It could also be a good idea to store differently frequently changing data (flags, mailbox) from almost constant data (the mail content).

src/adr/0010-enable-elasticsearch-routing.md
chibenwa: The low level example I provided is not relevant in an ADR.

Please refer to the pull request comment explaining that: https://github.com/linagora/james-project/pull/2757#issuecomment-545366619

src/adr/0010-enable-elasticsearch-routing.md
chibenwa: Yes. And it is already the case.

src/adr/0011-remove-elasticsearch-document-source.md
chibenwa: Parent-child relationships comes with it's own sets of problems. Parent on child needs to live on the same index, thus needs to have similar structure.

Also we need evaluation of the "JOIN" cost upon flags searches/filtering.

Such a proposal impacts _source decision, but is beneficial even outside this context. We can rechallenge the _source debate once your proposal is implemented.

Note please that _source allows admin flexibility like **performing the following operations without downtime**:
 - Increase shards
 - Modifying replication factor
 - Changing analysers (IE allows an admin to configure FR analyser instead of EN analyser)

Amongst others.

If we *ever* consider removing the **_source** field, we significantly harm possible admin maintainance over our work. I find it hard to push it for just a 20% size enhancement. I would really prefer exposing an option in the James configuration to state if we want to use **_source** or not.

src/adr/0011-remove-elasticsearch-document-source.md
```diff
@@ -27,6 +27,11 @@ document.
 Without `_source` field, flags update is two time slower, 99 percentile 4 time slower, and this impact negatively other 
 requests.
 
+Note please that `_source` allows admin flexibility like performing index level changes without downtime, amongst others:
+ - Increase shards
+ - Modifying replication factor
```
mbaechler: are you sure these two things are not doable without `_source`? It sounds strange

chibenwa: I'm sure that ES built in reindexing do not work without `_source`.

To quote https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html :

```
Think before disabling the _source field

Users often disable the _source field without thinking about the consequences, and then live to regret it. If the _source field isn’t available then a number of features are not supported:

-   The update, update_by_query, and reindex APIs.
-   On the fly highlighting.
-   The ability to reindex from one Elasticsearch index to another, either to change mappings or analysis, or to upgrade an index to a new major version.
-   The ability to debug queries or aggregations by viewing the original document used at index time.
 -   Potentially in the future, the ability to repair index corruption automatically.
```

BTW beloved ElasticSearch evangelists quote in this discussion reindex as the only way to change shard&replica count.

mbaechler: > BTW beloved ElasticSearch evangelist quote in this discussion reindex as the only way to change shard&replica count.

The link between reindex and shard+replica was not obvious but OK.

chibenwa: Not my fault.

chibenwa:
Merged

PR #172, opened 2019-10-23
JAMES-2929 JDBCMailRepository broken: SQL and JDBC code out of sync

jtconsol:
As [discussed on Gitter](https://gitter.im/apache/james-project?at=5daee7dca3f0b1784969e164) and described in a [Jira issue](https://issues.apache.org/jira/browse/JAMES-2929), this aligns and fixes the SQL resources only and changes the column type for `per_recepient_headers` to OID (for postgres only), so that it can work on postgres at all. The OID part is the quick workaround described in the related issue [JAMES-2828](https://issues.apache.org/jira/browse/JAMES-2828), for which I am looking into a proper solution.

Got rid of some trailing whitespace along the way.

### DIFF
--- a/dockerfiles/run/spring/destination/conf/sqlResources.xml
+++ b/dockerfiles/run/spring/destination/conf/sqlResources.xml
@@ -1,20 +1,20 @@
 <!--
-  Licensed to the Apache Software Foundation (ASF) under one   
-  or more contributor license agreements.  See the NOTICE file 
-  distributed with this work for additional information        
-  regarding copyright ownership.  The ASF licenses this file   
-  to you under the Apache License, Version 2.0 (the            
-  "License"); you may not use this file except in compliance   
-  with the License.  You may obtain a copy of the License at   
-                                                               
-    http://www.apache.org/licenses/LICENSE-2.0                 
-                                                               
-  Unless required by applicable law or agreed to in writing,   
-  software distributed under the License is distributed on an  
-  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       
-  KIND, either express or implied.  See the License for the    
-  specific language governing permissions and limitations      
-  under the License.                                           
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
  -->
 
 <!--
@@ -497,7 +497,8 @@
             remoteUser varchar (64) NOT NULL,
             remoteHost varchar (255) NOT NULL,
             PRIMARY KEY (localUser, localHost, remoteUser, remoteHost)
-            )    </sql>
+            )
+        </sql>
         <sql name="createWhiteListTable" db="hsqldb">
             CREATE CACHED TABLE ${table} (
             CREATE TABLE whitelist (
@@ -506,7 +507,8 @@
             remoteUser varchar (64) NOT NULL,
             remoteHost varchar (255) NOT NULL,
             PRIMARY KEY (localUser, localHost, remoteUser, remoteHost)
-            )    </sql>
+            )
+        </sql>
         <sql name="createWhiteListTable" db="mysql">
             CREATE TABLE whitelist (
             localUser varchar (64) NOT NULL,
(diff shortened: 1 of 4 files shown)


dockerfiles/run/spring/destination/conf/sqlResources.xml
```diff
@@ -258,458 +326,466 @@
...
             recip varchar (255) NOT NULL,
             count int NOT NULL,
             create_time timestamp NOT NULL,
             PRIMARY KEY (ipaddress,sender,recip)
-            )
-        </sql>
-    </sqlDefs>
+        )
+    </sql>
+</sqlDefs>
```
Arsnael: I don't think the indent of the block was wrong, as it's still wrapped by the `sqlResources` tag. Can you check the indent in this file again please?

jtconsol: The reason for this is that i took `server/app/src/main/resources/sqlResources.xml` as a "master" file, since it seemed like it would make sense - that's why I called this an "alignment" of the SQL resources. Consequently, both files' indentation should be fixed, right?

server/app/src/main/resources/sqlResources.xml
```diff
@@ -251,9 +259,10 @@
             recipients text NOT NULL ,
             remote_host varchar (255) NOT NULL ,
             remote_addr varchar (20) NOT NULL ,
+            per_recipient_headers oid NULL ,
```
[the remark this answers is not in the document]
jtconsol: As noted in the PR description, this is only an intermittant workaround for postgres, the proper type `bytea` returns with [this PR](https://github.com/apache/james-project/pull/173).

server/data/data-jdbc/src/test/resources/sqlResources.xml
```diff
@@ -251,10 +251,10 @@
             recipients text NOT NULL ,
             remote_host varchar (255) NOT NULL ,
             remote_addr varchar (20) NOT NULL ,
+            per_recipient_headers oid NULL ,
```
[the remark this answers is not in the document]
jtconsol: The answer is the same as well :)

server/app/src/main/resources/sqlResources.xml
```diff
@@ -160,6 +160,9 @@
     <!-- Statements used to list all messages stored in this repository. -->
     <sql name="listMessagesSQL">SELECT message_name, message_state, last_updated FROM ${table} WHERE repository_name = ? ORDER BY last_updated ASC</sql>
 
+    <!-- Statements used to count messages stored in this repository. -->
+    <sql name="countMessagesSQL">SELECT COUNT(*) FROM ${table}</sql>
```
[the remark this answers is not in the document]
jtconsol: Have a look at the `size` method in `org.apache.james.mailrepository.jdbc.JDBCMailRepository`. It's referenced there.

jtconsol:
@Arsnael @trantienduchn I'll try to add tests. Do you have any pointers for that? From a quick look I'll be adding to `org.apache.james.mailrepository.MailRepositoryContract`?

dockerfiles/run/spring/destination/conf/sqlResources.xml
Arsnael: I get your point... Then I believe indentation was probably wrong on those other files from the start, but that's not a big problem... could be done in an other quick PR in worse case. I'm ok with that for now

Arsnael:
@jtconsol : My guess is `MailRepositoryContract` could be it yes. Thank you !

jtconsol:
@Arsnael As I understand it now, it's not possible to test against this issue, because the SQL resources used by the tests - `server/data/data-jdbc/src/test/resources` did in fact match the JDBC code - but the other variants of the SQL resources did not. Since these reside in `server/app/src/main/resources` and `dockerfiles/run/spring/destination/conf`, their erroneous settings were not detected by the tests, only when you run the server on JDBC. Wouldn't it be ideal to only store this file once and copy it around somehow? But that's definitively beyond the scope of this issue (and my maven-foo).

dockerfiles/run/spring/destination/conf/sqlResources.xml
jtconsol: Added the indentation to the JAMES-2828 branch, as it has already got the OID change ([commit](https://github.com/apache/james-project/pull/173/commits/a69a00077ea2cd0fb294b7de0be798ce625e4882)).

Arsnael:
@jtconsol Alright thanks for investigating this !

Arsnael:
Hi @jtconsol ! Sorry to bother you but could you do one last little thing? On the comment regarding the xml indent, you did it actually in the other PR ( https://github.com/apache/james-project/pull/173/commits/a69a00077ea2cd0fb294b7de0be798ce625e4882 ). Could you move that commit in this PR instead? Then both of your PRs would be merged without problem. Thank you for your understanding :)

jtconsol:
@Arsnael ...shouldn't be a problem - Just to be 100% sure: So I'll remove the XML indentation commit from JAMES-2828 / PR-173 and add it to JAMES-2929 / PR-172?

Arsnael:
Yes exactly. Just a little rebase :) Or you can even I think just do a cherry-pick in this one and keep the commit in #173 . Shouldn't be a problem

jtconsol:
@Arsnael Tried to go the rebase route, looks like it worked OK!

rouazana:
> @Arsnael Tried to go the rebase route, looks like it worked OK!

The rebase is fine, but your first commit is still full of unrelated space changes. Do you think you could move all these space changes into the second commit?

Arsnael:
Thanks @jtconsol :)

chibenwa:
Hi,

Also, should we close https://issues.apache.org/jira/browse/JAMES-2929 ?

Thanks again,

Cheers,

Benoit

jtconsol:
Yes, [JAMES-2929](https://issues.apache.org/jira/browse/JAMES-2929) can be closed.

Closing this issue here as well.

PR #173, opened 2019-10-23
JAMES-2828 JDBCMailRepository fails to message into PostgreSQL when per recipient headers are absent

jtconsol:
This fixes the aforementioned problem by going from ` insertMessage.setNull(9, Types.BLOB);` to `insertMessage.setObject(9, null);` - the latter works on PostgreSQL. I've positively tested it on Oracle and MySQL, and it also runs on Derby used in James' tests, so it seems pretty portable. See [this SO answer](https://stackoverflow.com/a/34128743/1529709) for details on PostgreSQL's BLOB handling.

### DIFF
--- a/dockerfiles/run/spring/destination/conf/sqlResources.xml
+++ b/dockerfiles/run/spring/destination/conf/sqlResources.xml
@@ -259,7 +259,7 @@
             recipients text NOT NULL ,
             remote_host varchar (255) NOT NULL ,
             remote_addr varchar (20) NOT NULL ,
-            per_recipient_headers oid NULL ,
+            per_recipient_headers bytea NULL ,
             last_updated timestamp NOT NULL,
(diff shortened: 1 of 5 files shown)


Arsnael:
I know I approved, but do you think you could write a test or two about that as well? Thanks !

jtconsol:
I'll try to write tests... as with my other PR, if you have any tips on that, send them my way please :)

jtconsol:
@Arsnael Unfortunately, [like I laid out here](https://github.com/apache/james-project/pull/172#issuecomment-546057176), this also can't be tested as such. It is a vendor specific issue and the Derby database used in the tests doesn't exhibit this problem, so it can't catch it.

jtconsol:
Removed XML indentation commit as per [instructions](https://github.com/apache/james-project/pull/172#issuecomment-547806998).

PR #174, opened 2019-10-24
propose adoption of java 11 as a runtime

mbaechler:


ieugen:
A fix for that direction https://github.com/apache/james-project/pull/190 .

PR #178, opened 2019-11-19
JAMES-2986 NetMatcher's comma-separation requires space to be present

jtconsol:
Fixes [JAMES-2986](https://issues.apache.org/jira/browse/JAMES-2986#).

Allows for "true comma-separation" as in, don't require a space to follow the comma.

Test is included.

### DIFF
--- a/server/dns-service/dnsservice-library/src/main/java/org/apache/james/dnsservice/library/netmatcher/NetMatcher.java
+++ b/server/dns-service/dnsservice-library/src/main/java/org/apache/james/dnsservice/library/netmatcher/NetMatcher.java
@@ -20,7 +20,6 @@ package org.apache.james.dnsservice.library.netmatcher;
 
 import java.net.InetAddress;
 import java.net.UnknownHostException;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.Iterator;
@@ -34,6 +33,8 @@ import org.apache.james.dnsservice.library.inetnetwork.model.InetNetwork;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.base.Splitter;
+
 /**
  * NetMatcher Class is used to check if an ipAddress match a network.
  * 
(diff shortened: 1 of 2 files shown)


server/dns-service/dnsservice-library/src/main/java/org/apache/james/dnsservice/library/netmatcher/NetMatcher.java
```diff
@@ -84,7 +83,7 @@ public NetMatcher(Collection<String> nets, DNSService dnsServer) {
 
     public NetMatcher(String commaSeparatedNets, DNSService dnsServer) {
         this.dnsServer = dnsServer;
-        List<String> nets = Splitter.on(NETS_SEPARATOR).splitToList(commaSeparatedNets);
+        List<String> nets = Arrays.asList(commaSeparatedNets.split(NETS_SEPARATOR_PATTERN));
```
chibenwa: Can't we just trim the netwok string before calling splitToList to get to the same result?

chibenwa: Well, I consider the use of String::split backed by a regular expression being a regression to be honset...

jtconsol: How is this fix a "regression" when the original code does not implement what is widely regarded as comma-separation?

chibenwa: > How is this fix a "regression" when the original code does not implement what is widely regarded as comma-separation?

The regrtession is about the use of regular expression in one place that was free of it.

Have you tried:

```
List<String> nets = Splitter.on(",")
    .trimResults()
    .splitToList(commaSeparatedNets);
```

?

This should solve your concerns, and pass your test suite, I guess... (without a regex)

jtconsol:
@chibenwa Got it, no regular expressions.

chibenwa:
Hi,

Thanks again,

Cheers,

Benoit

PR #179, opened 2019-12-03
JAMES-2813 ADR about the new distributed workqueue design

rouazana:


src/adr/0003-distributed-workqueue.md
```diff
@@ -6,6 +6,8 @@ Date: 2019-10-02
 
 Accepted (lazy consensus)
 
+Superceded by [16. Distributed WorkQueue](0016-distributed-workqueue.md)
```
mbaechler: Ah, very good case: do we keep old stuff in the ADR or do we drop them as soon as they don't describe the current state of things?

src/adr/0016-distributed-workqueue.md
```diff
@@ -0,0 +1,29 @@
...
+
+By switching the task manager to a distributed implementation, we need to be able to run a `Task` on any node of the cluster.
+
+## Decision
+
+  For the time being we will keep the sequential execution property of the task manager.
+  This is an intermediate milestone toward the final implementation which will drop this property.
+
+ * Use a RabbitMQ queue as a workqueue where only the `Created` events are pushed into.
+   Instead of using the brittle exclusive queue mechanism described in [3. Distributed WorkQueue](0003-distributed-workqueue.md) ,we will
```
mbaechler: s/ ,we/, we/

src/adr/0003-distributed-workqueue.md
rouazana: I have understood that the principle is to keep old stuff.
And it's natively handled by the tool, here I did `adr new -s 3 Distributed WorkQueue` to generate the new ADR and have automatically the `Superceded` comment added to the previous one. So it seems to be the way to go.

src/adr/0016-distributed-workqueue.md
```diff
@@ -0,0 +1,29 @@
...
+  This is an intermediate milestone toward the final implementation which will drop this property.
+
+ * Use a RabbitMQ queue as a workqueue where only the `Created` events are pushed into.
+   Instead of using the brittle exclusive queue mechanism described in [3. Distributed WorkQueue](0003-distributed-workqueue.md), we will
+   now use the natively supported [Single Active Consumer](https://www.rabbitmq.com/consumers.html#single-active-consumer) mechanism. 
+
+
+## Consequences
+
+ * This solution is safer to use in production: if the active consumer dies, an other one is promoted instead.
```
remk: s/safer/safe ?

src/adr/0003-distributed-workqueue.md
mbaechler: great, thank you for the explanation, let's use the same rules as other people.

src/adr/0016-distributed-workqueue.md
rouazana: are you sure? :trollface:

chibenwa:
This is merged.

Can you close this pull request?

PR #188, opened 2020-04-29
[ADR] Against conditional statements in guice modulesand alternatives

chibenwa:


chibenwa:
ML thread https://www.mail-archive.com/server-dev@james.apache.org/msg65530.html

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+ ObjectStorageDependenciesModule::selectBlobStoreBuilder and Cassandra Blob Store Cache conditional statement.
+ - [S3 native blobStore implementation](https://github.com/linagora/james-project/pull/3099) along side with S3 endpoints
+ support as part of Swift removes the need to select the Object Storage implementation.
+ - Follow up work needs to be plan concerning `BlobExportMechanismModule` and `TikaMailboxModule::provideTextExtractor`.
+
+## Consequences
+
+Integration testing can not offer conditional, configuration based module composition. This is because:
+
+ - Integration test don't rely on Main classes but on GuiceJamesServer class with similar guice modules
```
Arsnael, line 104: s/test/tests

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+ - We decided to keep supporting Cassandra for blob storing as an upgrade solution from Cassandra-guice to Distributed 
+James for existing users.
+ - Proposals such as [HybridBlobStore](0014-blobstore-storage-policies.md) and then 
+[Cassandra BlobStore cache](0025-cassandra-blob-store-cache.md) proposed to leverage Cassandra as a performance 
+(latency) enhancer for ObjectStorage technologies.
+
+Yet again it had been decided to use conditional statements in modules in order to lower the cardinality.
+
+However, [Cassandra BlobStore cache](0025-cassandra-blob-store-cache.md) requires expensive resource initialization
+requiring to perform upgrade procedure (usage of an additional cache keyspace) that represents a cost we don't want to
```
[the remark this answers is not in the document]
chibenwa: No

resource initialization requiring to perform upgrade procedure

Here I mean "give rights on a cache keyspace"

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+
+We should no longer rely on conditional statements in Guice module.
+
+We should within James main method, upon James startup read the configuration and select the modules that should be 
+selected to run it, before calling the Guice injector to perform its full startup.
+
+This enables easy diagnose of the running components via the selected module list. It exposes tested, safe choices to 
+the user while limiting the Guice products count.
+
+Basic minimalistic integration tests will be written to cover the possibilities exposed to the user, by statically
```
mbaechler, line 89: Ok, so we have 6 products, 2 blob exports, 3 object storages, 2 content extractors, it means we have 72 (6*2*3*2) combinations to tests?
It looks doable but the problem is: it doesn't scale as each option doubles the number of tests, whether it's done with guice-less condition or not.
Do we agree?

chibenwa:
> Finally, this ADR raise the issue that we probably need to refine the policy about what makes a product vs what can be a product option and what we test and what not.

Could be a nice other ADR topic for sure, pretty fundamental for sure.

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+Some components however do have several subtly diverging implementations a user might choose to rely on independently 
+of the product he uses. This is the case for:
+
+ - BlobExport: Exporting a blob from the blobStore to an external user. Two implementations are currently supported: 
+ localFiles and LinShare.
+ - Text extraction: Extracting text from attachment to enable attachment search. There is a Tika implementation, but 
+ lighter JSOUP based options are also available.
+
+In order to keep the cardinality of Guice products low, we decided to use conditional statements in modules based on the 
+configuration to select which one to enable. Eventually defeating the Guice adoption goals mentioned above.
```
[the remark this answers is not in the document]
chibenwa:
```
This is a specific point if disagreement: we decided to restrict combinations to be able to have good QA and support on the combination we value the most.
Whether a combination is built at runtime or at compile time: the combination exists.
We stated that we 6 well defined products previously. The question is: do we want to have 12 to support export of blob to Linshare? To have 24 to disable Tika support?
We answered "no" to that question.
Yet we have 24 combinations using 6 products.

Does it hurts QA and support? In my experience, not much.

Should we run our testsuites on these 24 combinations? I don't think so.
```

:+1: 

```
Does it defeat the usage of Guice as a technology? No, we still define products by combining Guice modules some of which are disable/enabled at runtime.
```

Me I want to turn:

The modules choose what to inject at runtime upon start.

Into:

The product choose which module to rely one.

It is not the module purpose to choose anything IMO.

chibenwa:
> I think, however, that the topic of QA/support of combinations will be unchanged.

But auditing what you actually run is easier.

A debug log upon startup could for example list the modules you start with.

With 'if' statements in modules, you basically don't know what you run, wich is a concern.

Also we can't:

```
void configure() {
    if (xyz) {
        install(new XYZModule());
    }
}
```

This proposition adresses the above concerns.

chibenwa:
I will reply more in detail on moday

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+Instead of having a single big application being able to instantiate each and every component application, we have 
+several products defining their dependencies in a minimalistic way, relying only on the components implementation that 
+are needed.
+
+Here is the list of products we provide:
+
+ - In-Memory: A memory based James server, mainly for testing purposes
+ - Distributed James: A scalable James server, storing data in various data stores. Cassandra is used for metadata, 
+ ElasticSearch for search, RabbitMQ for messaging, and ObjectStorage for blobs.
+ - Cassandra: An implementation step toward Distributed James. It does not include messaging and ObjectStorage and should not be run in a cluster way but is still relevant for good performance.
```
rouazana, line 35: can we state that it has chance to be deprecated/removed soon?

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+ - JPA: A JPA and Lucene based implementation of James. Only Derby driver is currently supported.
```
rouazana, line 36: extension-jars is not working?

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+ - JPA with SMTP only using derby: A minimalist SMTP server based on JPA storage technology and Derby driver
```
rouazana, line 37:
```suggestion
 - JPA with SMTP only using Derby: A minimalist SMTP server based on JPA storage technology and Derby driver
```

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+ - JPA with SMTP only using mariadb: A minimalist SMTP server based on JPA storage technology and MariaDB driver
```
rouazana, line 38:
```suggestion
 - JPA with SMTP only using MariaDB: A minimalist SMTP server based on JPA storage technology and MariaDB driver
```

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+
+Some components however do have several implementations a user can choose from in a given product. This is the case for:
+
+ - BlobExport: Exporting a blob from the blobStore to an external user. Two implementations are currently supported: 
+ localFiles and LinShare.
+ - Text extraction: Extracting text from attachment to enable attachment search. There is a Tika implementation, but
```
rouazana, line 44: and also no text extraction?

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+ - Text extraction: Extracting text from attachment to enable attachment search. There is a Tika implementation, but 
+ lighter JSOUP based options are also available.
+
+In order to keep the number of products low, we decided to use conditional statements in modules based on the 
+configuration to select which one to enable at runtime. Eventually defeating the Guice adoption goals mentioned above.
+
+Finally, Blob Storing technology offers a wide combination of technologies:
+
+ - ObjectStorage in itself could implement either Swift APIs or Amazon S3 APIs
+ - We decided to keep supporting Cassandra for blob storing as an upgrade solution from Cassandra product to Distributed
```
rouazana, line 53: do we? is there some documentation about this?

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+Finally, Blob Storing technology offers a wide combination of technologies:
+
+ - ObjectStorage in itself could implement either Swift APIs or Amazon S3 APIs
+ - We decided to keep supporting Cassandra for blob storing as an upgrade solution from Cassandra product to Distributed 
+James for existing users.
+ - Proposals such as [HybridBlobStore](0014-blobstore-storage-policies.md) and then 
+[Cassandra BlobStore cache](0025-cassandra-blob-store-cache.md) proposed to leverage Cassandra as a performance 
+(latency) enhancer for ObjectStorage technologies.
+
+Yet again it had been decided to use conditional statements in modules in order to lower the number of product.
```
rouazana, line 59:
```suggestion
Yet again it had been decided to use conditional statements in modules in order to lower the number of products.
```

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+
+However, [Cassandra BlobStore cache](0025-cassandra-blob-store-cache.md) requires expensive resource initialization
```
rouazana, line 61: maybe you can start this paragraph with the design issue, and then explain that Cassandra cache is an example.

rouazana:
Could we try to define when we need an other product vs when we need an other option?

For example I'm not sure a product is really needed for LDAP vs local users, WDYT?

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
chibenwa: They are not on the classpath thus jdbc cannot load alternative drivers. Extension jars enhance the class loader used toinvocate user defined extensions only.

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
chibenwa: I remember it is the main reason.

No. Nothing more than blob.properties

chibenwa:
The solution is simple thus the decision part is small.

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
mbaechler: is it the same soon as last year?

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
mbaechler: > I remember it is the main reason.

What about actually wanting to store data in Cassandra? (For a dataset < 1 TB it makes perfect sense IMO)

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Context
+
+James products rely historically on Spring for injections. Spring version is outdated (4.3.25 instead of 5.3 release line).
+Spring enables overriding any component via a configuration file thus endangering overall correctness by giving too much 
+power to the user.
+
+James propose several implementations for each of the interfaces we define. We don't run tests for each possible interface
+combination. We rather run integration tests for combinations that make sense. By overriding any components, Spring defeats 
+this testing logic. Knowing which component implementation combines well with which other one is not intuitive to users,
```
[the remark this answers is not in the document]
chibenwa: It mostly is a repetition of the previous sentence.

`and allowing arbitrary component combination` express it better IMO

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+James for existing users.
+ - Proposals such as [HybridBlobStore](0014-blobstore-storage-policies.md) and then 
+[Cassandra BlobStore cache](0025-cassandra-blob-store-cache.md) proposed to leverage Cassandra as a performance 
+(latency) enhancer for ObjectStorage technologies.
+
+Yet again it had been decided to use conditional statements in modules in order to lower the cardinality.
+
+However, [Cassandra BlobStore cache](0025-cassandra-blob-store-cache.md) requires expensive resource initialization
+requiring to perform upgrade procedure (usage of an additional cache keyspace) that represents a cost we don't want to
+pay if we don't rely on that cache. Not having the cache module thus enables quickly auditing that the caching cassandra
```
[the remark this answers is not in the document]
chibenwa: GuiceJamesServer will emit an INFO log with the list of modules used to create its Guice injector.

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+ blobStore be wrapped by a caching layer?
+
+## Decision
+
+We should no longer rely on conditional statements in Guice module.
+
+We should within James main method, upon James startup read the configuration and select the modules that should be 
+selected to run it, before calling the Guice injector to perform its full startup.
+
+This enables easy diagnose of the running components via the selected module list. It exposes tested, safe choices to
```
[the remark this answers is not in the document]
chibenwa: Via an INFO log?

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Decision
+
+We should no longer rely on conditional statements in Guice module.
+
+We should within James main method, upon James startup read the configuration and select the modules that should be 
+selected to run it, before calling the Guice injector to perform its full startup.
+
+This enables easy diagnose of the running components via the selected module list. It exposes tested, safe choices to 
+the user while limiting the Guice products count.
```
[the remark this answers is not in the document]
chibenwa: I removed this part from the decision

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
chibenwa: Idem

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
chibenwa: As metionned by Matthieu:

```
We decided to keep supporting Cassandra for blob storing as an upgrade solution from Cassandra product to Distributed 
James for existing users. This option also makes sense for small data-sets (less than a TB) where storage cost are less 
of an issue and don't need to be taken into account when reasoning about performance.
```

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -12,20 +12,24 @@ James products rely historically on Spring for dependency injection. It doesn't
...
 
-James propose several implementations for each of the interfaces it defines. The number of possible combinations of implementations is thus really high (like factorial(n) with n > 10). It makes it unpractical to run tests for each possible interface
-combination. We rather run integration tests for combinations that we decide brings the more value to the users. By overriding any class, Spring defeats 
-this testing logic. Knowing which component implementation combines well with which other one is not intuitive to users,
-hard to document and test.
-
-We thus decided to rely on Guice. Components are defined by code in a static fashion. Overriding components requires 
-explicit code modification and recompilation, thus warning the user about the impact of the choices he does, and lowering 
-the project's responsibility.
+James propose several implementations for each of the interfaces we define. The number of possible combinations of
```
mbaechler, line 15: why did you choose to replace `it` with `we`? In this sentence, as it says "James propose" something, it makes little sense, don't you think?

mbaechler, line 15:
```suggestion
James proposes several implementations for each of the interfaces we define. The number of possible combinations of
```

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -12,20 +12,24 @@ James products rely historically on Spring for dependency injection. It doesn't
...
+implementations is thus really high (like factorial(n) with n > 10). It makes it unpractical to run tests for each 
+possible interface combination. We rather run integration tests for combinations that we decide brings the more value to
```
mbaechler, line 17: we don't combine interfaces

mbaechler, line 17:
```suggestion
possible interface combination. We run integration tests for combinations that we decide brings the more value to
```

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -12,20 +12,24 @@ James products rely historically on Spring for dependency injection. It doesn't
...
+the users. We rather run integration tests for combinations that make sense. By overriding any components, and
```
mbaechler, line 18: > We rather run integration tests for combinations that make sense

It's redundant, don't you think?

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -12,20 +12,24 @@ James products rely historically on Spring for dependency injection. It doesn't
...
-hard to document and test.
-
-We thus decided to rely on Guice. Components are defined by code in a static fashion. Overriding components requires 
-explicit code modification and recompilation, thus warning the user about the impact of the choices he does, and lowering 
-the project's responsibility.
+James propose several implementations for each of the interfaces we define. The number of possible combinations of
+implementations is thus really high (like factorial(n) with n > 10). It makes it unpractical to run tests for each 
+possible interface combination. We rather run integration tests for combinations that we decide brings the more value to
+the users. We rather run integration tests for combinations that make sense. By overriding any components, and 
+allowing arbitrary component combination, Spring defeats this testing logic.
```
mbaechler, line 19: Spring is not about components but about classes, I prefer previous wording,

mbaechler, line 19: It's not clear with this wording why changing classes breaks the principle of "test only combinations that make sense"

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -36,17 +36,19 @@ Here is the list of products we provide:
  - In-Memory: A memory based James server, mainly for testing purposes
  - Distributed James: A scalable James server, storing data in various data stores. Cassandra is used for metadata, 
  ElasticSearch for search, RabbitMQ for messaging, and ObjectStorage for blobs.
- - Cassandra: An implementation step toward Distributed James. It does not include messaging and ObjectStorage and should not be run in a cluster way but is still relevant for good performance.
+ - Cassandra: An implementation step toward Distributed James. It does not include messaging and ObjectStorage and 
+ should not be run in a cluster way but is still relevant for good performance. This product is about to be deprecated
```
mbaechler, line 40: I still don't agree to include "This product is about to be deprecated" here, it's a useless out-of-context addition.

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -55,17 +57,19 @@ Finally, Blob Storing technology offers a wide combination of technologies:
 
  - ObjectStorage in itself could implement either Swift APIs or Amazon S3 APIs
  - We decided to keep supporting Cassandra for blob storing as an upgrade solution from Cassandra product to Distributed 
-James for existing users.
+James for existing users. This option also makes sense for small data-sets (less than a TB) where storage cost are less
```
mbaechler, line 60: I note that 1TiB of mails is a small dataset (:

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,118 @@
...
+
+We should no longer rely on conditional statements in Guice module.
+
+Guice modules combination choice should be decided before starting the dependency injection stage.
+
+An INFO log with the list of modules used to create its Guice injector. This enables easy diagnose of the running 
+components via the selected module list. It exposes tested, safe choices to the user while limiting the Guice products 
+count.
+
+The following conditional statements in guice modules needs to be removed :
```
mbaechler, line 97: I think it relates to Consequences

src/adr/0036-against-use-of-conditional-statements-in-guice-modules.md
```diff
@@ -0,0 +1,118 @@
...
+ - [S3 native blobStore implementation](https://github.com/linagora/james-project/pull/3099) along side with S3 endpoints
+ support as part of Swift removes the need to select the Object Storage implementation.
+ - Follow up work needs to be plan concerning `BlobExportMechanismModule` and `TikaMailboxModule::provideTextExtractor`.
+
+## Consequences
+
+Component combination count keeps unchanged for Guice products, but the run combination is explicit. QA needs are 
+unchanged.
+
+Integration testing can not offer conditional, configuration based module composition. This is because:
```
mbaechler, line 110: I would really like we don't  do that to prevent diverging from what happens in production.
I'd like that:
* we define POJOs for configurations involving module choosing with a nice programmatic interface to build them
* we create a method that takes a module combination + POJOs to start the server and rely on that for both tests and production
* we write parsers for config file to POJOs and call the method to start the server once the POJOs are built

WDYT?

chibenwa:
Merged

PR #191, opened 2020-05-11
[JAMES-3187] Updates to the "Try" section of the User Manual

dleangen:
Now that the Antora documentation stub is available in the master branch, I wanted to create a place to host my proposed updates to the "User Manual" portion of the documentation.

**This is not yet ready to be merged.**

Questions will follow. :-)

### DIFF
--- a/docs/modules/user/pages/try/15_minutes.adoc
+++ b/docs/modules/user/pages/try/15_minutes.adoc
@@ -1,0 +1,232 @@
+= Try James in 15 minutes
+
+In this tutorial, we will set up a runing James demo using a prepared Docker image.
+We will then test the server by connecting with an email client. Finally, we
+will connect to the server via the REST-based Admin API.
+
+
+Requirements: 
+
+ * Docker
+ * curl
+ * Thunderbird client (optional)
+
+== Set up the demo server
+
+Before starting the server, because we will be connecting to the Admin API we need to set some configuration parameters.
+We will create a `webadmin.properties` file that James will use in order to allow us to connect to the API.
+
+Run this command to create the `webadmin.properties` file:
+
+[source,bash]
+----
+printf 'enabled=true\nport=8000\nhost=localhost' >> webadmin.properties
+----
+
+Explanation:
+
+ * `enabled=true` instructs James to run the Admin API service
+ * `port=8000` configures the Admin API to be made available via port 8000
+ * `host=localhost` configures the Admin API to respond on localhost
+
+
+Now run the James demo server using this command:
+
+[source,bash]
+----
+docker run -d -p "25:25" -p "143:143" -p "127.0.0.1:8000:8000" -v $(pwd)/webadmin.properties:/root/conf/webadmin.properties --name james linagora/james-jpa-sample:3.4.0
+----
+
+Explanation:
+
+ * `docker run` runs the provided image with the given parameters
+ * The `-d` parameter runs the container in "detached" mode
+ * `-p "25:25" -p "143:143"` attaches the image ports to the ports 25 (SMTP) and 143 (IMAP) on the host machine
+ * `-p "127.0.0.1:8000:8000" attaches the port 8000 to localhost on the host machine to allow us to connect to the Admin API
+ * `-v $(pwd)/webadmin.properties:/root/conf/webadmin.properties` mounts the webadmin.properties file to configure the Admin API
+ * The `--name james` parameter gives the running container a name to make it easier to manipulate
+ * `linagora/james-jpa-sample:3.4.0` is the image that is used for this demo
+
+Docker will pull the image and start the container.
+
+****
+To run commands using the James CLI, you can use the running container via Docker:
+
+----
+docker exec james java -jar /root/james-cli.jar \
+    -h \<<HOST>> -p \<<PORT>> \<<COMMAND>>
+----
+
+In this tutorial, we are using host 127.0.0.1 and port 9999, so every command looks like:
+
+----
+docker exec james java -jar /root/james-cli.jar \
+    -h 127.0.0.1 -p 9999 \<<COMMAND>>
+----
+
+Host 127.0.0.1 is of course localhost, and the use of port 9999 is completely arbitrary.
+
+To make this tutorial a little easier, set this up as a bash script by copying and pasting this script:
+
+[source,bash]
+----
+printf '#!/bin/bash\n\ndocker exec james java -jar /root/james-cli.jar -h 127.0.0.1 -p 9999 $@' >> james ; chmod +x james
+----
+****
+
+As an example, list all the domains currently in use by James:
+
+[source,bash]
+----
+./james listDomains
+----
+
+You should notice that a default domain, james.local, has been created
+
+List all the current users:
+
+[source,bash]
+----
+./james listUsers
+----
+
+You should see users ``user01@james.local``, ``user02@james.local``, and ``user03@james.local``.
+
+Create a new "test.local" domain:
+
+[source,bash]
+----
+./james addDomain test.local
+----
+
+List the domains again to ensure that "test.local" has successfully been added:
+
+[source,bash]
+----
+./james listDomains
+----
+
+Add the user "testuser" to the "test.local" domain with password "password":
+
+[source,bash]
+----
+./james addUser testuser@test.local password
+----
+
+You should now see your newly created user:
+
+[source,bash]
+----
+./james listUsers
+----
+
+== Connect to the server with an email client
+
+****
+For this tutorial, we will use Thunderbird, as it is available in multiple languages
+on Windows, Mac, and Linux.
+
+Go to https://www.thunderbird.net to download Thunderbird.
+****
+
+After you have installed Thunderbird,
+https://support.mozilla.org/en-US/kb/manual-account-configuration[manually set up an account]
+for user01@james.local using the following parameters:
+
+ * Account name: user01
+ * Your name: User 01
+ * Email address: user01@james.local
+ * SMTP server name: localhost
+ * SMTP port: 25
+ * SMTP connection security: none
+ * SMTP authentication: password
+ * IMAP server name: localhost
+ * IMAP user name: user01@james.local
+ * IMAP port: 143
+ * IMAP connection security: none
+ * IMAP authentication: none
+
+Repeat the above for testuser@test.local:
+
+ * Account name: testuser
+ * Your name: Test User
+ * Email address: testuser@test.local
+ * SMTP server name: localhost
+ * SMTP port: 25
+ * SMTP connection security: none
+ * SMTP authentication: password
+ * IMAP server name: localhost
+ * IMAP user name: testuser@test.local
+ * IMAP port: 143
+ * IMAP connection security: none
+ * IMAP authentication: none
+
+Now have fun sending emails back and forth to yourself. :-)
+But don't spend all day doing this. You have too much work to do.
+
+== Connect to the server via the Admin API
+
+We will use curl to connect to localhost on port 8000, as the Admin API
+is made available on this port.
+
+List the available domains:
+
+[source,bash]
+----
+curl http://localhost:8000/domains
+----
+
+You should see a response similar to this:
+
+[source,json]
+----
+[
+  "edfce41c55e6",
+  "james.linagora.com",
+  "james.local",
+  "localhost",
+  "test.local",
+  "172.17.0.2"
+]
+----
+
+Test that a domain exists:
+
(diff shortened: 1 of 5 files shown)


dleangen:
I was given [this information](https://github.com/apache/james-project/pull/188/files) from the mailing list.

Now that I see there is an in-memory product, I would like to use this for the “5 minute introduction” instead of the current docker-compose used for the example. I think that it would be preferable to have the lightest-weight version possible, with minimal dependencies, when introducing James for the first time.

Would somebody be able to show me where there is some explanation I can use to try to figure out how to get the in-memory product up and running so that I may update the User Manual to use this?

chibenwa:
Hi @dleangen 

>  I think that it would be preferable to have the lightest-weight version possible, with minimal dependencies, when introducing James for the first time.

I think a dockerized distributed james makes little sense for a 5 minute try too.

I would suggest JPA-guice image. We even provide a pre-provisionned image the could be used straight away:

`docker run -p "25:25" -p "143:143" linagora/james-jpa-sample:3.4.0`

We use it on the https://james.apache.org/ `Getting started` page.

Initial data can be found here: https://github.com/apache/james-project/blob/master/dockerfiles/run/guice/provisioned/initialdata.sh

IMO it makes more sense as the data would survive reboot, what is not the case with memory. Memory is mostly here for testing purposes.

Would it actually suite your needs?

dleangen:
Thanks @chibenwa. That should work. I'll give it a try.

dleangen:
Oops. Just realized that I can't change the name of the branch when a PR is open.

dleangen:
Hello!

I am almost finished with the "Try James" section, but I have a question...

I noticed that the sample sets up 3 users: user01@james.local, user02@james.local, and user03@james.local.

What would you recommend as the **simplest** way to allow a user to send and receive an email with this user so they can see that James is working for them.

Instead of setting up a new account with their email client, are there some CLI commands I can use to send and email to this user, then view the contents of the email?

On that topic, are there any more detailed documents sitting around somewhere about the CLI commands?

rouazana:
Hello,

The simpler user friendly way is probably to explain how to configure Thunderbird. Configuring local SMTP and IMAP server is not very complicated, and allows to test all the process.

If you expect CLI commands, I see to tracks:
- the hard way using telnet on SMTP and IMAP: pretty straightforward if and only if you know the protocols
- the soft way using curl which is supporting SMTP and IMAP (at the cost of some magic command lines arguments)

rouazana:
> On that topic, are there any more detailed documents sitting around somewhere about the CLI commands?

The James CLI is documented here: http://james.apache.org/server/manage-cli.html

But we plan to deprecate it (in fact to replace it to remove JMX and using REST instead) by: http://james.apache.org/server/manage-webadmin.html

chibenwa:
Well the plan is more to port the cli to rely on webadmin REST calls instead of JMX, we will not drop the CLI

rouazana:
> Well the plan is more to port the cli to rely on webadmin REST calls instead of JMX, we will not drop the CLI

That's what I wanted to say, but you better express it :)

dleangen:
Ready for review / merge, please. Thanks!

docs/modules/user/pages/try/15_minutes.adoc
```diff
@@ -0,0 +1,229 @@
...
+== Set up the demo server
+
+Before starting the server, because we will be connecting to the Admin API we need to set some configuration parameters.
+We will create a `webadmin.properties` file that James will use in order to allow us to connect to the API.
+
+Run this command to create the `webadmin.properties` file:
+
+[source,bash]
+----
+printf 'enabled=true\nport=8000\nhost=localhost' >> webadmin.properties
```
chibenwa, line 23: Shouldn't we enable "webadmin" by default so that "override the configuration" step is no longer needed.

Note: with docker the port can be bound but not exposed

docs/modules/user/pages/try/15_minutes.adoc
```diff
@@ -0,0 +1,229 @@
...
+----
+printf '#!/bin/bash\n\ndocker exec james java -jar /root/james-cli.jar -h 127.0.0.1 -p 9999 $@' >> james ; chmod +x james
+----
+****
+
+As an example, list all the domains currently in use by James:
+
+[source,bash]
+----
+./james listdomains
```
chibenwa, line 81: How about adding this little bash script in the docker image? Looks convenient...

This would turn the command into `docker exec james james-cli addDomain blah blah` which should be easy enough for a 15 minutes tutorial...

docs/modules/user/pages/try/15_minutes.adoc
```diff
@@ -0,0 +1,229 @@
...
+----
+
+You should notice that a default domain, james.local, has been created
+
+List all the current users:
+
+[source,bash]
+----
+./james listusers
```
chibenwa, line 90: I propose to rely on `./james listUsers`, easier to read ;-)

(Adding case sensitivity was a previous CLI readability improvment)

Idem for `addDomain`, `listDomains`, `addUsers`

docs/modules/user/pages/try/15_minutes.adoc
```diff
@@ -0,0 +1,229 @@
...
+----
+
+Now retest `test.local`:
+
+[source,bash]
+----
+curl -I -X GET http://localhost:8000/domains/test.local
+----
+
+This time you will receive a `404` code because the `test.local` domain no longer exists.
```
chibenwa, line 216: We could state that many more API are exposed via webadmin and are documented in detail https://james.apache.org/server/manage-webadmin.html , people can of course play longer with the content documented there ;-)

docs/modules/user/pages/try/15_minutes.adoc
dleangen: That would be great, as it would indeed simplify the demo.

Small picky point, but would it also be possible to change the name of the image from "james-jpa-sample" to "james-demo"?

Reason: JPA is not relevant, and "demo" matches the documentation better.

Regarding the port, do you mean even without the `-p "8000:8000"` it is possible to connect to the docker image on port 8000? Or did you mean something else?

docs/modules/user/pages/try/15_minutes.adoc
dleangen: Great! Yes!

docs/modules/user/pages/try/15_minutes.adoc
dleangen: Ok, I will make the change.

docs/modules/user/pages/try/15_minutes.adoc
dleangen: Ok, good point.

dleangen:
@chibenwa, thank you very much for the comments. I will make the changes soon.

If the image could be updated, that would be very nice.

chibenwa:
I'm about to open a pull request for provisionned demo image updates.

I realize there is no JIRA ticket open for this documentation effort. Could we open one?

(I would be happy to prefix my commits with it!)

dleangen:
Do you mean open a JIRA issue for the doc, and match the PR to the Jira Issue?

If that is what you mean, of course I don't mind at all. If you mean something else, could you please elaborate?

chibenwa:
> If that is what you mean, of course I don't mind at all. If you mean something else, could you please elaborate?

Sorry if I was not explicit enough.

`Do you mean open a JIRA issue for the doc, and match the PR to the Jira Issue?`

Yes, this is exactly what I meant.

Then from reading github / doing git blames in the code, people can go up to the intent (JIRA issue), find the discussions in github via a simple search, etc...

chibenwa:
I did a quick draft from `james-cli` helper packaged as part of provisionned image here: https://github.com/apache/james-project/pull/193

dleangen:
Ok, so IIUC, this should be the process to submit a PR correctly?

1. Create a JIRA issue
2. Create a PR and reference the JIRA Issue
3. Keep the discussions / comments in JIRA, not here in GitHub

Is that correct? Sorry I have not followed the proper procedure so far. I will fix my other PRs to do this once you confirm that I have correctly understood.

docs/modules/user/pages/try/15_minutes.adoc
chibenwa: > Small picky point, but would it also be possible to change the name of the image from "james-jpa-sample" to "james-demo"?

:+1: 

> Regarding the port, do you mean even without the -p "8000:8000" it is possible to connect to the docker image on port 8000? Or did you mean something else?

`-port 8000:8000` means that port 8000 of your container is bound with port 8000 of your computer.

We don't want people to disclose unauthenticated webadmin access by default for obvious security reasons.

We don't want people to do extra configuration because it is a pain.

So a good middle ground would be to enable always webadmin in docker images, and **warn** them that access to port 8000 should be limited for security reasons unless JWT authentication is setted up. 

Maybe we should document the port binding to loopback inerface to limit the risks.

Something like `docker run -p 127.0.0.1:8000:8000 ...` according to https://docs.docker.com/config/containers/container-networking/#published-ports. Maybe we might want to update existing README as well.

Also JWT authentication is not convenient enough. We don't have alternatives for webAdmin authentication and it is sad. I propose to add a mechanism to `None` and `Jwt` modes:  `Basic` using the `administratorId` and corresponding credentials as configured in `usersrepository.xml`. I will create a JIRA ticket for this.

chibenwa: See https://issues.apache.org/jira/browse/JAMES-3186 ;-)

chibenwa:
I would :

 1. Create a JIRA issue
 2. Rebase the open pull request to reword commits to add the ticket number as a prefix
 3. Click the edit button top right to add the JIRA ticket in the PR title.

So you don't need to open/close any pull request ;-)

> Is that correct? Sorry I have not followed the proper procedure so far. 

I'm sorry that we did not enforced the "correct" workflow so far....

> 3. Keep the discussions / comments in JIRA, not here in GitHub

My opinion is that we should handle discussions where it is the most convenient. What is important is that we reference pull requests in JIRA so that readers can go from JIRA to github PRs. A little extra description on pull requests would complete it perfectly.

People should read both, but I at least tend to be more responsive on github (and thus prefer it).

Feel free to do what you prefer.

docs/modules/user/pages/try/15_minutes.adoc
chibenwa: I opened a few pull requests on that webadmin in docker topic:

 - https://github.com/apache/james-project/pull/195 Add webadmin enabled by default to jpa-guice docker image -> we can deliver this as part of 3.5.0 release IMO

 - https://github.com/linagora/james-project/pull/3387 Document webadmin port binding to loopback interface (just documentation)

dleangen: Updated exposure of port 8000 to be `-p "127.0.0.1:8000:8000"`.

docs/modules/user/pages/try/15_minutes.adoc
dleangen: This is not yet documented in the new docs, so I included a placeholder instead.

docs/modules/user/pages/try/15_minutes.adoc
dleangen: Updated

dleangen:
I have made the requested updates. There are still a few outstanding issues:

 * Update the demo image
 * Include a reference of the webadmin in the new documentation (and link to it)

Should we keep this as a draft until those are completed? Or go ahead with a merge and open separate issues?

chibenwa:
> Should we keep this as a draft until those are completed? Or go ahead with a merge and open separate issues?

IMO go ahead.

I can address image related issues of your work in images pull requests.

Regarding the full webadmin documentation, I would prefer to link the existing documentation, and replace it with the new one once available.

dleangen:
> Regarding the full webadmin documentation, I would prefer to link the existing documentation, and replace it with the new one once available.

Ok, done.

I have switched off Draft mode in preparation for the merge.

docs/modules/user/pages/try/15_minutes.adoc
dleangen: These will be captured as issues elsewhere, so I am resolving here.

docs/modules/user/pages/try/15_minutes.adoc
dleangen: This will be captured as an issue elsewhere, so I am resolving here.

chibenwa:
I merged this work.

I believe this pull request can be closed now.

dleangen:
Thanks for all your support.

PR #192, opened 2020-05-13
[JAMES-3187] Updates to the "Run" section in the User Manual

dleangen:
This is the "Run James in Production" section (formerly the "build your own production server" section).

I changed to "run" because it is not actually built. It is run with Docker. The "build" section will only be available for developers.

All comments welcome. This is a work in progress and **is not yet ready to be merged**.

### DIFF
--- a/docs/modules/user/nav.adoc
+++ b/docs/modules/user/nav.adoc
@@ -3,4 +3,4 @@
 *** xref:try/5_minutes.adoc[]
 *** xref:try/15_minutes.adoc[]
 ** xref:learn.adoc[]
-** xref:build.adoc[]
+** xref:run.adoc[]
--- a/docs/modules/user/pages/build.adoc
+++ b/docs/modules/user/pages/build.adoc
@@ -1,4 +1,0 @@
-= Build your own custom mail server
-:navtitle: Build
-
-(TODO)
--- a/docs/modules/user/pages/index.adoc
+++ b/docs/modules/user/pages/index.adoc
@@ -18,6 +18,6 @@ Operator Manual for additional information about James from different perspectiv
 
 == How to...
 
- * xref::try.adoc[Try James in 5 minutes]
- * xref::build.adoc[Learn more about the architecture]
- * xref::build.adoc[Build your own custom mail server]
+ * xref::try.adoc[]
+ * xref::learn.adoc[]
+ * xref::run.adoc[]
--- a/docs/modules/user/pages/run.adoc
+++ b/docs/modules/user/pages/run.adoc
@@ -1,0 +1,94 @@
+= Run James in Production
+:navtitle: Run
+
+James provides a number of *profiles* out of the box.
+These profiles have been tested and verified so you can
+get set up quickly with a production mail server that you
+can use with confidence.
+
+The available profiles are:
+
+ * <<minimal,James Minimal Profile>>
+ * <<basic,James Basic Profile>>
+ * <<advanced,James Advanced Profile>>
+ * <<distributed,James Distributed Profile>>
+
+
+
+[#minimal]
+== James Minimal Profile
+
+If you are not sure which profile you should be using, then
+you probably ought to be using the *James Minimal Profile*.
+
+This profile is intended to be the simplest to set up and use in production.
+It has the least amount of dependencies and complexities. If you do not yet
+have a huge amount of emails to process, then usually the simplicity is
+well worth the loss of some functionality. The last thing you need is
+to have to resolve difficult issues on a production server when you have not
+yet acquired the requisite knowledge to deal with those issues. Using the
+*James Minimal Profile* will help you reduce the risk of running into 
+production issues.
+
+This profile is:
+
+ * Suggested for use with smaller deployments
+ * Appropriate for use by most operators
+ * The preferred choice for most installations
+ * Endowed with fewer dependencies, which makes it simpler and less risky to use in production
+ * Only dependent on your local file system for data storage, so very easy to manage
+
+(TODO: Add link to details page)
+
+
+[#basic]
+== James Basic Profile
+
+This profile is intended to be relatively easy to use, but provides additional features with regards
+to data storage and management. Of course, these additional features come at the cost of more
+operational complexity. You will have to now deal with a relational database and all the fun
+that comes with it. However, if you are already an expert at relational database management,
+then this profile should be a natural choice for you.
+
+This profile is:
+
+ * Suggested for smaller deployments that require data persisted in a relational database
+ * Appropriate for most operators provided that you understand relational databases and data indexing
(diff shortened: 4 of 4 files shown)


rouazana:
That's pretty nice. Just a side note: JMAP is for now only supported in the advanced and distributed profiles (and the memory one for those who just want to quickly test the server). Is it worth to mention it?

dleangen:
Thanks.

I think it is indeed worth mentioning in the details. This page is intended to just give an overview, so it's probably ok not to mention it here.

chibenwa:
Hello, 

Thanks a lot for that work.

I really like:

 - Specifying for whom a profile is intended, with use cases
 - Presenting all of them by gradual adoption complexity

This is a really nice suggestion that we should adopt.

I would add to it details (data store dependencies, and main architecture knowledge)

Then we tend to not use the word *profile* but *product*. Do you think it is an appropriate terminology?

Regarding existing products:

 - Your `Minimal` and `Basic` product should be JPA: we provide a JPA product with local storage (embedded derby driver + DB), feeting both needs of simple single user no-dependancy deployments, and DB backed single server deployments. We should mention that data is store in a database and that a user can configure its jdbc driver. We should mention mailqueue relies on a embedded activeMQ. We should mention Lucene is used for search.

 - For `Advanced` (Distributed product) I would add topology details: several james server works together as a cluster, stores metadata in Cassandra, relies on RabbitMQ for messaging, stores Blobs in a configurable location (Cassandra, ObjectStorage or a combinaition of both).

 - We need to still speak about the `Legacy` Spring product, not under active development, allowing arbitrary component combinaison configuration.

Would that make sense?

Koodo for the work!

Benoit

dleangen:
Hi @chibenwa. Thank you for the comments.

> I would add to it details (data store dependencies, and main architecture knowledge)

Ok, will do. I have not yet arrived at the detailed sections. It may still take me a while to get there.

> Then we tend to not use the word profile but product. Do you think it is an appropriate terminology?

Personally, I have a problem with "product". However, it could be my personal bias. I tend to think of a "product" as something a little different than what is intended here. Mainly, "product" seems to me to have a very commercial connotation in English. I think that is not what is intended here. Secondly, I feel that "product" for the average person probably refers to "Apache James" itself. If anything, these would be more like "sub-products" to me.

I understand that this is debatable. If it is very difficult to change the term, we could keep it. However, "profile" just feels so much better to me, and I think would make more sense to others new to James. :-)

> Your Minimal and Basic product should be JPA

My understanding was that there is also a file system version. Is that not so? My thought is that by not depending on a RDB, and by only depending on a filesystem, it eliminates some complexity for those (like me, hehe) who would rather simply store as a Maildir directly on the file system than have to manage an RDB. I have been using qmail for almost 20 years now I think, and the filesystem has been just fine for me. I am actually happy that I never had to deal with a RDB. It has made my life much simpler, given my limited needs.

> For Advanced (Distributed product) I would add topology details: several james server works together as a cluster, stores metadata in Cassandra, relies on RabbitMQ for messaging, stores Blobs in a configurable location (Cassandra, ObjectStorage or a combinaition of both).

Yes, absolutely. My thought was to do this in the detailed section. The reasoning is: those who "understand" will still be able to see those details, but those who are coming in for the first time don't need to be overwhelmed with information they will probably never need. The approach is: everything the "experts" need will be there, but new users should feel more welcome (by not being overwhelmed and by having clearer instructions).

> We need to still speak about the Legacy Spring product, not under active development, allowing arbitrary component combinaison configuration.

Yes, definitely. Again, the thought is to keep this only in the "developer" section.

The "user" section is for those who just want to use James without having to worry too much about the details. The "developer" section will be more for those who want to get inside James and make their own custom solution.

A simplistic way to make the distinction is: can I just run it with Docker? If so, then I am a "user". Do I need to compile? If so, then I am a "developer".

Anyway, I think it is better to separate the two.

chibenwa:
> Personally, I have a problem with "product". However, it could be my personal bias. I tend to think of a "product" as something a little different than what is intended here. Mainly, "product" seems to me to have a very commercial connotation in English. I think that is not what is intended here. Secondly, I feel that "product" for the average person probably refers to "Apache James" itself. If anything, these would be more like "sub-products" to me.

We definitly have to start a discussion about vocabulary we rely on, for sure.

Your point stands.

I think a disctussion on the mailing list should be started around this topic, which is fundamental for James as a project.

> My understanding was that there is also a file system version. Is that not so?

Maildir storage can be leveraged on top of (Legacy) Spring *profile*. It did not get it's own standalone "profile". The underlying implementation is maintained in best effort mode, integration test coverage is limited, and it suffers from well known bugs. See MAILBOX-389, MAILBOX-393 for instance.

I personally don't want to promote it as the simplest product as long as these limitation remains.

> I am actually happy that I never had to deal with a RDB. 

An embedded database is just another local file format, just exploring it via the filesystem is not doable. From an operating perspective, I'm unsure if it changes much. Maybe I am missing something?

> Yes, absolutely. My thought was to do this in the detailed section. The reasoning is: those who "understand" will still be able to see those details, but those who are coming in for the first time don't need to be overwhelmed with information they will probably never need. The approach is: everything the "experts" need will be there, but new users should feel more welcome (by not being overwhelmed and by having clearer instructions).

:+1: I have a tendency to go too quickly into technical details.

> Yes, definitely. Again, the thought is to keep this only in the "developer" section.

Huge :+1: here. I did not understood it. It reduces the reach of the legacy product without removing it.

> A simplistic way to make the distinction is: can I just run it with Docker? If so, then I am a "user". Do I need to compile? If so, then I am a "developer".

To dig into details, we do have a ready to use dockerized Spring image. But arbitrary component combinaison is that should be exposed only to "developers".

dleangen:
Ok, I see your point about Maildir. That's too bad, but I guess that's life.

Please note that this PR is **not yet ready to merge**.

Outstanding items:

1.  Resolve discussion about "Product" vs. "Profile" on the mailing list
2. Include a small demo using the Admin API via curl (it is not yet working for me)
3. Include details about how to connect to the sample app using Thunderbird (also no success yet for me)

mbaechler:
> Ok, I see your point about Maildir. That's too bad, but I guess that's life.
> 
> Please note that this PR is **not yet ready to merge**.

you can flag the PR as a Draft in this case

dleangen:
Oh, thanks! Did not know that. From what I just read, I need to do that when I create the PR. Do you know if it's possible to change this existing one to a draft? Other options are to leave this one and create as a draft next time, or to delete this one and make a new one.

Any preference?

mbaechler:
Screenshot from 2020-05-15 11-26-15

It's a really new feature you can find it on the right of the screenshot above : "Still in progress? convert to draft"

dleangen:
Cool, thanks!

dleangen:
This PR was folded into #199, so closing.

PR #193, opened 2020-05-18
JAMES-3187 Provisionned docker: add a cli helper script

chibenwa:
Turns the pain of

```
docker exec james java -jar /root/james-cli.jar -h 127.0.0.1 -p 9999 listUsers
```

Into:

```
docker exec james james-cli listUsers
```

See https://github.com/apache/james-project/pull/191#discussion_r426521321

Question:

 - Shouldn't we generalize the adoption to all of our docker images ?
 - Doing so will simplyfy the README by making no difference for CLI invocation between Guice and spring profiles...
 - Also, packaging this change as part of upcoming 3.5.0 wouldcompose well with the ongoing documentation effort.

### DIFF
--- a/dockerfiles/run/guice/provisioned/Dockerfile
+++ b/dockerfiles/run/guice/provisioned/Dockerfile
@@ -2,7 +2,7 @@
 #
 # VERSION	1.0
 
-FROM linagora/james-jpa-guice
+FROM linagora/james-jpa-guice:james-project-3.4.0
 
 # Install git
 RUN apt-get update
@@ -15,6 +15,7 @@ RUN cp /root/wait-for-it/wait-for-it.sh /usr/bin/wait-for-it.sh
 
 COPY startup.sh /root
 COPY initialdata.sh /root
+COPY james-cli /usr/local/bin/
 
 RUN chmod +x /root/startup.sh
 RUN chmod +x /root/initialdata.sh
--- a/dockerfiles/run/guice/provisioned/james-cli
+++ b/dockerfiles/run/guice/provisioned/james-cli
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+java -jar /root/james-cli.jar -h 127.0.0.1 -p 9999 "$@"


Arsnael:
> Question:
> 
>     * Shouldn't we generalize the adoption to all of our docker images ?
>
>     * Doing so will simplyfy the README by making no difference for CLI invocation between Guice and spring profiles...
> 
>     * Also, packaging this change as part of upcoming 3.5.0 wouldcompose well with the ongoing documentation effort.

I think we can generalize this adoption yes, I don't see any reason to not to

chibenwa:
Fair enough then.

Don't you think the "convenience VS distroless benefits" is not the same for a DEMO image?

Maybe we can have that bash script only in this image foor the sake of easing the tutorials?

mbaechler:
> Fair enough then.
> 
> Don't you think the "convenience VS distroless benefits" is not the same for a DEMO image?
> 
> Maybe we can have that bash script only in this image foor the sake of easing the tutorials?

Oh sorry, I checked twice but failed at understanding it's for demo image.
Go ahead and merge it.

chibenwa:
Merged

dleangen:
Hi! Has this image been deployed? Is it safe for me to update the docs?

chibenwa:
> Hi! Has this image been deployed? Is it safe for me to update the docs?

No it will be part of 3.5.0, I will update your documentation once it is available.

dleangen:
Ok, thanks! By the way, was the image name changed? Or should I file an issue for that? Ideally it should be something simple like "james-demo" or "james-demo-server" or something like that.

What do you think?

dleangen:
Never mind: https://github.com/linagora/james-jenkins/pull/190

chibenwa:
> By the way, was the image name changed? 

It is in progress. I will ping you hen done. `james-demo` is great. See https://github.com/linagora/james-jenkins/pull/190 to change the name the linagora org relies on.

dleangen:
Thank you!!

PR #195, opened 2020-05-18
JAMES-3187 Add webadmin enabled by default to jpa-guice docker image

chibenwa:
Relates to https://github.com/apache/james-project/pull/191/commits/5905a1dfad8c4e97a7a0fb34b5201b66a23bec73#r426520378

### DIFF
--- a/dockerfiles/run/guice/jpa/destination/conf/webadmin.properties
+++ b/dockerfiles/run/guice/jpa/destination/conf/webadmin.properties
@@ -1,0 +1,48 @@
+#  Licensed to the Apache Software Foundation (ASF) under one
+#  or more contributor license agreements.  See the NOTICE file
+#  distributed with this work for additional information
+#  regarding copyright ownership.  The ASF licenses this file
+#  to you under the Apache License, Version 2.0 (the
+#  "License"); you may not use this file except in compliance
+#  with the License.  You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


chibenwa:
> I don't understand why we want jpa image to be insecure by default.

1. All other images enables Webadmin by default because...
2. ... as long as you don't expose the port the service is not available outside.

Note that port binding can be bound to a specific interface for instance loopback, allowing us to achieve both user convenience and safety. See https://github.com/apache/james-project/pull/196

Why do something different for JPA?

Should we disable WebAdmin everywhere instead?

mbaechler:
> > I don't understand why we want jpa image to be insecure by default.
> 
>     1. All other images enables Webadmin by default because...
> 
>     2. ... as long as you don't expose the port the service is not available outside.
> 
> 
> Note that port binding can be bound to a specific interface for instance loopback, allowing us to achieve both user convenience and safety. See #196

A bit optimistic ...
 
> Why do something different for JPA?
> 
> Should we disable WebAdmin everywhere instead?

... but if it's already the case, let's use it here too

chibenwa:
(Please note that along with this work, I open a ticket to add basic auth to webadmin, based on usersrepository.xml aministratorId  enabling us a easier aythentication setup and eventually addressing this issue)

chibenwa:
Merged

PR #198, opened 2020-05-22
[JAMES-3193] Updates to the Developer Guide

dleangen:
I have created this as a draft PR for this first iteration as per the [recent email thread](https://www.mail-archive.com/server-dev@james.apache.org/msg65814.html).

Please let me know if this is on the right track or not before I continue further.

Thanks!


docs/modules/dev/partials/domain/core-intro.adoc
```diff
@@ -0,0 +1,6 @@
+The xref:dev:domain/core/index.adoc[core component]s of James are:
+
+ * xref:domain/core/server.adoc[Server]
```
chibenwa, line 3: I am unsure that the "server" is a core component in itself, but rather the entity that assembles "core components" together.

IMO **mailQueue**, **taskManager**, **mailRepository**, **mailetContainer** makes more sense.

docs/modules/dev/partials/domain/core-intro.adoc
```diff
@@ -0,0 +1,6 @@
...
+ * xref:domain/core/mailets.adoc[Mailets]
```
chibenwa, line 4: A **mailet** is more like a plugin conforming to a standard API, but whose implementation can vary.

In my opinion the component here is more the **mailet container** that provides a "mail processing service".

docs/modules/dev/partials/domain/core-intro.adoc
dleangen: Thanks! These are exactly the kinds of comments I am hoping to get.

Would you mind making this comment in [the ongoing email thread](https://www.mail-archive.com/server-dev@james.apache.org/msg65859.html)?

I'd really like some clarity before moving forward. 😀

docs/modules/dev/pages/index.adoc
```diff
@@ -1,4 +1,12 @@
...
-(TODO)
+This section is aimed at developers, meaning anybody who needs to interact
+with the source code in any way.
+
+The resources available in this section are:
+
+ * xref:architecture/index.adoc[The system architecture]
+ * xref:domain/index.adoc[The domain model]
+ * xref:code/index.adoc[The code base]
+
```
chibenwa, line 12: I would add a section: James allow a user to register custom code for various plugins, allowing to extend the behaviour of the James server.

Is that what you meant by `External Components`?

docs/modules/dev/partials/domain/external-intro.adoc
```diff
@@ -0,0 +1,12 @@
+There are a number of xref:dev:domain/external/index.adoc[external components] that, 
+although not part of the core, are essential to make the 
+James system work as a whole:
+
+ * Mailet implementations
+ * Protocol-based applications
+ * Administration tools
+ * Storage implementations
+ * Mime4j
```
chibenwa, line 9: In my opinion, Mime4J and the like should be referenced as libraries and not as external components.

docs/modules/dev/partials/domain/core-intro.adoc
chibenwa: Sure

docs/modules/dev/pages/index.adoc
dleangen: Well, in the recent email thread, @mbaechler mentioned that the hexagonal architecture was an extremely important concept, or at least that is how I interpreted the message. I am trying to reconcile that statement with what I see. The [document about hexagonal architecture](https://alistair.cockburn.us/hexagonal-architecture/) that I referenced refers to "External Devices" as the things that connect to the adapters.

Also, this image mentions "External Projects", which are external to the core, so I just named them "External Components".

Should I try to fit all this with the hexagonal architecture, or should I drop the idea of that architecture? My feeling is that if it is used, it should be clarified and used in a way that is not confusing.

What do you think?

docs/modules/dev/partials/domain/core-intro.adoc
dleangen: That is normally how I would describe it as well, but I am trying to fit it in the context of the hexagonal architecture.

docs/modules/dev/partials/domain/external-intro.adoc
dleangen: To me, the vocabulary that I am used to (meaning that I am not claiming to be "right"), a "library" is something that gets embedded into a "component". Both are just ways of packaging up code, but the library is "disposable" (meaning that it is just part of the implementation and is not seem from the outside, nor cared about), while the "component" is something that is actually seen and worth understanding.

If users or maybe even developers of James don't have to care about Mime4j, then I would agree that it is a library. It would just be part of some other components, and would only come into scope when developing that component. So for this documentation project, I probably wouldn't even have to mention it at all.

docs/modules/dev/partials/domain/core-intro.adoc
chibenwa: Answered ;-)

dleangen:
Folded into #199, so closing.

PR #199, opened 2020-05-25
[JAMES-3187] Restructuring new documentation

dleangen:
Please see [related email thread](https://www.mail-archive.com/server-dev@james.apache.org/msg65873.html).

This is a work in progress.


dleangen:
I have completed the restructuring.

If this PR is accepted, I will continue in this direction.

docs/modules/ROOT/pages/index.adoc
```diff
@@ -1,15 +1,32 @@
...
+    and how it is designed.
+* xref:main:servers:index.adoc[James Servers]
+**    We provide a few out-of-the-box servers that you can 
+    choose from depending on your needs. Start here if you just
+    want to get up and running quickly with a working mail server.
+* xref:main:toolkit:index.adoc[James Toolkit]
+**    Describes how you can assemble your own specialized server by 
+    using the various James components as a toolkit.
+    This section is aimed at experienced James users.
+* xref:main:customization:index.adoc[Customizing James]
```
chibenwa, line 23: I would call this section `Customizing James servers` and would move it above the `toolkit` section

docs/modules/servers/pages/demo.adoc
```diff
@@ -0,0 +1,13 @@
+= Server Demo
+:navtitle: Demo
+
+The James Demo Server sets up  a light-weight, in-memory mail server
```
chibenwa, line 4: `in-memory` is false, docker do save stuff on the disk. Maybe "zero dependency"?

docs/modules/servers/pages/index.adoc
```diff
@@ -0,0 +1,92 @@
...
+tested and verified so you can get set up quickly with a production-grade
+mail server that you can use with confidence.
+
+The available James Servers are:
+
+ * <<demo,James Demo Server>>
+ * <<minimal,James Minimal Server>>
+ * <<basic,James Basic Server>>
+ * <<advanced,James Advanced Server>>
+ * <<distributed,James Distributed Server>>
```
chibenwa, line 15: IMO we should have a goal to 'limit' servers that we provide.

More servers = more work.

I make little distinction between minial & basic and advanced & distributed, that would get implemented by the way currently by the same technologies.

dleangen: Yeah, not a bad idea.

So what about offering only these 3:

 * Demo
 * Basic
 * Advanced

As was pointed out by @rouazana, both Basic and Advanced could have different "Flavours".

The advantage of having only 3 servers is that it would be much easier to understand for the users. It would likely be much less confusing.

If the "flavours" are eliminated, it would actually be less work for the James team. So maybe just keep what is currently the "distributed" flavour for advanced, and a relatively simple server for the Basic server (not distributed, just the basic email features that most users would expect).

Could that work?

chibenwa: Yes I think so :-)

docs/modules/ROOT/pages/index.adoc
dleangen: @chibenwa, to help me understand, could you please explain to me what, in your mind, is the difference between "customization" and "assembly"?

They both are a way of "changing" James to accomplish a specific agenda.

Customization is a way of "changing" James from the "outside". Assembly is a way of using existing parts to put them together to form a certain solution.

Customization can work on any assembly. (Is that true??)

Why would somebody want to do one over the other? Why do you suggest that customization be first on the list?

docs/modules/servers/pages/demo.adoc
dleangen: Yes, I like "zero-dependency". Maybe "command-line" would also be appropriate?

I am trying to convey that it is to be used on the command line just for demo purposes. It is not intended to be set up as a "real" server.

docs/modules/servers/pages/index.adoc
dleangen: Could these "flavours" be configured somehow? If I were a user who wanted to set up a "Basic Server" with, say, the filesystem instead of RDB (I know it's not supported, but just for the sake of an example), how would I, as the user, configure my "Basic Server" to get this result?

docs/modules/concepts/pages/mail/index.adoc
```diff
@@ -0,0 +1,18 @@
...
+to understand and decipher. Fortuneately, James has your back.
+
+We have built up our own view of emails that allows us to process messages in a
+consistent and comprehensible way.
+
+To help you understand how James approaches emails, we have divided this section
+up into the following parts:
+
+ * xref:mail/mime.adoc[Message Contents] (MIME)
+ * xref:mail/smtp.adoc[Message Envelopes] (SMTP)
```
chibenwa, line 15: Message Transport would seem more correct to me here

dleangen:
This branch includes changed from #192. Putting reference here in order to keep a link to the discussion contents.

docs/modules/concepts/pages/mail/index.adoc
dleangen: Good point!

dleangen:
This branch includes changes from #198. Putting reference here in order to keep a link to the discussion contents.

docs/modules/concepts/pages/mail/index.adoc
chibenwa: Also we should add a section about JMAP (https://jmap.io/) (but I agree to contribute it later on)

dleangen: I can add a placeholder now.

docs/modules/servers/pages/index.adoc
dleangen: Note: I have updated this part as suggested by @rouazana to use {Demo, Local, Redundant, Distributed} instead of {Demo, Minimal, Basic, Advanced, Distributed}.

docs/modules/concepts/pages/mail/index.adoc
dleangen: Made suggested changes, including additional of stub for JMAP.

docs/modules/servers/pages/demo.adoc
dleangen: Changed to "disposable".

docs/modules/concepts/pages/mail/mime/james.adoc
```diff
@@ -0,0 +1,13 @@
...
+:navtitle: James MIME
+
+While the xref:mail/mime/general.adoc[general description] of MIME is
+not sufficient for building a complex system like James, the
+xref:mail/mime/technical.adoc[technical specifications] are unfortunately
+very messy and overly-complex due to their history and the context in which
+they were developed.
+
+To make dealing with emails possible and practical, James has defined its own
+version of what it means to be an email.
```
dleangen, line 11: Could somebody kindly tell me what this is?

How **does** James define a MIME Message?

The subject was touched on by @chibenwa [in this thread](https://www.mail-archive.com/server-dev@james.apache.org/msg65869.html).

docs/modules/ROOT/pages/index.adoc
rouazana: Customization is available on pre-built products as far as assembled products. It mainly some XML configuration. In some cases in can require to write one or 2 java classes (mailet / matcher) and put them in a directory to accomplish a particular customized workflow.

Assembly need to create a new Guice product, choose a particular set en Guice modules, and build this particular product. It's more for developpers than for experienced users (note that it was not the case with Spring, were with some XML magic you could reach the same objective).

docs/modules/concepts/pages/mail/index.adoc
```diff
@@ -0,0 +1,19 @@
+= Email Domain Model
+:navtitle: Emails
+
+The https://en.wikipedia.org/wiki/History_of_email[history of email] has resulted
+in a convoluted patchwork of standards and practices, which can be very difficult
+to understand and decipher. Fortuneately, James has your back.
```
rouazana, line 6:
```suggestion
to understand and decipher. Fortunately, James has your back.
```

docs/modules/concepts/pages/mail/mime/general.adoc
```diff
@@ -0,0 +1,33 @@
...
+Often, postal mail (i.e. a letter) is used as an analogy for email. You write a message
+on a piece of paper, put the paper in an envelope, and drop the envelope into a letterbox
+to be picked up by the post office. There are indeed some similarities with postal mail, 
+but there are also some important
+differences. For instance, when you write a letter, you are not required to annotate it
+with meta information. You could include a date, your address, and other information, and
+indeed, it is customary to do so. However, your letter would not be rejected if you 
+failed to include this this type of descriptive information. It would also be strange 
+not to include a message when writing a letter.
+A MIME message is almost the opposite. It *must* include a number of valid headers,
```
rouazana, line 26: are you sure? which headers are mandatory?

docs/modules/concepts/pages/mail/mime/general.adoc
```diff
@@ -0,0 +1,33 @@
...
+and the inclusion of a body is optional.
+Nonetheless, it is still a useful comparison.
```
rouazana, line 28: another difference is that without some security, the envelop is transparent and anyone carrying it can see the letter inside.

docs/modules/concepts/pages/mail/mime/index.adoc
```diff
@@ -0,0 +1,11 @@
+= Message Contents (MIME)
+:navtitle: MIME
+
+What people casually call "email" usually refers more specifically to a
+test message written in a specified format called https://en.wikipedia.org/wiki/MIME["MIME"].
```
rouazana, line 5:
```suggestion
text message written in a specified format called https://en.wikipedia.org/wiki/MIME["MIME"].
```

docs/modules/concepts/pages/mail/mime/james.adoc
rouazana: I'm not sure it's so important in an introduction. For me it's meaningful only for customizing James.

docs/modules/concepts/pages/mail/mime/technical.adoc
```diff
@@ -0,0 +1,109 @@
...
+https://tools.ietf.org/html/rfc822[RFC822] ("Standard for the Format of ARPA Internet Text Messages")
+was the original standard that defined the format of an email. It was obsoleted by
+<<RFC2822>>. The definition of an email under this standard was an attempt to take the lessons
+learned from the ARPANET and extend the use of text messaging to a broader context.
+
+Electronic mail messages are defined as having contents and an envelope. The contents
+consist of header fields and, optionally, a body. The body is nothing more than
+a (potentially empty) sequence of lines of text.
+
+Although this sounds like a an extremely simple concept, to get two completely separate systems
```
rouazana, line 25:
```suggestion
Although this sounds like an extremely simple concept, to get two completely separate systems
```

docs/modules/ROOT/pages/index.adoc
dleangen: Switched the order of the sections as suggested and updated the contents consequently.

docs/modules/concepts/pages/mail/mime/james.adoc
dleangen: Point taken... but I feel that I still need to understand. :-)

docs/modules/concepts/pages/mail/mime/general.adoc
dleangen: Not sure at all. I will check. :-)

docs/modules/concepts/pages/index.adoc
```diff
@@ -0,0 +1,29 @@
...
+xref:mail/index.adoc[domain model for email].
+
+
+[#processing]
+== James Email Processing
+
+There are many ways to peel a potato. James has its own particular
+xref:processing/index.adoc[model for processing mail],
+based mostly on the concept of a "Mailet". The idea of Mailet-based
+processing was heavily inspired by the https://en.wikipedia.org/wiki/Java_servlet[Servlet]
```
mbaechler, line 28: funnily, it's not, in my opinion, despite what the name and some documentation states

It could be matched to Server Filter but in that the Matcher/Mailet does not match.
And you don't really have a type for the final processor

It's not important but maybe we could just not say it

chibenwa: Well, I had that discussion with Danny Angus at ApacheCON Europe 2019, hence I have prime historical data :-)

Here I would suggest a vote to re-evaluate the goal as a PMC to deliver a standard for mail processing and the opportunity to retro-fit mailet-api as a James server extension mechanism.

docs/modules/concepts/pages/mail/mime/general.adoc
dleangen: I think this is now outdated, so I will resolve the comment.

docs/modules/concepts/pages/mail/mime/general.adoc
dleangen: I think this is now outdated, so I will resolve the comment.

docs/modules/concepts/pages/mail/mime/james.adoc
dleangen: This code is now outdated, so I will resolve the comment.

dleangen:
Hello! I think now may be a good time to commit these changes, pending acceptance of this restructuring.

I would like to start using more fine-grained PRs for the various sections instead of continuing to update this PR.

Thanks!

chibenwa:
> Hello! I think now may be a good time to commit these changes, pending acceptance of this restructuring.
> 
> I would like to start using more fine-grained PRs for the various sections instead of continuing to update this PR.

Let me help you then ;-)

chibenwa:
This work is now merged. Can you close the pull request?

dleangen:
Thank you!!

PR #200, opened 2020-06-01
[JAMES-3187] Update the "Concepts/Protocols" section of the new documentation

dleangen:
Within the new document structure, the "Concept" section requires numerous updates.

*This is a work in progress, and is currently set to "draft" mode.*


docs/modules/concepts/pages/glossary.adoc
```diff
@@ -0,0 +1,31 @@
...
+xref:mail/messages/imf.adoc[IMF]::
+*Internet Message Format* (IMF) is the base specification that defines the
+syntax of an email message.
+
+IMAP::
+
+JMAP::
+
+xref:mail/messages/mime.adoc[MIME]::
+*Multipurpose Internet Mail Extension* (MIME) is a specification that extends
```
mbaechler, line 12: we should link to RFCs here in my opinions. what do you think?

docs/modules/concepts/pages/glossary.adoc
```diff
@@ -0,0 +1,31 @@
...
+IMF, defining most of the "advanced" features of email.
+
+MTA::
+A *Mail Transfer Agent* (MTA) is a host that is capable of providing a mail
+transport service compliant with the SMTP protocol.
+
+MUA::
+A *Mail User Agent* (MUA) is either an agent that hands over a message to an initial
```
mbaechler, line 20: Well, I think that it's not clear from that description it's what people uses, it's the visible part of the email infrastructure

docs/modules/concepts/pages/glossary.adoc
dleangen: It does link, but indirectly. The original idea is to link to our description of MIME, which has links to the RFCs. There are many RFCs, so providing a single link is not as easy as it sounds. I didn't want the glossary description to be too long. The intent was to make it more like a jumping point to a more in-depth description, with just a short text here.

docs/modules/concepts/pages/glossary.adoc
dleangen: Agree. I'm not sure where to link to. The only description in the text so far is in the SMTP section.

The idea of the glossary is to provide only a very short-basic description with a link to a more in-depth description. I need to figure out if MUA is "important" enough to warrant its own section.

docs/modules/concepts/pages/protocols/smtp.adoc
```diff
@@ -71,6 +71,67 @@ SMTP relates to:
 
 == Try It!
 
-Add a section to the demo, and connect manuall via an SMTP session
+You can connect manually to the James SMTP server using telnet.
```
mbaechler, line 74: I don't really like that section: one should not run non-starttls servers and if you configure yours to use starttls you won't be able to test with telnet.
There's a solution with openssl to handle that situation but it makes things more complex than it's worth IMO.
What about providing a test tool instead?

dleangen: Note that this is *only* for the demo server, which should never be used in production.

mbaechler: I could not find why it's demo server only from the structure or the text in this file.

dleangen: It is implied in the "Requirements" list: "Run the long demo (without stopping)". (That item still needs a link, of course.)

Can you suggest something else I could do to make it more obvious?

mbaechler: Ok, I see now, I would say:

```suggestion
You can try speaking SMTP with your demo server by connecting to it using telnet. It would not work on a real world server as it would requires encrypted traffic. See "using openssl to talk to your server"
```

And I would write that section about using openssl (:

I can help if needed for that section, don't hesitate to ask

dleangen:
This is of course still a work in progress, but I am going to merge this section for now.

PR #201, opened 2020-06-05
JAMES-3187 Updates to "Concepts/Storage" section of new documentation

dleangen:
Updates to the Storage section.

Would be great to get some help with how James handles storage.

Please also comment on any concepts that I have missed, or anything that may not be correct.

Thanks!


ieugen:
Helllo David, 

I like the history lesson. I think we should mark that as History. I believe most of those RFC's are obsolete. I also believe (but did not check) that some RFC's that James implements are obsolete by other RFC's. Especially in the email format. 

Regarding your contributions I think we should add it. We can always improve it later. I can help you with more technical docs about the storage part. 

Did you check out this information? https://james.apache.org/mailbox/index.html .
We can start by migrating that first and then improve on it. 
The sources are here https://git.apache.org/repos/asf?p=james-site.git;a=tree;f=www;h=1c66d9f2f309f030a75a3bb7970c0ed8d6f19638;hb=HEAD .

I know there is this list here that should be migrated: https://james.apache.org/server/rfcs.html .
I would also like to improve it. Make it a table and mention what is implemented in James and from which version.

docs/modules/concepts/pages/storage/index.adoc
```diff
@@ -0,0 +1,112 @@
...
+   messages on an IMAP Server
+ * *Email clients*, as the client will store mail locally, usually in the form
+   of a "mailbox"
+ * *xref:protocols/smtp.adoc[SMTP]* as the protocol is related to transmitting
+   messages from one mailbox to another
+
+
+== James Model
+
+(TODO)
```
chibenwa, line 110: Should this speak of the mailbox-api, & the various implementation we have of it?

I would be happy to write such a section.

dleangen: That would be really nice if you could give it a go!

dleangen: I will add this information to a different section, and keep this section reserved for more general concepts.

dleangen:
> Did you check out this information? https://james.apache.org/mailbox/index.html

I do not find it useful, and even if it were useful, it does not belong in this section.

> I know there is this list here that should be migrated: https://james.apache.org/server/rfcs.html
> I would also like to improve it. Make it a table and mention what is implemented in James and from which version.

I don't mind adding this in some other section, but the list in its current form is not useful.

A list showing which RFCs are actually implemented could be useful, and even more so if it can actually point to the right places in the code.

dleangen:
Ready for review before merging. Will wait a few days for comments. (Even after merge we can always update at any time.)

docs/modules/concepts/pages/configuration.adoc
```diff
@@ -0,0 +1,34 @@
...
+
+== Overview
+
+An important aspect to understand about James is the idea of "configuration".
+Sometimes "configuration" means different things to different people, so we
+wanted to ensure that what we mean by the concept is clear and precise, 
+because in order operate a James server, you will necessarily need to 
+configure it.
+
+A "*Configuration*" is a https://martinfowler.com/bliki/ValueObject.html[value object]
```
mbaechler, line 11: It's not factually wrong that it's a Value Object but I think it's not a good description of it. Value Object usually represents a single value (in Martin Fowler article, a Coordinate is a single value for example).
Configuration is IMO an immutable data structure.

docs/modules/concepts/pages/configuration.adoc
```diff
@@ -0,0 +1,34 @@
...
+that communicates to a James server how the xref:user/index.adoc#_operator[Operator] wishes the
+server to behave. It is very similar to an interface, but yet it is not quite an interface
```
mbaechler, line 13: I don't find that at all similar to an interface or I don't understand what you mean by "interface"

docs/modules/concepts/pages/configuration.adoc
```diff
@@ -0,0 +1,34 @@
...
+(which is why we call it something different).
+
+An "interface" provides a means for an Operator to interact with a server as part of the
+normal course of ongoing operations. Interfacing with the system essentially means either
+reading the server's current state, or updating the server's current state. A typical example
+would be adding a new mailbox via the xref:concepts:???.adoc[Useradmin API]. Each time the Operator
+interacts with the system via an interface, although the *state* of the system may change, 
+the *behavior* of the system is not expected to change.
+
+By contrast, "configuration" of a system (also a noun, but yet not the same as the
```
mbaechler, line 23: sometimes the difference is artificial: for example you can, with some implementation, define what are your domains in the configuration. For others, it's part of the server state.

docs/modules/concepts/pages/configuration.adoc
dleangen: I know you know this, but just to establish a baseline for understanding...

Value objects and immutable objects can be used to have the same effect. In one case, you update the same instance of the object (immutable), in the other case the entire object gets replaced with a fresh new instance, so any given instance, after instantiation, never changes value.

Often the immutable object is called an "entity". The main difference between an entity and a value object is that an entity has an ID, and is identifiable.

The configuration systems I know use the latter method, i.e. it is an "anonymous" and disposable object that just carries the desired value, not an entity, with a lifecycle and identity.

I have trouble imagining why one wouldn't use value objects for configuration.

So how do we resolve our differences of opinion?

(BTW, I'm glad I brought it up here, because I think understanding configuration is actually quite important.)

docs/modules/concepts/pages/configuration.adoc
dleangen: Ok, I'll give it a shot.

An "interface" is essentially, conceptually speaking, some "surface" between which the inside of a system communicates with the outside. Except for an interface, normally it should not be possible to communicate with the system. Maybe this dates me a bit, but take the example of a DVD player, and for this example pretend we lost the remote. The "interface" is the collection of buttons and stuff that you would press. It is in this case literally the surface that connects the inside of the machine from the outside.

A configuration also connects the inside from the outside, but it is used for a different purpose and has a different lifecycle.

Does that make sense?

docs/modules/concepts/pages/configuration.adoc
dleangen: Perhaps, but allow me to dig in to your example.

I would argue that the state of the domains (i.e. the list of domains that are in the system) can be seeded by an "initial list". This initialDomainList is indeed a configuration, as you mention, but it does not represent the state. Rather, the state gets initialized with the initialDomainList.

So although they are related, in my mind the different roles are quite clear. The domainList itself is the state, while the initialDomainList is just a configuration that seeds the state.

But I think it's good to test this concept further.

Perhaps you have other examples that may invalidate or challenge this concept?

docs/modules/concepts/pages/configuration.adoc
ieugen: If we step back a bit, are these extra notions really necessary or are they creating more confusion by building more and more context?

dleangen: Which "extra notions" are you referring to?

I don't think you mean the idea of configuration itself, because I don't think you mean either of:
 - We remove all configuration from James 👎🏻
 - We provide a means of interacting with the system that either we don't fully understand or we can't explain 👎🏻

Can you explain a bit more?

docs/modules/concepts/pages/storage/users.adoc
```diff
@@ -0,0 +1,59 @@
+= User Directory
+:navtitle: Users
+
+== Overview
+
+Whereas xref:storage/mailbox.adoc[mailboxes] are clearly a core component of
```
chibenwa, line 6: We might be defining more clearly what a user is.

I propose:

`As user is someone entitled to receive, store and retrieve email on this given server.`

docs/modules/concepts/pages/storage/users.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+=== User Management
+
+A User has at least one mail box, and will usually have a canonical mailbox
+that matches the user's identity in the system.
+For each domain that is managed by the system, it is common to be able to:
+
+* List all users for the domain
+* Add a user
+* Remove user
```
chibenwa, line 47: * Change the password of a user

docs/modules/concepts/pages/storage/users.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+A User has at least one mail box, and will usually have a canonical mailbox
+that matches the user's identity in the system.
+For each domain that is managed by the system, it is common to be able to:
+
+* List all users for the domain
+* Add a user
+* Remove user 
+
+
```
chibenwa, line 49: == Alias and redirections

 * Add/remove email aliases for a given user
 * Add/remove email forwards for a given user
 * Add/remove a user to a "group" mail address
 * Other more advanced kind of Mail address rewritting are also available.

(I wrote that before seeing you mentioned it below.)

docs/modules/concepts/pages/storage/users.adoc
```diff
@@ -0,0 +1,59 @@
...
+=== Other Common Operations
+
+It is common to allow user's to have one or more
+https://en.wikipedia.org/wiki/Email_alias[aliases].
+
+Some systems provide an https://en.wikipedia.org/wiki/Autoresponder[autoresponder]
+(or "vacation") facility to its users.
```
chibenwa, line 56: This "auto responder" is either implemented via SIEVE scripts, or via JMAP. Maybe it should be coming after SIEVE scripts.

docs/modules/concepts/pages/storage/users.adoc
```diff
@@ -0,0 +1,59 @@
...
+are some operations that are relatively common to any email system.
+
+=== Domain Management
+
+A mailbox belongs to a https://en.wikipedia.org/wiki/Domain_name[domain]. For a mail
+system that can process multiple domains, the system will usually have operations to:
+
+* List available domains
+* Add a domain
+* Remove a domain
```
chibenwa, line 36: * Add a domain alias

(Email sent to xxx@alias.org will be rewritten to xxx@original.org)

docs/modules/concepts/pages/storage/users.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+== Overview
+
+Whereas xref:storage/mailbox.adoc[mailboxes] are clearly a core component of
+any email system, user management is a lighter shade of gray.
+
+On the one hand, mailboxes ought to match up with actual users, so there is
+a clear relationship between mailbox management and user management.
+Because of this relationship, it is not at all strange to include a user directory
+directly within an email system.
```
chibenwa, line 12: We need to mention that "user management" is abstracted with an API, and can be configured to follow an external user management tool.

IE there is an LDAP implementation for this.

docs/modules/concepts/pages/storage/users.adoc
```diff
@@ -0,0 +1,59 @@
...
+
+However, on the other hand email is only one of many applications that require
+user management. It is therefore also completely reasonable to use some kind of
+centralized user directory, and ensure that mailboxes somehow synchronize with
+this external system.
```
chibenwa, line 17: `and ensure that mailboxes somehow synchronize with
this external system.`

This somehow you are looking for is LDAP as a UsersRepository implementation ?

chibenwa: Of course we need to mention that JAMES will "read" the LDAP anuary, not modify it.

docs/modules/concepts/pages/storage/index.adoc
```diff
@@ -1,4 +1,34 @@
...
+The system of storage must work intimately with the various 
+xref:protocols/index.adoc[protocols].
+
+In this section, we describe the two principle types of storage that James manages:
+
+ * xref:storage/mailbox.adoc[Mailboxes] for email storage, and
+ * xref:storage/users.adoc[User Directory] for data about Mailbox users.
+
+James also optionally includes an
+xref:storage/indexing.adoc[optional indexer] for providing a better search experience.
```
chibenwa, line 34: This should be mentioned as a sub-part of mailbox IMO

docs/modules/concepts/pages/storage/users.adoc
```diff
@@ -0,0 +1,59 @@
+= User Directory
```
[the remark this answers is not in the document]
dleangen: Ok, I will make that change.

docs/modules/concepts/pages/storage/users.adoc
dleangen: Ok. I will check the description of User in the "User Model" section, update if necessary, and link that that. Would that be ok?

docs/modules/concepts/pages/storage/users.adoc
dleangen: Right!

docs/modules/concepts/pages/storage/users.adoc
dleangen: This list is more detailed, so maybe I'll take this instead (and move it down to the section below).

docs/modules/concepts/pages/storage/users.adoc
dleangen: Ok. I was kinda grasping at straws by this point anyway. :)

docs/modules/concepts/pages/storage/users.adoc
dleangen: Ok, thanks!

docs/modules/concepts/pages/storage/users.adoc
dleangen: Yes, I agree. However, my intention for the "Concepts" section is to try to remain general, then have a "James specific" section later that is more specific to James.

docs/modules/concepts/pages/storage/index.adoc
dleangen: Are you suggesting that indexing is **only** for mailboxes? (That would be good to know, as it has not been clear to me.) In that case, then yes, I would agree.

docs/modules/concepts/pages/storage/users.adoc
dleangen: Again, my intent here is to be more general. How James works would be left for a different section. That is my thought now, anyway. We'll see how that actually works out.

docs/modules/concepts/pages/storage/index.adoc
```diff
@@ -1,4 +1,34 @@
...
+
+Email is useful because any one person can send a message to any other person,
+anywhere in the world. The heart of email is the ability to transfer some message
+based on a  commonly-understood message format between two different systems that
+may have been developed entirely independently of each other. The concept is relatively
+simple, but the fact that email is so ubiquitious is simply a miracle.
+
+However, there is more to email than just the transfer of messages.
+
+Although it is possible to read a message, then discard it, never to be seen again,
```
chibenwa, line 14: Email is by design asynchronous. People need to be able to send mail when I'm climbing, I need to review it afterward. 

In the meantime it needs to be stored somewhere.

docs/modules/concepts/pages/storage/index.adoc
```diff
@@ -1,4 +1,34 @@
-= Mailboxes and Email Storage
+= Storage
```
chibenwa, line 1: Mail storage in mailboxes makes sens for a mail delivery agent aka IMAP / POP3 / JMAP server.

However it don't make sense for mail servers that only relay emails.

This consideration IMO needs to appear here.

docs/modules/concepts/pages/storage/index.adoc
chibenwa: Well, maybe it could make sense for other usages but so far in james it is intimately linked to the mailboxes (it is a part of the mailbox-api) hence my suggestion.

docs/modules/concepts/pages/storage/users.adoc
chibenwa: > Again, my intent here is to be more general. 

Overall +1. Sorry I failed at extracting the general part of this statement, but I guess there's still something to get out of it.

docs/modules/concepts/pages/storage/index.adoc
dleangen: I agree, and that is kinda the point I am trying to make. Your example is a good one. I could add something like that. In any case I'll think about how to update this so it makes more sense.

docs/modules/concepts/pages/storage/index.adoc
dleangen: Ok, good point.

docs/modules/concepts/pages/storage/index.adoc
dleangen: Ok, thanks.

docs/modules/concepts/pages/storage/users.adoc
dleangen: Probably it needs to be explained better. The idea is to make a gentle introduction into the various concepts so that when we later explain how James works, it won't seem so alien. At least, that is the idea. Since this is the first iteration, like many parts could be greatly improved.

docs/modules/concepts/pages/configuration.adoc
mbaechler: I had a second thought and Value Object is more precisely "a datastructure for which identity is its entire content", if 2 value objects has the same content they should be considered not only equals but the same.
So I'm ok with this use of Value Object

docs/modules/concepts/pages/configuration.adoc
mbaechler: I have a good one that breaks the initial vs living state view: Look at the wonderful configuration entry :
```
# Use to distribute the emails of a given slice within your cassandra cluster
# A good value is 2*cassandraNodeCount
# This parameter can only be increased.
# Optional, default 1
mailqueue.view.bucketCount=1
```
This configuration will change the state of the server by adding buckets. You can increase but never decrease this count.

To implement that we have to keep track of the previous values and check if an invariant is broken when this configuration key change.

We could very well have implemented that we a default value / REST API couple and it would still be configuration.

dleangen: Yeah, that's a good one. So does this invalidate the concept? Or should we consider this as an edge case and live with it?

docs/modules/concepts/pages/storage/users.adoc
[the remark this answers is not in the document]
dleangen: I have made the change in the docs for consistency, and I filed the issue [JAMES-3360](https://issues.apache.org/jira/browse/JAMES-3360).

docs/modules/concepts/pages/storage/users.adoc
dleangen: Done

docs/modules/concepts/pages/storage/users.adoc
dleangen: Done

docs/modules/concepts/pages/storage/users.adoc
dleangen: Updated

docs/modules/concepts/pages/storage/users.adoc
dleangen: Information about aliases was updated later in the text.

docs/modules/concepts/pages/storage/users.adoc
dleangen: This information is indeed necessary. As a solution, I am going to leave the text as is here in the concepts, and instead include this information in the more technical section.

docs/modules/concepts/pages/storage/users.adoc
dleangen: I updated the text. Hopefully this works.

docs/modules/concepts/pages/storage/users.adoc
dleangen: I'm going to leave this as is for now. I will be doing another review later.

docs/modules/concepts/pages/storage/index.adoc
dleangen: I removed the indexing.adoc page and moved the (edited) text as a subsection of Mailbox.

docs/modules/concepts/pages/storage/index.adoc
dleangen: Updated text using this example.

docs/modules/concepts/pages/storage/index.adoc
dleangen: This change has already been made.

dleangen:
I would like to merge this PR. Please let me know if you have any objections. I will go ahead by tomorrow morning otherwise, as the risk is very low and the text can always be changed.

PR #202, opened 2020-06-05
Feature/james 3159 web key directory protocol

ManuelB:
Hi guys,
I implemented a proof-of-concept for the Web Key Directory protocol (https://tools.ietf.org/id/draft-koch-openpgp-webkey-service-07.html) including the possibility to submit a key using the auth-submit policy. This means implementation of the confirmation email is still missing.

A corresponding client is implemented here:
https://github.com/ManuelB/patientcare/blob/master/frontend/patientcare/manager/Onboarding.js

The demo is currently hosted here but I can't promise that I will host it forever:
https://patientcare.incentergy.de/

A video can be seen here:
https://www.youtube.com/watch?v=pQreLOkn6lM

Unfortunately I don't have the time to raise the quality of the code to the requested contribution guidelines: https://james.apache.org/contribute.html

Currently at least the following is missing:
 * Confirmation email workflow
 * Javadoc for all methods
 * Test Cases for all methods
 * Documentation for installation
 * More persistence implementation like JPA etc. 

Nevertheless I wanted to create this pull request maybe this is useful for somebody else.

Thanks a lot
    Manuel


mbaechler:
Even if not mergeable because of missing parts, your work is very interesting, thank you for demonstrating it's doable.

ManuelB:
The main feature is the complete interaction free set up of GPG encryption for a person that enters the web site. The following happens in the client the first time somebody comes to the website:

 # Generation of an OpenGPG Key Pair that is saved in the local storage
 # Creation of a James Mailbox using the  admin api using the fingerprint of the public key as the local part in the email address and the sha256 sum of the private key as password e.g. EMail: ef076a5103d46d4c00811aecc1949584bdfbfc53@patientcare.incentergy.de Password: e90c0813a17e3785e330c8677d7f80562016468f7a72e55bb5e980eba0b375b1
 # Connecting the client via the JMap protocoll to the inbox
 # Sending an public key submission email via JMap to the server
 # The server saves the public key
 # The server publishes the key under the correspondig URL

After that it is possible to send and received encrypted emails with the created inbox and email address.

server/container/guice/memory-guice/src/main/java/org/apache/james/MemoryJamesServerMain.java
```diff
@@ -100,6 +101,9 @@
         new JmapTasksModule(),
         new MemoryDataJmapModule(),
         new JMAPDraftServerModule());
+    
+    public static final Module WKD = Modules.combine(
```
chibenwa, line 105: You likely don't need to combine a single module

server/container/guice/protocols/wkd/src/main/java/org/apache/james/wkd/WebKeyDirectoryModule.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.james.wkd;
```
chibenwa, line 1: Most of the files here miss the Apache license header

server/protocols/wkd/src/main/java/org/apache/james/wkd/store/WebKeyDirectoryStore.java
```diff
@@ -0,0 +1,11 @@
+package org.apache.james.wkd.store;
+
+public interface WebKeyDirectoryStore {
```
chibenwa, line 3: This should likely be in `server/data/data-wkd`

Implementation of this class might need to be extracted in other maven modules like `server/data/data-wkd-memory`

server/container/guice/protocols/wkd/src/main/java/org/apache/james/modules/protocols/WebKeyDirectoryServerModule.java
```diff
@@ -0,0 +1,98 @@
...
+        routesBinder.addBinding().to(SubmissionAddressRoutes.class);
+    }
+
+    @ProvidesIntoSet
+    InitializationOperation startWebKeyDirectoryServer(WebKeyDirectoryServer server,
+        WebKeyDirectoryConfiguration configuration) {
+        return InitilizationOperationBuilder.forClass(WebKeyDirectoryServer.class).init(() -> {
+            if (configuration.isEnabled()) {
+                server.start();
+                registerPEMWithSecurityProvider();
```
chibenwa, line 71: What happens if the server receives requests before the PEM security info are registered? Should we switch them?

server/data/data-wkd/src/main/java/org/apache/james/wkd/store/PublicKeyEntry.java
```diff
@@ -0,0 +1,115 @@
...
+public class PublicKeyEntry {
+    /**
+     * The so mapped local-part is hashed using the SHA-1 algorithm. The resulting
+     * 160 bit digest is encoded using the Z-Base-32 method as described in
+     * [RFC6189], section 5.1.6. The resulting string has a fixed length of 32
+     * octets.
+     */
+    private String hash;
+    private Status status;
+    private byte[] publicKey;
```
chibenwa, line 22: It could be interesting to have final fields to make this POJO more immutable?

server/pom.xml
```diff
@@ -96,6 +97,7 @@
         <module>protocols/protocols-smtp</module>
         <module>protocols/webadmin</module>
         <module>protocols/webadmin-integration-test</module>
+	<module>protocols/wkd</module>
```
chibenwa, line 100: There is an indent typo :-)

server/protocols/wkd/src/main/java/org/apache/james/wkd/WebKeyDirectoryServer.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Inject
+    public WebKeyDirectoryServer(WebKeyDirectoryConfiguration configuration,
+        Set<WebKeyDirectoryRoutes> webKeyDirectoryRoutes, FileSystem fileSystem,
+        org.apache.james.server.core.configuration.Configuration jamesConfiguration) {
+        this.configuration = configuration;
+        this.webKeyDirectoryRoutes = webKeyDirectoryRoutes;
+        this.server = Optional.empty();
+        this.fileSystem = fileSystem;
+        this.configurationPrefix = jamesConfiguration == null ? ""
```
chibenwa, line 66: When is jamesConfiguration null?

server/protocols/wkd/src/main/java/org/apache/james/wkd/WebKeyDirectoryServer.java
```diff
@@ -0,0 +1,148 @@
...
+            // Load certificate chain
+            Certificate[] chain = keyStore.getCertificateChain(alias);
+
+            SslContextBuilder sslContextBuilder = SslContextBuilder.forServer(myPrivateKey,
+                new X509Certificate[] { (X509Certificate) chain[0] });
+            return sslContextBuilder;
+        } catch (KeyStoreException | NoSuchAlgorithmException | CertificateException | IOException
+            | UnrecoverableEntryException e) {
+            e.printStackTrace();
+            return null;
```
chibenwa, line 136: I would suggest propagating the error rather than returning null.

server/protocols/wkd/src/main/java/org/apache/james/wkd/WebKeyDirectoryServer.java
```diff
@@ -0,0 +1,148 @@
...
+                .of(HttpServer.create().secure((spec) -> spec.sslContext(sslContextBuilder))
+                    .port(configuration.getPort().map(Port::getValue).orElse(RANDOM_PORT))
+                    .route(routes -> webKeyDirectoryRoutes
+                        .forEach(webKeyDirectoryRoute -> webKeyDirectoryRoute.define(routes)))
+                    .wiretap(wireTapEnabled()).bindNow());
+        }
+    }
+
+    private SslContextBuilder createSslContextBuilder() {
+        if (fileSystem == null) {
```
chibenwa, line 91: When does the filesystem is null?

server/protocols/wkd/src/main/java/org/apache/james/wkd/mailet/WebKeyDirectoryMailet.java
```diff
@@ -0,0 +1,185 @@
...
+import org.bouncycastle.openpgp.PGPUtil;
+import org.bouncycastle.openpgp.jcajce.JcaPGPObjectFactory;
+import org.bouncycastle.openpgp.jcajce.JcaPGPPublicKeyRingCollection;
+import org.bouncycastle.openpgp.operator.PublicKeyDataDecryptorFactory;
+import org.bouncycastle.openpgp.operator.bc.BcKeyFingerprintCalculator;
+import org.bouncycastle.openpgp.operator.bc.BcPublicKeyDataDecryptorFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class WebKeyDirectoryMailet extends GenericMailet {
```
chibenwa, line 39: We generaly write a piece of java doc explaining what the mailet does and also detailing how to configure it. A maven plugin then generate built in plugin documentation from the code.

server/protocols/wkd/pom.xml
```diff
@@ -0,0 +1,219 @@
...
+		<groupId>org.apache.james</groupId>
+		<artifactId>james-server</artifactId>
+		<version>3.5.0</version>
+		<relativePath>../../pom.xml</relativePath>
+	</parent>
+
+	<artifactId>james-server-wkd</artifactId>
+	<packaging>jar</packaging>
+
+	<name>Apache James :: Server :: Web Directory Protocol</name>
```
chibenwa, line 27: Can we add a description field linking the specification?

chibenwa:
Thanks again for the proof of concept.

This work had not been touched for over 2 years, I'll assume it will stay... a proof of concept. As such I close the related pull request.

Don't hesitate to reopen if I am wrong. 

To get this further than a POC we could:
 - Have a specialized James distribution as a web key directory
 - Or repackage it as an extension.
 
 If you are interested in any of those, please re-open.

PR #204, opened 2020-06-14
JAMES-3187 Updates to "Demo Server" section of new documentation

dleangen:
Work in progress


dleangen:
Hello!

Is there generally anything else that should be added about the "Demo Server"? Or can we merge and move on to other things?

dleangen:
Merging, but still a work in progress so comments are always welcome.

PR #205, opened 2020-06-14
JAMES-3187 Updates to "Concepts/Processing" section of new documentation

dleangen:
Work in progress


docs/modules/concepts/pages/processing/elements/index.adoc
```diff
@@ -0,0 +1,13 @@
...
+
+== Overview
+
+If we take the view of a mail server as a mail processing engine, then
+we can list some of the elemental processes that a typical mail server
+will likely perform:
+
+ * xref:processing/elements/receiving.adoc[Receiving] email from a remote system
+ * xref:processing/elements/queuing.adoc[Queuing] and spooling emails
+ * xref:processing/elements/chaining.adoc[Chaining] processes
```
chibenwa, line 12: I don't understand what "chaining means" maybe we should keep "Mail Processing' wording to fit in the James wording?"

docs/modules/concepts/pages/processing/elements/receiving.adoc
```diff
@@ -0,0 +1,35 @@
...
+not as obvious, email can also be received via <<IMAP>>.
+
+
+
+=== SMTP
+
+The most obvious and easily-understood method that an email server handles "email reception"
+is via an SMTP session. When receiving email, the server is acting in the role of an SMTP server,
+responding to requests by a remote SMTP client.
+
```
chibenwa, line 18: == LMTP

docs/modules/concepts/pages/processing/elements/receiving.adoc
```diff
@@ -0,0 +1,35 @@
...
+must still be delivered to a local xref:concepts/storage/mailbox.adoc[mailbox]. 
+xref:protocols/smtp.adoc#RFCXXXX[RFCXXXX] does not mandate any particular protocol for local
+email handling. The system is free to use any method at its discretion. In the case of James,
+mail is delivered locally simply via an in-process Java method call, making it simple and
+extremely fast.
+
+
+=== IMAP
+
+Although less common, it is possible for an email server to receive emails via IMAP. This is
```
chibenwa, line 32: IMO we should not call that receiving as such "Messages" (and not email) do not undergo the mail processing pipelines.

Mentionning it here IMO brings confusion.

docs/modules/concepts/pages/processing/elements/index.adoc
dleangen: I am having trouble with this one. In order to explain what "processing" means, I have broken it down into the "elements" that I was able to identify. If I try to change "chaining" to "email processing", not only does it take the level of abstraction up to the same level I am trying to define (which, in order to define requires breaking it down), it also creates a circular definition.

I will try to actually fill in the "Chaining" section. Perhaps after reading it you could comment again? If you still think the same way, I will try to to conceive of a different approach.

docs/modules/concepts/pages/processing/elements/receiving.adoc
dleangen: Right! I will add this subsection.

docs/modules/concepts/pages/processing/elements/receiving.adoc
dleangen: Ok, I have removed it.

docs/modules/concepts/pages/processing/elements/receiving.adoc
dleangen: Actually, did you mean in your comment that the "Local" section should be entitled "LMTP"? Or did you mean that a separate subsection should be added for LMTP? (I am still having a bit of trouble understand what LMTP even is, so it is difficult for me to write about it.)

dleangen: @chibenwa, I just added the content to the "chaining" section that this link points to. Please let me know if you still think this is problematic.

docs/modules/concepts/pages/processing/elements/index.adoc
dleangen: @chibenwa, I just added the content to the "chaining" section that this link points to. Please let me know if you still think this is problematic.

docs/modules/concepts/pages/processing/elements/receiving.adoc
dleangen: I am reading the spec now to try to understand it better.

dleangen: Updated.

PR #206, opened 2020-06-14
JAMES-3187 Updates to "Basic Server" section of the new documentation

dleangen:
Work in progress. Relates to the "Local Server". This is intended to be the "easy, entry-level" server that Users are encouraged to use.
Will take some work to make it so.

Related JIRA issue: [JAMES-3216](https://issues.apache.org/jira/browse/JAMES-3216)


docs/modules/servers/pages/local/conf/index.adoc
```diff
@@ -0,0 +1,23 @@
+= James Local Mail Server Configuration
+:navtitle: Configuration
+
+ * Domain names (domainlist.xml)
+ * Database (james-database.properties)
+ * LMTP (lmtpserver.xml)
+ * POP3 (pop3server.xml)
+ * DNS (dnsservice.xml)
+ * Loggin (logback.xml)
+ * ??? (jmx.properties)
```
chibenwa, line 10: Jmx protocol is used by James cli to interactwith the server. Jmx is known to be insecure, and we have as a project to rewrite the cli for james servers in order to rely on webadmin.

docs/modules/servers/pages/local/conf/index.adoc
```diff
@@ -0,0 +1,23 @@
...
+ * ??? (recipientrewritetable.xml)
```
chibenwa, line 11: Configures storage for recipient rewrite rules.

docs/modules/servers/pages/local/conf/index.adoc
```diff
@@ -0,0 +1,23 @@
...
+ * ??? (extensions.properties)
```
chibenwa, line 12: Allows loading of user defined guice modules to support injections of user definedcomponants in their nailets or other plugins. 

For instance, you could define a blacklist api with a jpa implementation, and use it in a mailet + administrate it via webadmin.

docs/modules/servers/pages/local/conf/index.adoc
```diff
@@ -0,0 +1,23 @@
...
+ * JWT (jwt_publickey)
+ * Mailets (mailetcontainer.xml)
+ * SMTP (smtpserver.xml)
+ * Health checks (healthcheck.properties)
+ * Mail storage (mailrepositorystore.xml)
+ * User storage (usersrepository.xml)
+ * IMAP (imapserver.xml)
+ * ??? (listeners.xml)
```
chibenwa, line 20: Mailbox listeners allows to react upon users mailbox events like "message added", "message deleted"...

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Noted, thanks!

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Thanks. Would be nice to have a little more in-depth information like [this](https://docs.microsoft.com/en-us/exchange/architecture/edge-transport-servers/address-rewriting?view=exchserver-2019) or [this](https://www.exim.org/exim-html-current/doc/html/spec_html/ch-address_rewriting.html).

dleangen: I wonder if this shouldn't be an "advanced" feature somehow. I would like to be able to simplify as much as possible the "default" server to make it less complex and likely to break for new users.

docs/modules/servers/pages/local/conf/index.adoc
dleangen: So IIUC, the Mailet part already has its own extension mechanism, but the webadmin part does not, so this extension mechanism would be for the webadmin part of the extension?

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Thanks!

docs/modules/servers/pages/local/conf/index.adoc
chibenwa: Oh I see.

The idea would be to package 2 JPA based server, one with (almost) no configuration, but close to 0 extra features. Easy to use, you don't have zillion of things to understand. And another one also using JPA but exposing complex configuration to allow customizing it. (the fact of using both JPA would be a coincidence bringed in for technical purposes).

I just got it!

Smart! :+1:

However regarding RRT it enables "alias" which is a very common featurethat even basic users could want (I will run out of time today for your "user email" that is really pertinent). We might not need any configuration for it for "basic users".

docs/modules/servers/pages/local/conf/index.adoc
chibenwa: `but the webadmin part does not` webadmin allow loading of custom routes actually

chibenwa:
Given https://github.com/apache/james-project/pull/206#discussion_r439926633 I would prefer to call it `Basic server` rather than `Local`. Local is still a technically oriented term that may be obscur to newcomer...

dleangen:
Sure, I can always change back.

So then we would have:

- Basic (was "Local", as described above, with minimal configuration)
- ???? (still the same as "Local", but with more advanced configurations)
- Advanced (as described here)
- Distributed

Should we consider merging "Advanced" and "Distributed"?? What should we name the new mid-range one?

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Ok... I'll get back to this part in more detail another time. :-)

docs/modules/servers/pages/local/conf/index.adoc
dleangen: I see. I will get back to this topic later. First I want to better understand the configuration options, then maybe later we can figure out what we should keep or drop for the "Default" (Basic) Server.

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Sorry, are these the same as SMTP Hooks that are [described on the old website](https://james.apache.org/server/dev-provided-smtp-hooks.html)?

Where can I find examples of these?

chibenwa:
I would actually merge ???? and advanced. (not being able to name things is generally symptomatic).

dleangen:
So, we effectively disallow most configurations in Basic and make it difficult for people to cause havoc. If they want to do any kind of configuration, extension, or whatever, then they would need to use Advanced, with the understanding that it is at their own risk.

Does that sound right?

docs/modules/servers/pages/local/conf/index.adoc
chibenwa: No it is different.

SMTP hooks are to add functionalities to the SMTP protocols.

Here MailboxListener allow to react upon any mailbox changes whatever the protocol.

See http://james.apache.org/server/config-listeners.html

dleangen: Ah, hehe. Ok, makes sense. I read the page at the url you pasted.

Are there any examples anywhere?

chibenwa:
> I 'm no longer sure that "Distributed" fits well as the name. But we'll cross that bridge later. :-)

Not that sure: advanced brings functional complexity, but 'distributed' brings 'technical difficulty', having a technical name to reflect that can somehow be nice.

dleangen:
Yeah... that makes sense. People who understand distributed should understand the consequences, else they should not be using it, really. Ok, let's stick with Basic, Advanced, Distributed!!

rouazana:
Sorry, I have some pain with this naming. For me Distributed is orthogonal to basic / advanced.

You can expect distributed to be advanced, but local could be basic or advanced, no?

I mean I don't see how we can make a choice from these 3 terms (apart distributed which for me makes a clear purpose of what is the goal).

dleangen:
@rouazana that is a fair point, and I had the same doubt.

So can you suggest how to resolve the current problem, then? The issue is that we now have these 4 Servers to name / conceptualize:

 * A "Local" Server with minimal configuration and default configurations (i.e. for newbies to James)
 * A "Local" Server that is configurable and customizable (requires more experience with James)
 * Advanced
 * Distributed

docs/modules/servers/pages/local/conf/index.adoc
mbaechler: If you look at example of users in the code, the quota-search feature is a good example : https://github.com/apache/james-project/blob/master/mailbox/plugin/quota-search-elasticsearch/src/main/java/org/apache/james/quota/search/elasticsearch/events/ElasticSearchQuotaMailboxListener.java

It allows to index quota usage per user into ElasticSearch

mbaechler:
> @rouazana that is a fair point, and I had the same doubt.
> 
> So can you suggest how to resolve the current problem, then? The issue is that we now have these 4 Servers to name / conceptualize:
> 
>     * A "Local" Server with minimal configuration and default configurations (i.e. for newbies to James)
> 
>     * A "Local" Server that is configurable and customizable (requires more experience with James)
> 
>     * Advanced
> 
>     * Distributed

I'll add my two cents here: in the first place we defined products based on their technical requirements. Now we are trying to define it another way (mainly) because we require too many configurations for beginners.

My opinion is: the problem is the configuration files. If you could configure your server with a simple configuration file containing a handful of parameters and when needed you could just add more complex configuration entries to add feature we would not be discussing how to name these products.

There's really no rational in requiring many configuration files with a lot of details but that's what we have right now.

What if we start from the strictly required parameters then build a configuration file in a decent format (say HOCON for example) and then we just implement the required stuff to support that simplified format?

We could then fill the gap with default values and incrementally support more customization parameters.

Would it solve the problem?

rouazana:
> @rouazana that is a fair point, and I had the same doubt.
> 
> So can you suggest how to resolve the current problem, then? The issue is that we now have these 4 Servers to name / conceptualize:
> 
>     * A "Local" Server with minimal configuration and default configurations (i.e. for newbies to James)
> 
>     * A "Local" Server that is configurable and customizable (requires more experience with James)
> 
>     * Advanced
> 
>     * Distributed

For me "Local"-minimal (aka newbies James) should be the same product than "Local"-configurable. We can achieve both having sane defaults in "Local"-configurable.

I still don't get what "Advanced" brings regarding that.

So for me the main products (flavors? servers? :)) should be:
* "Local", or "Standalone", or eventually "Basic" if you like this term to appeal new comers: the configuration should default to sane and simple defaults, but could be extended at will.
* "Distributed"

We could also add:
* "Testing" (in memory), which could be interesting by 2 kinds of people: developpers to have a handy mail server to integrate with in tests or even in their application; operators who would like to test quickly how behave James (not sure about this one, maybe it's worth to still recommend "Local"/"Basic" in this case).

dleangen:
> I'll add my two cents here: in the first place we defined products based on their technical requirements. Now we are trying to define it another way (mainly) because we require too many configurations for beginners.

Yes, this is true.

If we define based on technical features, then I think we are targeting developers only.

If we define based on usability features, then I think we are targeting users. One thing to consider is that when making a decision to adopt James, unless the adopter is just an individual, in an Enterprise there are more people involved in the decision than just the technical people.

Do we want to "sell" to just the tekkies? Or do we want to "sell" to people who actually make the decisions?

> My opinion is: the problem is the configuration files.

I don't disagree, but I think that this is only a part of the problem. Even if we make the configuration files easier to use (with something like HOCON, which looks very cool BTW), I think there is also another problem to consider. When approaching James for the first time, there are waaaaay too many concepts to learn. It is quite daunting.

What I am attempting to do is to minimize the cognitive overload for the "entry" product.

So any configuration that gets exposed becomes another thing that the user basically **has** to learn. By reducing choice, they have less to learn. If they want something different, then they have to actively choose themselves to use the configurable product.

Do you see what I mean.

So I think we are actually talking about slightly different things.

> I still don't get what "Advanced" brings regarding that.

Sorry, that was an old reference. Currently it is "Redundant": https://james.leangen.net/main/3.5/servers/index.html#redundant

> We could also add: "Testing" (in memory), which could be interesting...

It does sound interesting. I will get back to this in a different thread.

For now I think the main point we need to resolve is whether we can get agreement or not about the concept I am proposing about a "Basic" server, which has **minimal** configuration for the purpose of eliminating as many choices as possible for the newcomer.

dleangen:
> > I still don't get what "Advanced" brings regarding that.
> Sorry, that was an old reference. Currently it is "Redundant": https://james.leangen.net/main/3.5/servers/index.html#redundant

Sorry, I was getting confused myself. 😅

The original thought was "Redundant", but then we starting proposing it more from the user's perspective, rather than the James Developer's perspective, which led to "Basic" (i.e. for me the User it is basic and easier to use but with fewer features), or "Advanced" (which means for me the User I can do more with it but I know I need to learn more stuff).

My preference is to use the user's perspective if the objective is to expand the James community.

docs/modules/servers/pages/local/conf/index.adoc
```diff
@@ -1,23 +1,99 @@
...
  * Sieve filters (managesieveserver.xml)
  * Web admin (webadmin.properties)
+
+
+== Notes
+Jmx protocol is used by James cli to interactwith the server. Jmx is known to be insecure, and we have as a project to rewrite the cli for james servers in order to rely on webadmin.
+
+
+
+Scenarios for address rewriting
```
chibenwa, line 54: https://github.com/apache/james-project/blob/master/src/site/xdoc/server/config-recipientrewritetable.xml (not yet deployed on current website :-( tries to provide wording about Recipient Rewriting and use cases.

docs/modules/servers/pages/local/conf/index.adoc
```diff
@@ -1,23 +1,99 @@
...
+
+The following scenarios are examples of how you can use address rewriting:
+
+Group consolidation: Some organizations segment their internal businesses into separate domains that are based on business or technical requirements. This configuration can cause email messages to appear as if they come from separate groups or even separate organizations.
+
+The following example shows how an organization, Contoso, Ltd., can hide its internal subdomains from external recipients:
+
+Outbound messages from the northamerica.contoso.com, europe.contoso.com, and asia.contoso.com domains are rewritten so they appear to originate from a single contoso.com domain. All messages are rewritten as they pass through Edge Transport servers that provide SMTP connectivity between the whole organization and the Internet.
```
chibenwa, line 62: We speak of a **domain alias** here. I would more formulate it:

 - External email can send emails to northamerica.contoso.com, europe.contoso.com, ...
 - Internal users can send email from northamerica.contoso.com, europe.contoso.com, ...

docs/modules/servers/pages/local/conf/index.adoc
```diff
@@ -1,23 +1,99 @@
...
+
+Inbound messages to user@contoso.com are rewritten and routed to user@fourthcoffee.com mailboxes. Inbound messages that are sent to user@fourthcoffee.com are routed directly to Fourth Coffee's email servers.
+
+Partners: Many organizations use external partners to provide services for their customers, other organizations, or their own organization. To avoid confusion, the organization might replace the email domain of the partner organization with its own email domain.
+
+The following example shows how Contoso, Ltd. can hide a partner's email domain:
+
+Contoso, Ltd. provides support for the larger Wingtip Toys organization. Wingtip Toys wants a unified email experience for its customers, and it requires all messages from support personnel at Contoso, Ltd. to appear as if they were sent from Wingtip Toys. All outbound messages that relate to Wingtip Toys are sent through their Edge Transport servers, and all contoso.com email addresses are rewritten to wingtiptoys.com email addresses.
+
+Inbound messages for support@wingtiptoys.com are accepted by Wingtip Toy's Edge Transport servers, rewritten, and then routed to the support@contoso.com email address.
```
chibenwa, line 80: You nicely developed real word usage for domain aliases :-)

Maybe we need to specifically detail that these examples are for Domain Aliases, and we might want to detail other Recipient Rewritting options as well.

docs/modules/servers/pages/local/conf/index.adoc
chibenwa: Maybe `address rewriting` needs to be moved in the concept part?

dleangen: Yes, definitely. Anything that requires knowledge of any concept should be there.

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Hahahaha. Actually I copy-pasted this from some website and forgot to update the text. I thought it was interesting and useful information. Either we should link to that source or write our own text, else it would be plagiarism. :-)

chibenwa: (plagiarism is only an issue if there is some licensing issues ;-) )

Well giving some insight about real world situation and associate it to our *recipient rewrite table* wording is a must have.

dleangen: Yes, for sure. I am planning on updating this text and, as you pointed out, adding a section to "Concepts".

This is also the reason why I want to have a minimalistic Basic server. If everything needs to be backed by a "Concepts" text, that is a lot of work! :-)

mbaechler:
> > I'll add my two cents here: in the first place we defined products based on their technical requirements. Now we are trying to define it another way (mainly) because we require too many configurations for beginners.
> 
> Yes, this is true.
> 
> If we define based on technical features, then I think we are targeting developers only.
> 
> If we define based on usability features, then I think we are targeting users. One thing to consider is that when making a decision to adopt James, unless the adopter is just an individual, in an Enterprise there are more people involved in the decision than just the technical people.
> 
> Do we want to "sell" to just the tekkies? Or do we want to "sell" to people who actually make the decisions?
> 
> > My opinion is: the problem is the configuration files.
> 
> I don't disagree, but I think that this is only a part of the problem. Even if we make the configuration files easier to use (with something like HOCON, which looks very cool BTW), I think there is also another problem to consider. When approaching James for the first time, there are waaaaay too many concepts to learn. It is quite daunting.
> 
> What I am attempting to do is to minimize the cognitive overload for the "entry" product.
> 
> So any configuration that gets exposed becomes another thing that the user basically **has** to learn. By reducing choice, they have less to learn. If they want something different, then they have to actively choose themselves to use the configurable product.
> 
> Do you see what I mean.

I think I do. I also think I didn't explained my opinion well enough because it seems you didn't understand my point.

When I adopt a software, I don't care how many features are included as long as I can have it fill my simple beginner needs. I don't see any reason to remove features from James as long as users don't have to worry about them at first.

The reason we split the product in several alternatives is that the *requirements* are not the same.

Think: I buy an e-bike with a removable battery because I don't have a way to plug the bike vs I buy a e-bike with integrated battery because I can plug it. Users understand very well if they want to make their server run without having to setup a database or if they want to deploy a Cassandra cluster.

BTW, what I propose is not to have a different configuration format but to be able to run James with 3 lines of configuration.

Every other configuration entries would have hardcoded values (not in the configuration file): we could propose a beginner oriented configuration file sample with just these 3 values (maybe there's a bit more than 3, I don't know yet).

Forget about HOCON, it just added noise to my previous message.

dleangen:
> BTW, what I propose is not to have a different configuration format but to be able to run James with 3 lines of configuration.
> 
> Every other configuration entries would have hardcoded values (not in the configuration file): we could propose a beginner oriented configuration file sample with just these 3 values (maybe there's a bit more than 3, I don't know yet).

Ok, yes, then I think we are on the same page.

I still think that you are approaching this from the technical perspective (i.e. how to build). I am approaching this from the "how to present" perspective.

What we discussed [in a private thread](https://github.com/ieugen/james-self-hosting-sandbox/issues/9) (which I will move back to the community) was something like this:

 * Basic Server (for self-hosting - should be very simple with very few options)
 * Extendable Server (for customizing)
 * Distributed Server (for a distributed environment and all the complexities that come with it)

This does not mean "Basic" and "Extendable" cannot be built from the same jar file (I assume). It would just be packaged / presented differently given that it targets a different audience.

Kind of like the difference between a car with automatic transition (don't want to know, let the machine just do it for me) and manual transmission (I want to control it).

That said, I think we are converging on a mutual understanding.

mbaechler:
> Kind of like the difference between a car with automatic transition (don't want to know, let the machine just do it for me) and manual transmission (I want to control it).

I like this metaphor: actually most modern automatic gear-box have two modes: "I don't care" and "I want to take control back because algorithms are too dumb for driving on a mountain road"

And it's exactly what I propose to do.

docs/modules/servers/pages/basic/self-hosting.notes
```diff
@@ -0,0 +1,90 @@
...
+I would say "self hosting" means individuals having bought a domain name and want to install their own mail server for hosting their personal emails. Volumetry is small. They want control on their data.
+
+One generally tricky point about this kind of set up is getting third parties (like eg GMail) trusting your setup. It would be important to document a checklist regarding Mail Exchange trust like:
+
+ I'm not an open spam relay
+ MX record registered for my domain
+ Reverse DNS setted up for MX servers
+ SPF records setted up
+ DKIM signature setted up
+(which is slightly external to James as DNS setup is required)
```
Arsnael, line 16: I remember once a user posting on the mailinglist some links to his own documentation regarding some of those topics to help someone, and I saved them as I found it interesting and useful:
* https://dmatthews.org/email_auth.html
* https://dmatthews.org/java_email.html

would that help?

dleangen: Yes, thank you!

chibenwa:
I wrote an ADR to better describe the server offer you describe.

https://github.com/apache/james-project/pull/222

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Added above as a note

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Added as a note

docs/modules/servers/pages/local/conf/index.adoc
dleangen: Added as comment for now

dleangen:
I think this PR has lived long enough, and I would like to merge the contents, even if it is still not quite yet complete. If there are no objections I will merge in the next few days. If I have some agreement, I can merge quicker.

Of course, if there are any objections we can continue the discussion here.

Arsnael:
> I think this PR has lived long enough, and I would like to merge the contents, even if it is still not quite yet complete. If there are no objections I will merge in the next few days. If I have some agreement, I can merge quicker.

I would agree for merging this. It's still full of comments and stuff but I guess we can keep them for now and tackle them one at a time.

I'm just not exactly sure what to think of what is written in the help page. We probably do deserve a little kick but I'm not sure it's that negative? :)

dleangen:
> I would agree for merging this. It's still full of comments and stuff but I guess we can keep them for now and tackle them one at a time.

Ok, thanks!

I have already included all the comments in this thread as inline comments in the source, so I think it should be ok.

> I'm just not exactly sure what to think of what is written in the help page. We probably do deserve a little kick but I'm not sure it's that negative? :)

😀

I completely agree that we should not make that public. So, I have commented out that entire section, as I don't find it post-worthy. I would like to redo that discussion in order to arrive at a better result that everybody is comfortable with.

Arsnael:
> I completely agree that we should not make that public. So, I have commented out that entire section, as I don't find it post-worthy. I would like to redo that discussion in order to arrive at a better result that everybody is comfortable with.

Yes it would be worthy to redo this discussion I think as well. Much appreciated, thanks !

PR #211, opened 2020-06-17
JAMES-2343 Add buffered DeferredFileOutputStream and use it in MimeMe…

chibenwa:
…ssageInputStreamSource

From the original work of @chikei

Opened as the repository used for https://github.com/apache/james-project/pull/105 was (apparently?) deleted...

### DIFF
--- a/server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
+++ b/server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
@@ -1,0 +1,264 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.server.core;
+
+import java.io.BufferedOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.io.output.ByteArrayOutputStream;
+import org.apache.commons.io.output.DeferredFileOutputStream;
+import org.apache.commons.io.output.ThresholdingOutputStream;
+
+/**
+ * An almost copy of {@link DeferredFileOutputStream} with buffered file stream.
+  * <p>
+  * This copy is done because 1. the original class is not written in the way that
+  * we can decorate file output stream. 2. when switched to FileOutputStream the
+  * original class does not buffer file output stream which could cause up to 50%
+  * performance loss compare to buffered stream (see JIRA JAMES-2343 for simple
+  * benchmark).
+  * <p>
+  * The only difference is in the {@link #thresholdReached()} where when updating
+  * {@link #currentOutputStream}, instead of directly assign FileOutputStream to
+  * it, here we wrap the file output stream with BufferedOutputStream.
+  *
+  * @link https://issues.apache.org/jira/browse/JAMES-2343
+ */
+public class BufferedDeferredFileOutputStream extends ThresholdingOutputStream {
+
+    /**
+     * The output stream to which data will be written prior to the theshold
+     * being reached.
+     */
+    private ByteArrayOutputStream memoryOutputStream;
+
+
+    /**
+     * The output stream to which data will be written at any given time. This
+     * will always be one of <code>memoryOutputStream</code> or
+     * <code>diskOutputStream</code>.
+     */
+    private OutputStream currentOutputStream;
+
+
+    /**
+     * The file to which output will be directed if the threshold is exceeded.
+     */
+    private File outputFile;
+
+    /**
+     * The temporary file prefix.
+     */
+    private final String prefix;
+
+    /**
+     * The temporary file suffix.
+     */
+    private final String suffix;
+
+    /**
+     * The directory to use for temporary files.
+     */
+    private final File directory;
+
+
+    /**
+     * True when close() has been called successfully.
+     */
+    private boolean closed = false;
+
+    /**
+     * Constructs an instance of this class which will trigger an event at the
+     * specified threshold, and save data to a file beyond that point.
+     *
+     * @param threshold  The number of bytes at which to trigger an event.
+     * @param outputFile The file to which data is saved beyond the threshold.
+     */
+    public BufferedDeferredFileOutputStream(final int threshold, final File outputFile) {
+        this(threshold,  outputFile, null, null, null);
+    }
+
+
+    /**
+     * Constructs an instance of this class which will trigger an event at the
+     * specified threshold, and save data to a temporary file beyond that point.
+     *
+     * @param threshold  The number of bytes at which to trigger an event.
+     * @param prefix Prefix to use for the temporary file.
+     * @param suffix Suffix to use for the temporary file.
+     * @param directory Temporary file directory.
+     */
+    public BufferedDeferredFileOutputStream(final int threshold, final String prefix, final String suffix, final File directory) {
+        this(threshold, null, prefix, suffix, directory);
+        if (prefix == null) {
+            throw new IllegalArgumentException("Temporary file prefix is missing");
+        }
+    }
+
+    /**
+     * Constructs an instance of this class which will trigger an event at the
+     * specified threshold, and save data either to a file beyond that point.
+     *
+     * @param threshold  The number of bytes at which to trigger an event.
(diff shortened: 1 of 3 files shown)


chibenwa:
## Before

Capture d’écran de 2020-06-18 09-08-50

## After

Capture d’écran de 2020-06-18 09-08-50

## Conclusions:

 - Twice the rate, which then impact other metrics (p99 & errors)
 - x2 improvement for p50, normal processing time improved. ~ 2 mean time reduction reflects that too.

Exceptions are due to ObjectStorage (signle node in this test setup) being overwhelmed:

```
 org.jclouds.http.HttpResponseException: Server rejected operation connecting to PUT http://swift:8080/v1/TMP_test/default/7d69559f2b230319d17101a92802d4bb13639e2c23bf61e7b1ec7378b59c8669 HTTP/1.1
```

chibenwa:
Build is :green_apple: , let's wait for reviews :-)

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,264 @@
...
+     * @param threshold  The number of bytes at which to trigger an event.
+     * @param prefix Prefix to use for the temporary file.
+     * @param suffix Suffix to use for the temporary file.
+     * @param directory Temporary file directory.
+     */
+    public BufferedDeferredFileOutputStream(final int threshold, final String prefix, final String suffix, final File directory) {
+        this(threshold, null, prefix, suffix, directory);
+        if (prefix == null) {
+            throw new IllegalArgumentException("Temporary file prefix is missing");
+        }
```
blackheaven, line 117: Preconditions

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,264 @@
...
+     * Switches the underlying output stream from a memory based stream to one
+     * that is backed by disk. This is the point at which we realise that too
+     * much data is being written to keep in memory, so we elect to switch to
+     * disk-based storage.
+     *
+     * @exception IOException if an error occurs.
+     */
+    @Override
+    protected void thresholdReached() throws IOException {
+        if (prefix != null) {
```
blackheaven, line 165: `Optional`

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,264 @@
...
+            outputFile = File.createTempFile(prefix, suffix, directory);
+        }
+        final FileOutputStream fos = new FileOutputStream(outputFile);
+        try {
+            memoryOutputStream.writeTo(fos);
+        } catch (IOException e) {
+            fos.close();
+            throw e;
+        }
```
blackheaven, line 174: `finally`

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,264 @@
...
+        currentOutputStream = new BufferedOutputStream(fos, getThreshold());
+        memoryOutputStream = null;
```
blackheaven, line 176: `Optional`

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
```diff
@@ -0,0 +1,264 @@
...
+     * Writes the data from this output stream to the specified output stream,
+     * after it has been closed.
+     *
+     * @param out output stream to write to.
+     * @exception IOException if this stream is not yet closed or an error occurs.
+     */
+    public void writeTo(final OutputStream out) throws IOException {
+        // we may only need to check if this is closed if we are working with a file
+        // but we should force the habit of closing wether we are working with
+        // a file or memory.
```
blackheaven, line 248: the strategy pattern should be used then

server/container/core/src/test/java/org/apache/james/server/core/BufferedDeferredFileOutputStreamTest.java
```diff
@@ -0,0 +1,384 @@
...
+        final BufferedDeferredFileOutputStream dfos =
+            new BufferedDeferredFileOutputStream(testBytes.length - 5,  testFile);
+        try
+        {
+            dfos.write(testBytes, 0, testBytes.length);
+            dfos.close();
+        }
+        catch (final IOException e) {
+            fail("Unexpected IOException");
+        }
```
blackheaven, line 121: Rely on the test case `Exception`

server/container/core/src/test/java/org/apache/james/server/core/BufferedDeferredFileOutputStreamTest.java
```diff
@@ -0,0 +1,384 @@
...
+        {
+            dfos.write(testBytes, 0, chunkSize);
+            dfos.write(testBytes, chunkSize, chunkSize);
+            dfos.write(testBytes, chunkSize * 2,
+                testBytes.length - chunkSize * 2);
+            dfos.close();
+        }
+        catch (final IOException e) {
+            fail("Unexpected IOException");
+        }
```
blackheaven, line 157: idem

server/container/core/src/test/java/org/apache/james/server/core/BufferedDeferredFileOutputStreamTest.java
```diff
@@ -0,0 +1,384 @@
...
+
+            assertFalse(testFile.exists());
+            assertTrue(dfos.isInMemory());
+
+            try {
+                dfos.writeTo(baos);
+                fail("Should not have been able to write before closing");
+            } catch (final IOException ioe) {
+                // ok, as expected
+            }
```
blackheaven, line 191: `assertThrownBy`

server/container/core/src/main/java/org/apache/james/server/core/BufferedDeferredFileOutputStream.java
chibenwa: The BufferedDefferedFileOutputStream is (mostly) taken from commons-io, I would like to get that code as close as possible from the original version, if you don't mind.

server/container/core/src/test/java/org/apache/james/server/core/BufferedDeferredFileOutputStreamTest.java
chibenwa: idem for the tests

Arsnael:
> Twice the rate, which then impact other metrics (p99 & errors)

Why do we get twice the rate?

chibenwa:
Because it is faster. Buffering things before writing them to a file is a nice idea...

mbaechler:
Please always describe the load testing scenario you run and why.
Here you should run not only for big mails but also for small ones as they should be kept in memory under a given threshold.

Regarding the analysis, you should try to inject only half the current load (or a third) so that object storage is not a problem anymore and you can measure the latency anyway. WDYT?

Arsnael:
> Because it is faster. Buffering things before writing them to a file is a nice idea...

Ok same amount of time, so it means it's faster. Was not sure the schenario was like that :)

chibenwa:
> Please always describe the load testing scenario you run and why.

SmtpMessage sending, big size.

The idea is to exercise enqueue speed, what the original contributor did tested in isolation.

Here is the link to the scenario: https://github.com/chibenwa/james-gatling/blob/smtp-benchs/src/test/scala-2.12/org/apache/james/gatling/simulation/smtp/SmtpNoAuthenticationNoEncryptionBigBodySimulation.scala (slightly adapted)

> Here you should run not only for big mails but also for small ones as they should be kept in memory under a given threshold.

> Regarding the analysis, you should try to inject only half the current load (or a third) so that object storage is not a problem anymore and you can measure the latency anyway. WDYT?

Let's run that then...

chibenwa:
I did launch the 4 tests (with a reduced user count) let's see the output on monday

chibenwa:
```
Simultation: org.apache.james.gatling.simulation.smtp.SmtpNoAuthenticationNoEncryptionSimulation
chibenwa/james-gatling master branch (btw could be nice to specify custom branches here)
```

## Before

Capture d’écran de 2020-06-21 11-29-00

## After

Capture d’écran de 2020-06-21 11-29-53


## Conclusion for small mail reception

 - Less errors but slightly slower

Which is good!

Regarding a lower rate for big mails, one of the simultation aborted, I relaunch it.

chibenwa:
Here is the result of the run with less throughtput and big messages:

## master

Capture d’écran de 2020-06-22 10-46-56

## This changeset

Capture d’écran de 2020-06-22 10-48-18

mbaechler:
We really need small messages too

chibenwa:
> We really need small messages too

https://github.com/apache/james-project/pull/211#issuecomment-647077562

mbaechler:
> ```
> Simultation: org.apache.james.gatling.simulation.smtp.SmtpNoAuthenticationNoEncryptionSimulation
> chibenwa/james-gatling master branch (btw could be nice to specify custom branches here)
> ```
> 
> ## Before
> 
> Capture d’écran de 2020-06-21 11-29-00
> ## After
> 
> Capture d’écran de 2020-06-21 11-29-53
> ## Conclusion for small mail reception
> 
>     * Less errors but slightly slower
> 
> 
> Which is good!
> 
> Regarding a lower rate for big mails, one of the simultation aborted, I relaunch it.

16% slow down is a lot, don't you think ?

chibenwa:
> 16% slow down is a lot, don't you think ?

That's a very specific workload, with only messages below 1 KB that don't represent a real production usage.

chibenwa:
I will rerun this

chibenwa:
Shall I take time rebase this work, or can I just discard it?

chibenwa:
I would actually REALLY be interested to know "how this plays" with https://github.com/apache/james-project/pull/309

I'll run benches <3

chibenwa:
We ended up merging this work.

Thanks for your time and the contribution, it ended up not being lost :-)

PR #212, opened 2020-06-18
JAMES-3187 Added User Model to Concepts Section of new documentation

dleangen:
Work in progress

### DIFF
--- a/docs/modules/concepts/pages/user/index.adoc
+++ b/docs/modules/concepts/pages/user/index.adoc
@@ -42,7 +42,7 @@ Apache James offers Operators a range of servers for various use cases, includin
    time to gain deep expertise
  * An xref:server/advanced/index.adoc[Advanced Server] well-suited for Operators with
    advanced skills and resources who are not daunted by getting their hands dirty
-   with configurations and extensions
+   with configurations or using built-in extensions
 
 
 
--- a/docs/modules/concepts/pages/user/index.adoc
+++ b/docs/modules/concepts/pages/user/index.adoc
@@ -59,7 +59,7 @@ An Integrator does **not** need to download the source code or compile James.
 == Developer
 
 A **Developer** uses the James codebase, executes a build using the provided
-build system (Maven), and may edit and deploy the source code.
+build system (Maven), may edit the source code and deploy the built artifacts.
 
 
 
--- a/docs/modules/concepts/pages/user/index.adoc
+++ b/docs/modules/concepts/pages/user/index.adoc
@@ -1,0 +1,74 @@
+= James User Model
+:navtitle: User Model
+
+
+For clarity, we have defined various roles for actors who interact 
+with James in one way or another.
+The deliverables we make available are produced with these roles in mind.
+Note that any given person may play more than one role, even at the same
+time.
+
+ * <<User>>
+ * <<Operator>>
+ * <<Integrator>>
+ * <<Developer>>
+ * <<Member>>
+ * <<Committer>>
+
+== User
+
+A **User** simply uses email without any notion of what server is providing the service.
+
+More precisely, in typical cases a User will have an email client for reading and writing email. The
+xref:glossary.adoc[email client] will connect to the James server via 
+xref:mail/protocols/smtp.adoc[SMTP], 
+xref:mail/protocols/pop.adoc[POP], or 
+xref:mail/protocols/imap.adoc[IMAP].
+The User will have no indication anywhere that the server is a James server.
+
+
+== Operator
+
+An **Operator** installs and runs James for the benefit of Users.
+
+The Operator does not compile any Java code, and does not require any software
+development environment. In fact, the objective of the James project is provide
+a mail server that does not even require any knowledge of software programming at all.
+
+
+
+== Integrator 
+
+An **Integrator** has some expertise in Java development. Based on this expertise
+the Integrator can perform more advanced server configurations and may
+develop xref:xxx.adoc[Extensions] to customize the behavior of a James server.
+
+An Integrator does **not** need to download the source code or compile James.
+
+
+
+== Developer
+
+A **Developer** uses the James codebase, executes a build using the provided
+build system (Maven), and may edit and deploy the source code.
+
+
+
+== Member
+
+A **Member** is anybody who has:
+
+ * Filed an issue on the issue tracking system
+ * Posted a question on the Mailing list
+ * Submitted a pull request (PR)
+
+
+== Committer
+
+A **Committer** is granted special rights (according to Apache practices) 
+to participate in the development of James by:
+
+ * Accepting (being assigned) issues in the issue tracking system
+ * Merging code into the git repository
+ * Being granted a voice when voting matters arise
+
--- a/docs/modules/concepts/pages/user/index.adoc
+++ b/docs/modules/concepts/pages/user/index.adoc
@@ -65,11 +65,16 @@ build system (Maven), and may edit and deploy the source code.
 
 == Contributor
 
-A **Contributor** is anybody who has:
+A **Contributor** refers to anyone involved with the Apache James community by:
 
- * Filed an issue on the issue tracking system
- * Posted a question on the Mailing list
- * Submitted a pull request (PR)
+ * Filing an issue on the https://issues.apache.org/jira/secure/Dashboard.jspa[issue tracking system]
+ * Posting a question on the https://james.apache.org/mail.html[mailing list] 
+   or on the https://gitter.im/apache/james-project[gitter channel]
+ * Submitting a https://github.com/apache/james-project/pulls[pull request] (PR)
+ * Fixes, refines and writes documentation
+ * Writes a tutorial or article about James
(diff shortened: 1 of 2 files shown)


docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,54 @@
...
+   - Anybody who compiles James
+   - Anybody who wants to contribute to James
+   - Anybody who wants to otherwise work with the James codebase
+
+
+A Basic Operator should not be assumed to know much (if anything) about Java and the Java ecosystem.
+
+
+
+Is this another dimension? Basic, Advanced etc.???
```
chibenwa, line 52: I do not understand the question

dleangen: Well, if "Operator" basically means "doesn't compile" (simplification), and "Developer" basically means "compiles", is there a difference between an Operator who doesn't want to configure anything (target of the Basic Server) and an Operator who is more advanced and doesn't mind getting her hands dirty?

Another thing to consider...

At what point to we force Operators to have knowledge of Java? Just because the system is written in Java, does that mean that the Operator has to know Java?

I am sure other questions will soon arise, too.

chibenwa: > At what point to we force Operators to have knowledge of Java?

Enough to run Java.

Basic server operators are expected almost none IMO.

Advanced server operator might need to deal with advanced concepts like Garbage collection and other JVM settings on bigger installations.

Does it makes sense?

Don't hesitate, these questions are valuable.

dleangen: Right now in the docs (not yet committed) I am going with this idea:

* Operator does not require any knowledge of Java
* Integrator must have some expertise in Java, either to do advanced configurations or to develop Extensions.

Roles are not mutually exclusive of course. I will update the text and commit soon so you can tell me what you think.

dleangen:
This content is now ready for review and merge. Thanks!

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,74 @@
+= James User Model
+:navtitle: User Model
+
+
+For clarity, we have defined various roles for actors who interact 
+with James in one way or another.
```
chibenwa, line 6: who interact with James **servers**.

We  need to keep some rooms for some other personna like libraryt users (let's keep it for later)

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,74 @@
...
+A **Member** is anybody who has:
+
+ * Filed an issue on the issue tracking system
+ * Posted a question on the Mailing list
+ * Submitted a pull request (PR)
+
+
+== Committer
+
+A **Committer** is granted special rights (according to Apache practices)
```
chibenwa, line 68: Can we have a link to https://www.apache.org/dev/committers.html ?

Maybe we could mention project member comittee as a special case of committer (and link relevant apache page https://www.apache.org/dev/pmc.html)

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+== Operator
+
+An **Operator** installs and runs James for the benefit of Users.
+
+The Operator does not compile any Java code, and does not require any software
+development environment. In fact, the objective of the James project is provide
+a mail server that does not even require any knowledge of software programming at all.
+
+
```
chibenwa, line 38: I would state: 

```
We offer several servers adapted to various cases of Operator. Operators interested in self-hosting, with limited time, 
skills and resources would benefit from the [link: Basic server] while operators willing to leverage complex James 
configuration will use the [link: Advanced Server]
```
What do you think?

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+== Integrator 
+
+An **Integrator** has some expertise in Java development. Based on this expertise
```
chibenwa, line 42: I would state JVM language instead of Java: They could write his extensions in kotlin if they wishes

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,74 @@
...
+
+
+== Developer
+
+A **Developer** uses the James codebase, executes a build using the provided
+build system (Maven), and may edit and deploy the source code.
+
+
+
+== Member
```
chibenwa, line 57: This is the definition of a `Contributor` IMO, in order to respect Apache wording

docs/modules/concepts/pages/user/index.adoc
dleangen: Updated. Will push in a few minutes.

docs/modules/concepts/pages/user/index.adoc
dleangen: Updated. Will push in a few minutes.

docs/modules/concepts/pages/user/index.adoc
dleangen: Updated. Will push in a few minutes.

docs/modules/concepts/pages/user/index.adoc
dleangen: Looks good! I reworded a bit and updated. Will push in a few minutes.

docs/modules/concepts/pages/user/index.adoc
dleangen: Updated. Will push in a few minutes.

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,90 @@
...
+development environment. 
+
+Apache James offers Operators a range of servers for various use cases, including:
+
+ * A xref:server/basic/index.adoc[Basic Server] ideal for busy Operators who 
+   are interested in an easy-to-use self-hosting solution, but who may not have the
+   time to gain deep expertise
+ * An xref:server/advanced/index.adoc[Advanced Server] well-suited for Operators with
+   advanced skills and resources who are not daunted by getting their hands dirty
+   with configurations and extensions
```
chibenwa, line 45: I propose `who are not daunted by getting their hands dirty with configurations and **built-in** extensions` to distinguish clearly them from the Integrator

dleangen: Right! Good point.

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,90 @@
...
+== Developer
+
+A **Developer** uses the James codebase, executes a build using the provided
+build system (Maven), and may edit and deploy the source code.
+
+
+
+== Contributor
+
+A **Contributor** is anybody who has:
```
chibenwa, line 68: Let's try to provide a self-sufficant, inclusive definition here, without relying on examples.

```
A **Contributor** is anyone getting involved within the Apache James community. A **Contributor** can:

 - ...
```

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,90 @@
...
+
+ * Filed an issue on the issue tracking system
+ * Posted a question on the Mailing list
```
chibenwa, line 71: ` * Posted a question on the Mailing list or on the gitter channel`

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,90 @@
...
+ * Submitted a pull request (PR)
```
chibenwa, line 72: ` * Fixes, refines and writes documentation`

chibenwa: ` * Writes tutorials, articles about James`
` * etc...` -> invites people to find their own way of contributing.

docs/modules/concepts/pages/user/index.adoc
dleangen: Updated. Will push in a minute.

docs/modules/concepts/pages/user/index.adoc
dleangen: Updated. Will push in a minute.

docs/modules/concepts/pages/user/index.adoc
dleangen: Updated. Will push in a minute.

docs/modules/concepts/pages/user/index.adoc
```diff
@@ -0,0 +1,95 @@
...
+develop xref:xxx.adoc[Extensions] to customize the behavior of a James server.
+
+An Integrator does **not** need to download the source code or compile James.
+
+
+
+== Developer
+
+A **Developer** uses the James codebase, executes a build using the provided
+build system (Maven), and may edit and deploy the source code.
```
mbaechler, line 62:
```suggestion
build system (Maven), may edit the source code and deploy the built artifacts.
```

chibenwa:
Let's merge this :-)

PR #213, opened 2020-06-19
JAMES-3252 DomainList autoDection should be turned off by default

chibenwa:
Also take this opportunity to clean default domains within all configurations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ of tasks being currently executed.
 - Differentiation between domain alias and domain mapping. Read upgrade instructions.
 - JAMES-3122 Log4J2 adoption for Spring product. Log file configuration needs to be updated. See upgrade instructions.
 - JAMES-2760 mailqueue.size.metricsEnabled should be false by default
+- JAMES-3252 DomainList autoDection should be turned off by default. Operators relying on implicit values for enabling DomainList autoDection now needs to explicitly configure it.
 
 ### Fixed
 - JAMES-2828 & JAMES-2929 bugs affecting JDBCMailRepository usage with PostgresSQL thanks to Jörg Thomas & Sergey B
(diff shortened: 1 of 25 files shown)


server/data/data-library/src/main/java/org/apache/james/domainlist/lib/DomainListConfiguration.java
```diff
@@ -93,13 +93,15 @@ public Builder addConfiguredDomains(Domain... domains) {
 
         public DomainListConfiguration build() {
             return new DomainListConfiguration(
-                autoDetectIp.orElse(true),
-                autoDetect.orElse(true),
+                autoDetectIp.orElse(false),
```
mbaechler, line 96: oh, a breaking change! (:

chibenwa: This, is intended! Maybe I should mention it in the changelog?

mbaechler: maybe

chibenwa:
Merged

CHANGELOG.md
```diff
@@ -38,6 +38,7 @@ of tasks being currently executed.
 - Differentiation between domain alias and domain mapping. Read upgrade instructions.
 - JAMES-3122 Log4J2 adoption for Spring product. Log file configuration needs to be updated. See upgrade instructions.
 - JAMES-2760 mailqueue.size.metricsEnabled should be false by default
+- JAMES-3252 DomainList autoDection should be turned off by default. Operators relying on implicit values for enabling DomainList autoDection now needs to explicitly configure it.
```
rouazana, line 41: autoDection? (twice in this line)

PR #214, opened 2020-06-19
JAMES-3255 Postmaster example values should be valid

chibenwa:


chibenwa:
Not really, it is configured in another place.

If we want that,wecan remove this property, but I wanted to keep this example and minimize the code diff.

Is it ok to you to keep it as is?

rouazana:
I hate receiving mails from postmaster@localhost, it really does not help to understand the issue. For me it's neither better nor less good from previous value.
I won't oppose to this change, but it does not really improve things.

chibenwa:
Maybe lets remove this and enforce default domain use then

rouazana:
that would be the better. Other better temporary workaround: replace by postmaster@example.com

chibenwa:
Not really as example.org would lead to warning as it is not a local domain... There will be a log.

rouazana:
> Not really as example.org would lead to warning as it is not a local domain... There will be a log.

```
ping example.com
PING example.com (127.0.0.1) 56(84) bytes of data.
64 bytes from localhost (127.0.0.1): icmp_seq=1 ttl=64 time=0.041 ms
64 bytes from localhost (127.0.0.1): icmp_seq=2 ttl=64 time=0.025 ms
```

chibenwa:
I think you need to check your /etc/hosts

For me it's clearly not local.

```
% ping example.com
PING example.com (93.184.216.34) 56(84) bytes of data.
64 bytes from 93.184.216.34 (93.184.216.34): icmp_seq=1 ttl=53 time=197 ms
64 bytes from 93.184.216.34 (93.184.216.34): icmp_seq=2 ttl=53 time=219 ms
64 bytes from 93.184.216.34 (93.184.216.34): icmp_seq=3 ttl=53 time=242 ms
64 bytes from 93.184.216.34 (93.184.216.34): icmp_seq=4 ttl=53 time=265 ms
```

rouazana:
my main concern about this parameter is that ops should be aware that they have to change it (if it cannot be set automatically to a sane value).
having
```
<postmaster>postmaster</postmaster>
```
does really not make this obvious, and one cannot even expect to think that it should be an email address
```
<postmaster>postmaster@localhost</postmaster>
```
has also chances to never been updated
maybe
```
<postmaster>postmaster.CHANGE.ME.PUT.THE.RIGHT.DOMAIN@localhost</postmaster>
```
is slightly better :) (or something less aggressive with some documentation just above)

chibenwa:
The value i prpose is a sane value as it relies on default domain

chibenwa:
Sure, but for the comment changing all these configuratoon files is painful...

chibenwa:
comments added for user facing configurations

chibenwa:
Merged

PR #217, opened 2020-06-20
JAMES-3260 Gradle as a build system

ieugen:
This is a work in progress PR to explore Gradle build system for James.

* Configured maven build scans https://scans.gradle.com/#maven


ieugen:
`mvn install -DskipTests` ->  https://scans.gradle.com/s/vmmyekhurnmme

ieugen:
First build scan with `mvn install` failed https://scans.gradle.com/s/sjbnibmspsohu

ieugen:
`mvn clean install` has gone further but stil failed  https://gradle.com/s/xwhrgkh4vk3uu

ieugen:
Initial scan with gradle `gradle build --scan` -> https://scans.gradle.com/s/fmmh2pm7dy3os

ieugen:
Latest `./gradlew clean build --scan` : 39 min -> https://scans.gradle.com/s/7t7xhr4lkshb2

Arsnael:
Hi @ieugen 

Firstly thanks a lot for this work, I don't know much about gradle yet but it looks like it could speed up our build time, and maybe make it more stable too, so I'm feeling quite excited about it.

Second, well I'm trying to check a bit and try your build. I was just wondering, maybe we should in a first step try to be able to build the project without running the tests? Which would maybe help having a base working gradle build without spending too much time between run builds. Then we could make it work with all test suites in a second step?

Like for example with what would be the equivalent of a `mvn clean install -DskipTests`

So I tried this: `./gradlew clean build -x test --scan` (it seems the option `-x test` skips the tests)

It took 7mins > https://scans.gradle.com/s/wvjcva227grqe

Arsnael:
It's ok for now as it's not the focus of the work too but we have for practice to order dependencies alphabetically per section. We should maybe try to do something similar with Gradle?

Arsnael:
Played a bit with it to familiarize myself and got the compilation of `james-server:james-server-data-jmap` working, if you are interested:  https://github.com/linagora/james-project/commit/4db205b8396cb7ab2ad856ac920f5d2df558766e

I didn't make a fixup! as you seemed interested of making the tests work (`compileTestJava`) but use it, squash it or in any way fitting you if you feel you need it.

> https://scans.gradle.com/s/jznjvqhpggbq6

We might want to use this when we get a stable build too: https://docs.gradle.org/current/userguide/migrating_from_maven.html#migmvn:using_boms , or deps management is gonna be hell. I guess you know it and it's not the focus now, but just to keep in mind :)

Well I might be able to spend some more time on it this week if you want, so let me know how I can help you without us stepping on each other's foot ;)

ieugen:
Thank you for your feedback :). 
It's valuable and yes, the current focus is just to make it build.
I tried to order things by section first, alphabetically - was too much at this time.
I do think we can merge a lot of the existing components so I did not spend too much time on this.
It seems the modules are built around the technology and we get 3-4 artifacts with name like, e.g.  *-protocols-smtp-* .
I'm sure there are reasons for doing so but it's very confusing when dealing with them: which one should I use ?

It makes sense to have the build without tests first.
Right now the PR it's not ready for contributions since it's in chaos, but once it will build, I will slow down and it will be open for contributions. 

We will need a developer guide so please post a list of the questions / tasks you do when building with maven so we can translate that to gradle and also, try them out.

chibenwa:
> It seems the modules are built around the technology and we get 3-4 artifacts with name like, e.g. -protocols-smtp- .
> I'm sure there are reasons for doing so but it's very confusing when dealing with them: which one should I use ?

On the example you take:

 - You have the protocol library
 - It's adaptation to the James server
 - Integration tests (mpt)
 - guice bindings

Does it clarifies the split? How would you simplify this?

chibenwa:
> Right now the PR it's not ready for contributions since it's in chaos

:-)

Arsnael:
> We will need a developer guide so please post a list of the questions / tasks you do when building with maven so we can translate that to gradle and also, try them out.

Well if I'm wrong others can correct me or help complete this, but when we buildon the CI with Maven I believe it's something like this:

* `mvn package -DskipTests -T 1C` -> compilation of the project without the test suite at first. There is different plugins in the root pom being used to enforce some things with the compilation, like the `maven-checkstyle-plugin` enforcing those coding rules https://github.com/apache/james-project/blob/master/checkstyle.xml. We will need to check the equivalent of some of those plugins with Gradle. It seems the option `--parallel` in Gradle can replace `-T 1C` in Maven, from what I tested.

* `mvn package -T 1C` => full compilation with tests. Using some plugins like surefire to parallelize the tests. Is it doing it natively with gradle? It looks fast already anyway to me with the little I tested

Might be more, but I think that's the two main maven command ones being used

chibenwa:
@ieugen 

Regarding playing tests there's a few tricks you need to be aware of to fasten the docker tests.

In order to fasten things up:

 - We configured surefire to reuse forks. This prevents testContainer initialization on each test class, and allow docker container reuse for an entire maven module.

 - We configured surefire for parrallel test execution within a single module. It allows ie to go down to 4 min for testing mailbox/cassandra instead of 25+ minutes without parallelism.

chibenwa:
> We will need a developer guide so please post a list of the questions / tasks you do when building with maven so we can translate that to gradle and also, try them out.

:+1:

ieugen:
We have compilation without tests. 
Time on my machine: 2m36s
`./gradlew clean build -x test --scan`  -> https://gradle.com/s/r2j7vzecltalo  .

ieugen:
I will add some sub-task to track the required gradle migration steps. 
Please feel free to add steps if you think they are missing.
Also, I would love some help on the docker migration, checkstyle, etc.

@chibenwa, @Arsnael  : Parallel builds is an optimization IMO and not in the current goal.  
If it's done in maven we can do it in gradle, but for now let's focus on getting this merged so we don't have two build systems.

chibenwa:
>  but for now let's focus on getting this merged so we don't have two build systems.

Should we set an acceptance criteria? 

In my opinion, we should:

 - Be able to run tests with the new build system in less than 4h (we can iterate from that value)
 - Be able to build docker images from the gradle build output (so that we can run deployment tests against it)

ieugen:
@chibenwa : I agree. I have created a task and assigned it do you :D .
Of course you don't have to do it but I think you are the best person to do it https://issues.apache.org/jira/browse/JAMES-3273 .
We should satisfy Linagora's current build process so the transition will be smooth. 

@Arsnael : Would you like to handle https://issues.apache.org/jira/browse/JAMES-3271 ? 
There are some suggestions there. 
Regarding code formatting, I've used the gradle google-code-format plugin and formatted + verified the code as part of the build. 
Spotless plugin seems to have integration with that.

Arsnael:
> @Arsnael : Would you like to handle https://issues.apache.org/jira/browse/JAMES-3271 ?

I can try to give it a shot yes !

ieugen:
@chibenwa : Pleas assess the impact of the Gradle migration inside Linagora and ask people to test the build.
We need to discover if there are any blockers and what those blockers are otherwise we will not be able to make the merge.

chibenwa:
I'm currently collecting feedbacks :-)

Arsnael:
All green for me :
`./gradlew clean build -x test --scan` ->https://scans.gradle.com/s/xvchoumy2msda

chibenwa:
@ieugen I did my homeworks.

We came up with:

 - tests runs, within ~4hours
 - Today behaviour of `dockerfiles/compilation/java-11` image is unaltered

ieugen:
Some context: 

The project has a lots of inter dependencies, especially on the tests side.
The modules depend on a lot of `tests` artifacts which creates some classpath issues for integration tests.
I had to disable some tests with `enabled = false` to complete the build - please help out with those (use search to find them ).
This is a rough migration, things can be improved once we adopt gradle as the build system. 
Tests can probably be improved to run in parallel.
I will take a break from this a bit, please issue PR's for this PR with fixes.

ieugen:
Disabled tests: `git grep 'enabled = false'`

```
mpt/impl/imap-mailbox/cyrus/build.gradle:    enabled = false
mpt/impl/imap-mailbox/external-james/build.gradle:    enabled = false
mpt/impl/smtp/cassandra-rabbitmq-object-storage/build.gradle:    enabled = false
mpt/mavenplugin/build.gradle:    enabled = false
server/container/cli-integration/build.gradle:    enabled = false
server/container/core/build.gradle:    enabled = false
server/container/guice/blob-objectstorage-guice/build.gradle:    enabled = false
server/container/guice/cassandra-rabbitmq-guice/build.gradle:    enabled = false
server/container/guice/cassandra-rabbitmq-ldap-guice/build.gradle:    enabled = false
server/container/guice/jpa-guice/build.gradle:    enabled = false
server/container/guice/memory-guice/build.gradle:    enabled = false
server/mailet/integration-testing/build.gradle:    enabled = false
server/protocols/fetchmail/src/main/java/org/apache/james/fetchmail/FetchScheduler.java:    private volatile boolean enabled = false;
server/protocols/jmap-draft-integration-testing/memory-jmap-draft-integration-testing/build.gradle:    enabled = false
server/protocols/jmap-draft-integration-testing/rabbitmq-jmap-draft-integration-testing/build.gradle:    enabled = false
server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/build.gradle:    enabled = false
server/protocols/jmap-rfc-8621-integration-tests/memory-jmap-rfc-8621-integration-tests/build.gradle:    enabled = false
server/protocols/webadmin-integration-test/distributed-webadmin-integration-test/build.gradle:    enabled = false
server/protocols/webadmin-integration-test/memory-webadmin-integration-test/build.gradle:    enabled = false
server/protocols/webadmin/webadmin-mailbox/build.gradle:    enabled = false
```

Arsnael:
Hi @ieugen 

Unfortunately, I can't succeed running the build with tests. I ran it three times, and every time it fails on the same test:
`CassandraSubscriptionManagerTest > user1ShouldBeAbleToSubscribeTwoMailbox() FAILED`

I ran this command : `./gradlew clean build --scan` > https://scans.gradle.com/s/kmifqf6h673fm/console-log?anchor=644

I might try to take a look if I have time. However if other people could try run it so we can have more input, it would be great too :)

ieugen:
Strange. The tests pass on my machine.
It may be environment related. 
Can you please check the reason they failed? 
Also what is your git commit? Mine is 
```
commit 55c719054e8fa8bb7b90daf07c757a0c8de8a534 (HEAD -> JAMES-3260-gradle-poc, origin/JAMES-3260-gradle-poc, apache/JAMES-3260-gradle-poc)
Author: Eugen Stan <ieugen@apache.org>
Date:   Fri Jun 26 00:15:18 2020 +0300

    JAMES-3260 Some more build fixes
```

I've ran the build again an also a fail but after 43 min. 
It failed because the build is not reproducible: it creates a var/ directory for local derby data and it does not clean it between builds. 

```
Task :apache-james-mailbox:apache-james-mailbox-spring:test FAILED

BUILD FAILED in 43m 36s
511 actionable tasks: 511 executed
```

ieugen:
I pushed some fixes for spring-mailbox tests - moved var inside build so it gets cleaned.
Moved database properties file from main classpath to tests classpath.
Configuration should be supplied at deployment.

Arsnael:
I did pull to your latest commit so commit was the same as you. Error:

```
CassandraSubscriptionManagerTest > user1ShouldBeAbleToSubscribeTwoMailbox() FAILED	
    com.datastax.driver.core.exceptions.DriverInternalError at DriverThrowables.java:36	
        Caused by: java.lang.IllegalStateException at SessionManager.java:693
```

```
Execution failed for task ':apache-james-mailbox:apache-james-mailbox-cassandra:test'.	
	
BUILD FAILED in 10m 58s	
355 actionable tasks: 355 executed
```

ieugen:
Good, but look for the html report: 
The line that looks like bellow.
It contains details about all the tests. 
```
* What went wrong:
Execution failed for task ':apache-james-mailbox:apache-james-mailbox-spring:test'.
> There were failing tests. See the report at: file:///home/ieugen/proiecte/james/james-project/mailbox/spring/build/reports/tests/test/index.html

````

**Comment by ieugen** [comment]
@Arsnael , @chibenwa :  In time I think I will manage the tests that fail because of classpath issues.
However I'm not as familiar with the code (cassandra implementation) and I would appreciate help with tests that fail from other reasons.

One such case is `WithCassandraBlobStoreTest.mailsShouldBeWellReceivedByBothRecipient()` that fails with   :
```
org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a lambda expression in org.apache.james.MailsShouldBeWellReceived that uses org.apache.james.GuiceJamesServer 
Expecting:
 <false>
to be equal to:
 <true>
but was not within 10 seconds.
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:145)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:122)
	at org.awaitility.core.AssertionCondition.await(AssertionCondition.java:32)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:902)
```

If it fails with maven, then we can Ignore it for the purpose of our migration.
If it passes with maven then why is it failing with Gradle.

It would be great if we find a solution for this case.

File: backends-common/cassandra/build.gradle [hunk 1]
```diff
@@ -0,0 +1,67 @@
...
+
+dependencies {
+    implementation project(':james-core')
+    implementation project(':james-json')
+
+    implementation project(':james-server:james-server-task-api')
+    implementation project(':james-server:james-server-task-json')
+    implementation project(':james-server:james-server-util')
+
+    implementation 'com.fasterxml.jackson.core:jackson-annotations:2.9.9'
```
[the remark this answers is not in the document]
Arsnael: If you read the comments you might have seen I spotted this, https://docs.gradle.org/current/userguide/migrating_from_maven.html#migmvn:using_boms

I'm no Gradle expert at all, but that looked like the answer to your question?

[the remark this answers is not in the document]
ieugen: We have several options, but let's do that later.
We can defined a property like in maven and use that.
We can use a BOM.
We can keep it like it is and do search and replace from IDE. 
I like this last one because we don't limit the module coupling. 

One option is like this: https://dev.to/jmfayard/configuring-gradle-with-gradle-properties-211k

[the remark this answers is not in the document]
chibenwa: > We can keep it like it is and do search and replace from IDE.

Likely we would end up with diverging dependencies, auditing them will be a pain.

Arsnael:
@ieugen here is an attempt for checkstyle with Gradle: https://github.com/apache/james-project/pull/226

Work still in progress, there is so many checkstyle errors on test classes as it was not working with maven plugin for test classes somehow... Will finish it later

backends-common/cassandra/build.gradle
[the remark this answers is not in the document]
mbaechler: None of the proposed solutions are good ones: 
* properties do not carry semantic, it's just a String and it would be a regression wrt to maven
* BOM are for release-train (when you ship a library in small pieces that should have compatible versions)
* duplication is a pain, specifically when merging stuff where it's so easy to bring an outdated version

I'm pretty sure there's a better solution, it's why I asked in the first place

rouazana:
that's great, I was able to build James with gradle with the tests in less than 2 hours.

Unfortunately that's not very representative, as the main integrations tests are currently disabled:
```
$ git grep TODO:\ @ieugen
mpt/impl/imap-mailbox/external-james/build.gradle:    //TODO: @ieugen: Disabled temporarily because they require an environment variable: JAMES_ADDRESS=127.0.0.1  https://issues.apache.org/jira/browse/JAMES-3260
mpt/impl/smtp/cassandra-rabbitmq-object-storage/build.gradle:    // TODO: @ieugen Disabled temporarily. Fails with class not found  https://issues.apache.org/jira/browse/JAMES-3260
server/container/cli-integration/build.gradle:    //TODO: @ieugen: I disabled these as they fail with NoClassDefFoundError . Tests are part of https://issues.apache.org/jira/browse/JAMES-3260 .
server/container/guice/blob-objectstorage-guice/build.gradle:    //TODO: @ieugen: Disabled temporarily. Fails with java.lang.NoClassDefFoundError . Part of  https://issues.apache.org/jira/browse/JAMES-3260
server/container/guice/cassandra-rabbitmq-guice/build.gradle:    //TODO: @ieugen: Temporarily disabled. Failes to load classes at runtime. Part of https://issues.apache.org/jira/browse/JAMES-3258
server/container/guice/cassandra-rabbitmq-ldap-guice/build.gradle:    //TODO: @ieugen: Temporarily disabled. Fails to load classes at runtime. Part of https://issues.apache.org/jira/browse/JAMES-3258
server/container/util/src/test/java/org/apache/james/util/ReactorUtilsTest.java:    //TODO: @ieugen: Disabled temporarily for gradle migration - https://issues.apache.org/jira/browse/JAMES-3260
server/mailet/integration-testing/build.gradle:    //TODO: @ieugen: I disabled these as they fail with NoClassDefFoundError . Tests are part of https://issues.apache.org/jira/browse/JAMES-3260 .
server/protocols/jmap-draft-integration-testing/memory-jmap-draft-integration-testing/build.gradle:    // TODO: @ieugen: Disabled temporarily. Related to https://issues.apache.org/jira/browse/JAMES-3260
server/protocols/jmap-draft-integration-testing/rabbitmq-jmap-draft-integration-testing/build.gradle:    // TODO: @ieugen: Disabled temporarily. Related to https://issues.apache.org/jira/browse/JAMES-3260
server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/build.gradle:    // TODO: @ieugen: Disabled temporarily. Related to https://issues.apache.org/jira/browse/JAMES-3260
server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/build.gradle:    //TODO: @ieugen: this uses code from netty 4 and we use netty 3 !!!!
server/protocols/jmap-rfc-8621-integration-tests/memory-jmap-rfc-8621-integration-tests/build.gradle:    // TODO: @ieugen: Disabled temporarily. Related to https://issues.apache.org/jira/browse/JAMES-3260
server/protocols/webadmin-integration-test/memory-webadmin-integration-test/build.gradle:    //TODO: @ieugen Disabled temporarily. https://issues.apache.org/jira/browse/JAMES-3260
server/protocols/webadmin/webadmin-mailbox/build.gradle:    //TODO: @ieugen: Disabled because some tests fail with    java.lang.NoClassDefFoundError . Part of https://issues.apache.org/jira/browse/JAMES-3260
```
from these `external-james` should be configured only to run on demand
and some integration tests should be run only when BasicFeature tag is present. See for example in `server/protocols/jmap-draft-integration-testing/cassandra-jmap-draft-integration-testing/pom.xml`
```
<groups combine.self="override">org.apache.james.jmap.categories.CassandraAndElasticSea
rchCategory,org.apache.james.junit.categories.BasicFeature,org.apache.james.jmap.categories.EnableCucumber</groups>
```
other related issue:
```
> Task :james-server:james-server-blob:blob-cassandra:test FAILED
```
failed one time on my computer (out of memory), and then succeeded. I don't know if it was already flacky or if Gradle needs some specific memory configuration.

ieugen:
Regarding the tests - they are disabled because of classpath issues that I did not have the patience to fix.
Well, the purpose of this migration is not to improve the build speed.
Build time improvement comes after the migration to gradle.
There are some immediate benefits: 
- better dependency management
- other builds will only run changed projects

Once we have gradle we can improve by playing with: 
- parallel builds 
- build caching 
- make integration tests opt-in via a parameter: e.g `./gradlew build -Pcassandra-integration -P jpa-integration` ...
- composite builds - separate mailbox into it's own build system (same repo) that integrates seamlessly into james-server 

But all the above require a bit of thinking and I would not want to prolong the migration because of that. 

I worked on this for ~5 days straight and I believe it will open a lot of new options for us.
Le's make the migration and then improve it.

backends-common/cassandra/build.gradle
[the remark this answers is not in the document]
ieugen: There are but IMO they are out of scope for now (like the one in the link I shared).
I would like to see the migration done and then improve it in time.
There are a lot of things we can do and in multiple ways. 
Let's take them one at a time.

Marking this as resolved, if you disagree please reopen.

Arsnael:
If it works for @rouazana then I believe maybe the laptop I use for work might be not powerful enough for the integration tests, or maybe I have other issues... Should investigate more

ieugen:
Hi,
I fixed some more issues with the integration tests and now we are in this situation that I need some help with:
- some tests fail because they produce wrong results - this requires some knowledge about the implementation
- some tests fail because with `java.lang.NoClassDefFoundError: com/google/inject/internal/util/$Maps` . 
  I believe it's because of guice 3 - guice 4 classpath issue.
  I saw in maven: `<!-- JCloud switch relies on incompatible guice 3.3.0 for multibindings -->` .
  Can someone explain better what causes this issue? 
  I think I might be able to find a fix.

p.s. I will push soon once I run through all the tests again.

ieugen:
We have the build passing with some extra integration tests: https://gradle.com/s/xtmjpzeun6uka .

The disabled tests are the ones that fail.
Most tests are failing because of guice classpath issue. 
1-2 fail because of timeouts.
Some fail bacuse they shoud.

 git grep ' enabled = false '
```
mpt/impl/imap-mailbox/cyrus/build.gradle:    enabled = false
mpt/impl/imap-mailbox/external-james/build.gradle:    enabled = false
mpt/impl/smtp/cassandra-rabbitmq-object-storage/build.gradle:    enabled = false
mpt/mavenplugin/build.gradle:    enabled = false
server/container/cli-integration/build.gradle:    enabled = false
server/container/core/build.gradle:    enabled = false
server/container/guice/blob-objectstorage-guice/build.gradle:    enabled = false
server/container/guice/cassandra-rabbitmq-guice/build.gradle:    enabled = false
server/container/guice/cassandra-rabbitmq-ldap-guice/build.gradle:    enabled = false
server/data/data-jmap/build.gradle:    enabled = false
server/mailet/integration-testing/build.gradle:    enabled = false
server/protocols/jmap-draft-integration-testing/rabbitmq-jmap-draft-integration-testing/build.gradle:    enabled = false
server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/build.gradle:    enabled = false
server/protocols/webadmin-integration-test/distributed-webadmin-integration-test/build.gradle:    enabled = false
```

.mvn/gradle-enterprise.xml
```diff
@@ -0,0 +1,8 @@
+<gradleEnterprise>
+  <buildScan>
+    <termsOfService>
+      <url>https://gradle.com/terms-of-service</url>
```
mbaechler, line 4: As far as I understand this is not compatible with an Apache project.

event-sourcing/event-sourcing-core/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+    id 'scala'
+}
+
+configurations {
+    tests
+}
+
+dependencies {
+    api project(':event-sourcing:event-sourcing-event-store-api')
+    api project(':event-sourcing:event-sourcing-pojo')
```
mbaechler, line 15: this one is already provided by `event-sourcing:event-sourcing-event-store-api` as an api dependency I guess

mpt/impl/imap-mailbox/cassandra/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+plugins {
+    id 'java-library'
+}
+
+configurations {
+    tests
+}
+
+dependencies {
+    implementation project(':james-backends-common:apache-james-backends-cassandra')
```
mbaechler, line 13: this kind of dependency should be provided by `apache-james-mailbox:apache-james-mailbox-cassandra`. We should not have to repeat it here IMO.

.mvn/gradle-enterprise.xml
ieugen: I've asked on legal-discuss@a.o for advice. 
If it's not ok.

event-sourcing/event-sourcing-core/build.gradle
ieugen: Yes. I've started 

> I think you use `implementation` deps too much and it avoids the good side of transitive dependencies (of course the wrong side would be to expose too much apis and create bad implicit dependencies)

Thanks for the review. 
This was a bit on purpose. 
There are too many modules IMO and I'm not familiar with their dependencies. 
We should improve this over time.
But we should be mindfull of what we consider API and what not.
In later modules I migrated I've put API all project modules, but not external libraries. 
I think this will help us make dependencies more visible - because we HAVE to deal and add each one. 
We can leave them like this and improve the situation when we re-organize the modules.

mpt/impl/imap-mailbox/cassandra/build.gradle
ieugen: Thanks for the review and suggestion.
It is the same in the pom.xml, and this was migrated by gradle.
I'm not very familiar with the dependencies and how they should be done so I stayed away from that.
My goal is and was to make the build work as it is in maven.
We can always improve once the migration is done. 
I think this can be done as part of the re-organization we have planned. 
We can drop and merge modules then and also cleanup dependencies as part of the process.

One thing I think we should avoid is dragging this migration process for too long for minor things. 
Let's have a good migration checklist https://issues.apache.org/jira/browse/JAMES-3273 and decide what is a migration blocker and what is not.

.mvn/gradle-enterprise.xml
ieugen: Legal is ok https://mail-archives.apache.org/mod_mbox/www-legal-discuss/202007.mbox/browser (Reply from Roy Fielding may not be yet in archive) 
Waiting on infra to agree to the TOS and we are set.

Arsnael:
Just to let you know, started investigating migrating our CI towards Gradle: #231

Still work in progress

ieugen:
> Just to let you know, started investigating migrating our CI build towards Gradle: #231
> 
> Still work in progress

Thanks for the update. 
Use the multibranch pipeline plugin with git / github sources.
Control the build from the Jenkinsfile. 
You can specify a path for the Jenkinsfile if you need anther build configuration.

p.s. Have you made any progress with donating a CI server for James?

Arsnael:
> Use the multibranch pipeline plugin with git / github sources.
> Control the build from the Jenkinsfile.
> You can specify a path for the Jenkinsfile if you need anther build configuration.

Just trying to adapt what we have currently running on our CI for now. If I can manage to build similar fat jars with Gradle compared to what we do with Maven, then the rest of the flow we have on our Jenkins should follow without changes

> p.s. Have you made any progress with donating a CI server for James?

I'm not sure of the status of this myself, and I think it will take time. Maybe @aduprat ?

ieugen:
> Just trying to adapt what we have currently running on our CI for now. If I can manage to build similar fat jars with Gradle compared to what we do with Maven, then the rest of the flow we have on our Jenkins should follow without changes

I believe I did part of that by using the application plugin in some guice modules.
You don't get a fat jar but you get the lib directory.
To also build a fat jar use the shadow plugin: 

https://github.com/johnrengelman/shadow
https://imperceptiblethoughts.com/shadow/

Arsnael:
> I believe I did part of that by using the application plugin in some guice modules.
> You don't get a fat jar but you get the lib directory.
> To also build a fat jar use the shadow plugin:
> 
> https://github.com/johnrengelman/shadow
> https://imperceptiblethoughts.com/shadow/

I do use the shadow library. I might get a closer look on what you did though, cause the lib folder was what we had until now with Maven. Might be easier to debug and see what dependency is possibly missing though (because I seem to have that issue actually)

ieugen:
> > I believe I did part of that by using the application plugin in some guice modules.
> > You don't get a fat jar but you get the lib directory.
> > To also build a fat jar use the shadow plugin:
> > https://github.com/johnrengelman/shadow
> > https://imperceptiblethoughts.com/shadow/
> 
> I do use the shadow plugin. I might get a closer look on what you did though, cause the lib folder was what we had until now with Maven. Might be easier to debug and see what dependency is possibly missing though (because I seem to have that issue actually)

See here how we do it for a james distribution https://github.com/ieugen/james-self-hosting-sandbox .
https://github.com/ieugen/james-self-hosting-sandbox/blob/master/james-basic-server/Dockerfile
https://github.com/ieugen/james-self-hosting-sandbox/blob/master/james-basic-server/docker-compose.yaml

See here how I build the distribution for server-app https://github.com/apache/james-project/blob/19e0a9a682de79875ef13b9e1055dd3f54b3694e/server/app/build.gradle

chibenwa:
So far the gradle migration do not make progress, for the past two years.

I will close related PRs to reflect this matter of fact.

Don't hesitate to re-open if you work on this.

PR #219, opened 2020-06-21
[ADR] Define quality levels

chibenwa:
See https://www.mail-archive.com/server-dev@james.apache.org/msg66909.html

### DIFF
--- a/src/adr/0040-quality-levels-definitions.md
+++ b/src/adr/0040-quality-levels-definitions.md
@@ -8,13 +8,23 @@ Accepted (lazy consensus)
 
 ## Context
 
+We hereby define as an artifact compiled artifact that external people consumes. This includes:
+
+ - libraries
+ - Mail servers
+ - Extensions for James Mail Servers
+ - Command line tools
+
+We designate as a feature an optional, opt-in behaviour of a James server that can be configured by 
+user willing to rely on it.
+
 James as a project delivers several artifacts, and features. In order for project users to better
 understand the underlying quality of the artifact they use, as well as the level of risk associated,
 we need to better define some quality levels.
 
 ## Decision
 
-By **supported** we mean that:
+For a given artifact or feature, by **mature** we mean that:
 
  - *interfaces* in components need a contract test suite
  - *implementation* of these interfaces need to pass this contract test suite which provides unit tests
@@ -29,20 +39,20 @@ artifacts or features.
 
 By **experimental** we designate a part of the code not matching yet the above requirements. However some
 active contributors are willing to raise the quality level of this component, and eventually make it 
-supported. Or at least are willing to support users.
+mature. Or at least are willing to support users.
 
-Users should have low expectations regarding experimental artifacts. They are encouraged to contribute to this 
-component in order to raise its code quality.
+Users should have low expectations regarding experimental artifacts or features. They are encouraged to contribute to them 
+in order to raise its code quality.
 
-By **unsupported** we mean that a component do not match most of the *supported* quality conditions. Active 
+By **unsupported** we mean that an artifact or feature do not match most of the *mature* quality conditions. Active 
 contributors do not feel confident delivering support for it. This artifact or feature might be deprecated and 
 removed from future James releases. Users are strongly encouraged to contribute to the artifact development.
 
 ## Consequences
 
-Quality levels needs to be mentioned explicitly in the documentation, per product, and for each product, per feature.
+Quality levels needs to be mentioned explicitly in the documentation, per artifact, and per feature.
 
-We need to audit existing artifacts and features to make a list of experimental and unsupported components. We will maintain
+We need to audit existing artifacts and features to make a list of experimental and unsupported artifacts. We will maintain
 JIRA tickets opened from them, detailing for each one of them expected actions to raise the quality level. Maintaining such 
 tickets will encourage contributions on experimental and unsupported components.
 
--- a/src/adr/0040-quality-levels-definitions.md
+++ b/src/adr/0040-quality-levels-definitions.md
@@ -42,7 +42,7 @@ active contributors are willing to raise the quality level of this component, an
 mature. Or at least are willing to support users.
 
 Users should have low expectations regarding experimental artifacts or features. They are encouraged to contribute to them 
-in order to raise its code quality.
+in order to raise its quality.
 
 By **unsupported** we mean that an artifact or feature do not match most of the *mature* quality conditions. Active 
 contributors do not feel confident delivering support for it. This artifact or feature might be deprecated and 
--- a/src/adr/0040-quality-levels-definitions.md
+++ b/src/adr/0040-quality-levels-definitions.md
@@ -51,7 +51,7 @@ removed from future James releases. Users are strongly encouraged to contribute 
 
 ## Consequences
 
-Quality levels needs to be mentioned explicitly in the documentation, per artifact, and per feature.
+Quality levels need to be mentioned explicitly in the documentation, per artifact, and per feature.
 
 We need to audit existing artifacts and features to make a list of experimental and unsupported artifacts. We will maintain
 JIRA tickets opened from them, detailing for each one of them expected actions to raise the quality level. Maintaining such 
--- a/src/adr/0040-quality-levels-definitions.md
+++ b/src/adr/0040-quality-levels-definitions.md
@@ -1,0 +1,51 @@
+# 40. Quality levels definition
+
+Date: 2020-06-21
+
+## Status
+
+Accepted (lazy consensus)
+
+## Context
+
+James as a project delivers several artifacts, and features. In order for project users to better
+understand the underlying quality of the artifact they use, as well as the level of risk associated,
+we need to better define some quality levels.
+
+## Decision
+
+By **supported** we mean that:
+
+ - *interfaces* in components need a contract test suite
+ - *implementation* of these interfaces need to pass this contract test suite which provides unit tests
+ - Decent integration tests coverage is needed
+ - Performance tests need to be conducted out
+ - Quality Assurance with external clients needs to be conducted out
+ - known existing production deployments/usages
+ - usable documentation
+
+This is the maximum quality level delivered by the James project. Users should feel confident using these
+artifacts or features.
+
+By **experimental** we designate a part of the code not matching yet the above requirements. However some
+active contributors are willing to raise the quality level of this component, and eventually make it 
+supported. Or at least are willing to support users.
+
+Users should have low expectations regarding experimental artifacts. They are encouraged to contribute to this 
+component in order to raise its code quality.
+
+By **unsupported** we mean that a component do not match most of the *supported* quality conditions. Active 
+contributors do not feel confident delivering support for it. This artifact or feature might be deprecated and 
+removed from future James releases. Users are strongly encouraged to contribute to the artifact development.
+
+## Consequences
+
+Quality levels needs to be mentioned explicitly in the documentation, per product, and for each product, per feature.
+
+We need to audit existing artifacts and features to make a list of experimental and unsupported components. We will maintain
+JIRA tickets opened from them, detailing for each one of them expected actions to raise the quality level. Maintaining such 
+tickets will encourage contributions on experimental and unsupported components.
+
+## References
+
+ - Mailing list discussion: https://www.mail-archive.com/server-dev@james.apache.org/msg66909.html
--- a/src/adr/0040-quality-levels-definitions.md
+++ b/src/adr/0040-quality-levels-definitions.md
@@ -27,6 +27,7 @@ we need to better define some quality levels.
 For a given artifact or feature, by **mature** we mean that:
 
  - *interfaces* in components need a contract test suite
+ - *interfaces* have several implementations
  - *implementation* of these interfaces need to pass this contract test suite which provides unit tests
  - Decent integration tests coverage is needed
  - Performance tests need to be conducted out
@@ -37,7 +38,7 @@ For a given artifact or feature, by **mature** we mean that:
 This is the maximum quality level delivered by the James project. Users should feel confident using these
 artifacts or features.
 
-By **experimental** we designate a part of the code not matching yet the above requirements. However some
+By **experimental** we designate an artifact or feature not matching yet the above requirements. However some
 active contributors are willing to raise the quality level of this component, and eventually make it 
 mature. Or at least are willing to support users.
 


dleangen:
Thank you for this initiative.

What is not clear to me is: is this aspirational, or descriptive?

What I am advocating for is something that is descriptive of the **actual** state of James, so people new to James know exactly what to expect (or not), both from the code as well as from the community.

chibenwa:
> What is not clear to me is: is this aspirational, or descriptive?

It is descriptive.

Note that some features of supported products are however "experimental" or "unsupported", leading to an unclear overview as a user.

An example would be fetchmail usage.

I believe that a better extension packaging would allow to take these experimental/unsupported features out of the default supported products, and would better educate our operators about what is supported or not, and the underlying risks.

> What I am advocating for is something that is descriptive of the actual state of James, so people new to James know exactly what to expect (or not), both from the code as well as from the community.

Note that as a consequence, we should open JIRA in order to maintain a clear description of James current status, and encourage contributions on experimental / unsupported components.

This ADR serve as a basis to build this exact diagnostic.

Does it makes sense?

dleangen:
Thank you. That clarifies a few points, but muddies a few others. Notably, you use the word "supported" in your description, but in the mail discussion there was a fierce reaction to any notion of "support".

From what I understood from the mail thread and if this is indeed descriptive and not aspirational, **everything** should be labeled as "unsupported", which kind of makes this whole initiative moot. At least **I** am quite confused.

But then again, producing this ADR is a very good idea because hopefully it will clarify all the confusion. :)

chibenwa:
> Notably, you use the word "supported" in your description, but in the mail discussion there was a fierce reaction to any notion of "support".

Good point. Then let's change that terms as it might bring unreasonable expectations from operators.

> From what I understood from the mail thread and if this is indeed descriptive and not aspirational, everything should be labeled as "unsupported", which kind of makes this whole initiative moot. At least I am quite confused.

Let's refine our vocabulary to make this cristal clear.

-------------------

Here are some proposals coming to my mind for a better wording:

 - stable / experimental / unsupported -> I don't like the stable wording as it might give false expectations regarding API stability (which imo is orthogonal to our problem)

 - mature / experimental / unsupported -> maybe better ?

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,51 @@
...
+
+Date: 2020-06-21
+
+## Status
+
+Accepted (lazy consensus)
+
+## Context
+
+James as a project delivers several artifacts, and features. In order for project users to better
```
Arsnael, line 11: What do you mean by artifacts here?

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,51 @@
...
+Users should have low expectations regarding experimental artifacts. They are encouraged to contribute to this 
+component in order to raise its code quality.
+
+By **unsupported** we mean that a component do not match most of the *supported* quality conditions. Active 
+contributors do not feel confident delivering support for it. This artifact or feature might be deprecated and 
+removed from future James releases. Users are strongly encouraged to contribute to the artifact development.
+
+## Consequences
+
+Quality levels needs to be mentioned explicitly in the documentation, per product, and for each product, per feature.
```
Arsnael, line 43:
```suggestion
Quality levels need to be mentioned explicitly in the documentation, per profile, and for each profile, per feature.
```

Arsnael: I am not sure that a decision has been taken yet regarding the misleading term `product`, but I think propositions made like `profile` or `flavor` should maybe better used instead?

src/adr/0040-quality-levels-definitions.md
chibenwa: `artifacts` => things James project users consumes.

 - Libraries as stated here: 
 - Mail Servers
 - Command line tools
 - Jars that can be used as an extension for a running James server

src/adr/0040-quality-levels-definitions.md
chibenwa: Let's keep the word "artifact" instead of product/profile given that I correctly defines it.

chibenwa:
> Do we have a more up to date version of this somewhere or is it something that needs to be done too?

We release 3.0.0, so the related roadmap had been carried out.

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,61 @@
...
+
+James as a project delivers several artifacts, and features. In order for project users to better
+understand the underlying quality of the artifact they use, as well as the level of risk associated,
+we need to better define some quality levels.
+
+## Decision
+
+For a given artifact or feature, by **mature** we mean that:
+
+ - *interfaces* in components need a contract test suite
```
mbaechler, line 29: I would add: 'at least two implementations of the interface'

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,61 @@
...
+
+This is the maximum quality level delivered by the James project. Users should feel confident using these
+artifacts or features.
+
+By **experimental** we designate a part of the code not matching yet the above requirements. However some
+active contributors are willing to raise the quality level of this component, and eventually make it 
+mature. Or at least are willing to support users.
+
+Users should have low expectations regarding experimental artifacts or features. They are encouraged to contribute to them 
+in order to raise its code quality.
```
mbaechler, line 45:
```suggestion
in order to raise its quality.
```

chibenwa:
I removed the only "code" mention.

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,62 @@
...
+Users should have low expectations regarding experimental artifacts or features. They are encouraged to contribute to them 
+in order to raise its quality.
+
+By **unsupported** we mean that an artifact or feature do not match most of the *mature* quality conditions. Active 
+contributors do not feel confident delivering support for it. This artifact or feature might be deprecated and 
+removed from future James releases. Users are strongly encouraged to contribute to the artifact development.
+
+## Consequences
+
+Quality levels needs to be mentioned explicitly in the documentation, per artifact, and per feature.
```
rouazana, line 54:
```suggestion
Quality levels need to be mentioned explicitly in the documentation, per artifact, and per feature.
```
or
```suggestion
Quality level needs to be mentioned explicitly in the documentation, per artifact, and per feature.
```

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,62 @@
...
+artifacts or features.
+
+By **experimental** we designate an artifact or feature not matching yet the above requirements. However some
+active contributors are willing to raise the quality level of this component, and eventually make it 
+mature. Or at least are willing to support users.
+
+Users should have low expectations regarding experimental artifacts or features. They are encouraged to contribute to them 
+in order to raise its quality.
+
+By **unsupported** we mean that an artifact or feature do not match most of the *mature* quality conditions. Active
```
rouazana, line 48:
```suggestion
By **unsupported** we mean that an artifact or feature does not match most of the *mature* quality conditions. Active 
```

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,62 @@
...
+ - Mail servers
+ - Extensions for James Mail Servers
+ - Command line tools
+
+We designate as a feature an optional, opt-in behaviour of a James server that can be configured by 
+user willing to rely on it.
+
+James as a project delivers several artifacts, and features. In order for project users to better
+understand the underlying quality of the artifact they use, as well as the level of risk associated,
+we need to better define some quality levels.
```
dleangen, line 23: The document defines "quality levels" without defining "quality".

I think that "level" is obvious enough that it doesn't require any further definition, but "quality" may be interpreted differently by different people.

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Decision
+
+For a given artifact or feature, by **mature** we mean that:
+
+ - *interfaces* in components need a contract test suite
```
dleangen, line 29: Would it be useful to describe "separation of API/implementation"? It is a term I have used a few times on the list, but I'm not sure I use it in the same way as others. (I have strongly corrupted by OSGi thinking, and it has a very clear meaning over there.)

"Interface" as you use it seems to suggest at the class level, whereas the real "interface" of a component is actualy at the package level, which may be why they call it "API" instead of "interface", to avoid confusion. Usually it is not just a single Java interface that defines behaviour, but rather an entire package. (Should generally not expand beyond a package.)

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,62 @@
...
+ - *interfaces* have several implementations
+ - *implementation* of these interfaces need to pass this contract test suite which provides unit tests
+ - Decent integration tests coverage is needed
+ - Performance tests need to be conducted out
+ - Quality Assurance with external clients needs to be conducted out
```
dleangen, line 34: What do you mean by "Quality Assurance"? Is that defined in a different ADR?

src/adr/0040-quality-levels-definitions.md
```diff
@@ -0,0 +1,62 @@
...
+
+ - libraries
+ - Mail servers
+ - Extensions for James Mail Servers
+ - Command line tools
+
+We designate as a feature an optional, opt-in behaviour of a James server that can be configured by 
+user willing to rely on it.
+
+James as a project delivers several artifacts, and features. In order for project users to better
```
dleangen, line 21: These seem to me like important and fundamental enough concepts to either expand the scope of this ADR, or to have a separate ADR that clearly defines (as a base concept for other ADRs like this one) the concepts of "artifact" and "feature".

src/adr/0040-quality-levels-definitions.md
dleangen, line 48: In the referenced email, @chibenwa wrote:

> This is a fair point, but 'quality' differs from 'support' concerns.

I agree with this point, which is why I have trouble with the "unsupported" label.

dleangen:
First, I want to mention again that I think this is a really good initiative. Personally, I think it's really important that people can "know" to some extent what they are getting, especially from a system as huge and important as James.

> Here are some proposals coming to my mind for a better wording:
> 
> stable / experimental / unsupported -> I don't like the stable wording as it might give false expectations regarding API stability (which imo is orthogonal to our problem)
> 
> 
> mature / experimental / unsupported -> maybe better ?

It's difficult nailing down a vocabulary that is commonly shared.

Maybe the orthogonal concepts would be something like:

* **Maturity** (mature vs. immature) - The idea that a feature that is either relatively new and in many respects "untested", vs. a feature that has been around for a while, has been seen and used by many eyes, and is considered to be fairly "battle hardened".
* **Formality** - Relating to the community requirements regarding whether a feature goes through some kind of formal RFQ process, or if it is just thrown together by a single committer without much oversight.
* **Quality** - Relating to whether or not a feature respects the stated design principles of the community (including testing, documentation, code review...).
* **Reliability** - Relating to whether or not a feature does what it is stated to do.
* **Support** (supported vs. unsupported) - the idea relating to the community's commitment to ensuring that a module / feature (or whatever) remains reliable.

(I just realized that this all implies that the contract be very clearly stated.)

I quickly tested these, and they look at first glance to be orthogonal. (High quality but low reliability vs low quality and high reliability, etc.) They are debatable, so feel free to suggest improvements.

The list "mature", "experimental", "unsupported" seems to mix these concerns a little. But it is difficult to come up with a practical list. "Quality" needs a lot of definition, and likely depends on the other terms.

So based on these, just a thought, but perhaps the first step is to clearly define the "Formality" and "Design Principles" of the community (which to some extent you already do in this ADR), then it would be very easy to map a quality level.

src/adr/0040-quality-levels-definitions.md
chibenwa: > The document defines "quality levels" without defining "quality".

At one point it becomes hard for me to betterdefines such a generic term. I believes the levels definition disambiguate clearly this.

src/adr/0040-quality-levels-definitions.md
chibenwa: > Would it be useful to describe "separation of API/implementation"?

:+1: and luckily it is the case almost everywhere.

> "Interface" as you use it seems to suggest at the class level, whereas the real "interface" of a component is actualy at the package level, which may be why they call it "API" instead of "interface", to avoid confusion. Usually it is not just a single Java interface that defines behaviour, but rather an entire package

If others don't complain, I will replace "interfaces" terms by "APIs".

src/adr/0040-quality-levels-definitions.md
chibenwa: I mean "Manual Testing". Quality Assurance is in my opinion a common practice of the industry. Defining it is in my opinion not rely needed.

src/adr/0040-quality-levels-definitions.md
chibenwa: I think the context part does it well enough. If it is not clear enough, please provide alternative wordings.

src/adr/0040-quality-levels-definitions.md
chibenwa: Propose an alternative please. 

I want a term that reflexes the "nothing to expect" concepts. "unsupported" convey this idea quite well to be honest.

src/adr/0040-quality-levels-definitions.md
dleangen: That is not quite what I meant. I was not commenting on the words, I was commenting on the scope. This ADR is about "quality". However, the terms "artifacts" and "features" seem to be so fundamental that they should not be hidden away in a document that describes something different.

So we should either:

1. Expand the scope of this document (i.e. it defines "Artifacts", "Features", and "Quality Levels" and not just "Quality Levels", or
2. Have a separate ADR that defines "Artifacts" and "Features".

Then this can be referred to from other discussions which will surely use these fundamental concepts.

Just something to consider, maybe.

src/adr/0040-quality-levels-definitions.md
dleangen: You may be surprised as to how something that **ought** to be clearly defined in the industry creates a lot of conflict. :)

If there is already a clear _de facto_ _modus operandi_ for the James community, or if nobody challenges it, then I guess it's fine. :)

src/adr/0040-quality-levels-definitions.md
dleangen: Writing is really hard, I agree. The hardest part is writing something "objective" that everybody understands the same way, especially in a project like this with people from different countries, backgrounds, cultures, education, generations... We'll never have a completely common understanding, so the trick is to decide the right point at which we should draw a boundary given the time and resources we have available. We will never completely agree on where that boundary should be. :)

src/adr/0040-quality-levels-definitions.md
dleangen: I wrote in a separate comment below how I think the concepts are a bit mixed up. But again, there comes a point at which a boundary has to be drawn, so even though IMO it is confusing, I leave the decision up to you. :)

src/adr/0040-quality-levels-definitions.md
chibenwa: I have a different approach here.

I started another document to explain which server we ship as part of the James project, then I hope that with @ieugen we come up with a list of libraries we deliver. Each one of them will get their own ADR, giving a de-facto definition (and exhaustive list of artifacts).

Regarding features, refining extension mechanisms and planning the work for their extraction would do the same.

src/adr/0040-quality-levels-definitions.md
chibenwa: Actually this document, and the "mature" quality level dis-ambiguify what we mean by quality level, no?

chibenwa:
What would you think about `mature / experimental / immature` ?

src/adr/0040-quality-levels-definitions.md
dleangen: Are you saying that "mature" implies "high quality"? In the definitions I proposed, a component could be mature, but unreliable.

I do agree that it's not easy figuring out how all the pieces fit together. Quality really depends on the objective. There is not "absolute". That is why context is so important.

ieugen:
Kubernetes uses some words to describe the API stability as well https://kubernetes.io/docs/concepts/overview/kubernetes-api/ . 
It's nice to llok at what others are doing. 
They have a process for moving an API from alfa -> beta -> stable. 
But they cover the API (interface) not the implementation. 
I believe they consider the implementation stability as part of the API stability.
But they do get a lot more use of their components.

chibenwa:
I would like to make a push and merge this.

Any comments you would like me to address before ?

ieugen:
Let's use it like this and see where it goes.

chibenwa:
Merged

PR #221, opened 2020-06-21
 JAMES-3261 Build a ZIP distribution for Guice memory 

chibenwa:


chibenwa:
I revive this work.

The goal: offer Apache downloads for Guice servers (memory, distributed, JPA, etc...)

Arsnael (approved):

chibenwa:
Let's wait a bit more for input, and then merge it in a few days ;-)

I'll then contribute ZIP packaging for other guice servers.

jeantil:
would this be the start of an answer to https://www.mail-archive.com/server-user@james.apache.org/msg16455.html ?

chibenwa:
... I missed that one. He definitely miss an answer.

Well on the long term plan, in the back of my head, I dream about removing Spring support, but then we need a downloadable artifacts for Guice.

So downloadable from 3.6.1 likely just for them (I propose myself)

Deprecate Spring in 3.7.0,

And target a removal in 3.8.0.

jeantil:
> He definitely miss an answer.

I haven't answered because I am unsure what a minimal binary would look like ... for a pure open SMTP relay, I guess that the SMTP protocol + an in memory queue + file based RRT would be enough, add in some user store  (file based? ) for authenticated smtp 

I'm all for removing spring, one dependency injection runtime is enough. In my understanding though, the spring runtime is the historical one, dropping that means breaking all the custom servers built on top of it. Sounds like a major breaking change to me...

chibenwa:
> I haven't answered because I am unsure what a minimal binary would look like ... for a pure open SMTP relay, I guess that the SMTP protocol + an in memory queue + file based RRT would be enough, add in some user store (file based? ) for authenticated smtp

Well we already have JPA-SMTP.

If it do not fit his needs he likely need to assemble his own Guice server...

> Sounds like a major breaking change to me...

Agree, the idea to have a 4.0 had been around. It would be an opportunity to also rename the servers according to their purpose (and no longer according to the underlying technologies):

 - Memory -> Testing server
 - JPA -> Simple/Basic server
 - JPA-SMTP -> Simple/Basic processing server
  - Cassandra-ElasticSearch-Rabbit-S3 -> Distributed server.
 
We might not be that far away from it.... A bit of doc, a bit of maven module renames, and we can have something convincing IMO.

chibenwa:
> Let's wait a bit more for input, and then merge it in a few days ;-)

3 days already.

I'll rebase that to get it in line for the merge.

PR #222, opened 2020-06-22
[ADR] Review James project server offering

chibenwa:


src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+as more features will be packaged as part of it, configuration shows via examples advanced usage possibilities, and the documentation is centered 
+around feature extensibility.
+
+Note that transitionning from basic server to the advanced one will be easy as they rely on the same technologies.
+
+ - **Distributed server** will remain unchanged.
+
+ - **Testing server** will rely on current `memory guice` server.
+
+ - **Mail processing server** will rely on current `JPA SMTP guice` server.
```
dleangen, line 88: Is there a link to more information about this one?

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+
+Also, we should separate our server application from the technical guice bindings, and group them together for better discoverability.
+
```
dleangen, line 91: Are you sure this should be bundled into the same ADR?

dleangen:
Very nice!

src/adr/0041-james-servers-offering.md
chibenwa: Let see, but this should be quite consensual and hopefully addressed at the same time...

src/adr/0041-james-servers-offering.md
chibenwa: Not really, it started as a proof of concept and is mostly undocumented, however I did included it as it satisfies an unmentioned use case that some server like Haraka do address. 

Maybe I should develop mail processing as a sub-concern of customization in this document.

dleangen: Ok, that's interesting. I've been wanting to ask a question relating to this, and now is my chance. I will do it on the list though, as it takes the discussion off into a different direction.

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+ - **JPA SMTP** (`server/container/guice/jpa-smtp`) demonstrate a Mail Transfer Agent, dedicated to mail
+processing, that do not ship the James Mailbox, as well as Mail Delivery protocols like IMAP, JMAP or POP3.
+
+We can notice that this offer is specified in technical terms, and do not directly map to identified user needs.
+
+Also the large amount of servers makes the choice harder to make for newcomers.
+
+## Decision
+
+We should deliver a subset of mail server directly adressing users needs, and brand them according to their use cases
```
mbaechler, line 63:
```suggestion
We should deliver a subset of mail server directly addressing users needs, and brand them according to their use cases
```

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+and not due to their technical details.
```
mbaechler, line 64: I agree but let not fool ourselves: where the mail are stored is not a technical detail for a mail server.

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+
+In order to get a clearer offer we should first reduce the server cardinality. This can be achieved via:
+
+ - **Spring server** deprecation and removal
+ - Merging Guice servers with their LDAP variations (see [ADR 36 about guice module choosing](0036-against-use-of-conditional-statements-in-guice-modules.md))
+ - **Cassandra-guice** deprecation and removal
```
mbaechler, line 70: We have to provide a decent substitution first (JPA is not a good substitution for now)

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+
+We should then rename our servers in order to match the use cases:
+
+ - **Basic server** addresses the self-hosting use case. It will be based on current *jpa-guice* server but with the clear intent to 
+lower adoption barriers. It exposes limited configurations in order not to confuse the user. Configuration and documentation is centered
```
mbaechler, line 75: I don't agree about "It exposes limited configurations in order not to confuse the user". I would really prefer a solution where "it requires very few configuration to start using it".
It would prevent the need for a switch to another server once people feel limited by the arbitrary feature selection we make.

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+ - Merging Guice servers with their LDAP variations (see [ADR 36 about guice module choosing](0036-against-use-of-conditional-statements-in-guice-modules.md))
+ - **Cassandra-guice** deprecation and removal
+
+We should then rename our servers in order to match the use cases:
+
+ - **Basic server** addresses the self-hosting use case. It will be based on current *jpa-guice* server but with the clear intent to 
+lower adoption barriers. It exposes limited configurations in order not to confuse the user. Configuration and documentation is centered 
+around self hosting use cases.
+
+ - **Advanced server** addresses the customization use cases. It will be based on the current *jpa-guice* server. It differs from the basic server
```
mbaechler, line 78: In my opinion this one is useless

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+ - **Testing server** will rely on current `memory guice` server.
+
+ - **Mail processing server** will rely on current `JPA SMTP guice` server.
+
+Also, we should separate our server application from the technical guice bindings, and group them together for better discoverability.
+
+We should adopt this folder structure:
+
+```
+server/apps
```
mbaechler, line 95: why server _and_ apps? why not choose only one?

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+\_____ mail-processing-server
+```
+
+## Consequences
+
+By better focussing our offer on community needs we expect a larger adoption of James project servers.
+
+Such a split will furthermore ease documentation efforts.
+
+The split between basic and advanced server will furthermore ease bug reports as the customization capabilities of the basic server are limited.
```
mbaechler, line 109: that's the main disagreement point.

dleangen: Perhaps we need to post a clear hypothesis and somehow test it in order to resolve these kinds of disagreements more objectively.

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+
+ - **Spring server** (`server/app`): barely maintained, Spring server exposes configuration allowing 
+class level overides, resulting in untested componant combinaison leading to unexpected runtime errors. 
+Its usage requires configuration to adapt it to the user use cases, wich can be 
+not obvious for people with limited knowledge/resources, and can discourage self hosting usages.
+
+Guice servers includes:
+
+ - **Cassandra - guice** (`server/container/guice/cassandra-guice`) and its LDAP counter-part
+(`server/container/guice/cassandra-guice-ldap`): about to be deprecated, this guice server relies on
```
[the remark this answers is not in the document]
chibenwa: If we find people willing to find interests in it and devote efforts maintaining it, then deprecating something have no meaning.

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+a Cassandra & ElasticSearch database without achieving distribution (as it misses a messaging component).
```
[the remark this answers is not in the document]
chibenwa: By messaging component, I mean something like rabbitMQ, allowing to have several James server working together. Hopefully I am speaking of:

 - The EventBus (distributed is need for IMAP & soon JMAP) - a nice to have is the workQueue pattern for the EventBus groups.
 - The TaskManager
 - The MailQueue (optional, I agree, it is not needed for correctness)

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+
+ - **Distributed James** (`server/container/guice/cassandra-rabbitmq-guice`) and its LDAP counter part
+(`server/container/guice/cassandra-rabbitmq-ldap-guice`) enabled to achieve the distributed use case. This
+server furthermore is extensible, and ships many extra features by default. It is however hard to operate.
```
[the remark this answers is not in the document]
chibenwa: I don't understand. 

(what I meant by extensible here is "I can implement arbitrary features on top of distributed james the same way that I can do it with the Advanced server")

src/adr/0041-james-servers-offering.md
chibenwa: >  We have to provide a decent substitution first (JPA is not a good substitution for now)

If it is non consensual then let's forget this deprecation. 

I would like us to think of a non technical branding for **Cassandra-guice**.

Also, I'm intending to present in this document a short term (3 month, with a delivery hopefully in 3.6.0) achievable offerings. Things not following this temporality will need IMO to be discussed separately.

src/adr/0041-james-servers-offering.md
chibenwa: I propose `server/distributions` instead. Would it be better?

src/adr/0041-james-servers-offering.md
chibenwa: Well, I do think "1 use case = 1 server" is quite nice.

Once we have separate products we can start working on configuration simplifications.

For the self hosting the mailetcontainer.xml file could be replaced with (middle term) : 

```
# (optional)
# smtp.gateway.host=smtpout.my.cloud.supplier
# smtp.gateway.password=xyz

spf.verify=true

dkim.verify=true
dkim.sign=true
dkim.key=/path/to/something

# We can think of other "ready to use features" that would just need simple "properties" configuration to enable
# Let's say: AntiSpam, Virus Scan, etc..
```

(Which focuses on trust issues self hosting users gets). A self hosting user should not even need to know what a mailet / matcher is in order to achieve is basic configuration. Only minimal opt-in knowledge of the email domain is needed.

While for the advanced one we want to keep the full power of mailet-container.

Another example is regarding keystore (SSL management) : 

 - An advanced users might want to specify a SSL certificate per protocol
 - A basic user will for sure want to have a single SSL certificate. From it we can derive all the protocols configuration

Splitting products helps you better address these needs. Keeping a single product don't allow it.

Furthermore, it would enable writing documentation addressing different use cases: self hosting and "extensibility" which, by definition, we don't know what the advanced operator would like to do.

src/adr/0041-james-servers-offering.md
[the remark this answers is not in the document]
rouazana: @mbaechler why is it better than the "Cassandra - JPA" one for a mid-size deployment?

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+
+ - **Guice memory** (`server/container/guice/memory-guice`) ships a server based on memory. Easy to boot, 
+without external dependencies, we mostly rely on it for testing purposes.
+
+ - **JPA Guice** (`server/container/guice/jpa-guice`) ships an extensible server based on JPA storage. With
+a default configuration using Derby embedded database, this server have no external dependencies by default.
```
rouazana, line 50:
```suggestion
a default configuration using Derby embedded database, this server has no external dependencies by default.
```

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+However the still large amount of configuration can make new comers confused. It currently satisfies operator 
+looking for self hosting solutions as well as those looking for an extensible server.
+
+ - **JPA SMTP** (`server/container/guice/jpa-smtp`) demonstrate a Mail Transfer Agent, dedicated to mail
+processing, that do not ship the James Mailbox, as well as Mail Delivery protocols like IMAP, JMAP or POP3.
```
rouazana, line 55:
```suggestion
processing, that does not ship the James Mailbox, as well as Mail Delivery protocols like IMAP, JMAP or POP3.
```

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+
+We can notice that this offer is specified in technical terms, and do not directly map to identified user needs.
```
rouazana, line 57:
```suggestion
We can notice that this offer is specified in technical terms, and does not directly map to identified user needs.
```

src/adr/0041-james-servers-offering.md
rouazana: > We have to provide a decent substitution first (JPA is not a good substitution for now)

@mbaechler Probably fit for an other ADR, but could you explain quickly why?

src/adr/0041-james-servers-offering.md
rouazana: agree with Matthieu, I would also prefer such a solution

src/adr/0041-james-servers-offering.md
rouazana: `server/flavor`? :trollface:
more seriously I'm fine with distributions

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,109 @@
...
+
+We should adopt this folder structure:
+
+```
+server/apps
+\_____ basic-server
+\_____ advanced-server
+\_____ distributed-server
+\_____ testing-server
+\_____ mail-processing-server
```
rouazana, line 100: is there a need to repeat `server`?

src/adr/0041-james-servers-offering.md
dleangen: Is it possible to have both? Meaning: by default a configuration is not shown at all to the Basic Operator, and a default is used without being exposed. However, if they want to config, then they can add it.

What I would like to avoid is showing something that will just cause more pain and a larger barrier to entry.

I think @chibenwa is right about focusing on use cases, and @mbaechler and @rouazana are talking about the implementation. Because we're talking about different things, it should be possible to do both.

The driver should be the use case, though. :)

src/adr/0041-james-servers-offering.md
dleangen: I think your comment is only relevant when taken together with the Basic Server, so your comment is about both simultaneously.

chibenwa:
> the conversation brings some example of what would be a configuration for the basic server. It's very interesting because it starts to tell the differences between the basic and the advanced server.
 Else how do you plan the migration from basic to advanced?

I took a lot of time to understand this, too, but now I clearly see the value of this proposal.

> Is the basic server some kind of builder/installer of the advanced server?

Not really. I would state "a subset of the advanced server" with reworked configuration & documentation.

Once we have configuration POJO we can defines different Configuration POJO generator and address the self-hosting James-naive use case.

src/adr/0041-james-servers-offering.md
chibenwa: @rouazana did you writte this comment ( https://github.com/apache/james-project/pull/222#discussion_r443544577 ) before or after https://github.com/apache/james-project/pull/222#pullrequestreview-434899849

Which one better reflects your position?

src/adr/0041-james-servers-offering.md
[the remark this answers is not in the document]
chibenwa: Nevermind I think I got it.

src/adr/0041-james-servers-offering.md
rouazana: as far as the basic server can be configured to be an advanced server, I'm fine :)

src/adr/0041-james-servers-offering.md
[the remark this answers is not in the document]
mbaechler: > If we find people willing to find interests in it and devote efforts maintaining it, then deprecating something have no meaning.

I will probably put it in production so I will ensure it keep working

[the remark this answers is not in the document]
mbaechler: > @mbaechler why is it better than the "Cassandra - JPA" one for a mid-size deployment?

I guess you ask me why I would rather use Cassandra than JPA. Cassandra uses ES for search which is a huge bonus IMO. Also, Cassandra provides a robust blob-store that I think could be used for the MailQueue, even when using ActiveMQ.

src/adr/0041-james-servers-offering.md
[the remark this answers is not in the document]
mbaechler: I'm not arguing, all you said is correct. "messaging component" is very overloaded in computer science and in James so I ask. You could add some precision to that.

src/adr/0041-james-servers-offering.md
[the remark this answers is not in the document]
mbaechler: I think you can extend almost any version of James so I found that weird to list that here.

src/adr/0041-james-servers-offering.md
mbaechler: I don't want my data in derby and my indices into Lucene files for a small org (around 100 users), even if Cassandra and ES are not perfect they are way more reliable than files.
And features are lagging (no blob-store, limited support for search, etc)

src/adr/0041-james-servers-offering.md
mbaechler: I think we all agree but fail at expressing ourselves (:

src/adr/0041-james-servers-offering.md
mbaechler: Distribution is how you put things for consumption (like a tar or a zip).
Here we put things together, so `apps` without `server` prefix (because James is a server anyway) would probably better fit.

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,159 @@
...
+
+The Apache James project is willing to provide ready-to-use Mail Servers for our users.
+
+There exists several use cases for such servers:
+
+ - **Self hosting**: a hobbyist, or a small organization is willing to host a mail service. They 
+have limited infrastructure, time and knowledge and are mostly looking for an easy to use 
+solution.
+
+ - **Customization use cases**: operators is willing to provide advanced, integrated features
```
Arsnael, line 19:
```suggestion
 - **Customization use cases**: operators are willing to provide advanced, integrated features
```

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,159 @@
...
+We should adopt this folder structure:
+
+```
+server/distributions
+\_____ basic
+\_____ advanced
+\_____ distributed
+\_____ testing
+\_____ mail-processing
+\_____ cassandra-guice (TODO rebrand me please)
```
Arsnael, line 108: Just a thought, couldn't that be grouped with basic and advanced servers, but, to retake a wording from @rouazana , being called a different flavor?

Like :
* basic server has 2 flavors -> JPA flavor and Cassandra flavor
* advanced server has 2 flavors -> JPA flavor and Cassandra flavor
etc

dleangen: For the Advanced Server, I think that would work just fine. My understanding is: we could say that there are "infinite" configurations possible (well, not really, but just to simplify), but that there are "common patterns" that we call "flavours".

For the Basic Server, I think it goes against the principle of what we are trying to achieve by the use case.

Arsnael: I agree with the fact that Basic server should stay very simple and not mingle different flavors thanks !

So yes the Cassandra guice version could be a flavor for the advanced server, alongside the JPA flavor

chibenwa:
I would like to make a push and merge this.

Any comments you would like me to address before ?

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,159 @@
...
+
+Date: 2020-06-22
+
+## Status
+
+Accepted (lazy consensus)
+
+## Context
+
+The Apache James project is willing to provide ready-to-use Mail Servers for our users.
```
dleangen, line 11: --> For our Operators (to remain consistent)

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,159 @@
...
+
+There exists several use cases for such servers:
+
+ - **Self hosting**: a hobbyist, or a small organization is willing to host a mail service. They 
+have limited infrastructure, time and knowledge and are mostly looking for an easy to use 
+solution.
```
dleangen, line 17: Again for consistency, maybe consider these changes?

is willing to host --> desires to operate ("host" is not really defined anywhere)
easy to use --> easy-to-operate

src/adr/0041-james-servers-offering.md
```diff
@@ -0,0 +1,159 @@
...
+ - Merging Guice servers with their LDAP variations (see [ADR 36 about guice module choosing](0036-against-use-of-conditional-statements-in-guice-modules.md))
+
+We should then rename our servers in order to match the use cases:
+
+ - **Basic server** addresses the self-hosting use case. It will be based on current *jpa-guice* server but with the clear intent to 
+lower adoption barriers. It exposes limited configurations in order not to confuse the user. Configuration and documentation is centered 
+around self hosting use cases. Configuration and documentation addresses the self-hosting usage without exposing underlying James concepts.
+
+The **Basic Server** is a subset of the advanced server, and differs from it from documentation and configuration. Users of the basic server feeling
+constraint by its minimalistic configuration capabilities and features would then be able to easily upgrade to the advanced server.
```
dleangen, line 81: When working on the Basic Server, I realized that it may be useful to allow a concept of "modes". I.e., either it is a fixed product (you get SMTP, IMAP, etc. with no choice), or you can set it to "Relay Only Mode", "Full Mode" etc.

Not sure if this is useful information here, or if it changes anything relating to this document, but thought I'd mention it just in case.

src/adr/0041-james-servers-offering.md
dleangen: I think it would be great to resolve this idea of mail processing. I think it is an essential part of the value that James offers, so it would be good to clarify the intent.

dleangen:
Hi, I just added a few additional comments. Generally I think this is a great initiative. Thank you very much!!

src/adr/0041-james-servers-offering.md
chibenwa: The "Basic server" is intended for self hosting so should target "SMTP and IMAP modes" while not addressing the others for simplicity?

dleangen: Yes, agree that simplicity is the primary objective here.

chibenwa:
While I still believe in the spirit of this proposal, let's be honnest, I will not make the effort to push is further.

PR #223, opened 2020-06-22
JAMES-3211 Extends mailbox-api EventBus interface ADR

LanKhuat:


src/adr/0040-generalize-event-bus.md
```diff
@@ -0,0 +1,33 @@
...
+We can create new events related to user management (UserCreated, UserRemoved and so on).
+
+To achieve that, we will extract the EventBus out of mailbox-api in order to make it a utility component (eventbus-api), then we will make both mailbox-api and data-api depend on that new module. 
+
+## Consequences
+
+Mailbox-api would leverage the EventBus to keep exposing the MailboxListener api without changes on top of the generified EventBus. 
+We need to define a common Event interface in eventbus-api, then each event-bus usage will define its own sealed event hierarchy implementing Event. 
+
+We could rely on the EventBus reliability for building any feature in James.
```
chibenwa, line 33: As a consequence dead-letter storage needs to be reworked in order to store events of various event bus separately (which is need for knowing which eventbus the event should be reprocessed on, for knowing which sealed hierarchy an event belongs to.)

As a consequence, we will need a Cassandra data migration to add the eventBus name as part of the Event DeadLetter primary key.

(Sorry I just thought about this consequence, can you add it?)

src/adr/0040-generalize-event-bus.md
```diff
@@ -0,0 +1,38 @@
...
+
+For quota-search to be initialized/removed for a given user while keeping this feature as a plugin, we decided to adopt the Event Driven pattern we already use in mailbox-api. 
+We can create new events related to user management (UserCreated, UserRemoved and so on).
+
+To achieve that, we will extract the EventBus out of mailbox-api in order to make it a utility component (eventbus-api), then we will make both mailbox-api and data-api depend on that new module. 
+
+## Consequences
+
+Mailbox-api would leverage the EventBus to keep exposing the MailboxListener api without changes on top of the generified EventBus. We need to define a common Event interface in eventbus-api, 
+then each event-bus usage will define its own sealed event hierarchy implementing Event.
```
rouazana, line 31:
```suggestion
then each EventBus usage will define its own sealed event hierarchy implementing Event.
```

src/adr/0040-generalize-event-bus.md
```diff
@@ -0,0 +1,38 @@
...
+
+Dead-letter storage needs to be reworked in order to store events of various event bus separately (which is needed for knowing which EventBus the event should be reprocessed on
```
rouazana, line 33: in the sentence you are using "event bus" and "EventBus". Is there a different meaning between both? Can you uniform the usage of this term in this ADR?

chibenwa:
IMO we should merge this as part of the grooming...

Arsnael:
Merged

PR #224, opened 2020-06-25
JAMES-3253 Remove Log warning for remoteDelivery configuration in def…

chibenwa:
…aultMailetContainer.xml for JPA Guice

This file is relied upon when configuration file is missing

### DIFF
--- a/server/container/guice/jpa-guice/src/main/resources/defaultMailetContainer.xml
+++ b/server/container/guice/jpa-guice/src/main/resources/defaultMailetContainer.xml
@@ -49,6 +49,7 @@
         <mailet match="SMTPAuthSuccessful" class="RemoteDelivery">
             <outgoingQueue>outgoing</outgoingQueue>
             <delayTime>5000, 100000, 500000</delayTime>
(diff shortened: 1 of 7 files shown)


server/container/guice/jpa-guice/src/main/resources/defaultMailetContainer.xml
```diff
@@ -49,7 +49,6 @@
         <mailet match="SMTPAuthSuccessful" class="RemoteDelivery">
             <outgoingQueue>outgoing</outgoingQueue>
             <delayTime>5000, 100000, 500000</delayTime>
-            <maxRetries>25</maxRetries>
```
rouazana, line 52: not `3` like in most configuration files?

chibenwa: Not mentioning it achieve the same.

rouazana: and so, any reason to not be coherent between our configuration files?

chibenwa:
Merged

PR #230, opened 2020-07-03
JAMES-2330: Give ability to override AbstractConfigurableAsyncServer.…

ieugen:
This is a redo of PR https://github.com/apache/james-project/pull/103 . 
The source from that has disapeared. 
Issue is https://issues.apache.org/jira/browse/JAMES-2330


chibenwa (approved):
Re-reading the ticket I think it make sense.

Let's allow people to write complex logic for how they want to manage SSL context. I do not have objections.

Arsnael (approved):

PR #236, opened 2020-07-16
[RELEASE] git-commit-id-plugin should not fail when no git directory

chibenwa:
People compiling downloaded sources don't have a git environment setted up,
we should not abort their compilation, while keeping useful git information
shipped into the JARs when there is a git environment.


chibenwa:
Merged

PR #240, opened 2020-07-27
Issue: Unable to resolve host address ‘downloads.apache.org’

bhaskar24:
Replace:
from
https://downloads.apache.org/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
to
http://apache.mirrors.pair.com/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz

### DIFF
--- a/dockerfiles/compilation/java-11/Dockerfile
+++ b/dockerfiles/compilation/java-11/Dockerfile
@@ -12,7 +12,8 @@ RUN apt-get install -y git wget unzip
 
 # Install Maven
 WORKDIR /root
-RUN wget https://downloads.apache.org/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
+RUN wget http://apache.mirrors.pair.com/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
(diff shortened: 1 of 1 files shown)


rouazana:
This is weird, the first link is still working for me, and I remember that Apache recommends now to use downloads.apache.org which will automatically find the best mirror. Are you sure we have to change this? By the way what is this mirror? Finally I'm a priori against using HTTP instead of HTTPS.

rouazana:
It's not even listed in the Apache Maven mirror list: https://maven.apache.org/download.cgi

bhaskar24:
I think below is the reason why it is working and why not:

Please have a look at availability report: http://www.apache.org/mirrors/

Arsnael:
Agreed with @rouazana. Thereis plenty of mirrors using https, which is safer than plain http. `downloads.apache.org` is up in your link by the way. Might have been just a temporary issue with the node (can happen), I don't see the point of changing this.


But thanks for your concern either way :)

ieugen:
Https works. 
I don't see a reason why we should downgrade to http.
Thanks for the concern.

PR #243, opened 2020-08-11
JAMES-3187 Community section of the new documentation

Arsnael:
For now just the index page of the community. I will contribute the other sub-sections bits by bits. But I think an early feedback could possibly help me fixing the structure of this section if necessary!


docs/modules/community/nav.adoc
```diff
@@ -1 +1,6 @@
 * xref:index.adoc[]
+** xref:mailing-lists.adoc[]
+** xref:contributing.adoc[]
+** xref:guidelines.adoc[]
+** xref:download.adoc[]
+** xref:support.adoc[]
```
chibenwa, line 6: In the nav we should have the following links appearant for ASF compliance:

http://www.apache.org/

http://www.apache.org/foundation/getinvolved.html

http://www.apache.org/foundation/faq.html

http://www.apache.org/licenses/

http://www.apache.org/foundation/sponsorship.html

http://www.apache.org/foundation/thanks.html

http://www.apache.org/security/

docs/modules/community/pages/index.adoc
```diff
@@ -1,5 +1,42 @@
...
+is governed by a number of xref:guidelines.adoc[guidelines] directly depending on the *ASF*.
+
+If you wish to get more involved with the *ASF* community in general, those links might interest you as well:
+
+ * https://www.apache.org/foundation/getinvolved.html[Get involved with the ASF]
+ * https://www.apache.org/foundation/faq.html[Apache FAQ]
+ * https://www.apache.org/licenses/[Apache Licenses]
+ * https://www.apache.org/security/[The Apache security team]
+ * https://www.apache.org/foundation/sponsorship.html[Sponsorship program]
+ * https://www.apache.org/foundation/thanks.html[Sponsors list]
```
chibenwa, line 42: (this should be in nav as well ;-)

docs/modules/community/nav.adoc
Arsnael: For the compliance really necessary to put them in nav? Ok now I understand better why it is in the old doc, I found that strange!

dleangen:
This is an important section. Thanks for adding this! I think some of the contents require a lot of discussion on the mailing list, but I will add a few comments here in this PR.

Arsnael:
> This is an important section. Thanks for adding this! I think some of the contents require a lot of discussion on the mailing list, but I will add a few comments here in this PR.

I mainly do a porting of what exists but yes you are right it would deserve more discussions on some points, totally opened to it !

docs/modules/community/pages/mailing-lists.adoc
```diff
@@ -0,0 +1,165 @@
+= Mailing lists
+
+== Guidelines
+
+A mailing list is an electronic discussion forum that anyone can subscribe to. When someone sends an email message to
+the mailing list, a copy of that message is broadcasted to everyone who is subscribed to that mailing list.
+Mailing lists provide a simple and effective communication mechanism for discussion and decision making.
+
+The *Apache Software Foundation* has well established reasons for using email and not other types of forum.
+We are *not* interested in hearing proposals to use NNTP (news groups) or web forums or any other medium.
```
chibenwa, line 10: I think we can remove that aggressive "NNTP" statement. (L 10-11) it never happened in years.

docs/modules/community/pages/mailing-lists.adoc
```diff
@@ -0,0 +1,165 @@
+= Mailing lists
+
+== Guidelines
```
chibenwa, line 3: How about listing mailing list first, then detailing the guidelines?

docs/modules/community/pages/mailing-lists.adoc
```diff
@@ -0,0 +1,165 @@
...
+
+=== Server user list
+
+*Low traffic*
+mailto:server-user-subscribe@james.apache.org[Subscribe]
+mailto:server-user-unsubscribe@james.apache.org[Unsubscribe]
+http://www.mail-archive.com/server-user@james.apache.org[Archive]
+
+This is the list where users of the Apache James (Server) meet and discuss issues. Developers are also expected to be
+subscribed to this list to offer support to users of Apache James (Server).
```
chibenwa, line 85: I think we could benefit from use cases:

 - I want to receive Apache James announcements
 - I run into xx problem and want some help. In which case I should search the mail archive first.

docs/modules/community/pages/mailing-lists.adoc
```diff
@@ -0,0 +1,165 @@
...
+=== Site developer list
+
+*Low traffic*
+mailto:site-dev-subscribe@james.apache.org[Subscribe]
+mailto:site-dev-unsubscribe@james.apache.org[Unsubscribe]
+http://www.mail-archive.com/site-dev@james.apache.org[Archive]
+
+This is the list where participating developers of the Apache James project discuss changes/additions to the James
+websites etc. Subscribers to this list get notifications of every commit of the website resources. Do not send mail to
+this list with James software problems -- that's what server-user@james is for.
```
chibenwa, line 116: Should we keep that mailing list?

docs/modules/community/pages/mailing-lists.adoc
```diff
@@ -0,0 +1,165 @@
...
+*Low traffic*
+mailto:general-subscribe@james.apache.org[Subscribe]
+mailto:general-unsubscribe@james.apache.org[Unsubscribe]
+http://www.mail-archive.com/general@james.apache.org[Archive]
+
+This is the list for general discussions related to the running of the project, it is the public list of the
+James project management committee (PMC) and is a public list opened to all. Do not send mail to this list with James
+software problems -- that's what server-user@james is for.
+
+=== Mailet API list
```
chibenwa, line 129: Should we keep that mailing list? It is not active.... Maybe it is better to discuss mailet pi concerns in server-dev....

docs/modules/community/pages/support.adoc
```diff
@@ -0,0 +1,16 @@
+= Professional support
+
+Here is a list of companies which can provide you some support on James.
+
+== Linagora
+
+image::linagora.png[link=https://linagora.com]
```
chibenwa, line 7: That's a side note for Linagora as an organization but we may want to have a detailed "James offering" page...

Arsnael: I don't believe we have such a dedicated page yet though right?

docs/modules/community/pages/mailing-lists.adoc
Arsnael: Hmmm... We can definitely discuss about this. I know on the legacy documentation website the list of mailing lists comes first. But I was thinking... Does people are really gonna notice and read the guidelines if we put the list first?

I don't say putting the guidelines first will make everybody read them, but at least they will notice it, and probably more will read. They can always easily jump to the list after. 

Is this necessary or did I overthink a bit here? :)

docs/modules/community/pages/mailing-lists.adoc
dleangen: If in doubt... my suggestion would be to remove it. IMO there are too many mailing lists. 😀

Also, since we're on the topic... I wonder if it would be possible to separate the notifications from posts by members. Like maybe: notifications@james.apache.org or whatever. wdyt?

docs/modules/community/pages/mailing-lists.adoc
dleangen: Remove! 😀

docs/modules/community/pages/mailing-lists.adoc
chibenwa: That is what general is for, apparently.  You are very right.

docs/modules/community/pages/mailing-lists.adoc
chibenwa: :-)

though removal might need ML discussions ^^

docs/modules/community/nav.adoc
```diff
@@ -1 +1,14 @@
...
+** Apache Software Foundation
+*** https://www.apache.org/[ASF]
+*** https://www.apache.org/foundation/getinvolved.html[Get involved]
+*** https://www.apache.org/foundation/faq.html[FAQ]
+*** https://www.apache.org/licenses/[Licenses]
+*** https://www.apache.org/security/[Security]
+*** https://www.apache.org/foundation/sponsorship.html[Sponsorship]
+*** https://www.apache.org/foundation/thanks.html[Thanks]
```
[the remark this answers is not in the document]
chibenwa: The apache links are an actual foundation requirements (if that is the question)

docs/modules/community/pages/mailing-lists.adoc
chibenwa: I think it delays the valuable information people look for. But I'm not really opinionated on this.

docs/modules/community/pages/mailing-lists.adoc
ieugen: I would also go for notifications on separate ML . 
We should have a separate channel for user vs developer discussions and make it clear in the docs.
Right now gitter for example has all discussions into once place.

Arsnael: I would agree on separating notifications to an other ML. It's hard sometimes to distinguish user mails from notifications in server-dev mailing list for example

@ieugen : not sure to get your point with Gitter though, to me it seems gitter discussions are all user oriented?

ieugen: I'll add a JIRA issue and make sure we get notifications@james.apache.org and then make the changes for the projects so we have both commits and JIRA comments ?! sent to notifications.

The main benefit is that commits tend to polute the ML and will create a lot of traffic for archive as well.

Regarding Gitter, I feel the need to talk to fellow developers on something else than ML - so we can have a quick discussion and then send the conclusions to ML.

ieugen: See https://issues.apache.org/jira/browse/INFRA-20735 , please comment there.

ieugen (approved):

docs/modules/community/pages/guidelines.adoc
```diff
@@ -137,3 +137,21 @@ A one-line summary of a change that has been committed to the repository since t
 
 All product changes to the currently active repository are subject to lazy consensus. All product changes to a
 prior-branch (old version) repository require consensus before the change is committed.
+
+////
+
+Notes and thoughts on the guidelines (opened to discussion)
```
chibenwa, line 143: Don't forget to remove the notes!

chibenwa (approved):

chibenwa (approved):

ieugen:
let's merge it !

Arsnael:
> let's merge it !

Will do later, thanks for the additional commits !

Arsnael:
Merged

dleangen:
@Arsnael, thank you so much for doing this!

As I [mentioned in the issue](https://issues.apache.org/jira/browse/JAMES-3187), if I can get more feedback from the community, I will continue working on this in a new PR.

PR #248, opened 2020-09-25
[ADR] Applicative read repairs (POC mailbox & mailbox-counters)

chibenwa:


rouazana (commented):
For me you are missing 2 points:
- task solving inconsistencies are potentially too long to be run on a big cluster. It's not just "time" consuming", it can be impossible to run them if you have enough data.
- why don't we use native Cassandra capabilities for that? Such as secondary indexes or materialized views?

src/adr/0042-applicative-read-repairs.md
```diff
@@ -0,0 +1,62 @@
...
+Completes [20. Cassandra Mailbox object consistency](0020-cassandra-mailbox-object-consistency.md),
+[23. Cassandra Mailbox Counters inconsistencies](0023-cassandra-mailbox-counters-inconsistencies.md)
+
+## Context
+
+Cassandra eventual consistency is all about "replication", but "denormalization" consistency needs
+to be handled at the applicative layer (due to the lack of transactions in a NoSQL database).
+
+In the past we did set up "Solve inconsistency" tasks that can be assimilated to Cassandra repairs. Such
+tasks, after being scheduled, ensures the according entity denormalization is correctly denormalized.
```
Arsnael, line 18:
```suggestion
tasks, after being scheduled, ensure that the according entity denormalization is correctly denormalized.
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -0,0 +1,62 @@
...
+
+However, the inconsistencies persist between runs. We experienced inconsistencies in some production platform
+for both the mailbox entity, and the mailbox counter entity (whose table structure is exposed in
+[these](0020-cassandra-mailbox-object-consistency.md), [ADRs](0023-cassandra-mailbox-counters-inconsistencies.md)).
+Monitoring is required to detect when to run them and is time consuming for the platform administrator.
+
+Another classic eventual consistency mechanism, that enables auto-healing and completes
```
Arsnael, line 25: Not sure to get what you mean here, the sentence looks incomplete or incorrect?

src/adr/0042-applicative-read-repairs.md
```diff
@@ -0,0 +1,62 @@
...
+
+In order to achieve denormalization auto-healing, we thus needs to implement "applicative read repairs".
```
Arsnael, line 27:
```suggestion
In order to achieve denormalization auto-healing, we thus need to implement "applicative read repairs".
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -0,0 +1,62 @@
...
+This enables read path simplification (and performance enhancements) for the mailbox object.
+
+IMAP LIST should not read mailbox counters. This information is uneeded and we should avoid paying the
+price of read repairs for this operation.
+
+Provide a comprehensive documentation page regarding "Distributed James consistency model".
+
+## Consequences
+
+The expect **auto-healing** inconsistencies on existing deployments (at a limited configuration cost).
```
Arsnael, line 42:
```suggestion
The expected **auto-healing** inconsistencies on existing deployments (at a limited configuration cost).
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -0,0 +1,62 @@
...
+This should ease operation of the Distributed James server.
+
+A configuration for James Distributed server will be added to control read repairs, per entity.
+
+## Alternatives
+
+We already propose several tasks to solve denormalization inconsistencies. "Applicative read repairs" should be
+seen as a complement to it.
+
+Another classical mechanism in eventual consistent system is called hinted-handoff. It consist at retries
```
Arsnael, line 52:
```suggestion
Another classical mechanism in eventual consistent system is called hinted-handoff. It consists at retries
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -0,0 +1,62 @@
...
+(during a given period) when "replicating" data to other replica. We also already have a similar mechanism
+in James as we retry several time failures when writing data to denormalization table. Hard shut-down however
```
Arsnael, line 54:
```suggestion
in James as we retry several times failures when writing data to denormalization table. Hard shut-down however
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -46,6 +47,15 @@ A configuration for James Distributed server will be added to control read repai
 
 ## Alternatives
 
+Cassandra provides some alternative by itself:
+
+ - Secondary indexes avoids the deormalization in the first place. However they are not efficient in
```
Arsnael, line 52:
```suggestion
 - Secondary indexes avoids the denormalization in the first place. However they are not efficient in
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -46,6 +47,15 @@ A configuration for James Distributed server will be added to control read repai
...
+ a distributed environment as each nodes needs to be queried, which limits ability to scale.
```
Arsnael, line 53:
```suggestion
 a distributed environment as each node needs to be queried, which limits ability to scale.
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -46,6 +47,15 @@ A configuration for James Distributed server will be added to control read repai
...
+ - Materialized view enables Cassandra to maintain a projection on the behalf of the application,
+ come with an expensive write cost, requiring synchronisation, don't fit for complex denormalization
```
Arsnael, line 55:
```suggestion
 coming with an expensive write cost, requiring synchronisation, not fit for complex denormalization
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -60,3 +70,24 @@ defeats this strategy that is otherwise efficient to limit inconsistencies acros
...
+Especially:
+
+```
+Materialized View Limitations:
+
+    All updates to the view happen asynchronously unless corresponding view replica is the same node.
+    We must do this to ensure availability is not compromised.  It's easy to imagine a worst case
+    scenario of 10 Materialized Views for which each update to the base table requires writing to 10
+    separate nodes. Under normal operation views will see the data quickly and there are new metrics to
+    track it (ViewWriteMetricss).
```
Arsnael, line 85:
```suggestion
    track it (ViewWriteMetrics).
```

chibenwa:
> so it has the same constraints than hand-made projection, so why do we use them for simple cases?

Not really.

You write to the table A and the write returns, then (maybe?) table B will be updated, outside of your applicative transaction.

This would lead to much headaches.

With manual replication, you write to A then to B, then you return the applicative transaction. (Wich means we can directly read, for sure, data from table B)

Arsnael (approved):

rouazana:
Interesting, we should also check if materialized view is still experimental: https://www.mail-archive.com/user@cassandra.apache.org/msg54073.html

Also what do you think about Batches? Could it help maintaining coherence between some requests?

A word about this in this ADR would be welcome.

chibenwa:
Batches won't play well with multi partition Lightweight transaction. 

Likely a good idea when lightweight transaction dont come into play

TODO add links regarding materialised view deprecation :+1: 
 - https://docs.datastax.com/en/cql-oss/3.3/cql/cql_using/faqMV.html
 - https://issues.apache.org/jira/browse/CASSANDRA-13959

src/adr/0042-applicative-read-repairs.md
```diff
@@ -54,7 +55,13 @@ Cassandra provides some alternative by itself:
  - Materialized view enables Cassandra to maintain a projection on the behalf of the application,
  coming with an expensive write cost, requiring synchronisation, not fit for complex denormalization
  (like the message one: the primary key of the originating table needs to appear in the materialized
- view primary key).
+ view primary key). Most of all, the updates are performed asynchronously. This mechanism is considered experimental.
+ - Cassandra BATCH suffers from the following downsides:
+   - A batch containing conditional updates can only operate within a single partition
+   - It is unadvised to update many partition in a single batch, and keep the cardinality low for performance reasons
```
Arsnael, line 61:
```suggestion
   - It is unadvised to update many partitions in a single batch, and keep the cardinality low for performance reasons
```

src/adr/0042-applicative-read-repairs.md
```diff
@@ -54,7 +55,13 @@ Cassandra provides some alternative by itself:
...
+
+BATCH could be a good option to keep to tables synchronized, but do not apply to mailboxes (conditional update) nor
```
Arsnael, line 63:
```suggestion
BATCH could be a good option to keep tables synchronized, but does not apply to mailboxes (conditional update) nor
```

Arsnael (approved):

nvduc91 (approved):

chibenwa:
Merged

PR #249, opened 2020-09-30
JAMES-3395 make examples not inherit from parent, move package of examples out from org.apache.james

aasaru:


Arsnael:
Thanks for the initiative :) I don't think I have issues with that, on the contrary

Just a question, is `org.apache.examples.james` as a groupId a good one? (Opened question, I don't pretend having the answer)

aasaru:
one alternative to that would be org.example.james

examples/pom.xml
```diff
@@ -36,4 +33,14 @@
         <module>custom-mailets</module>
     </modules>
 
-</project>
\ No newline at end of file
+    <properties>
+        <james.groupId>org.apache.james</james.groupId>
+        <james.baseVersion>3.6.0-SNAPSHOT</james.baseVersion>
```
chibenwa, line 38: I understand the concerns however it likely makes the release process hard.

How do we ensure the version use by the example matches the released version? How does it integrate with the maven-release plugin?

I I was to tag a 3.6.0 tomorrow, I would likely forget that. Even if I don't forget, how do I ensure the released exemple JAR matches the 3.6.0 version? Would I have to do a pre-release commit switching it to the not yet released 3.6.0 - in which case the whole process might fail at publishing snapshots?

Maybe a wiser value here would be :

```
<james.baseVersion>${project.version}</james.baseVersion>
```

That way as the version of the example and James are implicitly aliagned, the maven-release prompt will enable me to do the switch painlessly...

Arsnael:
I don't think you can omit `apache` in there, as those modules belong to the apache foundation right?

examples/pom.xml
```diff
@@ -20,14 +20,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.apache.james</groupId>
-        <artifactId>james-project</artifactId>
-        <version>3.6.0-SNAPSHOT</version>
-    </parent>
+    <groupId>org.apache.examples.james</groupId>
```
chibenwa, line 23: I'm unsure that the James project is allowed to publish JARs to maven central  with a `org.apache.examples.` prefix. 

Can you check the Apache foundation policies regarding this?

chibenwa (commented):
Globally agree, though release details needs to be sorted out.

examples/pom.xml
chibenwa: https://infra.apache.org/publishing-maven-artifacts.html don't seem helpful regarding this concern...

chibenwa: Neither do https://infra.apache.org/release-distribution.html ...

chibenwa: Maybe we need to ask the INFRA team?

chibenwa: Also using org.apache.examples prefix don't seem to be a standard practice amongst other project.

See https://mvnrepository.com/search?q=org.apache.examples

Maybe we should keep the `org.apache.james.examples` FQDN while switching it to a top level POM not depending on James prent POM. ?

examples/pom.xml
```diff
@@ -20,14 +20,11 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <parent>
-        <groupId>org.apache.james</groupId>
-        <artifactId>james-project</artifactId>
-        <version>3.6.0-SNAPSHOT</version>
-    </parent>
```
chibenwa, line 27: We at least need to depend on the root Apache POM as a parent .

See https://infra.apache.org/publishing-maven-artifacts.html

examples/pom.xml
aasaru: Thanks for good feedback. I didn't release the examples are get published so of course it makes sense to use this property as a default.

examples/pom.xml
aasaru: After thinking about this I realised that if we have all example code under org.apache.james.examples then we cannot use package level access anywhere (as core james won't have the 'examples' part in the package name). Sorry for creating confusion - reverted back this part - let's stay to org.apache.james.examples

examples/pom.xml
aasaru: Good point. Made root apache pom as parent.

README.adoc
```diff
@@ -401,7 +401,8 @@ Beware : you will have concurrency issues if multiple containers are running on
 
 == Develop on James
 
-James requires at least JDK 11 and maven 3.6.0 to build.
+James requires at least JDK 11 and Maven 3.6.0 to build.
+Some parts of James are written in Scala so one might need to enable Scala plugin in IDE.
```
chibenwa, line 405: <3

chibenwa (approved):

Arsnael (approved):

Arsnael:
Merged. Can you please help close the PR @aasaru ? Thank you very much for your contribution :)

PR #254, opened 2020-10-17
Redirect to JMAP session resource from /.well-known/jmap

antonok-edm:
Section 2.2 of RFC 8620 is as follows:

> ## 2.2. Service Autodiscovery
>
> There are two standardised autodiscovery methods in use for Internet protocols:
>
> - **DNS SRV** (see [@!RFC2782], [@!RFC6186], and [@!RFC6764])
> - **.well-known/servicename** (see [@!RFC8615])
>
> A JMAP-supporting host for the domain `example.com` SHOULD publish a SRV record `_jmap._tcp.example.com` that gives a *hostname* and *port* (usually port `443`). The JMAP Session resource is then `https://${hostname}[:${port}]/.well-known/jmap` (following any redirects).
>
> If the client has a username in the form of an email address, it MAY use the domain portion of this to attempt autodiscovery of the JMAP server.

James' implementation exposes the JMAP Session Resource at `/jmap/session`. Clients which expect it at `/.well-known/jmap` are unable to find it.

The spec language ("SHOULD") is a strong recommendation and not a requirement, so it's acceptable if there's a good reason to disregard it - but I'm not personally aware of one, and in any case the choice ought to be documented somewhere. I tried to set up a James server for testing a JMAP client implementation and couldn't figure out why it wasn't working until I dug into the source code.

### DIFF
--- a/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/http/SessionRoutes.scala
+++ b/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/http/SessionRoutes.scala
@@ -29,7 +29,7 @@ import org.apache.james.jmap.HttpConstants.JSON_CONTENT_TYPE_UTF8
 import org.apache.james.jmap.JMAPRoutes.CORS_CONTROL
 import org.apache.james.jmap.JMAPUrls.AUTHENTICATION
 import org.apache.james.jmap.exceptions.UnauthorizedException
-import org.apache.james.jmap.http.SessionRoutes.{JMAP_SESSION, LOGGER}
+import org.apache.james.jmap.http.SessionRoutes.{JMAP_SESSION, WELL_KNOWN_JMAP, LOGGER}
 import org.apache.james.jmap.http.rfc8621.InjectionKeys
 import org.apache.james.jmap.json.ResponseSerializer
 import org.apache.james.jmap.model.Session
@@ -43,6 +43,7 @@ import reactor.netty.http.server.HttpServerResponse
 
 object SessionRoutes {
   private val JMAP_SESSION: String = "/jmap/session"
+  private val WELL_KNOWN_JMAP: String = "/.well-known/jmap"
   private val LOGGER = LoggerFactory.getLogger(classOf[SessionRoutes])
 }
 
@@ -58,6 +59,8 @@ class SessionRoutes @Inject() (@Named(InjectionKeys.RFC_8621) val authenticator:
       .subscribeOn(Schedulers.elastic())
       .asJava()
 
+  private val redirectToSession: JMAPRoute.Action = JMAPRoutes.redirectTo(JMAP_SESSION)
+
   override def routes: Stream[JMAPRoute] =
     Stream.of(
       JMAPRoute.builder()
@@ -65,7 +68,15 @@ class SessionRoutes @Inject() (@Named(InjectionKeys.RFC_8621) val authenticator:
         .action(generateSession)
         .corsHeaders,
       JMAPRoute.builder()
-        .endpoint(new Endpoint(HttpMethod.OPTIONS, AUTHENTICATION))
+        .endpoint(new Endpoint(HttpMethod.OPTIONS, JMAP_SESSION))
+        .action(CORS_CONTROL)
+        .noCorsHeaders,
+      JMAPRoute.builder()
+        .endpoint(new Endpoint(HttpMethod.GET, WELL_KNOWN_JMAP))
+        .action(redirectToSession)
+        .corsHeaders,
+      JMAPRoute.builder()
+        .endpoint(new Endpoint(HttpMethod.OPTIONS, WELL_KNOWN_JMAP))
         .action(CORS_CONTROL)
         .noCorsHeaders)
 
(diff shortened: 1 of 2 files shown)


ieugen (approved):

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/http/SessionRoutes.scala
```diff
@@ -42,7 +42,7 @@ import reactor.core.scheduler.Schedulers
 import reactor.netty.http.server.HttpServerResponse
 
 object SessionRoutes {
-  private val JMAP_SESSION: String = "/jmap/session"
+  private val JMAP_SESSION: String = "/.well-known/jmap"
```
chibenwa, line 45: Can we use a redirect as suggested hy the specs instead?

chibenwa:
(But I agree with implementing that should with a redirect)

antonok-edm:
@chibenwa I've implemented it as a redirect now, although I'll probably need some assistance with testing - when I try the build command from the README without the `-s` argument, I get a bunch of failures from the `DockerCassandraSingleton` class not being found.

```
> docker run -v $PWD/.m2:/root/.m2 -v $PWD:/origin -t james/project well-known-jmap

...

[ERROR] Errors:
[ERROR]   CassandraClusterTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   TestingSessionTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   CassandraTableManagerTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   CassandraTypeProviderTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   CassandraTypesCreatorTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   ClusterFactoryTest » ExceptionInInitializer
[ERROR]   ResilientClusterProviderTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   SessionWithInitializedTablesFactoryTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   CassandraHealthCheckTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   PaggingTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
[ERROR]   CassandraSchemaVersionDAOTest » NoClassDefFound Could not initialize class org.apache.james.backends.cassandra.DockerCassandraSingleton
```

chibenwa:
Thanks a lot, I will take a look.

I will launch tests for this on monday.

There needs to be a docker environment available for some tests to pass - likely an issue to be addressed.

chibenwa (approved):

chibenwa:
Thinking about it, I think we might need the cors to be handled for .well-known/jmap as well (OPTIONS verb).

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/http/SessionRoutes.scala
```diff
@@ -58,12 +59,18 @@ class SessionRoutes @Inject() (@Named(InjectionKeys.RFC_8621) val authenticator:
...
       JMAPRoute.builder()
         .endpoint(new Endpoint(HttpMethod.GET, JMAP_SESSION))
         .action(generateSession)
         .corsHeaders,
+      JMAPRoute.builder()
+        .endpoint(new Endpoint(HttpMethod.GET, WELL_KNOWN_JMAP))
+        .action(redirectToSession)
+        .corsHeaders,
       JMAPRoute.builder()
         .endpoint(new Endpoint(HttpMethod.OPTIONS, AUTHENTICATION))
```
antonok-edm, line 75: @chibenwa is `AUTHENTICATION` here an error? this should be `JMAP_SESSION` instead?

chibenwa: Of course it is ....

chibenwa: Here is a fix ;-)

https://github.com/linagora/james-project/pull/3918

Arsnael (approved):
Good catch. To be honest, that was in the backlog, but thanks for contributing to it, really appreciated :)

antonok-edm:
Ok, updated with CORS on `/.well-known/jmap` as well, and applied on top of the change from https://github.com/linagora/james-project/pull/3918

chibenwa (approved):

chibenwa:
Looks great!

O added a little test for this as well https://github.com/linagora/james-project/pull/3919/commits/ba25c8e6679eecaa70948a4d0e26b493a747d9a6

remk (approved):

Arsnael:
Hi @antonok-edm , your work has been merged. Can you help me close your PR? (I do not have that right sorry)

Thanks again for your contribution to the project :)

antonok-edm:
For sure, thanks all for the quick reviews :smile:

PR #256, opened 2020-10-23
JAMES-3389, add api to webadmin server for receiving mail over rest i…

andrevka:


server/container/guice/pom.xml
```diff
@@ -73,6 +73,7 @@
         <module>protocols/webadmin-mailrepository</module>
         <module>protocols/webadmin-rabbitmq-mailqueue</module>
         <module>protocols/webadmin-swagger</module>
+        <module>protocols/webadmin-mail-over-web</module>
```
chibenwa, line 76: I propose to not add yet-another module but add it to `protocols/webadmin-mailqueue`

server/container/guice/protocols/webadmin-mail-over-web/src/main/java/org/apache/james/modules/server/WebAdminMailOverWebModule.java
```diff
@@ -0,0 +1,16 @@
+package org.apache.james.modules.server;
```
chibenwa, line 1: The Apache V2 license header is compulsory.

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/request/BodyPartProps.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private String filename;
+
+    private String content;
+
+    private String encoding;
+
+    private String disposition;
+
+    private Map<String, String> headers = new HashMap<>();
```
chibenwa, line 39: We tend to use immutable POJOs, using final fields (no setters) and immutable datastructures like ImmutableMap (guava)

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/request/BodyPartProps.java
```diff
@@ -0,0 +1,89 @@
...
+
+        Optional.ofNullable(type).ifPresent(this::setType);
+        this.filename = filename;
+        Optional.ofNullable(content).ifPresent(this::setContent);
+        this.encoding = encoding;
+        this.disposition = disposition;
+        Optional.ofNullable(headers).ifPresent(this::setHeaders);
+    }
+
+    private String type = "text/plain";
```
chibenwa, line 29: That's a constant.

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
```diff
@@ -0,0 +1,86 @@
...
+import spark.Service;
+
+@Api(tags = "OverWebMailReceiver")
+@Path(ReceiveMailOverWebRoutes.BASE_URL)
+@Produces("application/json")
+public class ReceiveMailOverWebRoutes implements Routes {
+
+    private final JsonExtractor<MailProps> mailPropsJsonExtractor;
+
+    public static final String BASE_URL = "/receiveMail";
```
chibenwa, line 32: We should POST emails to a dedicated mailqueue.

I propose to add the endpoint in MailQueueRoutes : 

`curl -XPOST http://127.0.0.1:8000/mailQueues/spool/mails -d 'payload'`

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
```diff
@@ -0,0 +1,86 @@
...
+    })
+    public void defineReceiveMailFromWebService(Service service) {
+        service.post(BASE_URL, (request, response) -> {
+            try {
+                //Parse the json object to org.apache.james.webadmin.request.MailProps and build the MailImpl object
+                MailImpl mail = mailPropsJsonExtractor.parse(request.body()).asMailImpl();
+                //Send to queue api for mail processing
+                queue.enQueue(mail);
+                response.body("ENQUEUED");
+                response.status(204);
```
chibenwa, line 72: Use 201 + no body content

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
```diff
@@ -0,0 +1,86 @@
...
+            } catch (Exception e) {
+                ErrorResponder.builder()
+                        .cause(e)
+                        .statusCode(500)
+                        .type(ErrorResponder.ErrorType.SERVER_ERROR)
+                        .message("The mail will not be sent: " + e.getMessage())
+                        .haltError();
```
chibenwa, line 79: We have default error handling, this is not required to handle it manually here

server/protocols/webadmin/webadmin-mail-over-web/src/test/java/org/apache/james/request/MailPropsTest.java
```diff
@@ -0,0 +1,131 @@
...
+                null
+        );
+
+        MailImpl mail = mailProps.asMailImpl();
+
+        assertThat(mail.getName(), Matchers.equalTo("test"));
+        assertThat(mail.getSender().asPrettyString(), Matchers.equalTo("<valid@address>"));
+        assertThat(mail.getRecipients(), Matchers.contains(new MailAddress("valid1@address"), new MailAddress("valid2@address"), new MailAddress("valid3@address")));
+        assertThat(mail.getMessage().getHeader("Subject"), Matchers.array(Matchers.equalTo("test-subject")));
+        assertThat(mail.getMessage().getHeader("Content-Type"), Matchers.array(Matchers.equalTo("text/plain; charset=us-ascii")));
```
chibenwa, line 41: When we do several assertions, we wrap them in a SoftAssertions:

```
SoftAssertions.assertThatSoftly(softly -> {
    softly.assertThat(...).isEqualTo("test");
    softly.assertThat(...).isEqualTo("test-bis");
});
```

Also we prefer relying on assertj fluent syntax instead of hamcrest matchers.

server/protocols/webadmin/webadmin-mail-over-web/src/test/java/org/apache/james/request/MailPropsTest.java
```diff
@@ -0,0 +1,131 @@
...
+    public void mailImplemantationHascorrectFieldsWhenBuiltFromMailProps() throws Exception{
+        MailProps mailProps = new MailProps(
+                "test",
+                "valid@address",
+                Arrays.asList("valid1@address"),
+                Arrays.asList("valid2@address"),
+                Arrays.asList("valid3@address"),
+                "test-subject",
+                null,
+                null
```
chibenwa, line 32: I cannot understand this test as there is way too many arguments to this constructor, and most of them have the same type.

We could have a builder?

Or the tested input can be a JSON (the test would then deserialize the JSON first then turn it into a mail?

server/protocols/webadmin/webadmin-mail-over-web/src/test/java/org/apache/james/routes/ReceiveMailOverWebRoutesTest.java
```diff
@@ -0,0 +1,104 @@
...
+                .setConfig(newConfig().encoderConfig(encoderConfig().defaultContentCharset(StandardCharsets.UTF_8)))
+                .build();
+    }
+
+    @Test
+    public void whenFromIsMissingInRequestThenRequestFails() {
+        given()
+                .body(ClassLoaderUtils.getSystemResourceAsString("json/withoutFrom.json"))
+                .when()
+                .post()
```
chibenwa, line 63: We generally indent given(), .when() and .then() at the same level in order to improve readability.

Here:

```
        given()
                .body(ClassLoaderUtils.getSystemResourceAsString("json/withoutFrom.json"))
        .when()
                .post()
        .then()
                .assertThat()
                .statusCode(500)
                .body("message", Matchers.containsString("Mail missing from"));
```

server/protocols/webadmin/webadmin-mail-over-web/src/test/resources/json/mail.json
```diff
@@ -0,0 +1,13 @@
+{
```
chibenwa, line 1: Email is a (really) complex thing, and defining a standard JSON document  to represent an email is harder than when you propose here. To give you what it could look like you can have a look at the JMAP specification Email object (https://jmap.io/spec-mail.html#emails) and emailSubmission object (https://jmap.io/spec-mail.html#email-submission), as it serve a similar purpose.

Note that this simplistic format might work for you but might be too restrictive for other users.

Hence I am reluctant to add yet another non standard JSON representation of an email.

Also, emails size can be too big to be represented by a JSON object.

Finally you do not distinguish the mail enveloppe from the mail headers in your format.

chibenwa (commented):
Thanks for this interesting contribution! That's a huge work, and thank you for putting that together.

I did a couple of remark, but globally the API of this new route had not been discussed. I have a couple of issues with the API proposed here, and I would like to make it evolve.

I believe that, by speaking together, we can get something fitting your purposes, as well as the needs of others.

Cheers,

Benoit

server/protocols/webadmin/webadmin-mail-over-web/src/test/resources/json/mail.json
andrevka: Thanks for the pointers. I'll start with fixing the coding style issues right away, and moving the code to the mailqueue module. After that i will try to implement the request payload as the emailSubmission object.

chibenwa: > After that i will try to implement the request payload as the emailSubmission object.

Don't. I points it to show you this is a complex topic.

I would prefer passing the mail envelope via URL query parameters, and having the request payload being the raw MIME message. (I think having the raw MIME EML submission format is really easy to use). The client system could be as simple as a curl command (good inter-operability, but client system have to generate mime messages itself).

If we want to support another format, JSON based, we could use the Accept header to specify the Mail to be sent, and could be friendlier to modern application (but we won't have a one format fit all use cases without some over-engineering - we should rather start with something simple). The client system would need to be adapted to this format (limited inter-operability).

I propose you write  an API proposition, with examples, in the JIRA  ticket before implementing more things.

(You can of course take care of stylish and location issues).

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
mbaechler: if you think about this feature as a "put things in queue" you are right. I'm rather in favor of a functional "send a mail" view of it and then talking about mailqueue doesn't make sense.

pom.xml
```diff
@@ -1860,6 +1860,11 @@
                 <artifactId>james-server-webadmin-jmap</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
+                <groupId>${james.groupId}</groupId>
+                <artifactId>james-server-webadmin-mail-over-web</artifactId>
```
mbaechler, line 1865: I personally like that separated module

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/request/BodyPartProps.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class BodyPartProps {
+
+    @JsonCreator
+    public BodyPartProps(
```
mbaechler, line 13: this ordering will not pass checkstyle rules. You should run checkstyle and adjust the coding style accordingly.
Rules are mainly Java official guidelines

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/request/MailProps.java
```diff
@@ -0,0 +1,189 @@
...
+import org.apache.james.server.core.MailImpl;
+import org.apache.james.server.core.MimeMessageCopyOnWriteProxy;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.github.fge.lambdas.Throwing;
+import com.google.common.annotations.VisibleForTesting;
+
+public class MailProps {
+
+    private String name = UUID.randomUUID().toString();
```
mbaechler, line 30: we tend to set values to attributes only in the constructor for a better readability

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
mbaechler: I suggest `POST /mail-transfer-service/` (you don't know if it will be delivered locally or not, right? you want it to go into the mailet pipeline I guess)

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public void define(Service service) {
+        defineReceiveMailFromWebService(service);
+    }
+
+    @POST
+    @Path("/receiveMail")
+    @ApiOperation(value = "Deleting an user")
```
mbaechler, line 54: it looks like you left some things after your copy/paste

server/protocols/webadmin/webadmin-mail-over-web/src/test/resources/json/mail.json
mbaechler: I agree that defining new formats is almost always a bad idea: it reduces the value of the feature by preventing interoperability with others tools, it will probably have new design flaws (existing formats have know design flow and people are able to deal with them).
So I would rather send the mail as a `message/rfc822` payload.

mbaechler (commented):
Thank you for this contribution, it looks like you invested in understanding James quite deeply as this PR is quite well design.
With a little help from the community it will eventually land as a useful afdition in the codebase, thanks.

pom.xml
chibenwa: What are the pros of this separate module?

I see it as closely related to the mail queue, hence I wonder what we gain from separating it.

mbaechler: From a user perspective, mailqueue is a technical detail, what matters to them/me is sending an email to some people. 

So having a component that allow to send emails to people looks good.

Cons:
* more module

Pros:
* easier to understand / find / document
* well separated features
* easier to audit dependencies

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
chibenwa: Fibe with me

server/container/guice/pom.xml
```diff
@@ -189,6 +190,46 @@
...
+            <dependency>
+                <groupId>${james.groupId}</groupId>
+                <artifactId>james-server-guice-webadmin-mail-over-web</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>${james.groupId}</groupId>
+                <artifactId>james-server-guice-webadmin-mail-over-web</artifactId>
+                <version>${project.version}</version>
+            </dependency>
```
chibenwa, line 232: Why adding the dependency 8 times?

server/container/guice/protocols/webadmin-mail-over-web/src/main/java/org/apache/james/modules/server/WebAdminMailOverWebModule.java
chibenwa: The Apache V2 license header is compulsory.

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
```diff
@@ -0,0 +1,80 @@
...
+import io.swagger.annotations.Api;
+import io.swagger.annotations.ApiImplicitParam;
+import io.swagger.annotations.ApiImplicitParams;
+import io.swagger.annotations.ApiOperation;
+import io.swagger.annotations.ApiResponse;
+import io.swagger.annotations.ApiResponses;
+import spark.Service;
+
+@Api(tags = "OverWebMailReceiver")
+@Path(ReceiveMailOverWebRoutes.BASE_URL)
```
chibenwa, line 30: I prefer a terminology enforcing the process (transfer) rather than on which side of the transfer we are (receive here)

Hence I propose to name this `TransferEmailRoutes`

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
```diff
@@ -0,0 +1,80 @@
...
+            @ApiResponse(code = HttpStatus.BAD_REQUEST_400, message = "Could not ceate mail from supplied body"),
+            @ApiResponse(code = HttpStatus.INTERNAL_SERVER_ERROR_500,
+                    message = "Internal server error - Something went bad on the server side.")
+    })
+    public void defineReceiveMailFromWebService(Service service) {
+        service.post(BASE_URL, (request, response) -> {
+            //parse MimeMessage from request body
+            MimeMessage mimeMessage = new MimeMessage(Session.getDefaultInstance(System.getProperties()), new ByteArrayInputStream(request.bodyAsBytes()));
+            //create MailImpl object from MimeMessage
+            MailImpl mail = MailImpl.fromMimeMessage(UUID.randomUUID().toString(), mimeMessage);
```
chibenwa, line 70: If we are fine about never having a different envelope that the mime headers, then that's fine with me!

server/protocols/webadmin/webadmin-mail-over-web/src/test/java/org/apache/james/routes/ReceiveMailOverWebRoutesTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import io.restassured.RestAssured;
+import io.restassured.builder.RequestSpecBuilder;
+import io.restassured.http.ContentType;
+import io.restassured.specification.RequestSpecification;
+
+public class ReceiveMailOverWebRoutesTest {
```
chibenwa, line 27: We only test the response code here.

Maybe testing the content of the mail queue can be valuable too?

chibenwa (commented):
Nice contribution!

I agree with the content of this work, that in my opinion only needs some minor corrections :-)

We need to add compulsory license file header though.

andrevka:
Made the neccessary changes

server/protocols/webadmin/webadmin-mail-over-web/src/test/java/org/apache/james/routes/ReceiveMailOverWebRoutesTest.java
andrevka: added checking mailqueue contents to the tests

server/protocols/webadmin/webadmin-mail-over-web/src/main/java/org/apache/james/webadmin/routes/ReceiveMailOverWebRoutes.java
andrevka: Changed

mbaechler (approved):
It looks fine. We really need to build a working CI to check the tests are ok.
@chibenwa : can you please check on Linagora's CI for this time ?

chibenwa:
> @chibenwa : can you please check on Linagora's CI for this time ?

Was my plan ;-)

chibenwa (approved):
I'm fine with this work too.

chibenwa:
Thinking about it, we might want, within the mailet container, to be able to identify emails submitted by this API - in order for instance to allow their relay to remote mail server.

The way we addressed this problem, for instance within the JMAP code base, is to position a mail attribute on the email, which can then be matched on.

See for instance MailSpool::send method && SentByJmap matcher.

I think this work would benefit to adapt a similar approach - this can of course be done as a separate PR.

chibenwa:
Hi @andrevka ,

I just merged this (great!) peace of work. This PR can thus be closed.

Would you find some energy to handle https://github.com/apache/james-project/pull/256#issuecomment-722141233 ?

Cheers,

Benoit

andrevka:
> Thinking about it, we might want, within the mailet container, to be able to identify emails submitted by this API - in order for instance to allow their relay to remote mail server.
> 
> The way we addressed this problem, for instance within the JMAP code base, is to position a mail attribute on the email, which can then be matched on.
> 
> See for instance MailSpool::send method && SentByJmap matcher.
> 
> I think this work would benefit to adapt a similar approach - this can of course be done as a separate PR.



> Hi @andrevka ,
> 
> I just merged this (great!) peace of work. This PR can thus be closed.
> 
> Would you find some energy to handle [#256 (comment)](https://github.com/apache/james-project/pull/256#issuecomment-722141233) ?
> 
> Cheers,
> 
> Benoit

I think i will find the energy

chibenwa:
>I think i will find the energy

Many thanks!

PR #257, opened 2020-10-28
Use RFC6570 Level 1 URI Template for JMAP downloadUrl

antonok-edm:
Section 2 of [RFC8620](https://tools.ietf.org/html/rfc8620) describes the `downloadUrl` property of the JMAP Session resource as follows:

> **downloadUrl**: `String` The URL endpoint to use when downloading files, in URI Template (level 1) format [RFC6570](https://tools.ietf.org/html/rfc6570). The URL MUST contain variables called `accountId`, `blobId`, `type`, and `name`. The use of these variables is described in Section 6.2. Due to potential encoding issues with slashes in content types, it is RECOMMENDED to put the `type` variable in the query section of the URL.

I've updated the `downloadUrl` property to be returned in the correct Level 1 URI Template format by surrounding each expression with curly braces rather than the previous leading-`$` format.


chibenwa (approved):
Thanks ;-)

chibenwa:
https://github.com/linagora/james-project/pull/3977/commits/9e873a8073aba2d222ccb1ce7efd4b8e39a75cfe do add accountId parameter for upload URL too.

Arsnael (approved):
Thanks for the contribution @antonok-edm :)

antonok-edm:
Closing - looks like this has been merged now.

chibenwa:
Yes you are right!

Much sorry for not telling it here...

Apologies.

Benoit

PR #258, opened 2020-10-30
James 1902 extract some guice modules

mbaechler:


server/container/guice/guice-common/src/main/java/org/apache/james/modules/server/MailStoreRepositoryModule.java
```diff
@@ -61,14 +59,17 @@ protected void configure() {
 
     @Provides
     @Singleton
-    MailRepositoryStoreConfiguration provideConfiguration(ConfigurationProvider configurationProvider, DefaultItemSupplier defaultItemSupplier) throws ConfigurationException {
+    MailRepositoryStoreConfiguration provideConfiguration(ConfigurationProvider configurationProvider,
+                                                          Provider<MailRepositoryStoreConfiguration.Item> defaultConfigurationItem)
+        throws ConfigurationException {
```
Arsnael, line 64: I would have put that on the line above, this looks odd when I read it

Arsnael (approved):

chibenwa (approved):
I'm fine with this work.

server/container/guice/guice-common/src/main/java/org/apache/james/modules/server/MailStoreRepositoryModule.java
mbaechler: It would make the line very long, I didn't found a good solution for it.

mbaechler:
@chibenwa be careful to synchronize linagora's branch when I merge PR's on apache repository.

PR #259, opened 2020-11-11
[ADR] JMAP: Avoid ElasticSearch on critical reads

chibenwa:


chibenwa:
https://issues.apache.org/jira/browse/JAMES-3440 is the JIRA entry for this...

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,55 @@
...
+ - `Email/query` to retrieve the list of emails. This call is nowadays resolved on ElasticSearch.
+ - `Email/get` to retrieve various levels of details. Depending of requested properties, this is either
+ resolved on Cassandra alone or on ElasticSearch.
+
+So, ElasticSearch is queried on every JMAP interaction. Administrators thus need to enforce availability and good performance
+for this component.
+
+Relying on more software for every read also harms our resiliency as ElasticSearch outages have major impacts.
+
+Also we should mention our ElasticSearch implementation in Distributed James suffer the following flows:
```
Arsnael, line 26: s/suffer/suffers

Arsnael: s/flows/flaws

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,55 @@
...
+ - Updates of flags leads to updates of the all Email object, leading to sparse segments
```
Arsnael, line 27: s/leads/lead

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,55 @@
...
+ - We currently rely on scrolling for JMAP (in order to ensure messageId uniqueness in the response while respecting limit & position)
+ - We noticed some very slow traces against ElasticSearch, even for simple queries.
+
+Regarding Distributed James data-stores responsibilities:
+ - Cassandra is the source of truth for metadata, its storage need to be adapted to known access patterns.
```
Arsnael, line 32: s/need/needs

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,55 @@
...
+ - ElasticSearch allows resolution of arbitrary queries, and perform full text search.
```
Arsnael, line 33: s/perform/performs

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Consequences
+
+A migration task will be provided for new adopters.
+
+Administrators would be offered a configuration option to turn this view on and off as needed.
+
+If enabled administrators would no longer need to ensure high availability and good performances for ElasticSearch.
+We thus expect a decrease in overall ElasticSearch load, allowing savings compared to actual deployments.
+Furthermore, we expected better performances by resolving such queries against Cassandra.
```
Arsnael, line 50: s/expected/expect

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Status
+
+Accepted (lazy consensus).
+
+Scope: Distributed James
+
+## Context
+
+James powers the JMAP protocol.
```
rouazana, line 13: what does this sentence mean?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+
+A user willing to use a webmail will end up doing the following operations:
+ - `Mailbox/get` to retrieve the mailboxes. This call is resolved against metadata stored in Cassandra.
+ - `Email/query` to retrieve the list of emails. This call is nowadays resolved on ElasticSearch.
+ - `Email/get` to retrieve various levels of details. Depending of requested properties, this is either
+ resolved on Cassandra alone or on ElasticSearch.
```
rouazana, line 19: and object storage ?
(btw I don't see in which case Email/get uses ES)

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+
+Scope: Distributed James
+
+## Context
+
+James powers the JMAP protocol.
+
+A user willing to use a webmail will end up doing the following operations:
+ - `Mailbox/get` to retrieve the mailboxes. This call is resolved against metadata stored in Cassandra.
+ - `Email/query` to retrieve the list of emails. This call is nowadays resolved on ElasticSearch.
```
rouazana, line 17: and Cassandra (in particular for rights/acl checking)

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Consequences
+
+A migration task will be provided for new adopters.
+
+Administrators would be offered a configuration option to turn this view on and off as needed.
+
+If enabled administrators would no longer need to ensure high availability and good performances for ElasticSearch.
+We thus expect a decrease in overall ElasticSearch load, allowing savings compared to actual deployments.
+Furthermore, we expect better performances by resolving such queries against Cassandra.
```
rouazana, line 50: do we have an idea of the load that will be added to Cassandra? and of the size of the data we add to Cassandra?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+
+## Alternatives
+
+Those not willing to adopt this view will not be affected. By disabling the listener and the view usage, they will keep
+resolving all `Email/query` against ElasticSearch.
+
+## Example of optimized JMAP requests
+
+### A: Email list sorted by sentAt, with limit
```
rouazana, line 59: where is the limit? implicit?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+ "c1"]
+```
+
+Draft:
+
+```
+[["getMessageList", {"filter":{"inMailboxes": ["abcd"]}, "sort": ["date desc"]}, "#0"]]
+```
+
+### B: Email list sorted by sentAt, with limit, after a given receivedAt date
```
rouazana, line 84: idem?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+       "inMailbox":"abcd",
+       "after": "aDate"
+   }
+   "comparator": [{
+     "property":"sentAt",
+     "isAscending": false
+   }]
+ },
+ "c1"]
+```
```
rouazana, line 102: no draft?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+
+### C: Email list sorted by sentAt, with limit, after a given sentAt date
+
+Draft:
```
rouazana, line 106: no rfc?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,148 @@
...
+
+```
+[["getMessageList", {"filter":{"after":"aDate", "inMailboxes": ["abcd"]}, "sort": ["date desc"]}, "#0"]]
+```
+
```
rouazana, line 111: what about pagination requests? will they be handled? how?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Very low. Reading 30 rows is a very small tasks.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Yes. JMAP have an implicit limit of 256 if nothing is provided.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Intended. Draft do have only a single "date" field, there is no distinction between sentAt and receivedAt

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
MichaelBailly: W3 RuL3Z

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: by reading position+limit items from the view then doing soft filtering as needed.

rouazana: I think that's a key information (maybe that is one key point causing slow down on ElasticSearch), so could you elaborate a little and explain it in the document please?

chibenwa: Well both solutions needs to gather position+limit records then filter them.

Ok to add a line.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+Scope: Distributed James
+
+## Context
+
+A user willing to use a webmail powered by the JMAP protocol will end up doing the following operations:
+ - `Mailbox/get` to retrieve the mailboxes. This call is resolved against metadata stored in Cassandra.
+ - `Email/query` to retrieve the list of emails. This call is nowadays resolved on ElasticSearch for Email search after
+ a right resolution pass against Cassandra.
+ - `Email/get` to retrieve various levels of details. Depending of requested properties, this is either
```
mbaechler, line 17:
```suggestion
 - `Email/get` to retrieve various levels of details. Depending on requested properties, this is either
```

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+ resolved on Cassandra alone or on ObjectStorage.
```
mbaechler, line 18:
```suggestion
 retrieved from Cassandra alone or from ObjectStorage.
```

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+
+So, ElasticSearch is queried on every JMAP interaction. Administrators thus need to enforce availability and good performance
```
mbaechler, line 20: > So, ElasticSearch is queried on every JMAP interaction

Not exactly. It's queried for listing emails.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+for this component.
+
+Relying on more software for every read also harms our resiliency as ElasticSearch outages have major impacts.
```
mbaechler, line 23:
```suggestion
Relying on more services for every read also harms our resiliency as ElasticSearch outages have major impacts.
```

mbaechler, line 23: What do you expect? If you loose any service you loose James availability: S3, Cassandra, RabbitMQ, ElasticSearch.
Why would we want to support unavailability of highly available services in the first place?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+
+Also we should mention our ElasticSearch implementation in Distributed James suffers the following flaws:
+ - Updates of flags lead to updates of the all Email object, leading to sparse segments
+ - We currently rely on scrolling for JMAP (in order to ensure messageId uniqueness in the response while respecting limit & position)
+ - We noticed some very slow traces against ElasticSearch, even for simple queries.
```
mbaechler, line 28: any clue why?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+
+Regarding Distributed James data-stores responsibilities:
+ - Cassandra is the source of truth for metadata, its storage needs to be adapted to known access patterns.
```
mbaechler, line 31: I don't understand this sentence

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+ - ElasticSearch allows resolution of arbitrary queries, and performs full text search.
+
+## Decision
+
+Provide an optional view for most common `Email/query` requests both on Draft and RFC-8621 implementations.
+This includes filters and sorts on 'sentAt'.
+
+This view will be stored on Cassandra, and updated asynchronously via a MailboxListener.
```
mbaechler, line 39:
```suggestion
This view will be stored into Cassandra, and updated asynchronously via a MailboxListener.
```

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Consequences
+
+A migration task will be provided for new adopters.
+
+Administrators would be offered a configuration option to turn this view on and off as needed.
+
+If enabled administrators would no longer need to ensure high availability and good performances for ElasticSearch.
```
mbaechler, line 47: untrue, you can't assume people won't trigger complex queries

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+We thus expect a decrease in overall ElasticSearch load, allowing savings compared to actual deployments.
+Furthermore, we expect better performances by resolving such queries against Cassandra.
+
+The expected added load to Cassandra is low, as the search is a simple Cassandra read. As we only store messageId,
+Cassandra dataset size will only grow of a few percents if enabled.
+
+## Alternatives
+
+Those not willing to adopt this view will not be affected. By disabling the listener and the view usage, they will keep
```
mbaechler, line 56: it's not an alternative to this decision.
Maintaining a memory cache would be another solution

mbaechler (approved):
Lot of small comments by I agree with this feature.
However, please ensure that paging is actually working with these new projections.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -156,4 +156,6 @@ PRIMARY KEY mailboxId
 CLUSTERING COLUMN messageId
 COLUMN sentAt
 COLUMN receivedAt
-```
\ No newline at end of file
+```
+
+Note that to handle position & limit, we need to fetch `position + limit` ordered items then removing `position` firsts items.
```
rouazana, line 161: so if I scroll quickly n times, I will generate 1+2+...+n = n*(n+1)/2 cassandra requests ~= O(n²)

that's pretty bad, no? Couldn't it be a cause of ElasticSearch slowness? Could it slow down Cassandra?

rouazana (commented):
the problem of position and limit is hard, it could have consequences on Cassandra.

ElasticSearch is meant to have some native position & limit capabilities. I know we don't use them essentially because of rights managements, but maybe we are doing a misusage here expecting that Cassandra behaves better in this use case.

Anyway I'm ok to experience it, but we should really care of the global performance in this case.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: > untrue, you can't assume people won't trigger complex queries

I agree with this statement.

However complex queries won't happen for carefully crafted "basic email display queries" our users are careful about using for their basic use cases <3

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: I'll let you do a PR for documenting this alternative in this ADR if you are willing to :-)

chibenwa:
> the problem of position and limit is hard, it could have consequences on Cassandra.

We are returning a full list on metadata on every IMAP synchronisation (that does a **full** fetch because we do not support QRSYNC). Clients trigger this every 15 minutes or so, and it get executed (with extra metadata on mutable data) in 1-2 seconds for mailboxes around 200.000 mails.

This is a VERY rare operation in JMAP.

I'm not scared ;-)

If you are (or other people are) they can turn that of.

If users run into issues in production plateform, they can disable this.

Of course if that turns out being a bad idea, that could be removed from the code base and this ADR abandonned. But let's give a chance to this experimental feature a chance first, because I really believe that is the best decision we can take about ElasticSearch.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: True for Cassandra.

True for ElasticSearch.

JMAP includes some limits concurent call, rate limiting - that can help mitigating these concenrs in the future.

>  Couldn't it be a cause of ElasticSearch slowness?

Maybe for some.

I succeeded to clearly link some to reindexing as well thanks to @tuanlc .

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: And clue why.

But ElasticSearch slow performance likely would require its own ADR. That's a lengthy topic.

Paging is one, there's many others. I described scrolling & data mutabilityu above.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Cassandra is the source of truth for metadata

 -> I think you have no problem understanding this 

its storage needs to be adapted to known access patterns.

 -> This come from Cassandra storage constraints. You need to plan your reads ahead (or allow filtering and kill your cluster)

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: If I loose ES, given that ADR content, I only loose advanced search.

My customers will be waaaay less complaining about "not having search" that "not being able to read their emails".

> Why would we want to support unavailability of highly available services in the first place?

I and the people I work with are human, we do software, there **will** be unavailability on some of those services.

The question now is how we deal with it.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
mbaechler: oh yes, i now understand. The ambiguity comes from the fact I expect responsibilities in this list, not details about how Cassandra works.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Thanks for the remark, that's right.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
mbaechler: What's the point having global queries if you don't expect to build valuable features based on that? Having only simple queries sounds like: we just want a simple webmail to me.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
```diff
@@ -0,0 +1,161 @@
...
+We thus expect a decrease in overall ElasticSearch load, allowing savings compared to actual deployments.
+Furthermore, we expect better performances by resolving such queries against Cassandra.
+
+The expected added load to Cassandra is low, as the search is a simple Cassandra read. As we only store messageId,
+Cassandra dataset size will only grow of a few percents if enabled.
+
+## Alternatives
+
+Those not willing to adopt this view will not be affected. By disabling the listener and the view usage, they will keep
+resolving all `Email/query` against ElasticSearch.
```
mbaechler, line 57:
```suggestion
Another solution is to implement the projecting using a in-memory datagrid such as infinispan. The projection would be computed using a MailboxListener and the data would be first fetched from this cache and fallback to ElasticSearch.
```

mbaechler (approved):

mbaechler:
For solving the scrolling issue, we can design a (git like) DAG to store entries and associate a DAG node to a scrolling state by using state feature of JMAP.
By the way, we'll need to implement `state` at one point.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: We just want to have a simple that always work **.** If that webmail exposes some advanced features that have a lower availability, then it is nice too.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Why would we prefer the Cassandra solution to it?

chibenwa:
What is a DAG ?

Or we can wait scrolling being a problem before over-engineering it.

So far, we just don't know if the current proposal is good enough or not.

Arsnael (approved):

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: It's responsibility is to handle known, common data access pattern, that's not mutually exclusive.

mbaechler:
The problem is, if you don't include the needed complexity from the start, you won't know how it will behave once you include the complexity and thus you may loose your time.

A DAG is a direct acyclic graph, like git.

Whatever the implementation (a DAG may not be the best idea), the idea is to have a "persistent structure" (every change creates a new immutable state) so that a scroll is bound to a given structure. RBDMS usually implements that using MVCC. JMAP state maps to this concept.

I don't know what is the best implementation for that in Cassandra to be honest.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
mbaechler: Cassanra:
+ already there
+ well mastered
In-Memory datagrid:
+ much faster
+ much less restrictions to what can be done (retrieving data and filtering with some code is less of a problem)

chibenwa:
> The problem is, if you don't include the needed complexity from the start, you won't know how it will behave once you include the complexity and thus you may loose your time.

I take the risk.

This proposal is a small implementation effort. Discarding it when needed won't be a problem

> A DAG is a direct acyclic graph, like git.

Thanks for the explanation.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Can you write it down on your proposal so I can one-click apply it?

Also memory is more expensive than disks.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
mbaechler: I don't agree at all but it's not a problem because which webmail you want to offer to your customer is not a James matter (and this feature is optimal).
Could we just drop this sentence as it's not a consensus?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
mbaechler: > Can you write it down on your proposal so I can one-click apply it?

Sorry no, I don't care enough to invest the time for that.

mbaechler:
> This proposal is a small implementation effort. Discarding it when needed won't be a problem

Exactly why writing an ADR before doing a PoC may not be the best idea

rouazana:
> Exactly why writing an ADR before doing a PoC may not be the best idea

And doing a PoC without a proper ADR is often misunderstood.

Here we have some kind of feature flag, so it can be easily tried and removed if not conclusive. The ADR is interesting because without it the first question I would have asked would have been: "why do you want to do this", and the second one "how do you handle pagination". And thus long debates which are really better explained here.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: No. That is one of the more important here.

It is opt-in: you do not necessarily have to consider es as non critical.

But if you want, you can.

What is the problem about alliwing diverging phylosophies?

mbaechler: This sentence comes after a lot of assumptions: we know the usage pattern of users, we know the client they use, we don't care about QoS of services (ES in this case), etc.

The discussion we just had make these things apparent but the ADR does not AFAICT. 

I'm afraid people draw false conclusion by reading the ADR without the comments.

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Done. Documenting alternative is important. Commit history will link to that thread for curious peoples.

chibenwa:
> JMAP state maps to this concept. (DAG)

@mbaechler I would be curious to know why you think that. Can you develop a bit?

I think before starting complicated developments, having a flat, ordered list of changes, served from oldest to newest is way easier to implement than the "from newest to oldest using some intermediate temporary states" documented as an optimization by the spec.

Would that be what you reference as a DAG?

src/adr/0043-avoid-elasticsearch-on-critical-reads.md
chibenwa: Clarified: `given clients following well defined Email/query requests` + `of basic usages (mailbox content listing)`

Out of these assumptions, of course ElasticSearch is a critical dependency.

Arsnael (approved):

chibenwa:
Merged

PR #261, opened 2020-11-16
JAMES-1902 reorganizes blob modules 

jeantil:
This PR also removes a couple duplicate dependencies that were present in the POM files.


server/container/guice/blob/blob-api-guice/pom.xml
```diff
@@ -25,7 +25,7 @@
         <groupId>org.apache.james</groupId>
         <artifactId>james-server-guice</artifactId>
         <version>3.6.0-SNAPSHOT</version>
-        <relativePath>../pom.xml</relativePath>
+        <relativePath>../../pom.xml</relativePath>
```
chibenwa, line 28: Does it makes sense to keep the blob- prefix for folders inside a blob directory? Idem for the guice suffix.

Shouldn't the naming scheme be guice/blob/api ?

chibenwa (commented):
That was a short follow up ;-)

We will launch tests on that tomorrow morning hanoi tz.

Thanks for your involvement.

server/container/guice/blob/blob-api-guice/pom.xml
jeantil: you are right, I wanted to minimize changes but should have included that bit.

chibenwa: Should we do the same for the queue-xx-guice modules, if any?

jeantil: the queue-xxx- modules were already in *clean*  directories ( neither prefix nor suffix)

jeantil:
Looking to make sure I had the proper sort order in the POM files I stumbled upon sortpom, wich led me to `site/src/xdoc/contribute` which helpfully provides: 
```
mvn com.github.ekryd.sortpom:sortpom-maven-plugin:sort -Dsort.keepBlankLines -Dsort.sortDependencies=groupId,artifactId -Dsort.nrOfIndentSpace=4 -Dsort.createBackupFile=false -Dsort.sortModules=true -Dsort.expandEmptyElements=false  -Dsort.predefinedSortOrder="recommended_2008_06"
```
I'm not sure if this recommandation is still applicable because running it returned over 150 incorrectly sorted pom, I applied the fixes to pom files from :
- `server/container/guice/queue/*`
- `server/container/guice/blob/*`
- `server/container/guice/pom.xml`

hopefully that's ok

chibenwa:
We did not run that for ages, but the rules still applies, and we should do a run... (conflicts likely expected!) You are very right...

chibenwa (approved):
A well invested lunch pause!

jeantil:
>We did not run that for ages, but the rules still applies, and we should do a run... (conflicts likely expected!) You are very right...

Should it be configured to run automatically on every compile ?
something like 
```
mvn compile -Psort
```
to make it easier to sort the modules ? then each new PR can commit the fixed poms it has modified

chibenwa:
> Should it be configured to run automatically on every compile ?

And abort the build if needed? => :+1:

Arsnael (approved):
Nice work, thanks!

mbaechler (approved):

rouazana (approved):

chibenwa:
I just merged this work. Can we close the PR?

PR #263, opened 2020-11-18
JAMES-2124 Enforces pom modules and dependencies ordering

jeantil:
Brace for impacts....

This PR adds systematic enforcement of the ordering of pom dependencies and modules using the sortpom maven plugin. The sorting engine when rewriting a pom also enforces top level tag ordering and tag ordering within the parent tag as they appear in the maven project descriptor (pom xsd). 

After applying this PR the build will **fail** (and fail early at the first violation) if dependencies or modules appear in the wrong order in a POM.

To make developer life easier this PR also adds the `-PsortPom` profile which will rewrite offending POM files to sort the dependencies. Note that because of how XML is specified, the rewrite can alter attribute order, remove whitespace etc. This means that a manual review is recommended after running a build with the profile.

The fastest way to test this is to use 
```
mvn package -DskipTests //will fail if a pom contains violations
//or
mvn package -DskipTests -PsortPom // will rewrite all the poms which contain violations
```

The order of the dependencies was significantly altered because the sorter considers that `${james.groupId}` starts with `$` and not with `j` and `$` has lexicographical precedence over alpha characters. I reviewed commits and found one occurrence which seemed to warrant bypassing lexicographical order to ensure a correct build (the netty override in `./server/queue/queue-activemq/pom.xml` ) but I may have missed something.    

`docker run -v $PWD/.m2:/root/.m2 -v $PWD:/origin -t james/project -s` passes 
`docker run -v $PWD/.m2:/root/.m2 -v $PWD:/origin -t james/project` doesn't but I am unsure why (currently running against  master as a sanity check)

> Here be conflicts


mailbox/api/pom.xml
```diff
@@ -45,10 +49,6 @@
             <artifactId>james-server-task-memory</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>${james.groupId}</groupId>
-            <artifactId>apache-mime4j-dom</artifactId>
```
rouazana, line 50: this one is trickier because it's not really an internal POM. Maybe we should replace ${james.groupId} by org.apache.james here?

server/queue/queue-activemq/pom.xml
```diff
@@ -35,9 +35,11 @@
     <dependencies>
         <!-- Forcing netty-transport version -->
         <dependency>
+            <?SORTPOM IGNORE?>
```
rouazana, line 38: why?

server/pom.xml
```diff
@@ -70,9 +70,9 @@
         <module>dns-service/dnsservice-library</module>
         <module>dns-service/dnsservice-test</module>
 
-        <module>mailet/integration-testing</module>
-
         <module>mailet/dkim</module>
+
```
rouazana, line 74: this blank line does not make sense now

rouazana (approved):

server/queue/queue-activemq/pom.xml
chibenwa: Prevent a reordering causing netty override to be lost i guess

server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/pom.xml
```diff
@@ -17,19 +17,17 @@
     specific language governing permissions and limitations
     under the License.
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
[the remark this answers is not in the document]
jeantil: it's not mandatory but because whitespace is considered non-significant in the XML specification parsers don't keep position information in many cases this is one of them. Since
- there were few pom affected by the whitespace removal 
- this will prevent spurious modifications that humans will have to think about
- the namespace declaration on the project tag are more or less noise to humans
I added the few normalizations to the PR

mailbox/api/pom.xml
jeantil: I don't understand : how can it not be internal if it is under our groupId ?

server/queue/queue-activemq/pom.xml
jeantil: exactly this is one of the few cases where dependency declaration order is important more information at https://github.com/Ekryd/sortpom/wiki/Parameters-that-can-affect-your-build

server/pom.xml
jeantil: fixedup

mailbox/api/pom.xml
rouazana: mime4j is a project of org.apache.james. But it's not in the same repository and has not the same versioning than James.
That's why I think here we should explicit "org.apache.james".

server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/pom.xml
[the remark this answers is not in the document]
rouazana: no problem as far as it is not mandatory

mailbox/api/pom.xml
jeantil: I will make the change.

However I would like to point out https://maven.apache.org/guides/mini/guide-naming-conventions.html
> groupId uniquely identifies your project across all projects. [...]  That is, if the current project is a multiple module project, it should append a new identifier to the parent's groupId.

rouazana: I guess it's historical. Probably we should rename mime4j's group id now, but it's not a trivial change :)

jeantil: I have added a commit which explicitely drops the `${james.groupId}` for all the `apache-mime4j-xxx` artifacts

jeantil:
I added 4 commits to address the comments and improve the PR a bit 
- 073b33c drops the incorrect blank line in server/pom.xml
- bcd0a26 switches to using explicit groupIds for apache-mime4j dependencies
- 7a062c3 reverts incorrectly changed empty tags
- defb95b switches pom validation and sorting from compile to earlier validate phase, meaning the build will fail even earlier and running `mvn validate -PsortPom` lets developpers fix poms in a few seconds instead of a several minutes compilation

mailbox/api/pom.xml
chibenwa: >  Probably we should rename mime4j's group id now

:+1:

chibenwa (approved):
I was expecting more changes ;-)

Huge thanks for putting that work together!

chibenwa:
```
The fastest way to test this is to use

mvn package -DskipTests //will fail if a pom contains violations
//or
mvn package -DskipTests -PsortPom // will rewrite all the poms which contain violations

....
```

This fit the https://github.com/apache/james-project/blob/master/src/site/xdoc/contribute.xml https://github.com/apache/james-project/blob/master/docs/modules/community/pages/contributing.adoc content.

Can you add that there?

chibenwa:
Compiling locally:

```
[ERROR] Failed to execute goal com.github.ekryd.sortpom:sortpom-maven-plugin:2.12.0:verify (strict-pom-ordering) on project apache-james-mpt-app: The file /home/benwa/Documents/apache/james-project/mpt/app/dependency-reduced-pom.xml is not sorted -> [Help 1]
```

Can you have a look? Is that an auto generated file? Can we ignore it?

jeantil:
>Can you have a look? Is that an auto generated file? Can we ignore it?

I'll look into this a quick run of mvn validate didn't fail but running the `docker run` based compilation from the readme did fail.

jeantil:
@chibenwa I don't know how I missed it but the latest fixup should fix the issue

jeantil:
@chibenwa I have updated the documentatiom. I also moved POM validation to the `validate` phase ( before compile ) so it can be executed very quickly.

chibenwa (approved):

chibenwa:
I merged this work quickly in order to limit conflicts, and prevent rebases here.

If there are some other feedback, we can handle it afterward.

Cheers, and thanks for this nice contribution!

Benoit

jeantil:
aweswome, I was unsure if I could to squash the commits to clean up the PR as it would reduce readability for review followup

chibenwa:
I squash the fixups unpo merges, and I also added a missing `JAMES-2124` prefix.

Overwize, I keep the history presented during the review, as you produced nice separated unitary commits ;-)

Does this answer some of your questions?

jeantil:
It was more an apology for letting you do the fixup work than a question :) I had no doubts the history would be preserved ( so little doubts that I didn't event check !)

PR #277, opened 2020-12-10
JAMES-3435 Make message read isolation configurable

chibenwa:
Disabling should be considered experimental.

If enabled, regular consistency level is used for
read transactions for message. This might result
in stale reads as the system.paxos table will not
be checked for latest updates. Better performance
are expected by turning it off.

Note that reads performed as part of write transactions
are always performed with a strong consistency.

### DIFF
--- a/backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/configuration/CassandraConfiguration.java
+++ b/backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/configuration/CassandraConfiguration.java
@@ -72,6 +72,7 @@ public class CassandraConfiguration {
     private static final String BLOB_PART_SIZE = "mailbox.blob.part.size";
     private static final String ATTACHMENT_V2_MIGRATION_READ_TIMEOUT = "attachment.v2.migration.read.timeout";
     private static final String MESSAGE_ATTACHMENTID_READ_TIMEOUT = "message.attachmentids.read.timeout";
+    private static final String MESSAGE_READ_STRONG_MESSAGE_READ_STRONG_CONSISTENCY = "message.read.strong.consistency";
     private static final String CONSISTENCY_LEVEL_REGULAR = "cassandra.consistency_level.regular";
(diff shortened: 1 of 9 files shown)


backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/configuration/CassandraConfiguration.java
```diff
@@ -72,6 +72,7 @@
     private static final String BLOB_PART_SIZE = "mailbox.blob.part.size";
     private static final String ATTACHMENT_V2_MIGRATION_READ_TIMEOUT = "attachment.v2.migration.read.timeout";
     private static final String MESSAGE_ATTACHMENTID_READ_TIMEOUT = "message.attachmentids.read.timeout";
+    private static final String MESSAGE_READ_STRONG_MESSAGE_READ_STRONG_CONSISTENCY = "message.read.strong.consistency";
```
Arsnael, line 75: Duplicate in the constant name

backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/configuration/CassandraConfiguration.java
```diff
@@ -300,7 +312,8 @@ public CassandraConfiguration build() {
                 consistencyLevelLightweightTransaction,
                 mailboxReadRepair.orElse(DEFAULT_MAILBOX_READ_REPAIR),
                 mailboxCountersReadRepairMax.orElse(DEFAULT_MAX_MAILBOX_COUNTERS_READ_REPAIR_CHANCE),
-                mailboxCountersReadRepairChanceOneHundred.orElse(DEFAULT_ONE_HUNDRED_MAILBOX_COUNTERS_READ_REPAIR_CHANCE));
+                mailboxCountersReadRepairChanceOneHundred.orElse(DEFAULT_ONE_HUNDRED_MAILBOX_COUNTERS_READ_REPAIR_CHANCE),
+                messageReadStrongConsistency.orElse(true));
```
Arsnael, line 316: should declare a default constant?

Arsnael (approved):

chibenwa:
Merged

PR #278, opened 2020-12-10
JAMES-3431 Optional DSN support for SMTP server

chibenwa:
I decided to write DSN support as a completely optional extension to the James SMTP server.

Why? Because RFC-3461 compliance needs the mail processing to be correctly configured (and handle DSN bounces correctly). Which we cannot assume. Hence, for the sake of compliance, this should be optional.

For the sake of the exercise, I decided to challenge leveraging SMTP hooks to implement new ESMTP extensions. Needless to say, I had bad surprises.

I needed to modify "core" SMTP server code in order to:
 - Pass new supported ESMTP extensions to EHLO command
 - Cary over RCPT parameters to the RcptHooks.
 - Rcpt hooks needs to announce which RCPT parameters they can handle. Rcpt command should not reject those. 

Default methods where used in order NOT to introduce braking changes.

To enable DSN in JAMES, add the following handler to smtpserver.xml : 

```
            <handler class="org.apache.james.smtpserver.dsn.DSNHeloHook"/>
            <handler class="org.apache.james.smtpserver.dsn.DSNMailParameterHook"/>
            <handler class="org.apache.james.smtpserver.dsn.DSNRcptParameterHook"/>
            <handler class="org.apache.james.smtpserver.dsn.DSNMessageHook"/>
```

Not only does this contribution enables opt-in DSN support,

But also it makes it easier for contributors to write their own ESMTP extensions.

### DIFF
--- a/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/HeloHook.java
+++ b/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/HeloHook.java
@@ -29,6 +29,9 @@ import com.google.common.collect.ImmutableSet;
  * Implement this interfaces to hook in the HELO Command
  */
 public interface HeloHook extends Hook {
+    /**
+     * @return ESMTP extensions to be advertised as part of EHLO answers
+     */
     default Set<String> implementedEsmtpFeatures() {
         return ImmutableSet.of();
     }
--- a/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/HeloHook.java
+++ b/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/HeloHook.java
@@ -19,12 +19,19 @@
 
 package org.apache.james.protocols.smtp.hook;
 
+import java.util.Set;
+
 import org.apache.james.protocols.smtp.SMTPSession;
+
+import com.google.common.collect.ImmutableSet;
 
 /**
  * Implement this interfaces to hook in the HELO Command
  */
 public interface HeloHook extends Hook {
+    default Set<String> implementedEsmtpFeatures() {
+        return ImmutableSet.of();
+    }
 
     /**
      * Return the HookResult after run the hook
--- a/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/RcptHook.java
+++ b/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/RcptHook.java
@@ -18,14 +18,25 @@
  ****************************************************************/
 package org.apache.james.protocols.smtp.hook;
 
+import java.util.Map;
+import java.util.Set;
+
 import org.apache.james.core.MailAddress;
 import org.apache.james.core.MaybeSender;
 import org.apache.james.protocols.smtp.SMTPSession;
+
+import com.google.common.collect.ImmutableSet;
 
 /**
  * Implement this interfaces to hook in the MAIL Command
  */
 public interface RcptHook extends Hook {
+    /**
+     * @return RCPT parameters supported by this hook
+     */
+    default Set<String> supportedParameters() {
+        return ImmutableSet.of();
+    }
     
     /**
      * Return the HookResult after run the hook
@@ -49,6 +60,10 @@ public interface RcptHook extends Hook {
      */
     default HookResult doRcpt(SMTPSession session, MaybeSender sender, MailAddress rcpt) {
         return doRcpt(session, sender.asOptional().orElse(null), rcpt);
+    }
+
+    default HookResult doRcpt(SMTPSession session, MaybeSender sender, MailAddress rcpt, Map<String, String> parameters) {
+        return doRcpt(session, sender, rcpt);
     }
 
 }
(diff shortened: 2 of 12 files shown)


protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/HeloHook.java
```diff
@@ -19,12 +19,19 @@
...
+
 import org.apache.james.protocols.smtp.SMTPSession;
 
+import com.google.common.collect.ImmutableSet;
+
 /**
  * Implement this interfaces to hook in the HELO Command
  */
 public interface HeloHook extends Hook {
+    default Set<String> implementedEsmtpFeatures() {
```
chibenwa, line 32: We need Java Doc here!

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/RcptHook.java
```diff
@@ -51,4 +62,8 @@ default HookResult doRcpt(SMTPSession session, MaybeSender sender, MailAddress r
         return doRcpt(session, sender.asOptional().orElse(null), rcpt);
     }
 
+    default HookResult doRcpt(SMTPSession session, MaybeSender sender, MailAddress rcpt, Map<String, String> parameters) {
```
chibenwa, line 65: We need Java doc here!

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/dsn/DSNHeloHook.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Set;
+
+import org.apache.james.protocols.smtp.SMTPSession;
+import org.apache.james.protocols.smtp.hook.HeloHook;
+import org.apache.james.protocols.smtp.hook.HookResult;
+
+import com.google.common.collect.ImmutableSet;
+
+public class DSNHeloHook implements HeloHook {
```
Arsnael, line 30: here too... Helo -> Ehlo ?

server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/DSNTest.java
```diff
@@ -0,0 +1,354 @@
...
+
+    @Before
+    public void setUp() throws Exception {
+        domainList = new MemoryDomainList(new InMemoryDNSService()
+            .registerMxRecord(Domain.LOCALHOST.asString(), "127.0.0.1")
+            .registerMxRecord(LOCAL_DOMAIN, "127.0.0.1")
+            .registerMxRecord("examplebis.local", "127.0.0.1")
+            .registerMxRecord("127.0.0.1", "127.0.0.1"));
+        domainList.setAutoDetect(false);
+        domainList.setAutoDetectIP(false);
```
Arsnael, line 111: https://github.com/linagora/james-project/pull/4115 just got merged, rebase using this syntax for DomainList

server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/DSNTest.java
```diff
@@ -79,6 +83,8 @@
 
 public class DSNTest {
     public static final String LOCAL_DOMAIN = "example.local";
+    public static final Username BOB = Username.of("bob@localhost");
```
Arsnael, line 86: this commit title misses the ticket number

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/hook/HeloHook.java
```diff
@@ -19,12 +19,19 @@
...
+import java.util.Set;
+
 import org.apache.james.protocols.smtp.SMTPSession;
 
+import com.google.common.collect.ImmutableSet;
+
 /**
  * Implement this interfaces to hook in the HELO Command
  */
 public interface HeloHook extends Hook {
```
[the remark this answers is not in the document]
chibenwa: I do not want to change that as it will cause breaking changes for SMTP hooks implementors.

But you are right, there is a typo.

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/dsn/DSNMailParameterHook.java
```diff
@@ -0,0 +1,63 @@
...
+                    LOGGER.debug("Invalid DSN RET value: {}", paramValue);
+                    return Optional.empty();
+                })
+                .ifPresent(ret -> session.setAttachment(DSN_RET, ret, Transaction));
+        }
+        if (paramName.equals(ENVID_PARAMETER)) {
+            DsnParameters.EnvId envId = DsnParameters.EnvId.of(paramValue);
+            session.setAttachment(DSN_ENVID, envId, Transaction);
+        }
+        return HookResult.DECLINED;
```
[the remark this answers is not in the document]
chibenwa: Declined = I'm not taking decisions on whether the hooks should accept or reject the command. 

I agree the naming is ambiguous here (but same thing I want to limit breaking changes in this part of the code!)

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/dsn/DSNHeloHook.java
chibenwa: :+1:  to change this one.

Arsnael (approved):

LanKhuat (approved):

chibenwa:
Merged

PR #280, opened 2020-12-11
JAMES-3477 demonstrates concurrency issue in MimeMessageCOW

chibenwa:
Thanks to @jeantil @mbaechler

### DIFF
--- a/server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
+++ b/server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
@@ -24,6 +24,8 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Date;
 import java.util.Enumeration;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Function;
 
 import javax.activation.DataHandler;
 import javax.mail.Address;
@@ -41,17 +43,40 @@ import javax.mail.search.SearchTerm;
 import org.apache.james.lifecycle.api.Disposable;
 import org.apache.james.lifecycle.api.LifecycleUtil;
 
+import com.google.common.base.Preconditions;
+
 /**
  * This object wraps a "possibly shared" MimeMessage tracking copies and
  * automatically cloning it (if shared) when a write operation is invoked.
  */
 public class MimeMessageCopyOnWriteProxy extends MimeMessage implements Disposable {
+    @FunctionalInterface
+    interface Read<T> {
+        T read(MimeMessage message) throws MessagingException;
+    }
+
+    @FunctionalInterface
+    interface ReadIO<T> {
+        T read(MimeMessage message) throws MessagingException, IOException;
+    }
+
+    @FunctionalInterface
+    interface Write {
+        void write(MimeMessage message) throws MessagingException;
+    }
+
+    @FunctionalInterface
+    interface WriteIO {
+        void write(MimeMessage message) throws MessagingException, IOException;
+    }
 
     /**
      * Used internally to track the reference count It is important that this is
      * static otherwise it will keep a reference to the parent object.
      */
     protected static class MessageReferenceTracker {
+
+        private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
 
         /**
          * reference counter
@@ -67,68 +92,162 @@ public class MimeMessageCopyOnWriteProxy extends MimeMessage implements Disposab
             wrapped = ref;
         }
 
-        protected synchronized void incrementReferenceCount() {
-            referenceCount++;
-        }
-
-        protected synchronized void decrementReferenceCount() {
-            referenceCount--;
-            if (referenceCount <= 0) {
-                LifecycleUtil.dispose(wrapped);
-                wrapped = null;
+        protected void dispose() {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
+            lock.lock();
+            try {
+                referenceCount--;
+                if (referenceCount <= 0) {
+                    LifecycleUtil.dispose(wrapped);
+                    wrapped = null;
+                }
+            } finally {
+                lock.unlock();
             }
         }
 
-        protected synchronized int getReferenceCount() {
-            return referenceCount;
+        protected void incrementReferences() {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
+            lock.lock();
+            try {
+                referenceCount++;
+            } finally {
+                lock.unlock();
+            }
         }
 
-        public synchronized MimeMessage getWrapped() {
+        protected <T> T wrapRead(Read<T> op) throws MessagingException {
+            ReentrantReadWriteLock.ReadLock lock = this.lock.readLock();
+            lock.lock();
+            try {
+                Preconditions.checkState(referenceCount > 0, "Attempt to read a disposed message");
+                return op.read(wrapped);
+            } finally {
+                lock.unlock();
+            }
+        }
+
+        protected <T> T wrapReadIO(ReadIO<T> op) throws MessagingException, IOException {
+            ReentrantReadWriteLock.ReadLock lock = this.lock.readLock();
+            lock.lock();
+            try {
+                Preconditions.checkState(referenceCount > 0, "Attempt to read a disposed message");
+                return op.read(wrapped);
+            } finally {
+                lock.unlock();
+            }
+        }
+
+        protected <T> T wrapReadNoException(Function<MimeMessage, T> op) {
+            ReentrantReadWriteLock.ReadLock lock = this.lock.readLock();
+            lock.lock();
+            try {
+                Preconditions.checkState(referenceCount > 0, "Attempt to read a disposed message");
+                return op.apply(wrapped);
+            } finally {
+                lock.unlock();
+            }
+        }
+
+        protected MessageReferenceTracker wrapWrite(Write op) throws MessagingException {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
+            lock.lock();
+            try {
+                Preconditions.checkState(referenceCount > 0, "Attempt to write a disposed message");
+                if (referenceCount > 1) {
+                    referenceCount--;
+                    MessageReferenceTracker newRef = new MessageReferenceTracker(new MimeMessageWrapper(wrapped));
+                    newRef.wrapWrite(op);
+                    return newRef;
+                } else {
+                    op.write(wrapped);
+                    return this;
+                }
+            } finally {
+                lock.unlock();
+            }
+        }
+
+        protected MessageReferenceTracker wrapWriteIO(WriteIO op) throws MessagingException, IOException {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
+            lock.lock();
+            try {
+                Preconditions.checkState(referenceCount > 0, "Attempt to write a disposed message");
+                if (referenceCount > 1) {
+                    referenceCount--;
+                    MessageReferenceTracker newRef = new MessageReferenceTracker(new MimeMessageWrapper(wrapped));
+                    newRef.wrapWriteIO(op);
+                    return newRef;
+                } else {
+                    op.write(wrapped);
+                    return this;
+                }
+            } finally {
+                lock.unlock();
+            }
+        }
+
+        private MimeMessage getWrapped() {
             return wrapped;
         }
 
+        protected MessageReferenceTracker newRef() {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
+            lock.lock();
+            try {
+                if (referenceCount > 0) {
+                    referenceCount++;
+                    // Still valid
+                    return this;
+                }
+                throw new IllegalStateException("Message was disposed so new refs cannot be obtained");
+            } finally {
+                lock.unlock();
+            }
+        }
     }
+
+    /**
+     * @return a MimeMessageCopyOnWriteProxy wrapping the message. We consider 'message' as an external reference and
+     * will avoid mutate it, rather cloning the email.
+     *
+     * Please note however that modifying 'message' itself will not lead to a correct reference tracking management
+     * unless it is a MimeMessageCopyOnWriteProxy
+     */
+    public static MimeMessageCopyOnWriteProxy fromMimeMessage(MimeMessage message) {
+        return new MimeMessageCopyOnWriteProxy(message, true);
+    }
+
+    /**
+     * @return a MimeMessageCopyOnWriteProxy wrapping the message. We do not consider 'message' as an external reference
+     * and will mutate it, rather than cloning the email.
+     *
+     * Please note however that modifying 'message' itself will not lead to a correct reference tracking management
+     * unless it is a MimeMessageCopyOnWriteProxy
+     */
+    public static MimeMessageCopyOnWriteProxy fromMimeMessageNoExternalReference(MimeMessage message) {
+        return new MimeMessageCopyOnWriteProxy(message, false);
+    }
+
+    private final ReentrantReadWriteLock readWriteLock = new ReentrantReadWriteLock();
 
     protected MessageReferenceTracker refCount;
 
-    public MimeMessageCopyOnWriteProxy(MimeMessage original) throws MessagingException {
-        this(original, false);
-    }
-
-    public MimeMessageCopyOnWriteProxy(MimeMessageSource original) throws MessagingException {
-        this(new MimeMessageWrapper(original), true);
-    }
-
-    /**
-     * Private constructor providing an external reference counter.
-     */
-    private MimeMessageCopyOnWriteProxy(MimeMessage original, boolean writeable) {
+    public MimeMessageCopyOnWriteProxy(MimeMessage original, boolean originalIsAReference) {
         super(Session.getDefaultInstance(System.getProperties(), null));
 
         if (original instanceof MimeMessageCopyOnWriteProxy) {
-            refCount = ((MimeMessageCopyOnWriteProxy) original).refCount;
+            refCount = ((MimeMessageCopyOnWriteProxy) original).newRef();
         } else {
             refCount = new MessageReferenceTracker(original);
-        }
-
-        if (!writeable) {
-            refCount.incrementReferenceCount();
+            if (originalIsAReference) {
+                refCount.incrementReferences(); // We consider 'original' to be a reference too
+            }
         }
     }
 
-    /**
-     * Check the number of references over the MimeMessage and clone it if
-     * needed before returning the reference
-     * 
-     * @throws MessagingException
-     *             exception
-     */
-    protected synchronized MimeMessage getWrappedMessageForWriting() throws MessagingException {
-        if (refCount.getReferenceCount() > 1) {
-            refCount.decrementReferenceCount();
-            refCount = new MessageReferenceTracker(new MimeMessageWrapper(refCount.getWrapped()));
-        }
-        return refCount.getWrapped();
+    public MimeMessageCopyOnWriteProxy(MimeMessageSource original) {
+        this(new MimeMessageWrapper(original), false);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +92,208 @@ public MessageReferenceTracker(MimeMessage ref) {
             wrapped = ref;
         }
 
-        protected synchronized void incrementReferenceCount() {
-            referenceCount++;
+        protected void dispose() {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
```
Arsnael, line 97: I know this is code correct, but maybe I would rename the local variable `writeLock` to avoid any confusion with the class variable?

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +92,208 @@ public MessageReferenceTracker(MimeMessage ref) {
...
+            }
         }
 
-        protected synchronized void decrementReferenceCount() {
-            referenceCount--;
-            if (referenceCount <= 0) {
-                LifecycleUtil.dispose(wrapped);
-                wrapped = null;
+        protected void incrementReferences() {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
```
Arsnael, line 111: idem

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +92,208 @@ public MessageReferenceTracker(MimeMessage ref) {
...
+                referenceCount++;
+            } finally {
+                lock.unlock();
             }
         }
 
-        protected synchronized int getReferenceCount() {
-            return referenceCount;
+        protected <T> T wrapRead(Read<T> op) throws MessagingException {
+            ReentrantReadWriteLock.ReadLock lock = this.lock.readLock();
```
Arsnael, line 121: idem with `readLock`

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +92,208 @@ public MessageReferenceTracker(MimeMessage ref) {
...
+                return op.read(wrapped);
+            } finally {
+                lock.unlock();
+            }
         }
 
-        public synchronized MimeMessage getWrapped() {
-            return wrapped;
+        protected <T> T wrapReadIO(ReadIO<T> op) throws MessagingException, IOException {
+            ReentrantReadWriteLock.ReadLock lock = this.lock.readLock();
```
Arsnael, line 132: idem

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +92,208 @@ public MessageReferenceTracker(MimeMessage ref) {
...
+                Preconditions.checkState(referenceCount > 0, "Attempt to read a disposed message");
+                return op.read(wrapped);
+            } finally {
+                lock.unlock();
+            }
         }
 
-    }
+        protected <T> T wrapReadNoException(Function<MimeMessage, T> op) {
+            ReentrantReadWriteLock.ReadLock lock = this.lock.readLock();
```
Arsnael, line 143: idem

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +92,208 @@ public MessageReferenceTracker(MimeMessage ref) {
...
+                Preconditions.checkState(referenceCount > 0, "Attempt to read a disposed message");
+                return op.apply(wrapped);
+            } finally {
+                lock.unlock();
+            }
+        }
 
-    protected MessageReferenceTracker refCount;
+        protected MessageReferenceTracker wrapWrite(Write op) throws MessagingException {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
```
Arsnael, line 154: idem

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +92,208 @@ public MessageReferenceTracker(MimeMessage ref) {
...
+        }
 
-    public MimeMessageCopyOnWriteProxy(MimeMessage original) throws MessagingException {
-        this(original, false);
+        private MimeMessage getWrapped() {
+            return wrapped;
+        }
+
+        protected MessageReferenceTracker newRef() {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
```
Arsnael, line 196: idem

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +92,208 @@ public MessageReferenceTracker(MimeMessage ref) {
...
+                    op.write(wrapped);
+                    return this;
+                }
+            } finally {
+                lock.unlock();
+            }
+        }
+
+        protected MessageReferenceTracker wrapWriteIO(WriteIO op) throws MessagingException, IOException {
+            ReentrantReadWriteLock.WriteLock lock = this.lock.writeLock();
```
Arsnael, line 173: idem

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
```diff
@@ -67,98 +93,208 @@ public MessageReferenceTracker(MimeMessage ref) {
...
-        protected synchronized void decrementReferenceCount() {
-            referenceCount--;
-            if (referenceCount <= 0) {
-                LifecycleUtil.dispose(wrapped);
-                wrapped = null;
+        protected void incrementReferences() {
+            ReentrantReadWriteLock.WriteLock writeLock = lock.writeLock();
+            writeLock.lock();
+            try {
+                referenceCount++;
```
jeantil, line 114: out of curiosity: what meaningful exception could possibly be thrown here ?

jeantil (commented):
I know all the effort that has been invested in all this, but I would very much like to see the benchmark that led to implement this copy on write scheme instead of simply copying on every instanciation. It feels like trying to preserver heap space at the cost of synchronization this is optimization territory and should be backed by measurements that are kept up to date for new releases. I wonder what the perf profile of the server would look like with a more naive copy approach.

chibenwa:
> I know all the effort that has been invested in all this, but I would very much like to see the benchmark that led to implement this copy on write scheme instead of simply copying on every instantiation. 

:+1: we need to know if it worth it

We need to see if this code complexity is worth it (I am not super fan of the proposed solution here that partially resolves the problem)

> It feels like trying to preserver heap space at the cost of synchronization this is optimization territory and should be backed by measurements that are kept up to date for new releases.

We are trying to avoid disk writes that could happen from this copy from what I understand.

> I wonder what the perf profile of the server would look like with a more naive copy approach.

We need to design a test, where emails are actually split (matcher splitting leads to several copies being hold)

We could do micro benchmark of 1000 processed emails of 100KB split 5 time against both implems. What would you think ?

chibenwa:
(for the record I see the double lock solution more as a personal challenge than necessary a long term solution)

server/container/core/src/main/java/org/apache/james/server/core/MimeMessageCopyOnWriteProxy.java
chibenwa: I'm becoming paranoid when locks are around I guess!

chibenwa:
Add a quick 'toy' micro-benchmark for evaluating mailetcon…
32d1fca

…tainer efficiency

On top of master: 16.691 ms

On top of the proposed changes: 17.126 ms

With "always copy" behavior: 15.442 ms

Implementation of "always copy" behavior:

```
     private MimeMessageCopyOnWriteProxy(MimeMessage original, boolean originalIsAReference) {
        super(Session.getDefaultInstance(System.getProperties(), null));

        try {
            refCount = new MessageReferenceTracker(new MimeMessageWrapper(original));
        } catch (MessagingException e) {
            throw new RuntimeException(e);
        }
    }
```

Performance gains of copy-on-write proxy are not obvious...

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/MimeMessageBenchTest.java
```diff
@@ -0,0 +1,146 @@
...
+
+import com.github.fge.lambdas.Throwing;
+import com.google.common.base.Stopwatch;
+
+/**
+ * This benches the efficiency at which James mailetcontainer splits emails.
+ */
+@Disabled
+public class MimeMessageBenchTest {
+    private static String CONTENT = "0123456789\r\n".repeat(1024 * 10); // 120KB message
```
jeantil, line 60: I would recommend using random contents, the JVM is  really good at optimizing things that always look the same

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/MimeMessageBenchTest.java
```diff
@@ -0,0 +1,146 @@
...
+    @After
+    public void tearDown() {
+        jamesServer.shutdown();
+    }
+
+    @Test
+    public void receivedMessagesShouldContainDeliveredToHeaders() throws Exception {
+        messageSender.connect(LOCALHOST_IP, jamesServer.getProbe(SmtpGuiceProbe.class).getSmtpPort());
+
+        Stopwatch stopwatch = Stopwatch.createStarted();
```
jeantil, line 96: I think it would be worth it to investigate jmh microbenchmarks in the future. Some useful resources: 
- https://github.com/openjdk/jmh
- https://stackoverflow.com/questions/30485856/how-to-run-jmh-from-inside-junit-tests

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/MimeMessageBenchTest.java
```diff
@@ -0,0 +1,146 @@
...
+                    .matcher(RecipientIs.class)
+                    .matcherCondition("rcpt4@" + DEFAULT_DOMAIN)
+                    .mailet(LogMessage.class)
+                    .addProperty("passThrough", "true"))
+                .addMailet(MailetConfiguration.builder()
+                    .matcher(RecipientIs.class)
+                    .matcherCondition("rcpt5@" + DEFAULT_DOMAIN)
+                    .mailet(LogMessage.class)
+                    .addProperty("passThrough", "true"))
+                .addMailetsFrom(CommonProcessors.deliverOnlyTransport()));
```
jeantil, line 144: while this configuration does split message, as far as I can tell, none of the mailets attempt to concurrenlty "write" to the message, hence the exercised concurrency is actually pretty low.

chibenwa: If "writes" are performed, then content are not shareable anymore, and copy will happen anyway with the cow proxy.

Doing only reads, where things can be shared, and avoiding copy, was my intention...

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/MimeMessageBenchTest.java
chibenwa: Thanks! Junit integration sounds really easy to use.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/MimeMessageBenchTest.java
jeantil: alright

chibenwa:
I played a bit with JMH, garbage collector reports & message size...

```
Scenario: Startup + 100 messages for 5 rcpts, 12KB each + Shutdown
Code version: Locking version of COW
Benchmark                                                                  Mode  Cnt            Score            Error   Units
MimeMessageBenchTest.benchmark1                                            avgt   10        5954.512 ±       110.288   ms/op
MimeMessageBenchTest.benchmark1:·gc.alloc.rate                             avgt   10         116.627 ±         1.936  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Eden_Space                    avgt   10         280.591 ±         4.747  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Old_Gen                       avgt   10           0.171 ±         0.098  MB/sec
MimeMessageBenchTest.benchmark1:·gc.count                                  avgt   10          60.000                  counts
MimeMessageBenchTest.benchmark1:·gc.time                                   avgt   10         302.000                      ms

Scenario: Startup + 100 messages for 5 rcpts, 12KB each + Shutdown
Code version: Always COPY
Benchmark                                                                  Mode  Cnt           Score           Error   Units
MimeMessageBenchTest.benchmark1                                            avgt   10        6024.282 ±       130.217   ms/op
MimeMessageBenchTest.benchmark1:·gc.alloc.rate                             avgt   10         115.991 ±         2.242  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Eden_Space                    avgt   10         277.597 ±         5.493  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Old_Gen                       avgt   10           0.178 ±         0.089  MB/sec
MimeMessageBenchTest.benchmark1:·gc.count                                  avgt   10          60.000                  counts
MimeMessageBenchTest.benchmark1:·gc.time                                   avgt   10         298.000                      ms

Scenario: Startup + 100 messages for 5 rcpts, 120KB each + Shutdown
Code version: Locking version of COW
Benchmark                                                                  Mode  Cnt            Score            Error   Units
MimeMessageBenchTest.benchmark1                                            avgt   10        11838.707 ±       1909.568   ms/op
MimeMessageBenchTest.benchmark1:·gc.alloc.rate                             avgt   10          167.785 ±         22.464  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Eden_Space                    avgt   10         1059.389 ±        154.302  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Old_Gen                       avgt   10            0.583 ±          0.915  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Survivor_Space                avgt   10            3.754 ±          3.913  MB/sec
MimeMessageBenchTest.benchmark1:·gc.count                                  avgt   10         1179.000                   counts
MimeMessageBenchTest.benchmark1:·gc.time                                   avgt   10         5129.000                       ms

Scenario: Startup + 100 messages for 5 rcpts, 120KB each + Shutdown
Code version: Always COPY
MimeMessageBenchTest.benchmark1                                            avgt   10        12177.190 ±       1711.754   ms/op
MimeMessageBenchTest.benchmark1:·gc.alloc.rate                             avgt   10          164.735 ±         20.484  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Eden_Space                    avgt   10         1046.940 ±        134.789  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Old_Gen                       avgt   10            0.586 ±          1.042  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Survivor_Space                avgt   10            3.900 ±          3.591  MB/sec
MimeMessageBenchTest.benchmark1:·gc.count                                  avgt   10         1202.000                   counts
MimeMessageBenchTest.benchmark1:·gc.time                                   avgt   10         5273.000                       ms

Scenario: Startup + 10 messages for 5 rcpts, 1.2MB each + Shutdown
Code version: Locking version of COW
Benchmark                                                                  Mode  Cnt            Score            Error   Units
MimeMessageBenchTest.benchmark1                                            avgt    3        12911.341 ±       3552.079   ms/op
MimeMessageBenchTest.benchmark1:·gc.alloc.rate                             avgt    3          135.412 ±         34.923  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Eden_Space                    avgt    3          799.061 ±        217.977  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Old_Gen                       avgt    3          173.397 ±         76.272  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Survivor_Space                avgt    3            9.891 ±          2.784  MB/sec
MimeMessageBenchTest.benchmark1:·gc.count                                  avgt    3          228.000                   counts
MimeMessageBenchTest.benchmark1:·gc.time                                   avgt    3         3845.000                       ms

Scenario: Startup + 10 messages for 5 rcpts, 1.2MB each + Shutdown
Code version: Always COPY
Benchmark                                                                  Mode  Cnt           Score           Error   Units
MimeMessageBenchTest.benchmark1                                            avgt    3        13718.479 ±       11832.688   ms/op
MimeMessageBenchTest.benchmark1:·gc.alloc.rate                             avgt    3          128.467 ±         101.018  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Eden_Space                    avgt    3          752.122 ±         633.602  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Old_Gen                       avgt    3          163.080 ±         175.520  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Survivor_Space                avgt    3            9.312 ±           4.526  MB/sec
MimeMessageBenchTest.benchmark1:·gc.count                                  avgt    3          229.000                    counts
MimeMessageBenchTest.benchmark1:·gc.time                                   avgt    3         3990.000                        ms
```

Conclusion:
 - Always copy generates slightly more GC
 - Runs sligthly slower

mbaechler:
AFAICT, "always copy" is always faster. GC time is always comparable.
Did you read the right column?

mbaechler:
BTW, your JMH work is  awesome

chibenwa:
> AFAICT, "always copy" is always faster. GC time is always comparable. Did you read the right column?

I do not understand...

13718ms VS 12911ms, 12177 ms vs 11838ms & 6024 VS 5954ms => Always copy is always slightly slower (3% - 1%)

chibenwa:
BTW here is a run for 10MB

```
Scenario: Startup + 3 messages for 5 rcpts, 10MB each + Shutdown
Code version: Locking version of COW
Benchmark                                                                  Mode  Cnt            Score              Error   Units
MimeMessageBenchTest.benchmark1                                            avgt    3        49901.215 ±        92917.167   ms/op
MimeMessageBenchTest.benchmark1:·gc.alloc.rate                             avgt    3           85.808 ±          156.009  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Eden_Space                    avgt    3          511.349 ±          977.955  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Old_Gen                       avgt    3           76.985 ±          165.456  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Survivor_Space                avgt    3            2.232 ±           13.451  MB/sec
MimeMessageBenchTest.benchmark1:·gc.count                                  avgt    3          549.000                     counts
MimeMessageBenchTest.benchmark1:·gc.time                                   avgt    3         8716.000                         ms

Scenario: Startup + 3 messages for 5 rcpts, 10MB each + Shutdown
Code version: Always COPY
Benchmark                                                                  Mode  Cnt            Score              Error   Units
MimeMessageBenchTest.benchmark1                                            avgt    3        55688.810 ±       163013.521   ms/op
MimeMessageBenchTest.benchmark1:·gc.alloc.rate                             avgt    3           77.816 ±          207.068  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Eden_Space                    avgt    3          462.497 ±         1271.321  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Old_Gen                       avgt    3           70.097 ±          189.172  MB/sec
MimeMessageBenchTest.benchmark1:·gc.churn.G1_Survivor_Space                avgt    3            1.939 ±           13.101  MB/sec
MimeMessageBenchTest.benchmark1:·gc.count                                  avgt    3          558.000                     counts
MimeMessageBenchTest.benchmark1:·gc.time                                   avgt    3         9021.000                         ms
```

Maybe we should plan running similar bench on real implems, not just memory, before taking decisions?

mbaechler:
> > AFAICT, "always copy" is always faster. GC time is always comparable. Did you read the right column?
> 
> I do not understand...
> 
> 13718ms VS 12911ms, 12177 ms vs 11838ms & 6024 VS 5954ms => Always copy is always slightly slower (3% - 1%)

Oh yes, smaller is better, I'm wrong.

jeantil:
Hello,

I know it's Christmas time and all so no hurry :)

I am pretty confident that the COW currently on master has a very negative impact on the tests on the CI. It also sounds like a critical issue for production systems (since the failures exhibit total dataloss for incoming emails after they have been successfully accepted by the server ). So I would like to know how this PR or a similar fix can proceed ? 

What I understand is we have to choose between :
- an always copy implementation which has a much higher GC alloc rate for larger mails (perfectly understandable since it copies the whole mail) and seems to be quite a bit slower for large mails >10MB  
- A COW version based on ReentrantLock, with more complex code but better performance in the benchmarks. 

My personal preference would be for the always copy because the code is simpler and it feels safer even though it is slower in micro benchmarks. But my opinion doesn't matter much and all I really care about is a safe implementation :)

chibenwa:
> I know it's Christmas time and all so no hurry :)

Not in VN, but avoid disturbance during new year ;-)

I would say:
 
 - 1. Have a always-copy viable PR (that equals to MimeCow stuff removal)
 - 2. Have a bench on real Mail Queue implems (today & always-copy) - the memory mail queue does extra copies.
 
 We may also want to bring the topic back on the mailing list now that we have gained more insight.

chibenwa:
And thanks for pushing on this topic ;-)

chibenwa:
Let's drop the COW.

See https://github.com/apache/james-project/pull/282

chibenwa:
I started a mailig list discussion https://www.mail-archive.com/server-dev@james.apache.org/msg69361.html

PR #270, opened 2020-12-03
JAMES-2543 Migrate  SMTP tests to  junit 5

jeantil:
* upgrades MPT SMTP Tests
* upgrades protocol SMTP tests

PR #283, opened 2020-12-30
JAMES-2543 Uses JamesServerBuilder in MPT tests

jeantil:
Followup of the Junit 5 migration in #270

This commit replaces the custom xxxSmtpTestRuleFactories and all their custom server initialization logic by the JamesServerBuilder-based builders defined alongside each product. 
This way the integration tests execute against the whole product as it is assembled instead of a semi-custom assembly built especially for the tests.


mpt/impl/smtp/cassandra-rabbitmq-object-storage/src/test/resources/imapserver.xml
```diff
@@ -0,0 +1,54 @@
+<?xml version="1.0"?>
```
chibenwa, line 1: My bet is that we don't need IMAP, LMTP, POP3, etc... while testing SMTP. Maybe we can omit unneeded files?

mpt/impl/smtp/core/src/main/java/org/apache/james/mpt/smtp/InMemoryDnsExtension.java
```diff
@@ -0,0 +1,60 @@
...
+    private InMemoryDNSService inMemoryDns;
+
+    @Override
+    public Module getModule() {
+        return binder -> binder.bind(DNSService.class).toInstance(inMemoryDns);
+    }
+
+    @Override
+    public void afterEach(ExtensionContext extensionContext) throws Exception {
+        inMemoryDns = null;
```
chibenwa, line 43: Is this needed?

chibenwa (commented):
I agree with this work.

These MPT projects had an implicit goal: demonstrate APP assembling with James, and showing we can easily write a mail server without the mailbox component. But... Undocumented, and it was merely a side effect. I fully endorse abstracting it away.

Arsnael (approved):
Nice work, thanks again :)

mpt/impl/smtp/cassandra-rabbitmq-object-storage/src/test/resources/imapserver.xml
jeantil: I tried !  If absent, the configuration mechanism will pick up the default configs in sever/protocols and try to bind on priviledged ports (such as 143), somehow despite having test-jar dependencies to the guice-cassandra and guice-cassandra-rabbit project it did not find the files there. I gave up after wasting a full day trying to trace it :(

mpt/impl/smtp/core/src/main/java/org/apache/james/mpt/smtp/InMemoryDnsExtension.java
jeantil: nope

mpt/impl/smtp/cassandra-rabbitmq-object-storage/src/test/resources/imapserver.xml
chibenwa: Ok. I though the default was disabled.

Maybe instead we can pass an empty configuration block?

jeantil: >Ok. I though the default was disabled.
It definitely isn't for IMAP I didn't try for the others

>Maybe instead we can pass an empty configuration block?
The idea never crossed my ind but it does work rather well and is much more satisfying ! fixedup

chibenwa (approved):

Arsnael:
Hi @jeantil 

This has been merged. Thanks again a lot for this!

Can you help me close this PR? Thank you!

PR #285, opened 2020-12-31
JAMES-3487 Java property to control MimeMessageInputStreamSource thre…

chibenwa:
…shold

This governs the threshold MimeMessageInputStreamSource relies
on for storing MimeMessage content on disk.

Below, data is stored in memory. Above data is stored on disk.

Lower values will lead to longer processing time but will
minimize heap memory usage. Modern SSD hardware
should however support a high throughput. Higher values
will lead to faster single mail processing at the cost
of higher heap usage.

### DIFF
--- a/docs/modules/servers/pages/distributed/configure/index.adoc
+++ b/docs/modules/servers/pages/distributed/configure/index.adoc
@@ -61,3 +61,26 @@ By omitting these files, no extra behaviour is added.
 ** xref:distributed/configure/collecting-contacts.adoc[This page] documents contact collection
 ** xref:distributed/configure/collecting-events.adoc[This page] documents event collection
 ** xref:distributed/configure/dsn.adoc[this page] specified how to support SMTP Delivery Submission Notification (link:https://tools.ietf.org/html/rfc3461[RFC-3461])
+
+== System properties
+
+Some tuning can be done via system properties. This includes:
+
+.System properties
+|===
+| Property name | explanation
+
+| james.message.memory.threshold
+| (Optional). String (size, integer + size units, example: `12 KIB`, supported units are bytes KIB MIB GIB TIB). Defaults to 100KIB.
+This governs the threshold MimeMessageInputStreamSource relies on for storing MimeMessage content on disk.
+Below, data is stored in memory. Above data is stored on disk.
+Lower values will lead to longer processing time but will minimize heap memory usage. Modern SSD hardware
+should however support a high throughput. Higher values will lead to faster single mail processing at the cost
+of higher heap usage.
+
+
+| james.message.usememorycopy
+|Optional. Boolean. Defaults to false. Recommended value is true.
+Should MimeMessageWrapper use a copy of the message in memory?
+
+|===
--- a/docs/modules/servers/pages/distributed/configure/index.adoc
+++ b/docs/modules/servers/pages/distributed/configure/index.adoc
@@ -80,7 +80,7 @@ of higher heap usage.
 
 
 | james.message.usememorycopy
-|Optional. Boolean. Defaults to false. Recommended value is true.
+|Optional. Boolean. Defaults to false. Recommended value is false.
 Should MimeMessageWrapper use a copy of the message in memory?
 
 |===
--- a/src/site/xdoc/server/config-system.xml
+++ b/src/site/xdoc/server/config-system.xml
@@ -166,7 +166,7 @@
                         </dd>
 
                     <dt><strong>james.message.usememorycopy</strong></dt>
-                    <dd>Optional. Boolean. Defaults to false. Recommended value is true. <br/>
+                    <dd>Optional. Boolean. Defaults to false. Recommended value is false. <br/>
                         Should MimeMessageWrapper use a copy of the message in memory?</dd>
                 </dl>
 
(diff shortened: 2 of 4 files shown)


docs/modules/servers/pages/distributed/configure/index.adoc
```diff
@@ -59,4 +59,27 @@ By omitting these files, no extra behaviour is added.
...
+| (Optional). String (size, integer + size units, example: `12 KIB`, supported units are bytes KIB MIB GIB TIB). Defaults to 100KIB.
+This governs the threshold MimeMessageInputStreamSource relies on for storing MimeMessage content on disk.
+Below, data is stored in memory. Above data is stored on disk.
+Lower values will lead to longer processing time but will minimize heap memory usage. Modern SSD hardware
+should however support a high throughput. Higher values will lead to faster single mail processing at the cost
+of higher heap usage.
+
+
+| james.message.usememorycopy
+|Optional. Boolean. Defaults to false. Recommended value is true.
```
jeantil, line 82: > Recommended value is true 
I think the recommended value should be false as is the default, copy via MimeMessageInputStreamSource which leverages OS disk write caches through DeferredFileOutputStream thus letting the OS decide to use memory or disk (trading speed for memory) base  on system load and memory pressure.

src/site/xdoc/server/config-system.xml
```diff
@@ -151,6 +151,27 @@
...
+                    <dd>(Optional). String (size, integer + size units, example: `12 KIB`, supported units are bytes KIB MIB GIB TIB). Defaults to 100KIB. <br/>
+                        This governs the threshold MimeMessageInputStreamSource relies on for storing MimeMessage content on disk. <br/>
+                        Below, data is stored in memory. Above data is stored on disk.
+                        Lower values will lead to longer processing time but will minimize heap memory usage. Modern SSD hardware
+                        should however support a high throughput. Higher values will lead to faster single mail processing at the cost
+                        of higher heap usage.
+                        </dd>
+
+                    <dt><strong>james.message.usememorycopy</strong></dt>
+                    <dd>Optional. Boolean. Defaults to false. Recommended value is true. <br/>
```
jeantil, line 169: same as above on recommended value

docs/modules/servers/pages/distributed/configure/index.adoc
chibenwa: that a typo. Of course MUST be false

chibenwa:
I'll see if I find some tine to do some bench ;-)

jeantil (approved):

chibenwa:
I had been running a quick benchmark, sending 2000 mails of 100KB over 10 threads, on top of the distributed server. Here are the results:

|----------------|----------------|--------------------|
| Code version | master | #286         |
|----------------|----------------|--------------------|
| 1 byte threshold |       |      |
| 100KB threshold |         |   |
| 500 KB threshold |        |   |
|----------------|----------------|--------------------|

I did not manage to evaluate heap impact.

mbaechler (commented):
I don't really like system properties.
Configuration files are easier to audit and make it more easy to discover what can be tuned.
We have configuration file interpolation so we can always fallback to using system properties if we want to.
Would you mind replacing this by a config file entry?

rouazana:
Could we have some real (aka Gatling) bench please? I'm feeling very concerned by the potential impact on production.

jeantil:
>Could we have some real (aka Gatling) bench please? I'm feeling very concerned by the potential impact on production.

can you detail your concerns  ? as far as I can see the default has not changed ...
in theory the computation of the value for threshold could have an impact but it should get inlined pretty quickly since in a production setup the method call is always going to return the same value

chibenwa:
@mbaechler sure no problem, but which conf file would you recomend? `core.properties` ?

chibenwa:
> Could we have some real (aka Gatling) bench please? I'm feeling very concerned by the potential impact on production.

I guess people ending up relying on non default properties would run suh benchs.

chibenwa:
> can you detail your concerns ? as far as I can see the default has not changed ...
in theory the computation of the value for threshold could have an impact but it should get inlined pretty quickly since in a production setup the method call is always going to return the same value

IMO @rouazana concern as far as I understand was more about a Linagora adoption of a non-default value than on the impact of the system property call.

Here the scope is "we want to onfigure this." not "we want to change the default value".

rouazana:
Maybe I just did not comment on the right PR. If I got it correctly we are changing the default behavior to be more reliable, but it could have potential big performance impact. I would like to be reassured. But that's not specific to this PR, I will add a message on the mailing list thread.

jeantil:
I see, your comment makes much more sense in the context of  https://github.com/apache/james-project/pull/282 and possibly  https://github.com/apache/james-project/pull/286

chibenwa:
I will not invest time to introduce yet-another configuration file - we already have too much of them.

@mbaechler I see no viable alternative proposed hence I will follow up with this approach.

chibenwa:
test this please

chibenwa:
To be honnest we fail to reach a consensus, there's no real alternative proposals formulated to the java properties so far.

As I personally do not really care, I will close that work.

Please, kind contributors, if you are interested in customizing these constants feel free to take over this work.

chibenwa:
While having a round at open tickets, I find back this one.

We encountered a similar problem in https://github.com/apache/james-project/pull/450 and addressed it via an environment variable.

I wonder if we could not do the same here...

Cc @jeantil @mbaechler

PR #293, opened 2021-02-12
JAMES-3500 Improve build stability

chibenwa:


server/mailet/mock-smtp-server/src/main/java/org/apache/james/mock/smtp/server/testing/MockSmtpServerExtension.java
```diff
@@ -43,7 +44,8 @@
 
         DockerMockSmtp() {
             mockSmtpServer = DockerContainer.fromName(Images.MOCK_SMTP_SERVER)
-                .withLogConsumer(outputFrame -> LOGGER.debug("MockSMTP: " + outputFrame.getUtf8String()));
+                .withLogConsumer(outputFrame -> LOGGER.debug("MockSMTP: " + outputFrame.getUtf8String()))
+                .waitingFor(Wait.forLogMessage(".*Mock SMTP server started.*", 1));
```
jeantil, line 48: hmm as far as I can tell, forListeningPorts waits for all exposed ports to answer I'm not sure why waiting for the log is better ...

server/protocols/webadmin/webadmin-mailbox-deleted-message-vault/src/test/java/org/apache/james/webadmin/vault/routes/DeletedMessagesVaultRoutesTest.java
```diff
@@ -216,7 +215,6 @@ void beforeEach() throws Exception {
             .start();
         RestAssured.requestSpecification = WebAdminUtils.buildRequestSpecification(webAdminServer)
             .setBasePath(DeletedMessagesVaultRoutes.ROOT_PATH)
-            .log(LogDetail.METHOD)
```
jeantil, line 219: thank you !

server/mailet/mock-smtp-server/src/main/java/org/apache/james/mock/smtp/server/testing/MockSmtpServerExtension.java
chibenwa: HostPortWaitStrategy -> `Waits until a socket connection can be established on a port exposed or mapped by the container.`

I understand `any` not `all`.

A look at InternalCommandPortListeningCheck however tends to confirm what you say.

I will switch back to HostPortWaitStrategy and propose a PR upstream for fixing the ambiguous javadoc.

chibenwa: (yet fixing the container logs is still valuable ;-) )

chibenwa: Correction, I got misleaded by uncompiled code, testcontainer uses a single port for live check (`GenericContainer::getLivenessCheckPort`) I gonna revert the fixup and switch back to the log check.

jeantil: that's weird: https://github.com/testcontainers/testcontainers-java/blob/17b4f6c136f6f2c7dc223bad407221f62a8f0088/core/src/main/java/org/testcontainers/containers/wait/strategy/HostPortWaitStrategy.java#L26 seems to use `getLivenessCheckPorts` (with an `s`) for the external checks and 
https://github.com/testcontainers/testcontainers-java/blob/17b4f6c136f6f2c7dc223bad407221f62a8f0088/core/src/main/java/org/testcontainers/containers/wait/strategy/HostPortWaitStrategy.java#L35 uses `getExposedPorts` for the internal checks and before suggesting this I checked that this was already the case in 1.15.1 
I don't see how you could use the container if both ports are not exposed since the tests call both the smtp and the http service ...

chibenwa: My fault I did clone and looked at my long outdated fork instead of recent code. Sorry for the mess.

chibenwa:
Merged along #291 (just 2 commits were not included in it)

PR #297, opened 2021-02-15
JAMES-3400 Add remove and list domain aliases

chibenwa:
I had been playing a bit with the WebAdmin CLI, and added support for Domain Aliasses.

The road for a full CLI support with likely be long...

### DIFF
--- a/server/protocols/webadmin-cli/src/main/java/org/apache/james/cli/quota/SetGlobalQuotaCountCommand.java
+++ b/server/protocols/webadmin-cli/src/main/java/org/apache/james/cli/quota/SetGlobalQuotaCountCommand.java
@@ -1,0 +1,57 @@
+/******************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one     *
+ * or more contributor license agreements.  See the NOTICE file   *
+ * distributed with this work for additional information          *
+ * regarding copyright ownership.  The ASF licenses this file     *
+ * to you under the Apache License, Version 2.0 (the              *
+ * "License"); you may not use this file except in compliance     *
+ * with the License.  You may obtain a copy of the License at     *
+ *                                                                *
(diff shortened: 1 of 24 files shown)


chibenwa:
test this please

server/protocols/webadmin-cli/src/main/java/org/apache/james/cli/quota/SetGlobalQuotaCountCommand.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @CommandLine.Parameters
+    Long count;
+
+    @Override
+    public Integer call() {
+        try {
+            QuotaClient quotaClient = parentCommand.parentCommand.quotaCommand.fullyQualifiedURL();
+            Response rs = quotaClient.setQuotaCount(count);
+            if (rs.status() == DELETED_CODE) {
```
Arsnael, line 47: DELETED_CODE for a set?

I see similar occurences as well... After rechecking the code, it seems those status code are generally declared locally in each class. However, I think they should be extracted and centralized in a proper class.

WDYT?

server/protocols/webadmin-cli/src/test/java/org/apache/james/cli/QuotaManageTest.java
```diff
@@ -0,0 +1,210 @@
...
+
+                SoftAssertions.assertSoftly(softly -> {
+                    assertThat(exitCode).isEqualTo(0);
+                    assertThat(outputStreamCaptor.toString().trim().toCharArray()).containsOnly("256 bytes".toCharArray());
+                });
+            }
+        }
+    }
+
+    int executeFluent(String... args) {
```
Arsnael, line 204: private?

chibenwa:
:green_apple:  build

Arsnael (approved):

chibenwa:
Merged

PR #299, opened 2021-02-17
JAMES-3431 Fix various issues reported on DSN

chibenwa:
Otto, Karsten Andreas reported the following issues so far regarding DSN

```
 1. When I send a mail to an unknown user, the DSN report (part 1) has no text after "Error message:". There is a comment in the mail MailDispatcher class that might explain it?
    // In order for this server to meet the requirements of the SMTP
    // specification, mails on the ERROR processor must be returned to
    // the sender. Note that this email doesn't include any details
    // regarding the details of the failure(s).
    // In the future we may wish to address this.
    Anyways, this is no big issue for us, as we can describe the problem in the <messageString>. But the empty text is confusing, I thought it to be a bug at first.
    => Please change the report generation to just hide the "Error message:" part if there is no error message to report.

 2. I see a strange date format in the DSN report (part 2), e.g.
    Last-Attempt-Date: Mon, 15 Feb 2021 15:45:20 XXXXX (CET)
    According to rfc3464#section-2.3.7 this header must use the numeric timezone format (+hh:mm/-hh:mm). Does the XXXXX serve any special purpose?
    => Please fix this date format here to comply with the RFC.

 3. We would also like to have the Arrival-Date (rfc3464#section-2.2.5) header in the DSN report (part 2) as well. In the past I found it a useful diagnostic in conjunction with Last-Attempt-Date, especially with delayed delivery cases. Does James per chance already record the timestamp on each incoming mail?
    => Please add the Arrival-Date header to the DSN report. Record the necessary timestamp on mail arrival if James does not have this data already. 
```

### DIFF
--- a/server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/DSNBounceTest.java
+++ b/server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/DSNBounceTest.java
@@ -1364,4 +1364,102 @@ public class DSNBounceTest {
         SharedByteArrayInputStream actualContent = (SharedByteArrayInputStream) content.getBodyPart(1).getContent();
         assertThat(IOUtils.toString(actualContent, StandardCharsets.UTF_8)).isEqualTo(expectedContent);
     }
+
+    @Test
+    void errorMessagePartShouldNotBeAddedWhenNoError() throws Exception {
+        FakeMailetConfig mailetConfig = FakeMailetConfig.builder()
+            .mailetName(MAILET_NAME)
+            .mailetContext(fakeMailContext)
+            .setProperty("defaultStatus", "4.0.0")
+            .build();
+        dsnBounce.init(mailetConfig);
+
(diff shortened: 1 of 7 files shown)


server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/DSNBounceTest.java
```diff
@@ -1276,6 +1279,50 @@ void envIdShouldBePositioned() throws Exception {
         assertThat(IOUtils.toString(actualContent, StandardCharsets.UTF_8)).isEqualTo(expectedContent);
     }
 
+    @Test
+    void ArrivalDateShouldBePositioned() throws Exception {
```
Arsnael, line 1283: `arrivalDateShouldBePositioned`

Arsnael (approved):

chibenwa:
Merged

PR #303, opened 2021-02-18
JAMES-3499 Use a module chooser for LDAP users repository

chibenwa:
* Enables easy reuse for JPA & memory servers to provide
   optional LDAP support out of the box.
 * Drops Cassanra-ldap and Distributed-ldap servers. The
   regular version should be used instead. Docker images
   are maintained as is for now, but we can consider dropping
   them as well (separated PR). Maven projects are kept for now
   but only hosts tests.
 * Ships the minimal LDAP dependency for everybody.

Reduction of server counts ease introducing new servers, for exemple
to let a grace period of one release to upgrade from ElasticSearch
version 6 to 7 (classpath clashes forces distinct maven projects;
the chooser trick can not be employed).

This is a drop-in, invisible replacement for people running ldap
docker images, as well for the CIs building these images.

If consensual:

 - [ ] We likely can get rid of the 2 testing LDAP projects...
 - [ ] We can discuss getting rid of the extra containers (will require changes to the soon to be deprecated linagora/james-jenkins CI)


dockerfiles/compilation/java-11/compile.sh
```diff
@@ -73,16 +73,16 @@ unzip $GLOW_ROOT_ZIP
 if [ $? -eq 0 ]; then
    if [ -d "$CASSANDRA_RABBITMQ_LDAP_DESTINATION" ]; then
       echo "Copying cassandra - rabbitMQ - Ldap JARs"
-      cp server/container/guice/cassandra-rabbitmq-ldap-guice/target/james-server-cassandra-rabbitmq-ldap-guice.jar $CASSANDRA_RABBITMQ_LDAP_DESTINATION || true
-      cp -r server/container/guice/cassandra-rabbitmq-ldap-guice/target/james-server-cassandra-rabbitmq-ldap-guice.lib $CASSANDRA_RABBITMQ_LDAP_DESTINATION || true
+      cp server/container/guice/cassandra-rabbitmq-guice/target/james-server-cassandra-rabbitmq-guice.jar $CASSANDRA_RABBITMQ_LDAP_DESTINATION || true
+      cp -r server/container/guice/cassandra-rabbitmq-guice/target/james-server-cassandra-rabbitmq-guice.lib $CASSANDRA_RABBITMQ_LDAP_DESTINATION || true
```
Arsnael, line 77: I don't think you want to change that. Seeing that you are basing your module chooser on whether or not you have an ldap conf in usersrepository.xml, if you point the build  towards the conf of `cassandra-rabbitmq-guice`, you will not end up with a ldap support server here. You will likely generate twice the same cassandra-rabbitmq-guice image here, right?

dockerfiles/compilation/java-11/compile.sh
```diff
@@ -98,8 +98,8 @@ if [ $? -eq 0 ]; then
       cp server/container/cli/target/james-server-cli.jar $CASSANDRA_DESTINATION || true
       cp -r server/container/cli/target/james-server-cli.lib $CASSANDRA_DESTINATION || true
 
-      cp server/container/guice/cassandra-ldap-guice/target/james-server-cassandra-ldap-guice.jar $CASSANDRA_DESTINATION || true
-      cp -r server/container/guice/cassandra-ldap-guice/target/james-server-cassandra-ldap-guice.lib $CASSANDRA_DESTINATION || true
+      cp server/container/guice/cassandra-guice/target/james-server-cassandra-guice.jar $CASSANDRA_DESTINATION || true
+      cp -r server/container/guice/cassandra-guice/target/james-server-cassandra-guice.lib $CASSANDRA_DESTINATION || true
```
Arsnael, line 102: idem as above

dockerfiles/compilation/java-11/compile.sh
chibenwa: Here we only copy binaries, not the conf.

The conf comes from the dockerfiles folder. We should be good.

Arsnael: You are right I mixed things up here

Arsnael (approved):

LanKhuat (approved):

chibenwa:
This is :green_apple: on the CI

chibenwa:
Merged

PR #311, opened 2021-02-28
JAMES-1717 VacationMailet should avoid logging when no ReplyTo headers

chibenwa:
Some instances of MimeMessage might return null when the requested header
do not exist resulting in a NPE, that got logged.

Behaviour is not altered, as in both case we do abort the processing.


Arsnael (approved):

chibenwa:
Merged

PR #312, opened 2021-03-01
JAMES-3509 Some tiny IMAP fixes

chibenwa:
I found them while lurking in the unfriendly `server/protocols/imap` trying to understand https://issues.apache.org/jira/browse/JAMES-3507 why IMAP APPEND fails when inMemorySizeLimit is exceeded...


Arsnael (approved):

LanKhuat (approved):

chibenwa:
Merged

PR #309, opened 2021-02-25
Do not copy messages in memory before saving them to the blob store

chibenwa:
- This reduce memory impact for enqueueing emails - SMTP, JMAP message submission
 - This reduce memory impact of dequeuing emails - mailetcontainer processing
 - This reduce overhead of CassandraMailRepository (minor gains)

 The memory gain will be major upon big messages.

 Inefficient memory usage of MailboxMessage is not addressed here.

Here is a short presentation of the pattern and concept underneath that I did for my team:

[efficient-io.pdf](https://github.com/apache/james-project/files/6073927/efficient-io.pdf)

PR #315, opened 2021-03-01
JAMES-3506 Avoid multiple mailbox reads upon mail receival

chibenwa:
As we later read the mailbox we can save a "exist" check (~4ms)


chibenwa:
Extracted from https://github.com/apache/james-project/pull/309

Arsnael (approved):

chibenwa:
Merged

PR #317, opened 2021-03-02
JAMES-3505 Make mail delivery multi-threaded

brokenshoebox:
Mail delivery is single threaded due to blocking code in DeliveryRunnable, which prevents reactor from scheduling mails
simultaneously.

**Testing**
Sending email at a rate 400 emails/minute to the server which then relays the emails to an external SMTP server. The external SMTP server adds an extra 300ms delay to the delivery to simulate a slow SMTP server. Ran the test for about 15-20min before and after the fix, these are the results:

_Before:_
image
Confirmed that only 2 RemoteDelivery threads were being used the entire time, one of which seemed to be idle.

_After:_
image
Confirmed that 20 RemoteDelivery threads were being used (running with 2 CPU cores).

### DIFF
--- a/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/remote/delivery/DeliveryRunnable.java
+++ b/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/remote/delivery/DeliveryRunnable.java
@@ -104,24 +104,31 @@ public class DeliveryRunnable implements Disposable {
         }
     }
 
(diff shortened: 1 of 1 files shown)


chibenwa (approved):
Awesome!

Huge thanks ;-)

Arsnael (approved):
Nice, big thanks for this contribution, and the bench showing results of it :) I'm for getting that merged quickly

brokenshoebox:
No problem, happy to help :)

jeantil (approved):

Arsnael:
Merged

PR #319, opened 2021-03-03
MAILBOX-405 Add a retry upon rename

chibenwa:
Cassandra do not leverage transactionnality, hence we should try
to minimize half performed operations. Retries is a way to achieve
this and could be seen as a short term mitigation.


Arsnael (approved):

LanKhuat (approved):

Arsnael (approved):

chibenwa:
Merged

PR #320, opened 2021-03-03
JAMES-3508 Avoid uneeded data copy upon IMAP append

chibenwa:
## Before

b

## After

a

Nice improvment for APPEND...

Each APPENDed message being ~1MB...


protocols/imap/src/main/java/org/apache/james/imap/message/FileBackedLiteral.java
```diff
@@ -0,0 +1,51 @@
+package org.apache.james.imap.message;
```
Arsnael: license missing

Arsnael (approved):

LanKhuat (approved):

chibenwa:
Merged

PR #321, opened 2021-03-04
JAMES-3497 Upgrade camel to latest version

chibenwa:
Most notably the 3.x release line claims shorter boot time.

I did not try to leverage dependency reductions.

### DIFF
--- a/dockerfiles/run/guice/cassandra-ldap/destination/conf/mailetcontainer.xml
+++ b/dockerfiles/run/guice/cassandra-ldap/destination/conf/mailetcontainer.xml
@@ -47,9 +47,12 @@
             <mailet match="All" class="MetricsMailet">
                 <metricName>mailetContainerErrors</metricName>
             </mailet>
-            <mailet match="All" class="Bounce"/>
+            <mailet match="All" class="Bounce">
+                <onMailetException>ignore</onMailetException>
+            </mailet>
             <mailet match="All" class="ToRepository">
                 <repositoryPath>cassandra://var/mail/error/</repositoryPath>
+                <onMailetException>ignore</onMailetException>
             </mailet>
         </processor>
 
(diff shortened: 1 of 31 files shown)


Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-321/1/testReport/junit/org.apache.james.app.spring/?cloudbees-analytics-link=scm-reporting%2Ftests%2Ffailed

Seems creating issues with Spring tests

mbaechler (approved):
It looks good

chibenwa:
To be faily honnest @mbaechler as Camel goal is easing integration, but it a James implementation detail, I wonder if we can drop it, and rewritte the logic in pure Java, gaining boot time as well as potential gains for mailet container...

mbaechler:
That's clearly another topic. I would like to write a spec for a new way of building pipeline (including changing Matcher/Mailet) now that I know that part of James well enough.
We could set up a workshop on that topic if you have similar goals.

chibenwa:
> That's clearly another topic.

Agree, it just makes me think of it.

>  I would like to write a spec for a new way of building pipeline (including changing Matcher/Mailet) now that I know that part of James well enough.

Mixing technical and functional changes... One could be carried over without the over one.

BTW having a plain java mailetcontainer would allow to have reactive mailet/matchers easily.

> I would like to write a spec for a new way of building pipeline

To be honnest you make me curious. :-)

> We could set up a workshop on that topic if you have similar goals.

For me the goal could be to enable writing reactive mailet/matchers.

I think that functionnally mailet/matcher are flexible and a real asset of the James project (but I should not be too close minded AND we could leverage modularization to conciliate diverging point of view. Like ok, you get the feature rich mailetcontainer with advanced configuration and sematic OR you could get the preconfigured one for a basic MDA use case. This reminds me of the Basic/Advanced server discussion.)

Arsnael (approved):

Arsnael (approved):

chibenwa:
Merged

PR #323, opened 2021-03-05
JAMES-3512 DigestUtil: close base64 encoding stream 

chibenwa:


server/data/data-library/src/main/java/org/apache/james/user/lib/model/Algorithm.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public Algorithm(String rawValue) {
+        this.rawValue = rawValue;
+    }
+
+    public String algorithmName() {
+        return rawValue;
+    }
+
+    public String rawValue() {
```
mbaechler, line 49: `asString`?

server/data/data-library/src/main/java/org/apache/james/user/lib/model/Algorithm.java
```diff
@@ -22,36 +22,52 @@
 import java.util.Objects;
 
 public class Algorithm {
+    private static final boolean LEGACY = true;
+    public static final String LEGACY_SUFFIX = "-legacy";
+
     public static Algorithm of(String rawValue) {
-        return new Algorithm(rawValue);
+        if (rawValue.endsWith(LEGACY_SUFFIX)) {
+            return new Algorithm(rawValue, rawValue.substring(0, rawValue.length() - LEGACY_SUFFIX.length()), LEGACY);
```
mbaechler, line 30: why don't we define a closed list of support Algorithm in an enum instead?

chibenwa: Because it would be a breaking change?

Because it is not related to the ISSUE addressed?

Because as a user I might want to use something stronger than SHA-512 ?

mbaechler: `blabla-legacy` is not a breaking change?
I commented on some code that didn't exist before this PR, so debating the design is part of the issue AFAIC.

chibenwa: Well I do not want to build an abstraction limiting the options offered by Java security implementations.

mbaechler: can we at least avoid to expose the `-legacy` string and use two different factory methods?

chibenwa: > can we at least avoid to expose the -legacy string

How would you do to achieve that?

upgrade-instructions.md
```diff
@@ -41,14 +41,15 @@ string being omitted, causing inter-operability issues when synchronizing passwo
...
-default to previous behaviour.
+If you wish to avoid this breaking change you can add the `hashingMode` configuration option in `usersrepository.xml` 
+configuration file, with the value `legacy`, which will ensure previous behaviour.
 
 Example `usersrepository.xml`:
 
 ```
 <usersrepository>
     <algorithm>SHA-512-legacy</algorithm>
+    <hashingMode>legacy</hashingMode>
```
chibenwa, line 52: This is more or less what I understood from your remark @mbaechler .

server/data/data-library/src/main/java/org/apache/james/user/lib/model/Algorithm.java
```diff
@@ -19,19 +19,57 @@
...
+    }
+
+    public interface Factory {
+        Algorithm of(String rawValue);
+    }
+
+    public static class LegacyFactory implements Factory {
+        @Override
+        public Algorithm of(String rawValue) {
+            return new Algorithm(rawValue, rawValue, LEGACY);
```
Arsnael, line 58: If the two arguments for Algorithm `rawValue` and `algorithmName` are the same now, it's probably better to drop one?

Arsnael (approved):

mbaechler (approved):
It's not exactly what I had in mind but it completely reach the goal. Thank you.

chibenwa:
Merged

PR #325, opened 2021-03-08
JAMES-3513 Wrong UID dispatched on the EventBus for StoreMessageIdMan…

chibenwa:
…ager::setInMailboxes

```
Given msg1 in mbx1 with uid 12 and in mbx2 with uid 58
When I set it in mbx3
Then two deletion events are dispatched, one in mbx1 with uid 12 ok, one in mbx2 and uid 12.
```


Arsnael (approved):
Nice catch!

LanKhuat (approved):

chibenwa:
Merged

PR #329, opened 2021-03-11
JAMES-3506 Avoid a full body read within VacationMailet

chibenwa:
In order to see if a part have the message/disposition-notification content type
a full parsing of the message is required.

MDNs falls in the category of automatically
generated reports described in RFC-1892

The Multipart/Report Content Type for the Reporting of Mail System Administrative Messages

The Multipart/Report MIME content-type is a general "family" or
   "container" type for electronic mail reports of any kind.

So:

 - Not only we are more generic
 - But also we save a full message read and parsing


LanKhuat (approved):

Arsnael (approved):

chibenwa:
Merged

PR #330, opened 2021-03-12
JAMES-3499 Distributed guice module should not depend on activemq

chibenwa:
When building custom servers using the James Guice toolkits, this allows having distributed modules without the ActiveMQ dependency.


Arsnael (approved):

LanKhuat (approved):

chibenwa:
Merged

PR #333, opened 2021-03-16
JAMES-3518: Method for docker file copy

LanKhuat:
We are currently using test-containers library with Docker for testing and sometimes, we want to add/overwrite files inside the container. James is currently lacking this option.

### DIFF
--- a/server/testing/src/main/java/org/apache/james/util/docker/DockerContainer.java
+++ b/server/testing/src/main/java/org/apache/james/util/docker/DockerContainer.java
@@ -41,6 +41,7 @@ import org.testcontainers.containers.Network;
 import org.testcontainers.containers.output.OutputFrame;
 import org.testcontainers.containers.wait.strategy.WaitStrategy;
 import org.testcontainers.images.builder.ImageFromDockerfile;
+import org.testcontainers.utility.MountableFile;
 
 import com.github.dockerjava.api.model.ContainerNetwork;
 import com.google.common.base.Strings;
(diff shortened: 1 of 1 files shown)


chibenwa:
No ticket number? No explanation why this is needed?

jeantil:
I'm sorry, I don't understand the purpose of this change. 

As it stands all it does is add an **unused** method to DockerContainer, and the description in  JAMES-3518 doesn't help understand why it is desirable to have a new unused method there. If sonar was setup on the project, the checks would definitely not pass for the PR as it stands.

If you need this change as part of a larger change, please include it in that larger PR. This way the PR will offer context on how and why you need this change ...

chibenwa:
I guess the context is reuse of this class in an external project.

Either the external use could:
- use genericContainer directly
- duplicate and adapt this class
- maybe the smartest way around is to expose a method to manipulate the raw container (consumer<genericContainer>¿) and refactor this class to use it? This one would imo be the best option...

We shall remove - and not write dead code.

But james is also a toolkit to assemble email server. Imo we can introduce genericity not needed as part of james server as long as:
 - we test it
 - we document it.

jeantil:
> I guess the context is reuse of this class in an external project.

Considering what the class does, I wonder if the test container project would be open to a contribution or if a : this class has nothing James specific that I can spot ...

LanKhuat:
Agreed. This change is for an external project that depending on James and does not bring anything useful to James itself. I found a work around and this is no longer needed.

PR #335, opened 2021-03-18
JAMES-3519: Allow ElasticSearchConfiguration customization

LanKhuat:
Builders related to ElasticSearchConfiguration are using default access modifier, preventing users from creating their own configuration in their product.


chibenwa:
You should add that building the POJO enables direct injections in guice, without needing to use commons-configuration abstractions.

A long terme move for the project is to move towards pojoified configuration that is more flexible. Having public builders allows third parties to leverage this design choice.

chibenwa (approved):
Only visibility so you have my 👍

Arsnael (approved):

chibenwa (approved):

Arsnael (approved):

PR #336, opened 2021-03-19
JAMES-3316 Allow custom properties payload for custom capabilities in…

chibenwa:
… JMAP session

Currently, extra capability needed to have empty properties as the JSON generation
of the session capability payload was not modular.

We add an indirection level (the capability properties becomes JSON aware) in order to
permit that.

Session in a conveniant mean to convey information regarding JMAP extensions thus
customizing its payload is critical.


Arsnael (approved):

Arsnael:
```
java.lang.NoClassDefFoundError: Could not initialize class org.apache.james.backends.es.v7.DockerElasticSearchSingleton
	at org.apache.james.mailbox.elasticsearch.v7.ElasticSearchIntegrationTest.<init>(ElasticSearchIntegrationTest.java:87)
```
I don't see how it could be related though... could it be?

LanKhuat (approved):

chibenwa:
Merged

PR #339, opened 2021-03-23
JAMES-3521 Fix unstable integration tests in distributed server

LanKhuat:
Some integration tests running upon distributed server are failing because of fixed ID string, generating ID using default factory...

### DIFF
--- a/server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/src/test/java/org/apache/james/jmap/rfc8621/distributed/DistributedMailboxSetMethodTest.java
+++ b/server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/src/test/java/org/apache/james/jmap/rfc8621/distributed/DistributedMailboxSetMethodTest.java
@@ -23,6 +23,7 @@ import org.apache.james.CassandraExtension;
 import org.apache.james.CassandraRabbitMQJamesConfiguration;
 import org.apache.james.CassandraRabbitMQJamesServerMain;
 import org.apache.james.DockerElasticSearchExtension;
+import org.apache.james.GuiceJamesServer;
 import org.apache.james.JamesServerBuilder;
 import org.apache.james.JamesServerExtension;
 import org.apache.james.jmap.rfc8621.contract.MailboxSetMethodContract;
@@ -32,6 +33,8 @@ import org.apache.james.modules.AwsS3BlobStoreExtension;
 import org.apache.james.modules.RabbitMQExtension;
 import org.apache.james.modules.TestJMAPServerModule;
 import org.apache.james.modules.blobstore.BlobStoreConfiguration;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
 import com.datastax.driver.core.utils.UUIDs;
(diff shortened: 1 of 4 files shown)


chibenwa (approved):

server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/src/test/java/org/apache/james/jmap/rfc8621/distributed/DistributedMailboxSetMethodTest.java
```diff
@@ -64,4 +67,14 @@ public MailboxId randomMailboxId() {
...
+
+    @Override
+    @Test
+    @Disabled("Distributed event bus is asynchronous, we cannot expect the newState to be returned immediately after Mailbox/set call")
+    public void newStateShouldBeUpToDate(GuiceJamesServer server) {}
+
+    @Override
+    @Test
+    @Disabled("Distributed event bus is asynchronous, we cannot expect the newState to be returned immediately after Mailbox/set call")
+    public void oldStateShouldIncludeSetChanges(GuiceJamesServer server) {}
```
Arsnael, line 79: Could not maybe rewrite it, by separating /set and /changes call? You get the old state from the /set response and put an await on your changes call?

server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/src/test/java/org/apache/james/jmap/rfc8621/distributed/DistributedMailboxSetMethodTest.java
```diff
@@ -64,4 +67,14 @@ public MailboxId randomMailboxId() {
     public String errorInvalidMailboxIdMessage(String value) {
         return String.format("%s is not a mailboxId: Invalid UUID string: %s", value, value);
     }
+
+    @Override
+    @Test
+    @Disabled("Distributed event bus is asynchronous, we cannot expect the newState to be returned immediately after Mailbox/set call")
+    public void newStateShouldBeUpToDate(GuiceJamesServer server) {}
```
Arsnael, line 74: Could not maybe rewrite it, by separating /set and /changes call? You get the new state from the /set response and put an await on your changes call?

server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/src/test/java/org/apache/james/jmap/rfc8621/distributed/DistributedMailboxSetMethodTest.java
LanKhuat: Thing is at the point when the update completes, the new state might not been stored yet so we can't reliably extract it from the first call. 

If we use await on the state storing process then this test will be a duplication of a similar one from the MailboxChangesMethodContract.

server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/src/test/java/org/apache/james/jmap/rfc8621/distributed/DistributedMailboxSetMethodTest.java
LanKhuat: idem

Arsnael (approved):

LanKhuat:
`
org.awaitility.core.ConditionTimeoutException: Condition with lambda expression in org.apache.james.webadmin.integration.vault.DeletedMessageVaultIntegrationTest was not fulfilled within 2 minutes.
`
how do i restart the build?

Arsnael:
Merged

PR #342, opened 2021-03-25
JAMES-3524 Write an AESBlobStoreDAO based on (feu) AESPayloadCodec

chibenwa:
Users might not wish to administrate themselves a S3 compatible blobStore and might rely on a third arty to do so. As such, in order to avoid a third party compromission to escalate to a data leak, a good practice is to encrypt the data symmetrically, the secret key generation secrets being stored on the application server.

Such a mechanism prevents data leak for a third party compromission, but do not deend against an application server compromission (as the attacker would then know the private key).

As part of his work on a Swift compatible blob store [1] , Jean Helou contributed an AES encryption mechanism for that very blob store [2]. However, changes in the blobStore design, dropping of the (non-reactive) JCloud driver, rewrite on top of S3 API, as well as modularization of the blobStore (extraction of the BlobStoreDAO, PassThough VS Deduplicating blobStore) [3] lead to this work being dropped, for the sake of simplicity in an effort to finish a long lasting refactoring.

Note that:
- Needs to encrypt blob payload had been requested on top of the Cassandra blob store [4] in order to prevents (full) data leaks from a Cassandra DB compromission.
 - Some optimizations (prior [3]) of the object storage when using S3 were incompatible with payload encryption [5]

By adoption design proposed in [3], reusing the job made by Jean in [2] we can write a generic AESBlobStoreDAO that wraps any other BlobStoreDAO, adding a security layer. Using the BlobStoreChooser, we then can re-enable this capability on top of the Distributed James server.

This pull request contributes the first part of this work.

[1] https://issues.apache.org/jira/browse/JAMES-2525
[2] https://github.com/linagora/james-project/pull/1865 & https://github.com/linagora/james-project/pull/1975 & https://issues.apache.org/jira/browse/JAMES-2589
[3] https://issues.apache.org/jira/browse/JAMES-3028
[4] https://issues.apache.org/jira/browse/JAMES-3023
[5] https://issues.apache.org/jira/browse/JAMES-2692


Arsnael (approved):
I guess guice bindings and related conf would come in a following PR right?

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
```diff
@@ -0,0 +1,155 @@
...
+
+    public byte[] encrypt(byte[] input) {
+        try {
+            return aead.encrypt(input, EMPTY_ASSOCIATED_DATA);
+        } catch (GeneralSecurityException e) {
+            throw new RuntimeException("Unable to build payload for object storage, failed to encrypt", e);
+        }
+    }
+
+    public byte[] decrypt(byte[] ciphertext) throws IOException {
```
jeantil, line 88: This PR sure brings back memories :D 

Sadly this implementation still loads the whole blob data in memory ... The corresponding stream based API had issues because of the way Tink is implemented and the bug is still open to this day https://github.com/google/tink/issues/128

docs/modules/servers/pages/distributed/configure/blobstore.adoc
```diff
@@ -35,6 +35,22 @@ Consequently, all the requested deletions will not be performed, meaning that bl
...
+
+Data can be optionally encrypted with a symmetric key using AES before being stored in the blobStore. As many user relies
+on third party for object storage, a compromised third party will not escalate to a data disclosure. Of course, a
+performance price have to be paid, as encryption takes resources.
+
+*encryption.aes.enable* : Optional boolean, defaults to false.
+
+If AES encryption is enabled, then the following properties MUST be present:
+
+ - *encryption.aes.password* : String
```
jeantil, line 48: it would be nice to provide sample commands to generate proper values for these 
something like: 
- generate password with : openssl rand -base64 64 
- generate salt with : openssl rand -hex 16 

note that I haven't checked if the provided sizes are actually sound or not ...

protocols/imap/src/main/java/org/apache/james/imap/decode/ImapRequestLineReader.java
```diff
@@ -372,7 +372,7 @@ public String consumeLiteral(Charset charset) throws DecodingException {
             return consumeLiteral(US_ASCII);
         } else {
             try {
-                ImmutablePair<Integer, Literal> literal = consumeLiteral(false);
+                ImmutablePair<Integer, Literal> literal = consumeLiteral(true);
```
jeantil, line 375: I'm not sure how this change relates to the rest ...

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: > Sadly this implementation still loads the whole blob data in memory ... 

I know.

I considered using a reactor pool to manage a collection of `Cypher`s and use `CypherInputStream` in our concurrent environment.

But it is a bit more than "restore what existed before" ie at the point of 3.5.0.

If you have a SHA-1 from which I can copy from, I'd be happy to work with the Piped input stream in the waiting for a better solution...

protocols/imap/src/main/java/org/apache/james/imap/decode/ImapRequestLineReader.java
chibenwa: Crap. You are fully right.

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
jeantil: hmm I don have a sha1 but I have this : 
https://github.com/google/tink/issues/128#issuecomment-420893021 and alternatively https://github.com/google/tink/issues/128#issuecomment-433933047 but I'm not convinced it is a better solution as the implementations in these examples starts an unbounded number of threads ... :(

chibenwa: Re-found it, last commit contributes an acceptable version.

https://github.com/apache/james-project/pull/342/commits/d7e7ad9e95be36df7b595d417f9dca2c26da5192

My proposal just ends up writing data potentially to disk... Acceptable as we prefer having large data on disk (not to trash the memory) - and encrypted data size is unknown unless we have it in some kind of byte source - as Object Storage would require the size...

chibenwa: @jeantil thoughts on this final patch?

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
```diff
@@ -41,71 +36,54 @@
...
-        PBEKeySpec spec = new PBEKeySpec(cryptoConfig.password(), saltBytes, PBKDF2_ITERATIONS, KEY_SIZE);
-        return skf.generateSecret(spec);
-    }
-
-    public byte[] encrypt(byte[] input) {
-        try {
-            return aead.encrypt(input, EMPTY_ASSOCIATED_DATA);
-        } catch (GeneralSecurityException e) {
+    public FileBackedOutputStream encrypt(InputStream input) {
+        try (FileBackedOutputStream encryptedContent = new FileBackedOutputStream(100 * 1024)) {
```
jeantil, line 54: 100*1024 feels like deja vu (in the incoming mail processing IIRC), should this be a constant so the values stay aligned ?

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
```diff
@@ -41,71 +36,54 @@
...
+        } catch (GeneralSecurityException | IOException e) {
             throw new RuntimeException("Unable to build payload for object storage, failed to encrypt", e);
         }
     }
 
-    public byte[] decrypt(byte[] ciphertext) throws IOException {
+    public InputStream decrypt(InputStream ciphertext) throws IOException {
         try {
-            return aead.decrypt(ciphertext, EMPTY_ASSOCIATED_DATA);
+            return streamingAead.newDecryptingStream(ciphertext, PBKDF2StreamingAeadFactory.EMPTY_ASSOCIATED_DATA);
```
jeantil, line 66: shouldn't the file backed output stream be used here too ? if the mail was big when it was saved to the store it will still be big when it is read from the store ...

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
jeantil: I like how it aligns with the way incoming emails are processed. I left a couple comments regarding the asymmetry of read and write operations ( not sure if it was deliberate or not).

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: Very true

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: That's deliberate.

 - We have nothing 'yet' to write an outputstream to the blobStore
 - We are able to easily achieve resource management on the write path (hence we will be able to clean the temporary files we allocate) however it is not the same story on the read path - as much as I would like to move resources locally before working on them we won't be able to clean up easily these resources and we end up allocating MailboxMessages in memory to have the GC doing our cleanup - with impacts it implies on the GC...

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: It should be a constant, but do we need it alligned?

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
jeantil: Thanks for the clear explanation, it could warrant a comment ( or an extended commit message with the details ) since the reason for it is not obvious without a lot of contextual knowledge.

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
jeantil: good question: I have no idea. gut feeling says yes but only detailed benchmarking could tell. maybe keep a separate constant but make it easy to discover the other one ( javadoc reference ? comment ?)  so this decision cam be changed more easily later ?

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: Commit message tend to be prefered. 

Having some mailbox related comments in the blobStore feels wrong.

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: Let's go for a one-line comment

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: :-) Detailed benchmarking would be needed on MimeMessageInputStreamSource file threshold anyway

To be fair, here my goal is to restore an interesting feature that got removed. If you really need it slow encryption is better that nothing, and one might agree deploying more James instances to mitigate that matter of fact.

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: :-) Detailed benchmarking would be needed on MimeMessageInputStreamSource file threshold anyway

To be fair, here my goal is to restore an interesting feature that got removed. If you really need it slow encryption is better that nothing, and one might agree deploying more James instances to mitigate that matter of fact.

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: To be really honest, this design only makes sense when coupled with the underlying S3BlobStoreDAO design.

But I feel stuck using it as we miss an `InputStream -> InputSream` efficient AES solution...

chibenwa:
Forced pushed to reword commit only.

chibenwa:
If there is no opposition toward merging this, I will proceed in the coming days...

jeantil (approved):

Arsnael (approved):

chibenwa:
Rebased, I just squashed commits...

PR #345, opened 2021-03-26
JAMES-1489 Failing test for UTF-8 IMAP search

chibenwa:


Arsnael (commented):
Are you gonna try to solve the issue in this PR or it's just to show we have a problem?

chibenwa:
So far I did not succeeded to solve it, showing that we have a problem is better than nothing.

Arsnael (approved):

LanKhuat (approved):

jeantil:
Hi, 

While I think having a failing test to demonstrate that there is an issue is good, I am surprised it has not been tagged unstable. It feel like I missed something ... As it is wouldn't this PR make master's build fail in a systematic manner ?

jeantil:
@Arsnael thanks for clarifying, I had seen the @disabled test but since the PR also added `searchingASingleUTF8CriterionShouldComplete`  which is not disabled I was confused.

chibenwa:
> @Arsnael thanks for clarifying, I had seen the @disabled test but since the PR also added searchingASingleUTF8CriterionShouldComplete which is not disabled I was confused.

@jeantil I did not just took a failing test on the JIRA that I disabled.

I tried to understand the root cause.

I first thought it was a charset handling, the second test proves it is not the case.

I then tought it was a charset cary over issue within the search command parser.

But the real root cause is that our Netty stack is not able to handle a command with several literals and here it gets really complicated with a buggy framing. Ouch. I did not win this battle yet.

Contributing a disabled test allows other contributors to have a look. (That was the approach we followed for the MimeMessage COW?)

chibenwa:
I squashed the fixups. 

:green_apple: build

Arsnael (approved):

chibenwa:
My renewed take on this issue: https://github.com/apache/james-project/pull/980

PR #346, opened 2021-03-26
JAMES-3529 Validate version when set Filter

quantranhong1999:
As a client I do not want to override over people changes.

If I reset based on an outdated filter version, those changes should be rejected.

### DIFF
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/FilteringManagement.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/FilteringManagement.java
@@ -21,6 +21,7 @@ package org.apache.james.jmap.api.filtering;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 
 import org.apache.james.core.Username;
 import org.reactivestreams.Publisher;
@@ -29,14 +30,14 @@ import com.google.common.collect.ImmutableList;
 
 public interface FilteringManagement {
 
-    Publisher<Void> defineRulesForUser(Username username, List<Rule> rules);
+    Publisher<Version> defineRulesForUser(Username username, List<Rule> rules, Optional<Version> ifInState);
 
-    default Publisher<Void> defineRulesForUser(Username username, Rule... rules) {
-        return defineRulesForUser(username, Arrays.asList(rules));
+    default Publisher<Version> defineRulesForUser(Username username, Optional<Version> ifInState, Rule... rules) {
+        return defineRulesForUser(username, Arrays.asList(rules), ifInState);
     }
 
-    default Publisher<Void> clearRulesForUser(Username username) {
-        return defineRulesForUser(username, ImmutableList.of());
+    default Publisher<Version> clearRulesForUser(Username username, Optional<Version> ifInState) {
+        return defineRulesForUser(username, ImmutableList.of(), ifInState);
     }
 
     Publisher<Rules> listRulesForUser(Username username);
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/FilteringManagement.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/FilteringManagement.java
@@ -36,8 +36,8 @@ public interface FilteringManagement {
         return defineRulesForUser(username, Arrays.asList(rules), ifInState);
     }
 
-    default Publisher<Version> clearRulesForUser(Username username, Optional<Version> ifInState) {
-        return defineRulesForUser(username, ImmutableList.of(), ifInState);
+    default Publisher<Version> clearRulesForUser(Username username) {
+        return defineRulesForUser(username, ImmutableList.of(), Optional.empty());
     }
 
     Publisher<Rules> listRulesForUser(Username username);
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/DefineRulesCommandHandler.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/DefineRulesCommandHandler.java
@@ -46,7 +46,7 @@ public class DefineRulesCommandHandler implements CommandHandler<DefineRulesComm
         FilteringAggregateId aggregateId = new FilteringAggregateId(storeCommand.getUsername());
 
         return Mono.from(eventStore.getEventsOfAggregate(aggregateId))
-        .map(history -> FilteringAggregate.load(aggregateId, history).defineRules(storeCommand.getRules()));
+            .map(history -> FilteringAggregate.load(aggregateId, history).defineRules(storeCommand));
     }
 
 }
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/EventSourcingFilteringManagement.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/EventSourcingFilteringManagement.java
@@ -20,6 +20,7 @@
 package org.apache.james.jmap.api.filtering.impl;
 
 import java.util.List;
+import java.util.Optional;
 
 import javax.inject.Inject;
 
@@ -27,6 +28,7 @@ import org.apache.james.core.Username;
 import org.apache.james.eventsourcing.EventSourcingSystem;
 import org.apache.james.eventsourcing.Subscriber;
 import org.apache.james.eventsourcing.eventstore.EventStore;
+import org.apache.james.eventsourcing.eventstore.History;
 import org.apache.james.jmap.api.filtering.FilteringManagement;
 import org.apache.james.jmap.api.filtering.Rule;
 import org.apache.james.jmap.api.filtering.Rules;
@@ -56,8 +58,12 @@ public class EventSourcingFilteringManagement implements FilteringManagement {
     }
 
     @Override
-    public Publisher<Void> defineRulesForUser(Username username, List<Rule> rules) {
-        return Mono.from(eventSourcingSystem.dispatch(new DefineRulesCommand(username, rules))).then();
+    public Publisher<Version> defineRulesForUser(Username username, List<Rule> rules, Optional<Version> ifInState) {
+        return Mono.from(eventSourcingSystem.dispatch(new DefineRulesCommand(username, rules, ifInState)))
+            .then(Mono.from(eventStore.getEventsOfAggregate(new FilteringAggregateId(username)))
+                .map(History::getVersionAsJava)
+                .map(eventIdOptional -> eventIdOptional.map(eventId -> new Version(eventId.value()))
+                    .orElse(Version.INITIAL)));
     }
 
     @Override
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/FilteringAggregate.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/FilteringAggregate.java
@@ -68,7 +68,7 @@ public class FilteringAggregate {
 
     public List<? extends Event> defineRules(DefineRulesCommand storeCommand) {
         Preconditions.checkArgument(shouldNotContainDuplicates(storeCommand.getRules()));
-        Preconditions.checkArgument(expectedState(storeCommand.getIfInState()));
+        Preconditions.checkArgument(expectedState(storeCommand.getIfInState()), "Provided state must be as same as the current state");
         ImmutableList<RuleSetDefined> events = ImmutableList.of(
             new RuleSetDefined(aggregateId, history.getNextEventId(), ImmutableList.copyOf(storeCommand.getRules())));
         events.forEach(this::apply);
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/FilteringAggregate.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/FilteringAggregate.java
@@ -20,6 +20,7 @@
 package org.apache.james.jmap.api.filtering.impl;
 
 import java.util.List;
+import java.util.Optional;
 
 import org.apache.james.eventsourcing.Event;
 import org.apache.james.eventsourcing.EventId;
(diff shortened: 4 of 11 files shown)


server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/DefineRulesCommandHandler.java
```diff
@@ -46,7 +46,7 @@ public DefineRulesCommandHandler(EventStore eventStore) {
         FilteringAggregateId aggregateId = new FilteringAggregateId(storeCommand.getUsername());
 
         return Mono.from(eventStore.getEventsOfAggregate(aggregateId))
-        .map(history -> FilteringAggregate.load(aggregateId, history).defineRules(storeCommand.getRules()));
+            .map(history -> FilteringAggregate.load(aggregateId, history).defineRules(storeCommand.getRules(), storeCommand.getIfInState()));
```
chibenwa, line 49: Can we refactor to:

```suggestion
            .map(history -> FilteringAggregate.load(aggregateId, history).defineRules(storeCommand));
```

As the goal of an aggregate is to handle command?

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/EventSourcingFilteringManagement.java
```diff
@@ -56,8 +58,12 @@ public EventSourcingFilteringManagement(EventStore eventStore) {
...
 
     @Override
-    public Publisher<Void> defineRulesForUser(Username username, List<Rule> rules) {
-        return Mono.from(eventSourcingSystem.dispatch(new DefineRulesCommand(username, rules))).then();
+    public Publisher<Version> defineRulesForUser(Username username, List<Rule> rules, Optional<Version> ifInState) {
+        return Mono.from(eventSourcingSystem.dispatch(new DefineRulesCommand(username, rules, ifInState)))
+            .then(Mono.from(eventStore.getEventsOfAggregate(new FilteringAggregateId(username)))
+                .map(History::getVersionAsJava)
+                .map(optionalVersion -> optionalVersion.map(eventId -> new Version(eventId.value()))
+                    .orElse(null)));
```
chibenwa, line 66: 1. Is that really a `null` value?

In which case is it better than an Optional?

Are we begging for bugs?

Also, shouldn't we have the `EventId -> Version` conversion centralized in a single place? (this would avoid divergences.)

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/impl/FilteringAggregate.java
```diff
@@ -81,6 +83,13 @@ private boolean shouldNotContainDuplicates(List<Rule> rules) {
         return uniqueIdCount == rules.size();
     }
 
+    private boolean shouldHaveTheSameVersionOrExactlyVersionInitialWhenFirstTimeSetRules(Optional<Version> ifInState) {
+        if (ifInState.isPresent() && ifInState.get().equals(Version.INITIAL) && history.getVersionAsJava().isEmpty()) return true;
+        if (history.getVersionAsJava().isPresent() && ifInState.isPresent()
+            && history.getVersionAsJava().get().value() == ifInState.get().getVersion()) return true;
+        return false;
```
chibenwa, line 90: - it do not pass the checkstylle

 - don't expect me to understand piece of code. even less to maintain it.
 
 
Please write comprehensive code for mere humans, use map filter, etc... Let's write functional code...

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/FilteringManagementContract.java
```diff
@@ -66,7 +66,7 @@ default void listingRulesShouldThrowWhenNullUser(EventStore eventStore) {
     default void listingRulesShouldReturnDefinedRules(EventStore eventStore) {
         FilteringManagement testee = instantiateFilteringManagement(eventStore);
 
-        Mono.from(testee.defineRulesForUser(USERNAME, RULE_1, RULE_2)).block();
+        Mono.from(testee.defineRulesForUser(USERNAME, Optional.of(Version.INITIAL), RULE_1, RULE_2)).block();
```
chibenwa, line 69: Just pass `Optional.empty()` for all existing tests!

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/FilteringManagementContract.java
```diff
@@ -133,11 +133,36 @@ default void definingEmptyRuleListShouldRemoveExistingRules(EventStore eventStor
...
 
         assertThat(Mono.from(testee.listRulesForUser(USERNAME)).block())
             .isEqualTo(new Rules(ImmutableList.of(RULE_FROM, RULE_RECIPIENT, RULE_SUBJECT, RULE_TO, RULE_1), new Version(0)));
     }
 
+    @Test
+    default void definingRulesFirstTimeWithWrongVersionInitialShouldFail(EventStore eventStore) {
+        FilteringManagement testee = instantiateFilteringManagement(eventStore);
+
+        assertThatThrownBy(() -> Mono.from(testee.defineRulesForUser(USERNAME, Optional.of(new Version(-2)), RULE_3, RULE_2, RULE_1)).block())
```
chibenwa, line 146: IMO `new Version(-2)` should throw an Illegal arguent exception in itsel (cn you add a precondition there?)

Also, use `new Version(1)` here...

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/FilteringManagementContract.java
```diff
@@ -133,11 +133,36 @@ default void definingEmptyRuleListShouldRemoveExistingRules(EventStore eventStor
...
+    }
+
+    @Test
+    default void setRulesWithEmptyVersionShouldFail(EventStore eventStore) {
+        FilteringManagement testee = instantiateFilteringManagement(eventStore);
+
+        assertThatThrownBy(() -> Mono.from(testee.defineRulesForUser(USERNAME, Optional.empty(), RULE_2, RULE_1)).block())
+            .isInstanceOf(IllegalArgumentException.class);
+    }
+
```
chibenwa, line 167: Can you add a test when modifying with the right version?

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/impl/DefineRulesCommandTest.java
```diff
@@ -40,13 +43,13 @@ void shouldMatchBeanContract() {
 
     @Test
     void constructorShouldThrowWhenNullUser() {
-        assertThatThrownBy(() -> new DefineRulesCommand(null, ImmutableList.of(RULE_1, RULE_2)))
+        assertThatThrownBy(() -> new DefineRulesCommand(null, ImmutableList.of(RULE_1, RULE_2), Optional.of(new Version(0))))
```
chibenwa, line 46: empty

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/impl/DefineRulesCommandTest.java
```diff
@@ -40,13 +43,13 @@ void shouldMatchBeanContract() {
...
             .isInstanceOf(NullPointerException.class);
     }
 
     @Test
     void constructorShouldThrowWhenNullRuleList() {
-        assertThatThrownBy(() -> new DefineRulesCommand(Username.of("adam@james.org"), null))
+        assertThatThrownBy(() -> new DefineRulesCommand(Username.of("adam@james.org"), null, Optional.of(new Version(0))))
```
chibenwa, line 52: empty

server/protocols/jmap-draft/src/test/java/org/apache/james/jmap/mailet/filter/JMAPFilteringExtension.java
```diff
@@ -116,7 +118,7 @@ public void defineRulesForRecipient1(List<Rule.Condition> conditions) {
                     .build())
                 .collect(ImmutableList.toImmutableList());
 
-            Mono.from(testSystem.getFilteringManagement().defineRulesForUser(RECIPIENT_1_USERNAME, rules)).block();
+            Mono.from(testSystem.getFilteringManagement().defineRulesForUser(RECIPIENT_1_USERNAME, rules, Optional.of(Version.INITIAL))).block();
```
chibenwa, line 121: empty

server/protocols/jmap-draft/src/test/java/org/apache/james/jmap/mailet/filter/JMAPFilteringTest.java
```diff
@@ -694,6 +695,7 @@ void mailDirectiveShouldSetLastMatchedRuleWhenMultipleRules(JMAPFilteringTestSys
             MailboxId mailbox3Id = testSystem.createMailbox(RECIPIENT_1_USERNAME, "RECIPIENT_1_MAILBOX_3");
 
             Mono.from(testSystem.getFilteringManagement().defineRulesForUser(RECIPIENT_1_USERNAME,
+                Optional.of(Version.INITIAL),
```
chibenwa, line 698: empty everywhere...

server/protocols/jmap-draft/src/main/java/org/apache/james/jmap/draft/methods/SetFilterMethod.java
```diff
@@ -155,7 +157,7 @@ public SetFilterMethod(MetricFactory metricFactory, FilteringManagement filterin
         ensureNoDuplicatedRules(rules);
         ensureNoMultipleMailboxesRules(rules);
 
-        return Mono.from(filteringManagement.defineRulesForUser(username, rules))
+        return Mono.from(filteringManagement.defineRulesForUser(username, rules, Optional.of(Version.INITIAL)))
```
chibenwa, line 160: Optional.empty

chibenwa:
https://issues.apache.org/jira/browse/JAMES-3332

    James ServerJAMES-3321 Makes James Extendable Server easier to useJAMES-3332

Clarify Users Configuration (usersrepository.xml)

This do not seem related at all to your work. Can we find (or open?) a more relevant JIRA?

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/FilteringManagementContract.java
quantranhong1999: You mean when the first time set filter, the state/version should be empty, right ? I got it.
But when we do Filter/get for user dont have filter yet, it return Version.INITIAL(-1).
Is that a kind of lack of consistency? Shouldn't it return Optional.empty ?

quantranhong1999:
```
    public Rules listRules() {
        return new Rules(state.rules,
            history.getVersionAsJava().map(EventId::value).map(Version::new).orElse(Version.INITIAL));
    }
```

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/FilteringManagement.java
```diff
@@ -29,14 +30,14 @@
 
 public interface FilteringManagement {
 
-    Publisher<Void> defineRulesForUser(Username username, List<Rule> rules);
+    Publisher<Optional<Version>> defineRulesForUser(Username username, List<Rule> rules, Optional<Version> ifInState);
```
chibenwa, line 33: Why not simply `Publisher<Version>` as a return types ? 

When don't we want to have a version?

Idem below...

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/FilteringManagementContract.java
```diff
@@ -76,8 +76,8 @@ default void listingRulesShouldReturnDefinedRules(EventStore eventStore) {
     default void listingRulesShouldReturnLastDefinedRules(EventStore eventStore) {
         FilteringManagement testee = instantiateFilteringManagement(eventStore);
 
-        Mono.from(testee.defineRulesForUser(USERNAME, RULE_1, RULE_2)).block();
-        Mono.from(testee.defineRulesForUser(USERNAME, RULE_2, RULE_1)).block();
+        Mono.from(testee.defineRulesForUser(USERNAME, Optional.empty(), RULE_1, RULE_2)).block();
+        Mono.from(testee.defineRulesForUser(USERNAME, Optional.of(new Version(0)), RULE_2, RULE_1)).block();
```
chibenwa, line 80: Optional.empty() here too

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/FilteringManagementContract.java
```diff
@@ -122,8 +122,8 @@ default void definingRulesShouldKeepOrdering(EventStore eventStore) {
     default void definingEmptyRuleListShouldRemoveExistingRules(EventStore eventStore) {
         FilteringManagement testee = instantiateFilteringManagement(eventStore);
 
-        Mono.from(testee.defineRulesForUser(USERNAME, RULE_3, RULE_2, RULE_1)).block();
-        Mono.from(testee.clearRulesForUser(USERNAME)).block();
+        Mono.from(testee.defineRulesForUser(USERNAME, Optional.empty(), RULE_3, RULE_2, RULE_1)).block();
+        Mono.from(testee.clearRulesForUser(USERNAME, Optional.of(new Version(0)))).block();
```
chibenwa, line 126: Optional.empty() here too?

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/filtering/FilteringManagementContract.java
```diff
@@ -133,11 +133,37 @@ default void definingEmptyRuleListShouldRemoveExistingRules(EventStore eventStor
...
+
+    @Test
+    default void modifyExistingRulesWithRightVersionShouldSucceed(EventStore eventStore) {
+        FilteringManagement testee = instantiateFilteringManagement(eventStore);
+
+        Mono.from(testee.defineRulesForUser(USERNAME, Optional.empty(), RULE_3, RULE_2, RULE_1)).block();
+
+        assertThat(Mono.from(testee.defineRulesForUser(USERNAME, Optional.of(new Version(0)), RULE_3, RULE_2)).block())
+            .isEqualTo(new Version(1));
+    }
```
chibenwa, line 168: I see a lot of missing test before I can give my trust to such a piece of code...

 - Given say a rules with version=1 Am I able to update it specifying ifInState=1 ?
 - Can I use ifInState=INITIAL when no state is defined ?
 - Does using  ifInState=INITIAL fails when a state is defined?
 - Is idInState=1 is well rejected when no rules are persisted yet?

chibenwa (approved):

Arsnael (approved):

quantranhong1999:
@https://github.com/apache/james-project/pull/346/commits/72b265a2578314fa79786e46f6f8e1306e4f27e3
I added an error message for easier handling exception (easily catch error message at filter/set).
Why don't we just add an error message and every thing will be easier to handle? 
for example: 
```
Preconditions.checkArgument(shouldNotContainDuplicates(storeCommand.getRules()), "Contains duplicated rules");
```
then when we call this API, we dont need to validate this again and just catch the error message no?

PR #350, opened 2021-03-29
JAMES-3435 Fine grain LWT configuration for the Cassandra mailbox

chibenwa:
```
| mailbox.read.strong.consistency
| Optional. Boolean, defaults to true. Disabling should be considered experimental.
If enabled, regular consistency level is used for read transactions for mailbox. Not doing so might result
in stale reads as the system.paxos table will not be checked for latest updates. Better performance are expected
by turning it off. Note that reads performed as part of write transactions are always performed with a strong
consistency.

| message.read.strong.consistency
| Optional. Boolean, defaults to true. Disabling should be considered experimental.
If enabled, regular consistency level is used for read transactions for message. Not doing so might result
in stale reads as the system.paxos table will not be checked for latest updates. Better performance are expected
by turning it off. Note that reads performed as part of write transactions are always performed with a strong
consistency.

| message.write.strong.consistency.unsafe
| Optional. Boolean, defaults to true. Disabling should be considered experimental and unsafe.
If enabled, Lightweight transactions will no longer be used upon messages operation (table `imapUidTable`).
As message flags updates relies so far on a read-before-write model, it exposes yourself to data races leading to
potentially update loss. Better performance are expected
by turning it off. Reads performed as part of write transaction are also performed with a relaxed consistency.
```

### DIFF
--- a/docs/modules/servers/pages/distributed/configure/cassandra.adoc
+++ b/docs/modules/servers/pages/distributed/configure/cassandra.adoc
@@ -185,16 +185,23 @@ Controls the number of messages to be retrieved in parallel.
 | Optional. Defaults to 102400 (100KB).
 Controls the size of blob parts used to store messages in the Cassandra blobStore.
 
+| mailbox.read.strong.consistency
+| Optional. Boolean, defaults to true. Disabling should be considered experimental.
+If enabled, regular consistency level is used for read transactions for mailbox. Not doing so might result
+in stale reads as the system.paxos table will not be checked for latest updates. Better performance are expected
+by turning it off. Note that reads performed as part of write transactions are always performed with a strong
+consistency.
+
 | message.read.strong.consistency
 | Optional. Boolean, defaults to true. Disabling should be considered experimental.
-If enabled, regular consistency level is used for read transactions for message. This might result
+If enabled, regular consistency level is used for read transactions for message. Not doing so might result
 in stale reads as the system.paxos table will not be checked for latest updates. Better performance are expected
 by turning it off. Note that reads performed as part of write transactions are always performed with a strong
 consistency.
 
 | message.write.strong.consistency.unsafe
 | Optional. Boolean, defaults to true. Disabling should be considered experimental and unsafe.
-If enabled, Lightweight transactions will no longer be used upon messages operation (table `imapUidTable`).
+If disabled, Lightweight transactions will no longer be used upon messages operation (table `imapUidTable`).
 As message flags updates relies so far on a read-before-write model, it exposes yourself to data races leading to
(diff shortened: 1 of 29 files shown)


docs/modules/servers/pages/distributed/configure/cassandra.adoc
```diff
@@ -191,4 +191,11 @@ If enabled, regular consistency level is used for read transactions for message.
 in stale reads as the system.paxos table will not be checked for latest updates. Better performance are expected
 by turning it off. Note that reads performed as part of write transactions are always performed with a strong
 consistency.
+
+| message.write.strong.consistency.unsafe
+| Optional. Boolean, defaults to true. Disabling should be considered experimental and unsafe.
+If enabled, Lightweight transactions will no longer be used upon messages operation (table `imapUidTable`).
```
Arsnael, line 204: Shouldn't it be if disabled? (same for below docs)

Arsnael: No I was thinking the way you turn it, should be disabled for `message.write.strong.consistency.unsafe`, not for the other parameters. Let me extend it.

Here if disabled makes sense, because then you say LWT are not used in this case. However, for `message.read.strong.consistency` and `mailbox.read.strong.consistency` it should stay if enabled, as you say regular consistency level (so LWT) is used.

When I said same for below docs, that was for the other doc where you mention `message.write.strong.consistency.unsafe`

src/site/xdoc/server/config-cassandra.xml
```diff
@@ -156,6 +156,13 @@
             by turning it off. Note that reads performed as part of write transactions are always performed with a strong
             consistency.</dd>
 
+        <dt><strong>message.write.strong.consistency.unsafe</strong></dt>
+        <dd>Optional. Boolean, defaults to true. Disabling should be considered experimental and unsafe.
```
Arsnael, line 167: Which is this one!

chibenwa: I don't understand. Can you develop?

Arsnael: cf https://github.com/apache/james-project/pull/350#discussion_r603829405

Arsnael (approved):

chibenwa:
Build was :green_apple: only squashed fixups.

chibenwa:
If there's some reviewers out there that would like to give some <3 to this work?

jeantil:
I gave it a cursory look but I haven't dug into mailbox code yet, so I have no idea what i'm looking at which is why I abstained to review further.

PR #358, opened 2021-04-03
[REFACTORING] JMAP: All ids should be backed by IdConstraint

chibenwa:


server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/VacationResponseGetMethodContract.scala
```diff
@@ -482,14 +482,16 @@ trait VacationResponseGetMethodContract {
       .asString
       .stripMargin
 
-    assertThatJson(response).isEqualTo(
+    assertThatJson(response)
+      .whenIgnoringPaths("methodResponses[0][1].description")
```
Arsnael, line 486: Why ignoring the description path here, and then why keeping it and changing the description in the assert below then?

chibenwa: Honestly I gave up trying to understand the escape sequence after 1 hour +.

(a Json in a JSON in a scala string).

I see little value in spending my time doing it.

If you wonna give it a shot though, contribs are welcome.

Arsnael: It just feels pointless of having ignoring description and then a description in the assert... Maybe better to drop the description field in the assert json answer then, and having potentially a comment explaining quickly why, if somebody wants to give it a shot later?

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/MailboxSetMethodContract.scala
```diff
@@ -1059,7 +1059,7 @@ trait MailboxSetMethodContract {
          |      "notCreated": {
          |        "C42": {
          |          "type": "invalidArguments",
-         |          "description": "'/parentId' property in mailbox object is not valid: Predicate isEmpty() did not fail."
+         |          "description": "'/parentId' property in mailbox object is not valid: Left predicate of ((!(0 < 1) && !(0 > 255)) && \\"\\".matches(\\"^[#a-zA-Z0-9-_]*$$\\")) failed: Predicate taking size() = 0 failed: Left predicate of (!(0 < 1) && !(0 > 255)) failed: Predicate (0 < 1) did not fail."
```
mbaechler, line 1062: this looks like an over-precise check: it expects an error specific to Refined but should not

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/mail/MailboxSet.scala
```diff
@@ -46,7 +47,7 @@ case class MailboxSetRequest(accountId: AccountId,
                              onDestroyRemoveEmails: Option[RemoveEmailsOnDestroy]) extends WithAccountId
 
 object MailboxSetRequest {
-  type MailboxCreationId = String Refined NonEmpty
+  type MailboxCreationId = String Refined IdConstraint
```
mbaechler, line 50: As far as I understand Refined, `MailboxCreationId` type is equal to `UnparsedVacationResponseId` and thus defeat the purpose of having different types. You probably need newtype here. (see https://stackoverflow.com/questions/62579057/how-to-ensure-type-safety-with-scalas-refined-library-when-using-the-same-predi for example)

chibenwa: Fair. But it is not a regression, as before everything was 'refined NonEmpty'

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/MailboxSetMethodContract.scala
chibenwa: This is over-precise because this error is generated at the JSON deserialization layer, and we likely have no to little context there.

Instead of relying on JSON refined implicit, then we would have to provide separate types deserialization wrapping our knowledge, I imagine.

Arsnael (approved):

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/mail/MailboxSet.scala
mbaechler: you are right, it's already an enhancement.

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/MailboxSetMethodContract.scala
mbaechler: It's not the message that I dislike, it's the assertion, sorry if my comment was confusing. Could you just ignore part of this message?

chibenwa: Yes.

(I don't like the assertion neither)

But you are right that the description is cryptic ad lacks context. Hopefully it is now prefixed with a human readable message.

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/mail/MailboxSet.scala
chibenwa: But thanks for pointing out this good practice that was of our radar...

2 hours fighting with the code base is still acceptable, no significant rework needed so I decided to give it a shot.

We might now have some "Id POJOs" that represents similar concepts and might be de-duplicated - but I will keep that for other PRs / refactorings.

chibenwa:
Force pushed to solve a rebase conflict...

PR #370, opened 2021-04-08
JAMES-3432 Upload routes test was unstable

chibenwa:
One time out of 25 the uploaded content was altered (Different SHA 256 in
the data store). Several Apache CI builds had been failing because of this.

Investigating this, I found that a piped input stream strategy was not
hitting this pitfall (RectorUtils::toInputStream). I thus incriminate
some weird data races on the Reactor side...

Other areas of the code are likely affected as both the Cassandra and the
S3 blobStores relies on those for streaming (only relied upon for
attachment upload).

Changing the BlobStore API should likely be considered:

```
public interface BlobStore {
    // ...
    Flux<ByteBuffer> read(BucketName bucketName, BlobId blobId);
}
```

Why:

 - Supported by both implementation (Cassandra S3)
 - Avoids intermediate transformations requiring blocking
 operations/dedicated thread resources
  - We convert a Flux[ByteBuffer] => InputStream (to conform to BlobStore API) => Flux[ByteBuffer] (reactor-netty layer)

See https://issues.apache.org/jira/browse/JAMES-3555

### DIFF
--- a/server/container/util/src/main/java/org/apache/james/util/ReactorUtils.java
+++ b/server/container/util/src/main/java/org/apache/james/util/ReactorUtils.java
@@ -33,6 +33,7 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 
 import org.reactivestreams.Publisher;
+import org.reactivestreams.Subscription;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


chibenwa:
https://github.com/linagora/james-project/issues/3896 I believe is related

> Sending 2 images as attachments on Mobile, on the recipient's side, one of them displays lossy.

95813838-fab37680-0d42-11eb-86f0-deda33f5c13c

chibenwa:
Performance tests shall be conducted...

Arsnael (approved):

chibenwa:
There is an issue with my proposal: it do not handle exception propagation with the piped stream.

chibenwa:
As weird as it sounds https://github.com/apache/james-project/pull/374 seems to be enough

PR #377, opened 2021-04-09
JAMES-3557 */changes: Fail explicitly when too much entries on a single change

chibenwa:


server/data/data-jmap/src/test/java/org/apache/james/jmap/api/change/MailboxChangeRepositoryContract.java
```diff
@@ -376,7 +376,7 @@ default void getChangesShouldNotReturnMoreThanMaxChanges() {
     }
 
     @Test
-    default void getChangesShouldReturnEmptyWhenNumberOfChangesExceedMaxChanges() {
+    default void getChangesShouldReturnThrowWhenNumberOfChangesExceedMaxChanges() {
```
LanKhuat, line 380: getChangesShouldThrow...

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailChangesMethodContract.scala
```diff
@@ -168,9 +168,9 @@ trait EmailChangesMethodContract {
       val response = `given`
         .header(ACCEPT.toString, ACCEPT_RFC8621_VERSION_HEADER)
         .body(request)
-      .when
+        .when
         .post
-      .`then`
+        .`then`
```
LanKhuat, line 173: indent

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailChangesMethodContract.scala
```diff
@@ -197,6 +197,110 @@ trait EmailChangesMethodContract {
...
+      .setBody("testmail", StandardCharsets.UTF_8)
+      .build
+    val messageId1: MessageId = mailboxProbe.appendMessage(BOB.asString(), path, AppendCommand.from(message)).getMessageId
+    val messageId2: MessageId = mailboxProbe.appendMessage(BOB.asString(), path, AppendCommand.from(message)).getMessageId
+    val messageId3: MessageId = mailboxProbe.appendMessage(BOB.asString(), path, AppendCommand.from(message)).getMessageId
+    val messageId4: MessageId = mailboxProbe.appendMessage(BOB.asString(), path, AppendCommand.from(message)).getMessageId
+    val messageId5: MessageId = mailboxProbe.appendMessage(BOB.asString(), path, AppendCommand.from(message)).getMessageId
+    val messageId6: MessageId = mailboxProbe.appendMessage(BOB.asString(), path, AppendCommand.from(message)).getMessageId
+
+    val oldState: State = waitForNextState(server, AccountId.fromUsername(BOB), State.INITIAL)
```
LanKhuat, line 219: Did you test this on the distributed environment? You append 6 messages so it should have 6 more states after the INITIAL one.

Arsnael (commented):
I'm a bit confused here... This looks like just a workaround, is there no way to fix this issue to turn into the correct behavior?

chibenwa:
> This looks like just a workaround, is there no way to fix this issue to turn into the correct behavior?

No not really. If the chunck is too large, the chunk is too large. The client should use a bigger /changes then.

An alternative would be to , upon insert, generate several smaller changes instead than one big. But I don't like this approach much, to be honnest.

Arsnael:
> No not really. If the chunck is too large, the chunk is too large. The client should use a bigger /changes then.

What is the point of having `hasMoreChanges` field and mechanic from the spec then if the chunk is too large we return an exception anyway? I'm trying to understand the logic behind this choice (and not about the technical limitations we might have from our code), which seems odd to me sorry

Arsnael:
```
02:30:07,088 [INFO] Running org.apache.james.mailbox.cassandra.CassandraMessageIdManagerStorageTest
02:31:24,808 [ERROR] Tests run: 60, Failures: 0, Errors: 60, Skipped: 0, Time elapsed: 77.705 s <<< FAILURE! - in org.apache.james.mailbox.cassandra.CassandraMessageIdManagerStorageTest
02:31:24,809 [ERROR] setInMailboxesShouldAllowCopyingMessageFromReadOnlySharedMailbox  Time elapsed: 0.594 s  <<< ERROR!
java.lang.NoClassDefFoundError: org/apache/james/events/EventBusTestFixture
	at org.apache.james.mailbox.cassandra.CassandraMessageIdManagerStorageTest.createTestingData(CassandraMessageIdManagerStorageTest.java:41)
Caused by: java.lang.ClassNotFoundException: org.apache.james.events.EventBusTestFixture
	at org.apache.james.mailbox.cassandra.CassandraMessageIdManagerStorageTest.createTestingData(CassandraMessageIdManagerStorageTest.java:41)
```
weird?

vttranlina (approved):

LanKhuat (approved):

Arsnael:
> What is the point of having `hasMoreChanges` field and mechanic from the spec then if the chunk is too large we return an exception anyway? I'm trying to understand the logic behind this choice (and not about the technical limitations we might have from our code), which seems odd to me sorry

Still waiting an answer regarding this concern?

chibenwa:
Say we mutate 5845839345345345 messages at the same time.

We will store the changes in a single row.

We should return all the changes.

But if the maxChanges parameter is too low (say 256) then we can not return 5845839345345345 ids as it would exceed the client maxChanges. So we don't.

Today behaviour is to stay stuck (ouch!). Clients not aware of the bug will thus not be able to resync themselves.

With that changes, we return `I can not sync given your max changes`. The client can then resync with a higher max changes, or do a full resync...

> What is the point of having hasMoreChanges field and mechanic from the spec then if the chunk is too large we return an exception anyway?

It is for when you page updates (eg 4098 changes but I resync 256 at a time - wich works if we do not have crazy large records).

> I'm trying to understand the logic behind this choice (and not about the technical limitations we might have from our code)

Fair.

Ideally we could "chunk changes" by batches of say 256 upon saving them in the changelog so that no changelog entry exceeds a decent maxChanges value.

Arsnael (approved):

PR #380, opened 2021-04-09
JAMES-3559 Allow to window push notifications

chibenwa:
## Why?

Today a JAMES action might trigger several events being dispatched on the event bus (eg moving 2 messages will end up firing 2 additions and 2 deletion) resulting in likely 4 StateChanges being pushed to the end user.

## How?

Using the reactor library, James can likely delay pushes for a given timewindow and merge the state changes together in order to supply the client with only one state change.

A likely time value might be 2 seconds.

 - Pros: avoid event storm and will lower re-synchronization request count (good for perf?)
 - Cons: delays of 2s for real time across devices.

We could of course make it configurable via a jmap.properties configuration.

```
# Optional. Omiting this property leads to windowing being disables. James will 
# then directly forward all StateChanges to the end user without attempting to 
# buffer and aggregating them.
# If specified, James will delay stateChanges for that given amout of time and will attempt 
# to aggregate subsequent state changes together before returning them to the client.
# Units: ms, s, min, h, d
push.aggregation.window.duration=2s
```

### DIFF
--- a/dockerfiles/run/guice/cassandra-ldap/destination/conf/jmap.properties
+++ b/dockerfiles/run/guice/cassandra-ldap/destination/conf/jmap.properties
@@ -16,3 +16,11 @@ jwt.publickeypem.url=file://conf/jwt_publickey
 # Should simple Email/query be resolved against a Cassandra projection, or should we resolve them against ElasticSearch?
 # This enables a higher resilience, but the projection needs to be correctly populated. False by default.
 # view.email.query.enabled=true
+
+# Optional. Omitting this property leads to windowing being disables. James will
+# then directly forward all StateChanges to the end user without attempting to
+# buffer and aggregating them.
+# If specified, James will delay stateChanges for that given amount of time and will attempt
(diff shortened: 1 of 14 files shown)


chibenwa:
I really want it for my deployments.

I don't mid marking it as experimental.

That would be a shame that I need to fork JMAP implem to get it...

I would like to make progress integrating this.

rouazana:
> I would like to make progress integrating this.

Can you answer my comment? https://issues.apache.org/jira/browse/JAMES-3559?focusedCommentId=17326375&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17326375

chibenwa:
> Can you answer my comment? https://issues.apache.org/jira/browse/JAMES-3559?focusedCommentId=17326375&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17326375

Sure. But a long term mitigation strategy would anyway not address my current needs.

Arsnael (approved):

PR #386, opened 2021-04-14
[Release] Fix SHA-512 download links

chibenwa:


chibenwa:
This is a trivial fix, and I need it as part of the release process hence the timely merge without feedback.

PR #399, opened 2021-04-22
JAMES-3261 Package memory docker images with JIB

chibenwa:
I took time to port what we have regarding James packaging with James. Includes:

 - The CLI. Strategy is to add a runtime dependency and have a custom script.
 - Glowroot is bundled. Downloaded by a maven plugin, and combined with its configuration files before being added to the image.
 - extra java options are supported via an environment variable
 
Let's review this one first, then I devote myself committing this for the other guice servers.

The image size is slightly bigger (573 MB vs 383MB for the 3.6.0 memory docker image), it might be worth it to investigate where the extra size is coming from.

I choose buildTar as a goal, but in the end dockerBuild goal might enable us  to run tests on top of the James docker images (what we used to call 'deployment tests' to check image validity, as testing dependencies could (had) been hiding problems). Not docker users could then use `-Djib.skip` to not be impacted by this possible choice.


Arsnael (commented):
Looks good so far

chibenwa:
I removed the no review commits...

jeantil (commented):
nice work, thanks a lot. Once this is done for all the assemblies, it will make moving the modules around much easier I guess... 

one note : server/container/guice/memory-guice/src/main/extensions-jars/README.md is empty, is it on purpose ?

Arsnael (approved):

chibenwa:
We are not in a hurry.

Let's give a chance for people to play with the new packaging and propose enhancements.

Let's wait to see if further reviews could be attempted.

chibenwa:
I will likely implement JIB stuff for other servers this weekend.

If by then there is no other suggestions/remarks, I will merge this work.

chibenwa:
Rebased so that I can merge this - I'll wait the merge result.

PR #400, opened 2021-04-22
JAMES-3567 Distributed server should not depend on ActiveMQ

chibenwa:
Not required, this brings in the JGroup dependency, vulnerable prior version 4.0.

This is a simple dependency cleanup.

### DIFF
--- a/server/container/guice/cassandra-rabbitmq-guice/pom.xml
+++ b/server/container/guice/cassandra-rabbitmq-guice/pom.xml
@@ -81,6 +81,11 @@
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
+            <artifactId>apache-mailet-icalendar</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
             <artifactId>blob-s3</artifactId>
             <type>test-jar</type>
             <scope>test</scope>
@@ -113,10 +118,6 @@
         <dependency>
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-cassandra-guice</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>${james.groupId}</groupId>
-            <artifactId>james-server-cassandra-guice</artifactId>
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
@@ -132,11 +133,19 @@
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-guice-data-ldap</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
             <artifactId>james-server-guice-distributed</artifactId>
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-guice-elasticsearch</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-guice-es-resporter</artifactId>
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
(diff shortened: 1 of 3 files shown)


server/container/guice/cassandra-rabbitmq-guice/pom.xml
```diff
@@ -110,10 +115,6 @@
             <artifactId>event-sourcing-event-store-memory</artifactId>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>${james.groupId}</groupId>
```
jeantil, line 114: If I understand correctly `jame-server-cassandra-guice` is in fact `jame-server-cassandra-activemq-guice` ?

server/container/guice/cassandra-rabbitmq-guice/pom.xml
```diff
@@ -156,6 +165,26 @@
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-guice-jmx</artifactId>
         </dependency>
+        <dependency>
```
jeantil, line 168: I feel there must be something wrong in the way we assemble servers. 

Shouldn't these dependencies be in a common guice server definition ? (by these dependencies I mean the ones that are not specific to a particular storage technology. 

I had a look and `james-server-guice-common` feels like a good candidate but for some reason it pulls in inmemory implementations of some components  such as `james-server-task-memory` `james-server-mailrepository-memory`

server/container/guice/cassandra-rabbitmq-guice/pom.xml
```diff
@@ -168,12 +197,48 @@
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-guice-webadmin</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-guice-webadmin-data</artifactId>
```
jeantil, line 202: james-server-guice-webmin-common ? I'm not saying to change it for this PR but I think it would be interesting.

server/container/guice/cassandra-rabbitmq-guice/pom.xml
chibenwa: Yes. A guice server bundling Cassandra, ES, ActiveMQ.

We suffer from having no clearer separation between our guice modules and our apps.

I would be in favor of something like this:

```
server/apps/spring
server/apps/memory
server/apps/distributed
server/apps/jpa
server/apps/jpa-smtp
server/apps/cassandra //-es-activemq
server/containers/guice/... // all module definitions would go here
```

server/container/guice/cassandra-rabbitmq-guice/pom.xml
chibenwa: >Shouldn't these dependencies be in a common guice server definition ? (by these dependencies I mean the ones that are not specific to a particular storage technology.

I could decide to build a server without a given protocol (say JMAP, IMAP, POP3) for instance, to build a pure MTA.

> I had a look and james-server-guice-common feels like a good candidate but for some reason it pulls in inmemory implementations of some components such as james-server-task-memory james-server-mailrepository-memory

Yes some cleanup still to be done, to be honnest.

We might have considered by the past that inMemory dependencies were light with no extra external dependencies thus having them sneaking in was acceptable.

server/container/guice/cassandra-rabbitmq-guice/pom.xml
chibenwa: Again one might compose only some subparts of the webAdmin. An example of this is jpa-smtp that drops all routes related to the mailbox, to JMAP, etc...

chibenwa:
@jeantil would you agree with this work "as is"?

jeantil:
yes, go ahead. sorry for the huge latency these days

Arsnael (approved):

chibenwa:
> yes, go ahead. sorry for the huge latency these days

No problem ;-)

PR #410, opened 2021-04-30
JAMES-3571 MimeMessageWrapper getSize was incorrect for empty messages

chibenwa:
- Added the edge case tests
 - If the header delimiter is missing, then the size would be negative, then the body is as a zero size.
 
 Master PR: https://github.com/apache/james-project/pull/407


chibenwa:
:green_apple: , on master already, thus I merge this straight away ;-)

PR #413, opened 2021-05-01
JAMES-3575 End to end reactive chain for Email update and JMAP RFC-8621

chibenwa:


chibenwa:
## Before

Screenshot from 2021-05-03 20-08-15

## After

Screenshot from 2021-05-03 20-09-42

Arsnael (approved):

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/EmailSubmissionSetMethod.scala
```diff
@@ -199,30 +199,38 @@ class EmailSubmissionSetMethod @Inject()(serializer: EmailSubmissionSetSerialize
...
+            case (creationResults, processingContext) =>
+              createSubmission(session, emailSubmissionCreationId, jsObject, processingContext)
+                .map {
+                  case (created, updatedProcessingContext) => CreationResults(creationResults.created :+ created) -> updatedProcessingContext
+                }
+              .switchIfEmpty(SMono.error(new RuntimeException("I should not be empty")))
+          }.cache()
         }
       }
+      .flatMap(x => x)
```
LanKhuat, line 219: x?

chibenwa: Yes `_` misbehaved in my IDE, this is clearly an identity function so I don't think the name matter.

LanKhuat (approved):

PR #417, opened 2021-05-05
JAMES-3578 Drop Cassandra schema support prior version 8

chibenwa:


LanKhuat (approved):

rouazana:
I ask it also here: can we have some documentation to reference the migration path. I mean how to get from any old version to the last version.

chibenwa:
Hello Raphael.

I don't understand what you ask for. We already document the upgrade path in the upgrade instructions. 

What is ambiguous in the existing doc?

Aren't people upgrading from say 3.4.0 read all the upgrade instrctions between their version and the target one?

Can we be more clear here?

rouazana:
Let's take an example. I want to migrate from 3.3.0 to 3.8.0.

For that, I need to read upgrade notes, but I probably need to only install 3.6.0 and 3.8.0.

Before this change, I would have need only to install 3.8.0. Is it clearer?

chibenwa:
I got the feeling the upgrade notes already contain that information.

What form would you give it then?

rouazana:
Ideally an array

Version to install for migration:

From/To | 3.3.0 | 3.4.0 | 3.5.0 | 3.6.0 | 3.7.0 | 3.8.0
------------|--------|---------|--------|--------|---------|--------
3.3.0 | / | 3.4.0 | 3.5.0 | 3.6.0 | 3.7.0 | 3.7.0,3.8.0
... |
3.8.0| / | / | / | / | / | /

chibenwa:
Ok, let's do that array in another PR.

Because I have several remarks:

 - It makes sense to have it per server. Eg JPA do not have the same upgrade path than Distributed...
 - Also I do not understand the meaning of the versions in the cells.

rouazana:
> It makes sense to have it per server. Eg JPA do not have the same upgrade path than Distributed...

Right.

> Also I do not understand the meaning of the versions in the cells.

The list of version you need to install to migration from A to B.

Here to migration from 3.3.0 to 3.8.0 you need to install 3.7.0 and 3.8.0. Maybe specifying the last version is useless.

You can also take inspiration from here for example: https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-upgrade.html if it is more clear for you.

docs/modules/servers/pages/distributed/operate/index.adoc
```diff
@@ -24,4 +24,38 @@ distributed James operators.
 
 We also included a guide for xref:distributed/operate/migrating.adoc[migrating existing data] into the distributed server.
 
-Read more about xref:distributed/operate/cassandra-migration.adoc[Cassandra data migration].
\ No newline at end of file
+Read more about xref:distributed/operate/cassandra-migration.adoc[Cassandra data migration].
+
+== Upgrading
+
+This section sum up and links upgrade instructions for people willing to upgrade their Distributed server to newer versions.
```
Arsnael, line 31:
```suggestion
This section sums up and links upgrade instructions for people willing to upgrade their Distributed server to newer versions.
```

docs/modules/servers/pages/distributed/operate/index.adoc
```diff
@@ -24,4 +24,38 @@ distributed James operators.
...
+
+.dnsservice.xml content
```
Arsnael, line 33: That's your table description? dnsservice.xml?

docs/modules/servers/pages/distributed/operate/cassandra-migration.adoc
```diff
@@ -17,13 +17,8 @@ Note that currently the progress can be tracked by logs.
 
 Here are the implemented migrations:
 
-* *From V5 to V6* : Goal is to no longer rely on an UDT partition key for mailboxPath tables. Entries will be migrated
-to mailboxPathV2 table relying on a composite primary key.
-* *From V6 to V7* : Goal is to populate mapping_sources projection table. This table allows finding the source of a given redirection, which is
-handy for things like mail aliases (I want to list aliases rewritting things to bob). Without this projection table being available,
-(ie we rely on schema version 6 or less) such information is obtained through a full table scan, unoptimized. From schema version 7,
-the optimized projection can safely be used.
-* *From V7 to V8* : Add UID_VALIDITY to mailboxPath table in order not to mandate mailbox table reads.
```
Arsnael, line 26: Hmmm did we ever test versioning in our antora doc yet? That could be a good candidate to start :)

upgrade-instructions.md
```diff
@@ -14,6 +14,23 @@ Note: this section is in progress. It will be updated during all the development
...
+Date 05/05/2021
+
+JIRA: https://issues.apache.org/jira/browse/JAMES-3578
+
+Concerned products: Distributed James, Cassandra-guice James server
+
+In an effort to simplify the code base, we dropped support for Cassandra schema version prior version 8.
+
+Installation running older schema version than version 5 needs to rely on releases 3.6.x to upgrade to schema version 8,
+before upgrading to an eventual newer version.
```
Arsnael, line 32: ah actually you mention it here :) But I guess we need to mention it in the Antora doc too, probably safer...

docs/modules/servers/pages/distributed/operate/cassandra-migration.adoc
chibenwa: :+1: but we only have the 3.6.0 tag yet

docs/modules/servers/pages/distributed/operate/index.adoc
chibenwa, line 33:
```suggestion
.Upgrade path
```

upgrade-instructions.md
chibenwa: I did not wanted to mention unreleased stuff in the Antora documentation just yet.

docs/modules/servers/pages/distributed/operate/cassandra-migration.adoc
Arsnael: but that change would be for the 3.7.0 right?

chibenwa: Yes.

chibenwa:
@rouazana @Arsnael satisfied by the last version of the doc?

Arsnael (approved):

docs/modules/servers/pages/distributed/operate/index.adoc
```diff
@@ -24,4 +24,47 @@ distributed James operators.
...
+| 3.5.0
+| 1. Upgrade, and apply link:https://github.com/apache/james-project/blob/master/upgrade-instructions.md#360-version[3.6.0 upgrade instructions]
+
+Note: Because of ElasticSearch upgrade (6.3 -> 7.10) this migration cannot be done in a rolling fashion manner.
+
+2. Rolling upgrade to 3.7.0 (unreleased). No specific upgrade instructions to perform.
+
+
+| 3.4.0
+| 1. Rolling upgrade to 3.5.0, and apply link:https://github.com/apache/james-project/blob/master/upgrade-instructions.md#350-version[3.6.0 upgrade instructions]
```
rouazana, line 49:
```suggestion
| 1. Rolling upgrade to 3.5.0, and apply link:https://github.com/apache/james-project/blob/master/upgrade-instructions.md#350-version[3.5.0 upgrade instructions]
```

docs/modules/servers/pages/distributed/operate/index.adoc
```diff
@@ -24,4 +24,47 @@ distributed James operators.
...
+
+2. Upgrade to 3.6.0, and apply link:https://github.com/apache/james-project/blob/master/upgrade-instructions.md#360-version[3.5.0 upgrade instructions]
```
rouazana, line 51:
```suggestion
2. Upgrade to 3.6.0, and apply link:https://github.com/apache/james-project/blob/master/upgrade-instructions.md#360-version[3.6.0 upgrade instructions]
```

rouazana (commented):
Thanks, that's better. Though I don't understand why you need so many upgrade steps.

PR #420, opened 2021-05-06
[REFACTORING] JMAP Avoid MIME re-parsing when sending messages

chibenwa:


chibenwa:
## Before

Screenshot from 2021-05-06 14-14-03

## After

Screenshot from 2021-05-06 14-14-49

Not game changing but nice bonus :-)

Arsnael (approved):

PR #422, opened 2021-05-07
JAMES-3491 Configurable websocket URL for JMAP configuration

LanKhuat:
Allow for more flexibility in advertising URLs in the session object.

### DIFF
--- a/server/protocols/jmap-rfc-8621-integration-tests/memory-jmap-rfc-8621-integration-tests/src/test/resources/jmap.properties
+++ b/server/protocols/jmap-rfc-8621-integration-tests/memory-jmap-rfc-8621-integration-tests/src/test/resources/jmap.properties
@@ -1,3 +1,4 @@
 # Configuration urlPrefix for JMAP routes.
 url.prefix=http://domain.com
+websocket.url.prefix=ws://domain.com
 upload.max.size=20M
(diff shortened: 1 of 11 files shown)


server/protocols/jmap-rfc-8621-integration-tests/memory-jmap-rfc-8621-integration-tests/src/test/resources/jmap.properties
```diff
@@ -1,3 +1,5 @@
 # Configuration urlPrefix for JMAP routes.
-url.prefix=http://domain.com
+protocol.jmap=http
+protocol.ws=ws
+url.prefix=domain.com
```
chibenwa, line 4: That is a breaking change. And breaking changes are bad.

I would propose:

```
url.prefix=http://domain.com
websocket.url.prefix=ws://domain.com
```

That way if you don't change your config, you will not end up with an unwanted behaviour.

Bonus: One could set up proxying to a dedicated websocket server, and advertise that through the session, using a completely distinct load balancer... <3

chibenwa (approved):

chibenwa:
@LanKhuat can you squash this?

chibenwa:
I did squash it myself ;-)

PR #423, opened 2021-05-10
JAMES-XXXX Handle JMAP WebSocket ping

LanKhuat:


server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/core/JmapRfc8621Configuration.scala
```diff
@@ -25,28 +25,33 @@ import org.apache.commons.configuration2.Configuration
...
 
+import scala.concurrent.duration.Duration
+
 object JmapRfc8621Configuration {
   val LOCALHOST_URL_PREFIX: String = "http://localhost"
   val UPLOAD_LIMIT_30_MB: MaxSizeUpload = MaxSizeUpload.of(Size.of(30L, Size.Unit.M)).get
-  val LOCALHOST_CONFIGURATION: JmapRfc8621Configuration = JmapRfc8621Configuration(LOCALHOST_URL_PREFIX, UPLOAD_LIMIT_30_MB)
+  val LOCALHOST_CONFIGURATION: JmapRfc8621Configuration = JmapRfc8621Configuration(LOCALHOST_URL_PREFIX, None, UPLOAD_LIMIT_30_MB)
   val URL_PREFIX_PROPERTIES: String = "url.prefix"
+  val WEBSOCKET_INTERVAL_PROPERTIES: String = "websocket.ping.interval"
```
chibenwa, line 35: We need to document this option in `docs/modules/servers/pages/distributed/configure/jmap.adoc` and in `src/site/xdoc/server/config-jmap.xml`

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/core/JmapRfc8621Configuration.scala
```diff
@@ -25,28 +25,33 @@ import org.apache.commons.configuration2.Configuration
...
   val UPLOAD_LIMIT_PROPERTIES: String = "upload.max.size"
 
   def from(configuration: Configuration): JmapRfc8621Configuration = {
     JmapRfc8621Configuration(
       urlPrefixString = Option(configuration.getString(URL_PREFIX_PROPERTIES)).getOrElse(LOCALHOST_URL_PREFIX),
+      pingInterval = Option(configuration.getString(WEBSOCKET_INTERVAL_PROPERTIES)).map(Duration.apply),
```
chibenwa, line 41: We have a `DurationParser`

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/WebSocketRoutes.scala
```diff
@@ -110,10 +111,17 @@ class WebSocketRoutes @Inject() (@Named(InjectionKeys.RFC_8621) val authenticato
...
 
-    out.sendString(
+    val applicativeMessages: SFlux[OutboundMessage] =
       SFlux.merge(Seq(responseFlux, sink.asFlux()))
-        .map(ResponseSerializer.serialize)
-        .map(Json.stringify))
+
+    val pings: SFlux[OutboundMessage] = configuration.pingInterval
+      .map(any => SFlux.interval(any)
+        .map(_ => PingMessage(???)))
```
chibenwa, line 119: We can use the value 'long' generated by the ping interval to generate the ping payload

chibenwa:
We need to check pong responses, and ensures that we receive one push response every `2 x` the ping interval.

chibenwa:
> We need to check pong responses, and ensures that we receive one push response every 2 x the ping interval.

I wonder if that is necessary. With TCP if the connection is closed, we would get an error while sending the ping to the client, which should be enough for detection (unless the client is a kinda zombie accepting TCP connections but nothing else).

This would likely be acceptable as "eventSource" send solely a ping, without feedback from the client.

chibenwa (commented):
No tests?

LanKhuat:
> No tests?

Still fail to compile. i'm working on it.

rouazana:
> I wonder if that is necessary. With TCP if the connection is closed, we would get an error while sending the ping to the client, which should be enough for detection (unless the client is a kinda zombie accepting TCP connections but nothing else).

It's a common case that middle network equipment drop connections instead of closing them. In this case, you won't get noticed unless you set correctly your timeouts (with is never the default unfortunately).

There are several ways to reproduce this behavior if you want. Here are some examples:
- configure iptables to drop tcp packets
- contact a server (humm in your case it a client) and pause it with Ctrl-Z

chibenwa:
Then I am very curious to know how it would be different from https://jmap.io/spec-core.html#event-source

With the keepalive being implemented with an uni-directional ping...

server/protocols/jmap-rfc-8621-integration-tests/distributed-jmap-rfc-8621-integration-tests/src/test/resources/jmap.properties
```diff
@@ -1,4 +1,4 @@
 # Configuration urlPrefix for JMAP routes.
 url.prefix=http://domain.com
-websocket.ping.interval=20s
+websocket.ping.interval=2s
```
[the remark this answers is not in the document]
LanKhuat: i'm afraid that a interval too short would affect other tests... will try anyway

Arsnael:
What is left to do here except maybe testing with Gatling? 

I reverted the last commit. Maybe on the CI it was fine but locally for the distributed tests some were unstable, regarding the way some tests are being built. Like we would sometimes expect 3 responses on the websocket after sending some requests threw it, but if your ping interval is low enough, you would get some ping responses mixed instead, which fails the assert of the test.

chibenwa:
> What is left to do here except maybe testing with Gatling?

Ensure that we receive pong responses to our pings. If we do not receive a pong during too ping intervals then we can close the connection.

Arsnael:
I have been thinking again... Do we need at all this PR? It feels like reactor netty is managing that out of the box somehow, if you send it a ping, it answers a pong... And when I tested with websocat, I had that issue of having a closing connection. But when I added the ping option I think it was not closing anymore... 

I will check again tomorrow but I think it works that way... Regarding Gatling though, could not find a way to send a ping... But Didn't have much time to look, will dig more tomorrow.

chibenwa:
> I have been thinking again... Do we need at all this PR?

We need it to support clients that do not send websocket PINGs. This is the case EG of Gatling during performance tests. Now it this a real need? Or a corner case we can ignore because all real clients maintains PING? The question seems open to me.

> It feels like reactor netty is managing that out of the box somehow, if you send it a ping, it answers a pong...

The problem is when the client do not send PING. Eg Gatling.

> I will check again tomorrow but I think it works that way... Regarding Gatling though, could not find a way to send a ping... But Didn't have much time to look, will dig more tomorrow.

Maybe we could send a dummy echo method ?

---------

Speaking with @tuanlc Javascript clients typically generate ping messages

He had been solving similar problems with Gatling, he investigates :-)

Arsnael:
Sweet to have Tuan's expertice for the client side :)

Also for gatling yeah there seems no mechanism of ping/pong (but then can we send a pong from gatling if we send it a ping? it feels like quite complicated to do and manage I think potentially). But there is a method `wsReconnect` though, that will reconnect everytime the connection drops. Not sure how it works, maybe can do the trick for Gatling?

chibenwa:
> But there is a method wsReconnect though, that will reconnect everytime the connection drops. Not sure how it works, maybe can do the trick for Gatling?

You can try yes.

Otherwise just do JMAP echoes on a 20 second basis?

Arsnael:
So will close this PR after discussions. The rational being:
- the RFC does not really states who is responsible for sending the ping
- the trend nowadays seems to be clients send a ping and server answers with a pong. And Reactor netty does that already
- we should adapt the gatling test to send regularly echo jmap methods (as Gatling doesn't seem to manage that mechanism), that could act as a ping/pong mechanism to keep the connection alive, in case infra running James has some limitations in terms of idle timeout

PR #425, opened 2021-05-11
JAMES-3579 verifyIdentity param should be rejected if authRequired is set to false in SMTP configuration

Arsnael:


chibenwa (approved):
Let see now what breaks!

LanKhuat (approved):

Arsnael:
Squashed commits

chibenwa (approved):

PR #432, opened 2021-05-15
[FIX] Solve build issue (TaskManagerContract::awaitShouldAwaitWaiting…

chibenwa:
…Task)

These subscribers are not tolerant to running on the elastic Scheduler.

I am reverting this piece of code to the previous behaviour.


chibenwa:
Merging this to fix the build.

I found the explanation: The explanation was that even in flatMap was used, on a "immediate" scheduler calls were processed sequencially.

Adding the "elastic" breaks that ordering logic that needs to be restored via concat map.

PR #435, opened 2021-05-16
[PERFORMANCE] Various small enhancements detected thanks to flame graphs

chibenwa:
Screenshot from 2021-05-16 17-45-01

Here we can see a call to `MessageResultImpl::getMessageId` responsible of `0.78%` of the total allocated memory... Rationals are that and intermediate (uneeded) on the fly messageMetadata is used, which includes an (innefficient) flag copy.

Screenshot from 2021-05-16 17-51-05

Here is a flame graph capture showing the costs of initializing JMAP routes for each requests.

### DIFF
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/FlagsFactory.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/FlagsFactory.java
@@ -30,8 +30,7 @@ import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableList;
 
 public class FlagsFactory {
-
-    private static Flags asFlags(MailboxMessage mailboxMessage, String[] userFlags) {
+    public static Flags createFlags(MailboxMessage mailboxMessage, String[] userFlags) {
         final Flags flags = new Flags();
         if (mailboxMessage.isAnswered()) {
             flags.add(Flags.Flag.ANSWERED);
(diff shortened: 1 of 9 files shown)


chibenwa:
Screenshot from 2021-05-16 22-41-50

0.26% CPU resources on formatting a validation message...

chibenwa:
Screenshot from 2021-05-17 07-46-15

0.56% running an inefficient algorithm to state which mailboxes have kids, in O(n2). While the percentage is low acting on it with likely improve p99 for `Mailbox/get` with many mailboxes and is thus worth writing.

Arsnael (approved):

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/FlagsFactory.java
```diff
@@ -60,10 +60,7 @@ private static Flags asFlags(MailboxMessage mailboxMessage, String[] userFlags)
     }
 
     public static Flags createFlags(MailboxMessage mailboxMessage, String[] userFlags) {
-        return builder()
-            .flags(asFlags(mailboxMessage, userFlags))
-            .addUserFlags(userFlags)
-            .build();
+        return asFlags(mailboxMessage, userFlags);
```
rouazana, line 63: why not removing createFlags and replace it by asFlags then?

Arsnael (approved):

chibenwa:
I confirm the last fixups regarding "computing metadata" do not cost performance.

Arsnael (approved):

jeantil (approved):

PR #436, opened 2021-05-17
JAMES-3586 CL one option for the Cassandra blob store

Arsnael:
Just a beginning, many things left to do...

### DIFF
--- a/server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
+++ b/server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
@@ -166,6 +166,23 @@ public class CassandraBlobStoreDAO implements BlobStoreDAO {
     }
 
     private Mono<ByteBuffer> readPart(BucketName bucketName, BlobId blobId, Integer partIndex) {
+        if (configuration.isOptimisticConsistencyLevel()) {
+            return readPartClOne(bucketName, blobId, partIndex)
+                .switchIfEmpty(readPartClDefault(bucketName, blobId, partIndex));
+        } else {
+            return readPartClDefault(bucketName, blobId, partIndex);
+        }
+    }
+
+    private Mono<ByteBuffer> readPartClOne(BucketName bucketName, BlobId blobId, Integer partIndex) {
+        if (isDefaultBucket(bucketName)) {
+            return defaultBucketDAO.readPartClOne(blobId, partIndex);
+        } else {
+            return bucketDAO.readPartClOne(bucketName, blobId, partIndex);
+        }
+    }
+
+    private Mono<ByteBuffer> readPartClDefault(BucketName bucketName, BlobId blobId, Integer partIndex) {
         if (isDefaultBucket(bucketName)) {
             return defaultBucketDAO.readPart(blobId, partIndex);
         } else {
@@ -174,6 +191,23 @@ public class CassandraBlobStoreDAO implements BlobStoreDAO {
     }
 
     private Mono<Integer> selectRowCount(BucketName bucketName, BlobId blobId) {
+        if (configuration.isOptimisticConsistencyLevel()) {
+            return selectRowCountClOne(bucketName, blobId)
+                .switchIfEmpty(selectRowCountClDefault(bucketName, blobId));
+        } else {
+            return selectRowCountClDefault(bucketName, blobId);
+        }
+    }
+
+    private Mono<Integer> selectRowCountClOne(BucketName bucketName, BlobId blobId) {
+        if (isDefaultBucket(bucketName)) {
+            return defaultBucketDAO.selectRowCountClOne(blobId);
+        } else {
+            return bucketDAO.selectRowCountClOne(bucketName, blobId);
+        }
+    }
+
+    private Mono<Integer> selectRowCountClDefault(BucketName bucketName, BlobId blobId) {
         if (isDefaultBucket(bucketName)) {
             return defaultBucketDAO.selectRowCount(blobId);
         } else {
--- a/server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
+++ b/server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
@@ -37,6 +37,8 @@ import org.apache.james.blob.api.BlobStoreDAO;
 import org.apache.james.blob.api.BucketName;
 import org.apache.james.blob.api.ObjectNotFoundException;
 import org.apache.james.blob.api.ObjectStoreIOException;
+import org.apache.james.metrics.api.Metric;
+import org.apache.james.metrics.api.MetricFactory;
 import org.apache.james.util.DataChunker;
 import org.apache.james.util.ReactorUtils;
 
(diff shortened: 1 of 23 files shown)


server/blob/blob-cassandra/pom.xml
```diff
@@ -65,6 +65,10 @@
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-util</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>metrics-logger</artifactId>
```
chibenwa, line 70: Why? MetricFactory API is not enough?

server/blob/blob-cassandra/src/test/java/org/apache/james/blob/cassandra/CassandraBlobStoreClOneTest.java
```diff
@@ -0,0 +1,216 @@
+package org.apache.james.blob.cassandra;
```
chibenwa, line 1: Oups

server/blob/blob-cassandra/src/test/java/org/apache/james/blob/cassandra/CassandraBlobStoreClOneTest.java
```diff
@@ -0,0 +1,216 @@
...
+        int repeatCount = MULTIPLE_CHUNK_SIZE * CHUNK_SIZE;
+        String longString = Strings.repeat("0123456789\n", repeatCount);
+        BlobId blobId = Mono.from(testee().save(testee().getDefaultBucketName(), longString, LOW_COST)).block();
+
+        when(defaultBucketDAO().readPartClOne(blobId, 1)).thenReturn(Mono.empty());
+        when(defaultBucketDAO().readPart(blobId, 1)).thenReturn(Mono.empty());
+
+        assertThatThrownBy(() -> IOUtils.toString(testee().read(testee().getDefaultBucketName(), blobId), StandardCharsets.UTF_8))
+            .isInstanceOf(ObjectStoreException.class)
+            .hasMessageContaining("Missing blob part for blobId");
```
chibenwa, line 89: No then we should fallback to CL QUORUM...

server/blob/blob-cassandra/src/test/java/org/apache/james/blob/cassandra/CassandraBlobStoreClOneTest.java
```diff
@@ -0,0 +1,216 @@
...
+        int repeatCount = MULTIPLE_CHUNK_SIZE * CHUNK_SIZE;
+        String longString = Strings.repeat("0123456789\n", repeatCount);
+        BlobId blobId = Mono.from(testee().save(testee().getDefaultBucketName(), longString, LOW_COST)).block();
+
+        when(defaultBucketDAO().readPartClOne(blobId, 1)).thenReturn(Mono.empty());
+        when(defaultBucketDAO().readPart(blobId, 1)).thenReturn(Mono.empty());
+
+        assertThatThrownBy(() -> Mono.from(testee().readBytes(testee().getDefaultBucketName(), blobId)).block())
+            .isInstanceOf(ObjectStoreException.class)
+            .hasMessageContaining("Missing blob part for blobId");
```
chibenwa, line 104: No then we should fallback to CL QUORUM...

server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
```diff
@@ -166,6 +166,23 @@ private boolean isDefaultBucket(BucketName bucketName) {
     }
 
     private Mono<ByteBuffer> readPart(BucketName bucketName, BlobId blobId, Integer partIndex) {
+        if (configuration.isOptimisticConsistencyLevel()) {
+            return readPartClOne(bucketName, blobId, partIndex)
```
Arsnael, line 184: Was not sure if we need to apply it on the read for readPart too? But my guess is yes? Just want to be sure :)

chibenwa (commented):
You miss...

 - [ ] Falback mechanisms to QUORUM to avoid all inconsistencies
 - [ ] Documentation of the config option

server/blob/blob-cassandra/pom.xml
Arsnael, line 70: I don't like that... at all... but I had to add something for CassandraBlobStoreFactory that is not in the test package and is being used by other test classes in other modules... Not sure what would be the best way here... or add the metrics as a param in forTesting method?

server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
```diff
@@ -168,7 +182,9 @@ private boolean isDefaultBucket(BucketName bucketName) {
     private Mono<ByteBuffer> readPart(BucketName bucketName, BlobId blobId, Integer partIndex) {
         if (configuration.isOptimisticConsistencyLevel()) {
             return readPartClOne(bucketName, blobId, partIndex)
-                .switchIfEmpty(readPartClDefault(bucketName, blobId, partIndex));
+                .doOnNext(any -> metricClOneHitCount.increment())
```
Arsnael, line 185: Should we use the same metric for hit and miss between readBlob and readPart?

server/blob/blob-cassandra/pom.xml
Arsnael: https://github.com/apache/james-project/pull/436#discussion_r634214463

server/blob/blob-cassandra/src/test/java/org/apache/james/blob/cassandra/CassandraBlobStoreClOneTest.java
Arsnael: Read above the when and you will see that I miss with quorum too. It's an override just to keep the test behavior and make it pass. The test you are talking about is below

server/blob/blob-cassandra/src/test/java/org/apache/james/blob/cassandra/CassandraBlobStoreClOneTest.java
Arsnael: idem as above

Arsnael:
> Falback mechanisms to QUORUM to avoid all inconsistencies

Are you sure you didn't review a bit fast???

server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
chibenwa: ReadPart is the most critical.

server/blob/blob-cassandra/pom.xml
chibenwa: How about moving the forTesting stuff in tests, and fix imports of tests that depends on it?

> or add the metrics as a param in forTesting method?

Good solution and likely easier so :+1:

server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
chibenwa: IMO yes.

chibenwa (approved):

Arsnael:
Build was green. Did a rebase and squashed commits

PR #438, opened 2021-05-18
[PERFORMACE] Apply 'Things You Should Do Anyway' section of Jackson performance guide

chibenwa:
https://github.com/FasterXML/jackson-docs/wiki/Presentation:-Jackson-Performance

Performance gains are currently being evaluated.


chibenwa:
Sad but no performance improvments.

Better filtering management in JmapResponseWriterImpl would be beneficial.

PR #443, opened 2021-05-20
JAMES-3588 MailetContainer error handling: allow error propagation

chibenwa:
Base use case: Executing the mailet container with LMTP I want to escalate some failures to the LMTP client

Nice bonus: It enables better error management eg by propagating exception upon mail processor we can trigger a nack and prevent message loss.

### DIFF
--- a/server/protocols/protocols-lmtp/src/main/java/org/apache/james/lmtpserver/MailetContainerHandler.java
+++ b/server/protocols/protocols-lmtp/src/main/java/org/apache/james/lmtpserver/MailetContainerHandler.java
@@ -20,9 +20,11 @@
 package org.apache.james.lmtpserver;
 
 import java.util.Collection;
+import java.util.Objects;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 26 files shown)


server/protocols/protocols-lmtp/src/main/java/org/apache/james/lmtpserver/MailetContainerHandler.java
```diff
@@ -81,4 +130,28 @@ protected Response processExtensions(SMTPSession session, Mail mail) {
...
+    private LMTPMultiResponse executeEachRecipientSeparately(Mail mail, Collection<MailAddress> recipients) {
+        return LMTPMultiResponse.of(
+            recipients.stream()
+                .map(recipient -> executeFor(mail, recipient))
+                .collect(Guavate.toImmutableList()));
+    }
+
+    private SMTPResponse executeFor(Mail mail, MailAddress recipient) {
+        try {
+             Mail newMail = mail.duplicate();
```
Arsnael, line 143: indent

Arsnael (approved):

PR #445, opened 2021-05-21
JAMES-3589 Matched mail resume processing from root

chibenwa:
## What

One of my customer reported me that a side effect was done two time upon MailetContainer execution.

What was not my surprise when writing integration tests counting executions, that they were right!

Tracking down the bug, I encountered that `MailImpl.duplicate` do not preserve state, hence processing resumes from ROOT processor (leaving the exchange).

## The fix

MatcherSplitter should preserve the state of the matched mail.

## The complications

Preserving several mails on the same exchange, we should not shut the exchange down before processing is finished. Doing so would result in following  I did set up reference counting for this.


Arsnael (approved):

chibenwa:
TODO rebase this on https://github.com/apache/james-project/pull/447

chibenwa:
I'm closing this.

https://github.com/apache/james-project/pull/448 should be considered the most serious mitigation to this issue.

PR #446, opened 2021-05-21
JAMES-3516 Add threadId property to MessageResult POJO, MailboxMessage POJO

quantranhong1999:
We don't have threadId guessing algorithm yet, so now threadId property is naive implemented equals MessageId.
I think I should have put threadId property to some Implemented MailboxMessage constructor/ builder. However I not sure which Message model JMAP is using...

### DIFF
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/model/ThreadId.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/model/ThreadId.java
@@ -1,0 +1,58 @@
+/******************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one     *
+ * or more contributor license agreements.  See the NOTICE file   *
+ * distributed with this work for additional information          *
+ * regarding copyright ownership.  The ASF licenses this file     *
+ * to you under the Apache License, Version 2.0 (the              *
+ * "License"); you may not use this file except in compliance     *
+ * with the License.  You may obtain a copy of the License at     *
+ *                                                                *
+ * http://www.apache.org/licenses/LICENSE-2.0                     *
+ *                                                                *
+ * Unless required by applicable law or agreed to in writing,     *
+ * software distributed under the License is distributed on an    *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY         *
+ * KIND, either express or implied.  See the License for the      *
+ * specific language governing permissions and limitations        *
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.model;
+
+import java.util.Objects;
+
+import com.google.common.base.MoreObjects;
+
+
+public class ThreadId {
+    private final MessageId baseMessageId;
+
+    public ThreadId(MessageId baseMessageId) {
+        this.baseMessageId = baseMessageId;
+    }
+
+    public MessageId getBaseMessageId() {
+        return baseMessageId;
+    }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (o instanceof ThreadId) {
+            ThreadId that = (ThreadId) o;
+            return Objects.equals(this.baseMessageId, that.baseMessageId);
+        }
+        return false;
+    }
+
+    @Override
+    public final int hashCode() {
+        return Objects.hashCode(baseMessageId);
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+            .add("baseMessageId", baseMessageId)
+            .toString();
+    }
+}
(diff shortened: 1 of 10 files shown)


mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/DelegatingMailboxMessage.java
```diff
@@ -112,6 +113,10 @@ public MessageId getMessageId() {
         return message.getMessageId();
     }
 
+    public ThreadId getThreadId() {
+        return (ThreadId) message.getMessageId();
```
Arsnael, line 117: I would not be surprised that this breaks if you try to really access it... Did you see the implementations of MessageId for memory and Cassandra?

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreMessageResultIterator.java
```diff
@@ -191,6 +192,11 @@ public MessageId getMessageId() {
             return messageMetaData().getMessageId();
         }
 
+        @Override
+        public ThreadId getThreadId() {
+            return (ThreadId) messageMetaData.getMessageId();
```
Arsnael, line 197: idem

mailbox/api/src/main/java/org/apache/james/mailbox/model/ThreadId.java
```diff
@@ -0,0 +1,62 @@
...
+ * specific language governing permissions and limitations        *
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.model;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Objects;
+
+public class ThreadId implements MessageId {
```
Arsnael, line 25: I'm not convinced about this... I will think about it.

Arsnael:
Your ThreadId is not equal to MessageId here. You just "implement" the MessageId interface in a complete different way that the mem and Cassandra implems. I' prettu sure if you write some quick tests and try to access it with your cast, it breaks? 

I don't think ThreadId should be implemented this way... Not sure what would be the best approach here. Maybe I would have just duplicated the MessageId code and renamed it ThreadId? And add a method to make ThreadId and MessageId conversions... but I'm not sure either...

@chibenwa have a better idea?

quantranhong1999:
Instead of
```
return (ThreadId) message.getMessageId();
```
Which I think it gonna fail
How about 
```
return new ThreadId(message.getMessageId().serialize());
```

Arsnael:
Sounds already better

mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/model/openjpa/AbstractJPAMailboxMessage.java
```diff
@@ -493,6 +494,11 @@ public MessageId getMessageId() {
         return new DefaultMessageId();
     }
 
+    @Override
+    public ThreadId getThreadId() {
+        return (ThreadId) getMessageId();
```
chibenwa, line 499: You then should built a ThreadId that wraps the messageId.

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/MailboxMessage.java
```diff
@@ -33,6 +34,8 @@
  */
 public interface MailboxMessage extends Message, Comparable<MailboxMessage> {
 
+    ThreadId getThreadId();
```
chibenwa, line 37: It may worth writing tests for this in SimpleMailboxMessageTest ?

mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageResultImpl.java
```diff
@@ -87,7 +88,12 @@ public MessageUid getUid() {
     public MessageId getMessageId() {
         return message.getMessageId();
     }
-    
+
+    @Override
+    public ThreadId getThreadId() {
+        return message.getThreadId();
```
chibenwa, line 94: Can we try in MailboxManagerTest to access the threadId of a message?

chibenwa (commented):
A good first step.

chibenwa:
`return new ThreadId(message.getMessageId().serialize());`

Underlying storing things as string is innefficient.

I would prefer `return new ThreadId(message.getMessageId());` as:

 - The first messageId seems like a good way to designate a thread
 - We can rely on the implementation-specific MessageId to avoid yet another generic POJO.
 
 (for me the alternative is having a generic `ThreadId` and a `CassandraThreadId`, `JpaThreadId`, `MemoryThreadId` but it seems pretty boiler plate to me....)

chibenwa (approved):
Looks good to me!

quantranhong1999:
I have a question: 
How should I know where I need to write the tests following the POJOs changes? 
Why MailboxManagerTest and SimpleMailboxMessageTest? I see many other tests related to MailboxMessage and MessageResult POJOs

chibenwa:
Good question.

IMO they are the most logical place. If I am looking to understand a behaviour I would go there first. Why not in other places: because we IMO do not want to duplicate things too much.

Arsnael (approved):

chibenwa:
Quan can you squash this branch so that we can merge it?

quantranhong1999:
> Quan can you squash this branch so that we can merge it?

I will do it

PR #448, opened 2021-05-22
JAMES-3589 Mailet processing without Camel

chibenwa:
Addresses issues of https://github.com/apache/james-project/pull/447

Special care had been taken to dispose splitted/duplicated emails.

The split away from Camel is partial - one might want to:

 - [ ] Get rid of the Camel dependency
 - [ ] Rename maven package, classes, etc...
 - [ ] Get rid of inheritance and favor composition
 
I opened it to get a full build and thus identify potential issues.


chibenwa:
Ok, got a :green_apple: build on this.

server/mailet/mailetcontainer-camel/src/main/java/org/apache/james/mailetcontainer/impl/camel/MatcherSplitter.java
```diff
@@ -155,6 +155,8 @@ public MatcherSplitter(MetricFactory metricFactory, CamelMailetProcessor contain
 
                     Mail newMail = MailImpl.duplicate(mail);
                     newMail.setRecipients(matchedRcpts);
+                    newMail.setState(mail.getState());
+                    mail.setRecipients(rcpts);
```
Arsnael, line 159: That's already set up few lines above (154). Any reason to do it again?

Arsnael (approved):

rouazana:
No problem for me for removing camel. Did you see some improvement in build or launch time by the way? As far as I remember it was slowing down James' startup.

chibenwa:
> Did you see some improvement in build or launch time by the way?

 - I saw a marginal improvments on the mailet runtime. Likely 1-5% according to the benchs async-profiler flame graphs.
 - JMAP RFC-8621 memory now runs sub-100 ms after warmup.
 - I will evaluate the gains in term of image gains

chibenwa:
(Just rebased to solve conflict)

Arsnael (approved):

Arsnael:
Rebased and squashed fixups

PR #450, opened 2021-05-22
JAMES-3591 Warn against CassandraBlobStore usage

chibenwa:


server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/CassandraBlobStoreDAO.java
```diff
@@ -80,6 +87,9 @@ public CassandraBlobStoreDAO(CassandraDefaultBucketDAO defaultBucketDAO,
 
         this.metricClOneMissCount = metricFactory.generate(CASSANDRA_BLOBSTORE_CL_ONE_MISS_COUNT_METRIC_NAME);
         this.metricClOneHitCount = metricFactory.generate(CASSANDRA_BLOBSTORE_CL_ONE_HIT_COUNT_METRIC_NAME);
+
+        LOGGER.warn("WARNING: JAMES-3591 Cassandra is not made to store large binary content, its use will be suboptimal compared to " +
```
jeantil, line 91: Only if this log can be disabled by a config flag. otherwise a log at warn will trigger potentially spurious alerting in various tools such as sentry and it's a pain to properly ignore.

chibenwa: Fair.

Arsnael (approved):

PR #451, opened 2021-05-22
JAMES-3589 Allow tailor-made servers to customize their mailet container

chibenwa:
This enables those who want to adopt alternatives to the existing `CamelMailetContainer`, allowing to bypass existing bugs like `JAMES-3589`.

The issue is that Mailet container is pre-bundled in GuiceJamesServer thus do not allow overrides.


Arsnael (approved):

chibenwa:
I am merging it as it unstuck me and allow to bypass the issue on my deployments. No behavioural changes either way.

PR #454, opened 2021-05-23
[PERFORMANCE] AmqpForwardAttribute should reuse connection

chibenwa:
Today one connection and one channel is opened per processed
emails. Connections and channels are long lived objects,
made to be reused.

Flame graphs showed that this AMQP mailet represents ~1%
of CPU occupation with just about 30 mail/sec. Note that
92% of this time is used opening/closing connections/channels.

Moving connection and channel initialization at mailet
initialization enables reuse and leads to dramatic enhancements
of the performances of that very mailet. Note that retries are
in place in case Rabbit is not up right at start - common with
IE docker-compose.

The testing strategy relying on mocks is brittle. Given that we
have full-fledged integration tests in place, we can get rid of
these annoying mock-based tests.

## Before

newcon

newcon2

## After

resuecon


chibenwa:
AMQP channels are not threadsafe and should not be shared across thread.

 - [ ] We should create a new `mailet-amqp` maven project
 - [ ] Depends on `bckens-rabbitmq` to reuse `SimpleConnectionPool` and `ReactorRabbitMQChannelPool`
 - [ ] Clean up AMQP dependency out of `mailet-standard`
 - [ ] Finally we should take care that we should be able to maintain a separated AMQP host than the main one.

pom.xml
```diff
@@ -958,6 +958,11 @@
                 <version>${project.version}</version>
                 <type>test-jar</type>
             </dependency>
+            <dependency>
+                <groupId>${james.groupId}</groupId>
+                <artifactId>apache-mailet-amqp</artifactId>
```
Arsnael: idem

Arsnael (approved):

PR #458, opened 2021-05-27
JAMES-3516 Using the MessageResult::getThreadId property at the JMAP level

quantranhong1999:
Using the MessageResult::getThreadId property at the JMAP level - in Email/get + Email/import

### DIFF
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/model/ThreadId.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/model/ThreadId.java
@@ -35,6 +35,10 @@ public class ThreadId {
         return baseMessageId;
     }
 
+    public String serialize() {
+        return baseMessageId.serialize();
+    }
+
     @Override
     public final boolean equals(Object o) {
         if (o instanceof ThreadId) {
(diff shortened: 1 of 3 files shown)


server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/mail/Email.scala
```diff
@@ -421,6 +420,7 @@ private class EmailMetadataViewFactory @Inject()(zoneIdProvider: ZoneIdProvider)
     val mailboxIds: MailboxIds = MailboxIds(message._2
       .map(_.getMailboxId)
       .toList)
+    val threadId: ThreadId = ThreadId(message._2.head.getThreadId.getBaseMessageId.serialize())
```
chibenwa, line 423: We should not expose at the JMAP level `getBaseMessageId` which is an implementation detail.

We should have a `ThreadId::serialize` method IMO

mailbox/api/src/main/java/org/apache/james/mailbox/model/ThreadId.java
```diff
@@ -35,6 +35,10 @@ public MessageId getBaseMessageId() {
         return baseMessageId;
     }
 
+    public String serialize() {
+        return baseMessageId.serialize();
```
chibenwa, line 39: Can we have at least one unit test for this method?

chibenwa (approved):
Now https://jmap.io/spec-mail.html#emailset turn ;-)

```
For successfully created Email objects, the created response contains the id, blobId, threadId, and size properties of the object.
```

Arsnael (approved):

PR #461, opened 2021-05-28
JAMES-2886 Fix configuration parsing for Guice extensions

chibenwa:


Arsnael (approved):

server/container/guice/guice-utils/src/main/java/org/apache/james/utils/ExtensionConfiguration.java
```diff
@@ -19,24 +19,25 @@
...
     public static ExtensionConfiguration from(Configuration configuration) {
-        List<String> list = Optional.ofNullable(configuration.getStringArray("guice.extension.module"))
-            .map(Arrays::asList)
-            .orElse(ImmutableList.of());
+        String rawString = configuration.getString("guice.extension.module", "");
 
-        return new ExtensionConfiguration(list.stream()
+        return new ExtensionConfiguration(Splitter.on(",")
+            .omitEmptyStrings()
+            .trimResults()
```
jeantil, line 38: maybe extract 
```
Splitter.on(",")
            .omitEmptyStrings()
            .trimResults()
 ```
to a local constant ? I don't think we need to build a new instance for every call (not that it matters much since this should only run once per process)

File: server/container/guice/protocols/webadmin/src/main/java/org/apache/james/modules/server/WebAdminServerModule.java [hunk 2]
```diff
@@ -138,6 +138,16 @@ public WebAdminConfiguration provideWebAdminConfiguration(FileSystem fileSystem,
         }
     }
 
+    @VisibleForTesting
+    ImmutableList<String> additionalRoutes(Configuration configurationFile) {
+        String rawString = configurationFile.getString("extensions.routes", "");
+
+        return ImmutableList.copyOf(Splitter.on(',')
```
jeantil, line 145: same as above: extract to local constant (note that I don't think mutualizing the two would be beneficial)

jeantil (approved):

server/container/guice/guice-utils/src/main/java/org/apache/james/utils/ExtensionConfiguration.java
chibenwa: This code path is only executed once.

Do we care? Sounds like early optimisation to me...

server/container/guice/protocols/webadmin/src/main/java/org/apache/james/modules/server/WebAdminServerModule.java
chibenwa: This code path is only executed once.

Do we care? Sounds like early optimisation to me...

server/container/guice/guice-utils/src/main/java/org/apache/james/utils/ExtensionConfiguration.java
jeantil: I feel that the splitter setup gets in the way of the actual parsing logic. 
The first 3 lines are type Splitter, then the flow changes to stream processing with the actual input value only appearing on line 4 

extracting to a constant would change the code to 
```
SPLITTER.splitToList(rawString)
            .stream()
            .map(ClassName::new)
            .collect(Guavate.toImmutableList())
```
An alternative could be extracting to a private method that event takes care of the list<->stream ceremony  

``` 
splitOnCommas(rawString)
 .map(ClassName::new)
            .collect(Guavate.toImmutableList())
```

chibenwa: I had a look, Joiner and Splitter are thread safe, thus can be safely shared as constants

The point is we never did it in the source code. I think it could be nice to conduct a global refactoring regarding this.

I guess the best way to move is thus fix these usage here and open another PR for doing this refctoring.

chibenwa:
(Forced pushed to squash them)

PR #462, opened 2021-05-28
JAMES-2157 HasMimeType only look at the top level mime type

chibenwa:
It do not try to go down the mime tree which makes it impractical for some use cases like `no subpart should have the content type application/zip`.

I understand that some use case do not need to walk down the entire mime tree (eg when categorizing MDN reports only top level content type is enough) however some over generic use cases implies checking the wall mime tree.

Thus I propose to add a `HasMimeTypeAnySubPart` mailet that complement HasMimeType and does just this.

### DIFF
--- a/docs/modules/servers/partials/HasMimeType.adoc
+++ b/docs/modules/servers/partials/HasMimeType.adoc
@@ -2,7 +2,7 @@
 
 This matcher checks if the content type matches.
 
-This matcher do not walk down the mime tree and stops at the top level mime part.
+This matcher does not walk down the mime tree and stops at the top level mime part.
 
 use:
 
--- a/docs/modules/servers/partials/HasMimeType.adoc
+++ b/docs/modules/servers/partials/HasMimeType.adoc
@@ -2,6 +2,8 @@
 
 This matcher checks if the content type matches.
 
+This matcher do not walk down the mime tree and stops at the top level mime part.
+
 use:
 
 ....
--- a/docs/modules/servers/partials/HasMimeTypeAnySubPart.adoc
+++ b/docs/modules/servers/partials/HasMimeTypeAnySubPart.adoc
@@ -1,0 +1,11 @@
+=== HasMimeTypeAnySubPart
+
+This matcher checks if the content type matches.
+
+This matcher walks down the mime tree and will look up at all the mime parts of this message.
+
+use:
+
+....
+<mailet match="HasMimeTypeAnySubPart=text/plain,text/html" class="..." />
+....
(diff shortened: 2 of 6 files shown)


docs/modules/servers/partials/HasMimeType.adoc
```diff
@@ -2,6 +2,8 @@
 
 This matcher checks if the content type matches.
 
+This matcher do not walk down the mime tree and stops at the top level mime part.
```
jeantil, line 5:
```suggestion
This matcher does not walk down the mime tree and stops at the top level mime part.
```

docs/modules/servers/partials/HasMimeTypeAnySubPart.adoc
```diff
@@ -0,0 +1,11 @@
+=== HasMimeTypeAnySubPart
+
+This matcher checks if the content type matches.
+
+This matcher walk down the mime tree and will look up at all the mime parts of this message.
```
jeantil, line 5:
```suggestion
This matcher walks down the mime tree and will look up at all the mime parts of this message.
```

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeType.java
```diff
@@ -42,6 +42,8 @@
 /**
  * <p>This matcher checks if the content type matches.</p>
  *
+ * <p>This matcher do not walk down the mime tree and stops at the top level mime part.</p>
```
jeantil, line 45:
```suggestion
 * <p>This matcher does not walk down the mime tree and stops at the top level mime part.</p>
```

mailet/standard/src/main/java/org/apache/james/transport/matchers/HasMimeTypeAnySubPart.java
```diff
@@ -0,0 +1,119 @@
...
+import com.github.fge.lambdas.Throwing;
+import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+
+
+/**
+ * <p>This matcher checks if the content type matches.</p>
+ *
+ * <p>This matcher walk down the mime tree and will look up at all the mime parts of this message.</p>
```
jeantil, line 52:
```suggestion
 * <p>This matcher walks down the mime tree and will try to match any of the mime parts of this message.</p>
```

jeantil (approved):

chibenwa:
(Force pushed to solve conflict)

PR #466, opened 2021-05-29
JAMES-3261 Migrate all docker images to JIB

chibenwa:
This follows https://github.com/apache/james-project/pull/399 that provided JIB image for memory guice.

This PR provides the same thing for all other Guice distributions as well as updates the documentation accordingly.


Arsnael (approved):

server/container/guice/cassandra-guice/pom.xml
```diff
@@ -273,6 +278,124 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>com.googlecode.maven-download-plugin</groupId>
+                <artifactId>download-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>install-jbpm</id>
```
jeantil, line 286:
```suggestion
                        <id>install-glowroot</id>
```
for all pom.xml in this PR :D

server/container/guice/cassandra-guice/pom.xml
```diff
@@ -273,6 +278,124 @@
...
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>com.google.cloud.tools</groupId>
+                <artifactId>jib-maven-plugin</artifactId>
+                <configuration>
+                    <from>
+                        <image>adoptopenjdk:11-jdk-hotspot</image>
```
jeantil, line 327:
```suggestion
                        <image>gcr.io/distroless/java:11</image>
```
along with https://martinheinz.dev/blog/49 for debugging ?

applies to all pom.xml which use jib :D

server/container/guice/jpa-smtp/src/main/extensions-jars/README.md
```diff
@@ -2,4 +2,4 @@
 
 The jar in this folder will be added to JAMES classpath when mounted under /root/extensions-jars inside the running container.
 
-You can use it to add your customs Mailets/Matchers.
+You can use it to add you customs Mailets/Matchers.
```
jeantil, line 5:
```suggestion
You can use it to add your custom Mailets/Matchers.
```

jeantil (commented):
It's a bit sad to see all the duplicated glowroot json files :/ a couple suggestions on the POM files but otherwise a very welcome change ! jib is the best

chibenwa:
> It's a bit sad to see all the duplicated glowroot json files :/

Was the case before ;-)

If you want to have a shot at factorising it ;-)

server/container/guice/cassandra-guice/pom.xml
chibenwa: I am unsure we should change the running images. What are the pros and cons?

Maybe we need more discussion and you could propose that change (and related fixes like JVM option passing, glowroot). Likely not a drop in replacement.

jeantil: pros: 
- less attack surface in the running container as there is way less  stuff installed that can be used to exploit
- much much lighter images (some report 75% reduction)
cons: 
- forget docker exec distroless containers only contain strictly necessary (hence ephemeral containers)

chibenwa:
Just rebased prior a merge.

PR #471, opened 2021-06-02
[PERFORMANCE] Migrate where possible from Mono.fatMap to flatMapIterable

chibenwa:
Before : 

Screenshot from 2021-06-02 11-58-18

After:

Screenshot from 2021-06-02 11-58-51

(We downgraded the James architecture, only 2 x 2 VCPU + 4 GB heap for the James, and still 3 x OVH b2-30 for the Cassandra cluster)

### DIFF
--- a/backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/utils/CassandraAsyncExecutor.java
+++ b/backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/utils/CassandraAsyncExecutor.java
@@ -22,6 +22,7 @@ package org.apache.james.backends.cassandra.utils;
 import static org.apache.james.util.ReactorUtils.publishIfPresent;
 
 import java.util.Optional;
+import java.util.function.Function;
 
 import javax.inject.Inject;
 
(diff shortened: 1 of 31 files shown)


backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/utils/CassandraAsyncExecutor.java
```diff
@@ -67,7 +68,7 @@ public CassandraAsyncExecutor(Session session) {
 
     public Flux<Row> executeRows(Statement statement) {
         return execute(statement)
-            .flatMapMany(Flux::fromIterable);
+            .flatMapIterable(Function.identity());
```
Arsnael, line 71: Can you explain what `Function.identity()` is for exactly?

chibenwa: We have a `Mono<ResultSet>` and `ResultSet` is an interable, so we do not need to transform it.

Arsnael (approved):

PR #473, opened 2021-06-04
JAMES-2683 SimpleConnectionPool should be a guice singleton

chibenwa:
Not doing so results in opening too many connections...

 - Two connections were opened for the Health checks
 - One connection for ReactorRabbitMQChannelPool
 - One connection for ReceiverProvider

Screenshot from 2021-06-04 10-34-04

With the proposed fix we are down to one.


Arsnael (approved):

chibenwa:
After: 1 connection per node!

Screenshot from 2021-06-05 10-38-58

PR #476, opened 2021-06-05
[PERFORMANCE] Reduce Cassandra chunk length for some read intensive t…

chibenwa:
…ables

 - The table should be read heavy
 - Have a small PARTITION size

Reducing the compression chunk size reduces the amount of data Cassandra have to
pull and deflate, at the expense of a lower compression and of more object creation.

The proposed values lead to a net performance improvement both for mean and p99
response time (106 ms mean time for JMAP queries to 85ms)

Testing infrastructure: 2*2 CPU James with 4GB RAM, 3 Cassandra being OVH B2-30,
 ~900 requests per second

## Before

Screenshot from 2021-06-05 19-16-51

## After

Screenshot from 2021-06-05 19-17-45

### DIFF
--- a/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/modules/CassandraMailboxModule.java
+++ b/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/modules/CassandraMailboxModule.java
@@ -42,7 +42,8 @@ public interface CassandraMailboxModule {
         .comment("Holds the mailboxes information.")
         .options(options -> options
             .caching(SchemaBuilder.KeyCaching.ALL,
-                SchemaBuilder.rows(CassandraConstants.DEFAULT_CACHED_ROW_PER_PARTITION)))
+                SchemaBuilder.rows(CassandraConstants.DEFAULT_CACHED_ROW_PER_PARTITION))
+            .compressionOptions(SchemaBuilder.lz4().withChunkLengthInKb(8)))
         .statement(statement -> statement
             .addPartitionKey(CassandraMailboxTable.ID, timeuuid())
             .addUDTColumn(CassandraMailboxTable.MAILBOX_BASE, SchemaBuilder.frozen(CassandraMailboxTable.MAILBOX_BASE))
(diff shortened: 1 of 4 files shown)


vttranlina (approved):

Arsnael (approved):

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/modules/CassandraMailboxModule.java
```diff
@@ -42,7 +42,8 @@
         .comment("Holds the mailboxes information.")
         .options(options -> options
             .caching(SchemaBuilder.KeyCaching.ALL,
-                SchemaBuilder.rows(CassandraConstants.DEFAULT_CACHED_ROW_PER_PARTITION)))
+                SchemaBuilder.rows(CassandraConstants.DEFAULT_CACHED_ROW_PER_PARTITION))
+            .compressionOptions(SchemaBuilder.lz4().withChunkLengthInKb(8)))
```
jeantil, line 46: how did you come to choose `8` Kb ? 
is it based on BufferedReader's defaultCharBufferSize to align with memory page size or is based on the estimated size of a mailbox row ? 
(https://stackoverflow.com/questions/37404068/why-is-the-default-char-buffer-size-of-bufferedreader-8192)

a comment would be nice :D

(if the answer is size of a mailbox row, the same question applies to all the following `8` that appear in the PR)

chibenwa: https://thelastpickle.com/blog/2018/08/08/compression_performance.html

TLDR:  Using out of the box settings for compression on read heavy or mixed workloads will almost certainly put unnecessary strain on your disk while hurting your read performance.

Basically I tried a couple of values and 8KB worked well.

jeantil: > TLDR: Using out of the box settings for compression on read heavy or mixed workloads will almost certainly put unnecessary strain on your disk while hurting your read performance.

I understand the logic behind the change, I challenge the arbitrary number that was chosen :D  

> Basically I tried a couple of values (4K, 8K, 16K, ...) and 8KB worked well.

Aren't you worried that it is highly dependendant on the cassandra infrastructure ? After all 640k is enough memory for everything ;p

chibenwa: I do think compression is more related to the workload than the infrastructure.

That being said I'm fine keeping defaults, and requiring expertise to come and tune these settings.

jeantil (approved):

PR #478, opened 2021-06-06
JAMES-3594 Implement ReadOnlyLDAPUsersDAO with UnboundID library

chibenwa:
I had a sunday shot at using [UnboundID](https://docs.ldap.com/ldap-sdk/docs/index.html), ~ 1H between cooking and washing dishes. It went surprisingly well!

Cc @rouazana 

All tests are passing.

Remains to implement:

 - [x] Group restrictions
 - [x] Retries (The plan might be to use reactor for this...)
 
 # Context
 
### What

While doing some profiling on IMAP, I discovered that we start a thread and open a connection for each authentication performed by IMAP, JMAP, SMTP users.

Here is the incriminated code:

```
    @Override
    public boolean verifyPassword(String password) {
        boolean result = false;
        LdapContext ldapContext = null;
        try {
            ldapContext = this.ldapContext.newInstance(null);
            ldapContext.addToEnvironment(Context.SECURITY_AUTHENTICATION,
                    LdapConstants.SECURITY_AUTHENTICATION_SIMPLE);
            ldapContext.addToEnvironment(Context.SECURITY_PRINCIPAL, userDN);
            ldapContext.addToEnvironment(Context.SECURITY_CREDENTIALS, password);
            ldapContext.reconnect(null);
            result = true;
        } catch (NamingException exception) {
            // no-op
        } finally {
            if (null != ldapContext) {
                try {
                    ldapContext.close();
                } catch (NamingException ex) {
                    // no-op
                }
            }
        }
        return result;
    }
```

Needless to say, underload opening so much connections, and starting so much threads is of course detrimental. (I wonder if it could have been the root cause of some LDAP incidents...)

### Why?

Because we use JNDI to read the LDAP, because the standard is to try to authenticate as the user against the LDAP, and because doing so requires a new connection.

Directory-api is only used to sanitize filter, but appart from that data-ldap fully relies on "JNDI".

### How to fix it?

Apparently there is little way around with JNDI. I tried to use subcontexts (failed), removing the 'Reconnect' (fails).

I believe the best way to fix this is to completly drop the JNDI usage.

Having (quickly) discussed the topic with [~rouazana] (James committer and LDAP expert) we could choose "directory-api" (apache folks), but the UnboundID LDAP SDK could be a good candidate for this too.


chibenwa:
Preliminary performance results of UnboundID usage:

## Before

bbbbef

## After

aaaaaaaafff

No longer thread and connection hungry, three time faster...

chibenwa:
Following a discusion with @rouazana, he expressed concerns regarding the lack of tests for group/role restrictions.

As such, I propose to consider the feature as experimental, and clearly document it as such...

chibenwa:
Without this fix:

```
================================================================================
2021-06-08 07:20:54                                         260s elapsed
---- Requests ------------------------------------------------------------------
> Global                                                   (OK=107714 KO=233   )
> prepare / obtainContinuationToken                        (OK=10166  KO=43    )
> prepare / obtainAccessToken                              (OK=8152   KO=187   )
> prepare / getMailboxes                                   (OK=8152   KO=0     )
> prepare / listMessages                                   (OK=8152   KO=0     )
> SelectMailbox / listMessages                             (OK=22905  KO=0     )
> updateFlags / markAsRead                                 (OK=3194   KO=0     )
> SelectMailbox / getMessages                              (OK=22902  KO=3     )
> updateFlags / markAsAnswered                             (OK=903    KO=0     )
> OpenMessage / getMessages                                (OK=13636  KO=0     )
> InboxHomeLoading / getMailboxes                          (OK=2261   KO=0     )
> updateFlags / markAsFlagged                              (OK=463    KO=0     )
> InboxHomeLoading / listMessages                          (OK=2261   KO=0     )
> InboxHomeLoading / getMessages                           (OK=2260   KO=0     )
> sendMessage / sendMessages                               (OK=2307   KO=0     )
---- Errors --------------------------------------------------------------------
> status.find.is(201), but actually found 401                       187 (63.61%)
> j.n.s.SSLException: handshake timed out                            43 (14.63%)
> obtainAccessToken: Failed to build request: No attribute named     43 (14.63%)
 'continuationToken' is defined 
> getMailboxes: Failed to build request: No attribute named 'acc     18 ( 6.12%)
essTokenHeader' is defined 
> jsonPath($[?(@[0] == 'error')]).find.notExists preparation cra      2 ( 0.68%)
shed: Jodd failed to parse into a valid AST: j.j.JsonException...
> jsonPath($[?(@[0] == 'error')]).find.notExists preparation cra      1 ( 0.34%)
shed: Jodd failed to parse into a valid AST: j.j.JsonException...

---- JmapSendMessages ----------------------------------------------------------
[--------------------------------------------------------------------------]  0%
          waiting: 0      / active: 10000  / done: 0     
================================================================================
```

Notice the 401 upon high authentication rate...

After the change of library no longer 401 are seen.

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/LdapRepositoryConfiguration.java
```diff
@@ -160,7 +163,6 @@ public static LdapRepositoryConfiguration from(HierarchicalConfiguration<Immutab
         String userIdAttribute = configuration.getString("[@userIdAttribute]");
         String userObjectClass = configuration.getString("[@userObjectClass]");
         // Default is to use connection pooling
-        boolean useConnectionPool = configuration.getBoolean("[@useConnectionPool]", NO_CONNECTION_POOL);
```
rouazana, line 163: you need to document (in upgrade note) this breaking change

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUser.java
```diff
@@ -38,14 +41,13 @@
...
- * 
- * @see SimpleLDAPConnection
+ *
  * @see ReadOnlyUsersLDAPRepository
  * 
  */
 public class ReadOnlyLDAPUser implements User, Serializable {
-    // private static final long serialVersionUID = -6712066073820393235L; 
-    private static final long serialVersionUID = -5201235065842464013L;
+    private static final long serialVersionUID = -5201235065842464014L;
```
rouazana, line 49: you can probably remove this

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUser.java
```diff
@@ -55,50 +57,42 @@
...
-        super();
-    }
+    private final LdapRepositoryConfiguration configuration;
 
     /**
      * Constructs an instance for the given user-details, and which will
      * authenticate against the given host.
-     * 
-     * @param userName
+     *  @param userName
```
rouazana, line 78:
```suggestion
     * @param userName
```

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUser.java
```diff
@@ -139,27 +133,19 @@ public boolean setPassword(String newPass) {
...
+        } catch (Exception e) {
+            LOGGER.error("Unexpected error upon authentication", e);
+            return false;
         }
-        return result;
+    }
+
+    private boolean doVerifyPassword(String password) throws LDAPException {
+        BindResult bindResult = connectionPool.bindAndRevertAuthentication(userDN, password);
+        return bindResult.getResultCode()
```
rouazana, line 148: you cannot compare to a constant LDAP_SUCCESS?

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -120,63 +94,34 @@ public void configure(LdapRepositoryConfiguration configuration) {
...
+                + '\n' + "maxRetries: " + ldapConfiguration.getMaxRetries() + '\n');
         }
-        // Setup the initial LDAP context
-        updateLdapContext();
-    }
-
-    protected void updateLdapContext() throws NamingException {
-        ldapContext = computeLdapContext();
-    }
+        filterTemplate = "(&({0}={1})(objectClass={2})" + StringUtils.defaultString(ldapConfiguration.getFilter(), "") + ")";
```
rouazana, line 106: you are prone to some ldap injection attacks by building you filter manually, please use Filter and the associated builders to generate correctly your filter and eventually escape needed characters

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -120,63 +94,34 @@ public void configure(LdapRepositoryConfiguration configuration) {
...
 
-            @Override
-            public Context newDelegate() throws NamingException {
-                return new InitialLdapContext(getContextEnvironment(), null);
-            }
-        };
+        URI uri = new URI(ldapConfiguration.getLdapHost());
+        SocketFactory socketFactory = null;
+        LDAPConnection ldapConnection = new LDAPConnection(socketFactory, connectionOptions, uri.getHost(), uri.getPort(), ldapConfiguration.getPrincipal(), ldapConfiguration.getCredentials());
+        ldapConnectionPool = new LDAPConnectionPool(ldapConnection, 4);
```
rouazana, line 115: why 4?

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -120,63 +94,34 @@ public void configure(LdapRepositoryConfiguration configuration) {
...
+        // TODO implement retries
```
rouazana, line 116: still a TODO?

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -209,29 +154,20 @@ private boolean userInGroupsMembershipList(String userDN,
...
-        while (sr.hasMore()) {
-            SearchResult r = sr.next();
-            result.add(r.getNameInNamespace());
-        }
+    private Set<String> getAllUsersFromLDAP() throws LDAPException {
+        LDAPConnection connection = ldapConnectionPool.getConnection();
+        try {
+            SearchResult searchResult = connection.search(ldapConfiguration.getUserBase(),
+                SearchScope.SUB,
+                filterTemplate);
```
rouazana, line 162: this cannot work if you search directly with a template, is it tested?

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -209,29 +154,20 @@ private boolean userInGroupsMembershipList(String userDN,
...
 
-        return result;
+            return searchResult.getSearchEntries()
+                .stream()
+                .map(entry -> entry.getObjectClassAttribute().getName())
```
rouazana, line 166: why do you get this? I really think this method has never been tested

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -209,29 +154,20 @@ private boolean userInGroupsMembershipList(String userDN,
...
+                .collect(Guavate.toImmutableSet());
+        } finally {
+            ldapConnectionPool.releaseConnection(connection);
```
rouazana, line 169: why do you need this?

rouazana: I mean if you are doing only one operation, you can issue it directly on the pool instead of getting a connection

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -260,133 +195,150 @@ private boolean userInGroupsMembershipList(String userDN,
...
-      Attributes userAttributes = ldapContext.getAttributes(userDN);
-      Optional<Attribute> userName = Optional.ofNullable(userAttributes.get(ldapConfiguration.getUserIdAttribute()));
-      return userName
-          .map(Throwing.<Attribute, String>function(u -> u.get().toString()).sneakyThrow())
-          .map(Username::of)
-          .map(username -> new ReadOnlyLDAPUser(username, userDN, ldapContext));
+    private Optional<ReadOnlyLDAPUser> buildUser(String userDN) throws LDAPException {
+        LDAPConnection connection = ldapConnectionPool.getConnection();
+        try {
+            SearchResultEntry userAttributes = connection.getEntry(userDN);
```
rouazana, line 234: idem you are doing only one operation, so no need to getConn/op/release, just use the pool

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -260,133 +195,150 @@ private boolean userInGroupsMembershipList(String userDN,
...
+            ldapConnectionPool.releaseConnection(connection);
+        }
     }
 
     @Override
     public boolean contains(Username name) throws UsersRepositoryException {
-        return getUserByName(name).isPresent();
+        try {
+            return Mono.fromCallable(() -> doContains(name))
+                .retryWhen(ldapConfiguration.retrySpec())
```
rouazana, line 248: I don't understand exactly when the retry is done (I mean on which condition), can you explain me?

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
chibenwa: Let's make it configurable with a `poolSize` argument

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
chibenwa: Nope

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
chibenwa: Just a mis-understanding of how the library is working, that is all...

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
chibenwa: Upon errors

rouazana: Which error? You should retry only when connection-related errors.

chibenwa: Then honnestly I am more in favour of dropping retries alltogether and only rely on:

```
        ldapConnectionPool.setRetryFailedOperationsDueToInvalidConnections(true);
```

Thoughts @rouazana ?

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -209,29 +171,33 @@ private boolean userInGroupsMembershipList(String userDN,
...
-        NamingEnumeration<SearchResult> sr = ldapContext.search(ldapConfiguration.getUserBase(), "(objectClass="
-                + ldapConfiguration.getUserObjectClass() + ")", sc);
-        while (sr.hasMore()) {
-            SearchResult r = sr.next();
-            result.add(r.getNameInNamespace());
-        }
+    private Set<String> getAllUsersDNFromLDAP() throws LDAPException {
+        SearchRequest searchRequest = new SearchRequest(ldapConfiguration.getUserBase(),
+            SearchScope.SUB,
+            createFilter(),
```
rouazana, line 177: it's at usage, I think you don't need to create it for each request.

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -209,29 +171,33 @@ private boolean userInGroupsMembershipList(String userDN,
...
+            ldapConfiguration.getUserIdAttribute());
```
rouazana, line 178: you can eventually use `SearchRequest.NO_ATTRIBUTES` here instead

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
```diff
@@ -260,133 +225,143 @@ private boolean userInGroupsMembershipList(String userDN,
...
+            if (e.getCause() instanceof UsersRepositoryException) {
+                throw (UsersRepositoryException) e.getCause();
+            }
             throw new UsersRepositoryException("Unable to retrieve user count from ldap", e);
         }
     }
 
+    private int doCountUsers() throws LDAPException {
+        return Math.toIntExact(getValidUsers().stream()
+            .map(Throwing.function(this::buildUser).sneakyThrow())
```
rouazana, line 297: n+1 ldap requests to get the number of users?

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUsersDAO.java
rouazana: This parameter seems nice, maybe we could also use it in LinID cc @remk 

You could add the matching tests to check it is doing what you expect?

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/ReadOnlyLDAPUser.java
```diff
@@ -55,50 +57,42 @@
...
      * <code>&quot;john.bold&#64;myorg.com&quot;</code>.
      */
-    private Username userName;
+    private final Username userName;
 
     /**
      * The distinguished name of the user-record in the LDAP directory.
      */
-    private String userDN;
+    private final String userDN;
```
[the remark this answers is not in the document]
chibenwa: Yeah but their API is not strongly typed... Hence I did not find it.

chibenwa:
I added a few tests regarding custom extra filters...

rouazana (approved):
2 little remarks:
- you can use .toString instead of .toNormalized... to get a String from a DN, no need to normalize it
- I would add a connection broken test to ensure retry is working (and continue to do so) if there is not one already (search / stop container / search)

But I'm overall ok with this MR, thanks!

Arsnael (approved):

chibenwa:
(Just rebased, I will wait a green build before merge)

PR #481, opened 2021-06-07
JAMES-3516 Implement naive ThreadIdGuessingAlgorithm

quantranhong1999:
Just a few ideas want to show to you, not really some proper code yet.
My idea is enabling setThreadId in MailboxMessage (just like MoqSeq). 
Then in MessageStorer we can use message.setThreadId(with ThreadId from ThreadIdGuessingAlgorithm)
```
MailboxMessage message = messageFactory.createMessage(messageId, mailbox, internalDate, size, bodyStartOctet, content, flags, propertyBuilder, attachments);
threadId = ThreadIdManager.setThreadId(MailboxMessage message,...);
message.setThreadId(threadId);
return Mono.from(messageMapper.addReactive(mailbox, message))
    .map(metadata -> Pair.of(metadata, Optional.of(attachments)));
```

### DIFF
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/NaiveThreadIdGuessingAlgorithmImpl.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/NaiveThreadIdGuessingAlgorithmImpl.java
@@ -1,0 +1,35 @@
+/******************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one     *
+ * or more contributor license agreements.  See the NOTICE file   *
+ * distributed with this work for additional information          *
+ * regarding copyright ownership.  The ASF licenses this file     *
+ * to you under the Apache License, Version 2.0 (the              *
+ * "License"); you may not use this file except in compliance     *
+ * with the License.  You may obtain a copy of the License at     *
+ *                                                                *
+ * http://www.apache.org/licenses/LICENSE-2.0                     *
+ *                                                                *
+ * Unless required by applicable law or agreed to in writing,     *
+ * software distributed under the License is distributed on an    *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY         *
+ * KIND, either express or implied.  See the License for the      *
+ * specific language governing permissions and limitations        *
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.store.mail;
+
+import java.util.List;
+
+import org.apache.james.core.Username;
+import org.apache.james.mailbox.model.MessageId;
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MimeMessageId;
+import org.apache.james.mailbox.store.mail.model.Subject;
+
+public class NaiveThreadIdGuessingAlgorithmImpl implements ThreadIdGuessingAlgorithm {
+    @Override
+    public ThreadId guessThreadId(Username username, MessageId messageId, MimeMessageId thisMimeMessageId, MimeMessageId inReplyTo, List<MimeMessageId> references, Subject subject) {
+        return ThreadId.fromBaseMessageId(messageId);
+    }
+}
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/ThreadIdGuessingAlgorithm.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/ThreadIdGuessingAlgorithm.java
@@ -1,0 +1,32 @@
+/******************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one     *
+ * or more contributor license agreements.  See the NOTICE file   *
+ * distributed with this work for additional information          *
+ * regarding copyright ownership.  The ASF licenses this file     *
+ * to you under the Apache License, Version 2.0 (the              *
+ * "License"); you may not use this file except in compliance     *
+ * with the License.  You may obtain a copy of the License at     *
+ *                                                                *
+ * http://www.apache.org/licenses/LICENSE-2.0                     *
+ *                                                                *
+ * Unless required by applicable law or agreed to in writing,     *
+ * software distributed under the License is distributed on an    *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY         *
+ * KIND, either express or implied.  See the License for the      *
+ * specific language governing permissions and limitations        *
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.store.mail;
+
+import java.util.List;
+
+import org.apache.james.core.Username;
+import org.apache.james.mailbox.model.MessageId;
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MimeMessageId;
+import org.apache.james.mailbox.store.mail.model.Subject;
+
+public interface ThreadIdGuessingAlgorithm {
+    ThreadId guessThreadId(Username username, MessageId messageId, MimeMessageId thisMimeMessageId, MimeMessageId inReplyTo, List<MimeMessageId> references, Subject subject);
+}
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/Subject.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/Subject.java
@@ -1,0 +1,58 @@
+/******************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one     *
+ * or more contributor license agreements.  See the NOTICE file   *
+ * distributed with this work for additional information          *
+ * regarding copyright ownership.  The ASF licenses this file     *
+ * to you under the Apache License, Version 2.0 (the              *
+ * "License"); you may not use this file except in compliance     *
+ * with the License.  You may obtain a copy of the License at     *
+ *                                                                *
+ * http://www.apache.org/licenses/LICENSE-2.0                     *
+ *                                                                *
+ * Unless required by applicable law or agreed to in writing,     *
+ * software distributed under the License is distributed on an    *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY         *
+ * KIND, either express or implied.  See the License for the      *
+ * specific language governing permissions and limitations        *
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.store.mail.model;
+
+import java.util.Objects;
+
+import com.google.common.base.MoreObjects;
+
+public class Subject {
+    private final String value;
+
+    public Subject(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (o instanceof Subject) {
+            Subject that = (Subject) o;
+
+            return Objects.equals(this.value, that.value);
+        }
+        return false;
+    }
+
+    @Override
+    public final int hashCode() {
+        return Objects.hash(value);
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+            .add("value", value)
+            .toString();
+    }
+}
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/MimeMessageId.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/MimeMessageId.java
@@ -1,0 +1,58 @@
+/******************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one     *
+ * or more contributor license agreements.  See the NOTICE file   *
+ * distributed with this work for additional information          *
+ * regarding copyright ownership.  The ASF licenses this file     *
+ * to you under the Apache License, Version 2.0 (the              *
+ * "License"); you may not use this file except in compliance     *
+ * with the License.  You may obtain a copy of the License at     *
+ *                                                                *
+ * http://www.apache.org/licenses/LICENSE-2.0                     *
+ *                                                                *
+ * Unless required by applicable law or agreed to in writing,     *
+ * software distributed under the License is distributed on an    *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY         *
+ * KIND, either express or implied.  See the License for the      *
+ * specific language governing permissions and limitations        *
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.store.mail.model;
+
+import java.util.Objects;
+
+import com.google.common.base.MoreObjects;
+
+public class MimeMessageId {
+    private final String value;
+
+    public MimeMessageId(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (o instanceof MimeMessageId) {
+            MimeMessageId that = (MimeMessageId) o;
+
+            return Objects.equals(this.value, that.value);
+        }
+        return false;
+    }
+
+    @Override
+    public final int hashCode() {
+        return Objects.hash(value);
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+            .add("value", value)
+            .toString();
+    }
+}
--- a/mailbox/store/src/test/java/org/apache/james/mailbox/store/mail/model/MimeMessageIdTest.java
+++ b/mailbox/store/src/test/java/org/apache/james/mailbox/store/mail/model/MimeMessageIdTest.java
@@ -1,0 +1,32 @@
+/******************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one     *
+ * or more contributor license agreements.  See the NOTICE file   *
+ * distributed with this work for additional information          *
+ * regarding copyright ownership.  The ASF licenses this file     *
+ * to you under the Apache License, Version 2.0 (the              *
+ * "License"); you may not use this file except in compliance     *
+ * with the License.  You may obtain a copy of the License at     *
+ *                                                                *
+ * http://www.apache.org/licenses/LICENSE-2.0                     *
+ *                                                                *
+ * Unless required by applicable law or agreed to in writing,     *
+ * software distributed under the License is distributed on an    *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY         *
+ * KIND, either express or implied.  See the License for the      *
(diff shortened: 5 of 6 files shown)


mailbox/memory/src/main/java/org/apache/james/mailbox/inmemory/mail/InMemoryThreadIdManager.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public static class InMemoryThreadIdGuessingAlgorithm implements ThreadIdManager.ThreadIdGuessingAlgorithm {
+        @Override
+        public ThreadId guessingThreadId(MailboxMessage mailboxMessage) {
+            return new ThreadId(mailboxMessage.getMessageId());
+        }
+    }
+
+    @Override
+    public ThreadId createThreadId(MailboxMessage mailboxMessage) {
```
chibenwa, line 36: Do we take the mime message as an input? If I remember correctly MailboxMessage already has a ThreadId allocated to it...

ThreadId guessing needs to happen before.

Please consider...

 - The subject (string)
 - The references ( `List<String>` ) including `In-Reply-To`
 - The `MimeMessageId (`String`)
 - The James mailbox   `MessageId`
 
https://jmap.io/spec-mail.html#threads

```
The exact algorithm for determining whether two Emails belong to the same Thread is not mandated in this spec to allow for compatibility with different existing systems. For new implementations, it is suggested that two messages belong in the same Thread if both of the following conditions apply:

    An identical message id [@!RFC5322] appears in both messages in any of the Message-Id, In-Reply-To, and References header fields.
    After stripping automatically added prefixes such as “Fwd:”, “Re:”, “[List-Tag]”, etc., and ignoring white space, the subjects are the same. This avoids the situation where a person replies to an old message as a convenient way of finding the right recipient to send to but changes the subject and starts a new conversation.

```

mailbox/memory/src/main/java/org/apache/james/mailbox/inmemory/mail/InMemoryThreadIdManager.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.apache.james.mailbox.inmemory.mail;
+
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.ThreadIdManager;
+import org.apache.james.mailbox.store.mail.model.MailboxMessage;
+
+public class InMemoryThreadIdManager implements ThreadIdManager {
+
+    public static class InMemoryThreadIdGuessingAlgorithm implements ThreadIdManager.ThreadIdGuessingAlgorithm {
```
chibenwa, line 28: I don't see what is specific for memory here just yet.

mailbox/memory/src/main/java/org/apache/james/mailbox/inmemory/mail/InMemoryThreadIdManager.java
```diff
@@ -0,0 +1,44 @@
...
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.inmemory.mail;
+
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.ThreadIdManager;
+import org.apache.james.mailbox.store.mail.model.MailboxMessage;
+
+public class InMemoryThreadIdManager implements ThreadIdManager {
```
chibenwa, line 26: We don't need a manager just yet.

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/ThreadIdManager.java
```diff
@@ -0,0 +1,33 @@
...
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.store.mail;
+
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MailboxMessage;
+
+public interface ThreadIdManager {
+    interface ThreadIdGuessingAlgorithm {
```
chibenwa, line 26: I think only the ThreadIdGuessingAlgorithm makes sense and that we can drop ThreadIdManager.

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/MailboxMessage.java
```diff
@@ -34,6 +34,8 @@
  */
 public interface MailboxMessage extends Message, Comparable<MailboxMessage> {
 
+    void setThreadId(ThreadId threadId);
```
chibenwa, line 37: No I would definitly prefer to have threadId final and add the required onstructor parameters!

quantranhong1999:
Not really fixed compilation, I want to received some reviews to make sure if I am going the right way.

quantranhong1999:
Can someone review this PR please?

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreMessageManager.java
```diff
@@ -939,4 +944,8 @@ public Flags getApplicableFlags(MailboxSession session) throws MailboxException
     public EnumSet<MessageCapabilities> getSupportedMessageCapabilities() {
         return messageCapabilities;
     }
+
+    public ThreadId getThreadIdOfAMessage(MailboxMessage mailboxMessage) {
```
chibenwa, line 948: Why do we need this getter?

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreMessageManager.java
```diff
@@ -164,12 +166,14 @@ private MediaType(String mediaType, String subType) {
     private final BatchSizes batchSizes;
     private final PreDeletionHooks preDeletionHooks;
     private final MessageStorer messageStorer;
+    private final ThreadIdGuessingAlgorithm threadIdGuessingAlgorithm;
```
chibenwa, line 169: Where do we use this field?

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/ThreadIdGuessingAlgorithm.java
```diff
@@ -0,0 +1,27 @@
...
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.store.mail;
+
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MailboxMessage;
+
+public interface ThreadIdGuessingAlgorithm {
+    ThreadId getThreadId(MailboxMessage mailboxMessage);
```
chibenwa, line 26:
```suggestion
    ThreadId getThreadId(MessageId messageId, List<References> references, Subject subject);
```

With

```
public class Reference {
    private final String value;
    
    // ...
}
```

and 

```
public class Subject {
    private final String value;
    
    // ...
}
```

Where references aggegates the In-Reply-To // References header fields.

quantranhong1999: Ok it looks like I understand wrong about naive implementation level. I will working on this.

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/ThreadIdGuessingAlgorithmImpl.java
```diff
@@ -0,0 +1,30 @@
...
+ * specific language governing permissions and limitations        *
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.store.mail;
+
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MailboxMessage;
+
+public class ThreadIdGuessingAlgorithmImpl implements ThreadIdGuessingAlgorithm {
```
chibenwa, line 25: NaiveThreadIdGuessingAlgorithm

quantranhong1999:
Hi, 
I think there are some we need to discuss about.
About the algorithm input, based on your suggestion, JMAP specs and what was proposed preliminary in GSoC proposal. I suppose the guessing algorithm method:
```
ThreadId guessThreadId(Username username, MessageId messageId, Message-ID message-id, List<References> references, Subject subject);
```
Where:
- MessageId to be the naive ThreadId for now
- Username so we can get all emails of that user => get all header fields like Message-ID, In-reply-to, References, Subject => compare to already having Message-ID, In-reply-to, References, Subject fields.

Can someone explain me about the create and saving a message process in James? I doubt that in the final implementation of guessing thread id algorithm, there should not be MessageId messageId in the params because MessageId only available when we created a message (so threadId also).  After we have threadId, we can pass it into some create message params and save it. After that I suppose we can use ThreadIdManager to save a trio of accountId(Username), MessageId, ThreadId to the ThreadId table. 
Beside that, I have some other question (Which could be nice if you can answer shortly and directly to me. Otherwise I think I can read RFC5322 about this by my self, I'm fine.)
- What is Message-ID meaning? The difference between Message-ID and MessageId? Who decided Message-ID (client/ server...)?
- What is References header field meaning?
- Why should we combine  In-Reply-To and References into a object?

I know this is really a long discuss with many questions and I would very appreciate for your answers. :')

chibenwa:
Hi Quan

> `ThreadId guessThreadId(Username username, MessageId messageId, Message-ID message-id, List<References> references, Subject subject);`

I agree with that.

> I know this is really a long discuss with many questions and I would very appreciate for your answers. :')

When I have 5 minutes I will answer you, but likely on the server-dev mailing list :-) 

This sounds like legitimate questions.

quantranhong1999:
> When I have 5 minutes I will answer you, but likely on the server-dev mailing list :-)
> About the difference between Message-ID and MessageId, Tung just sent me this : https://www.mail-archive.com/server-user@james.apache.org/msg16459.html?fbclid=IwAR11h5J8N2JzWntoPX-bsEXKnkliuvvc_QCPgs6hAomYVXcskPX1sEr-SoM

chibenwa:
> About the difference between Message-ID and MessageId, Tung just sent me this : https://www.mail-archive.com/server-user@james.apache.org/msg16459.html?fbclid=IwAR11h5J8N2JzWntoPX-bsEXKnkliuvvc_QCPgs6hAomYVXcskPX1sEr-SoM

Do you need more explanations regarding this?

> What is In-Reply-To? 

In reply to is a field holding the Mime message-Id of the message this mail is a rely of. This is a strong indicator that they are in the same thread... You can find explanations here: https://datatracker.ietf.org/doc/html/rfc4021#section-2.1.9.

> What is References header field meaning?

Idem references are emails being related. This could be previous replies of the thread for instance. https://datatracker.ietf.org/doc/html/rfc4021#section-2.1.10

This could be used for instance also by forwards that technically are not replies.

> Why should we combine In-Reply-To and References into a object?

I personnally think we can wrap that into a `List<MimeMessageId> relatedMimeMessageIds` field instead of two fields.

Arsnael (commented):
Read it, nothing more to add

quantranhong1999:
> Do you need more explanations regarding this?

No. That's enough for me.

I am researching about MIME. In JMAP specs:
```
An Email object is a representation of a message [@!RFC5322], which allows clients to avoid the complexities of MIME parsing, transfer encoding, and character encoding.
```
As i understand, JMAP email object based on RFC5322. 
When an user client sends an email: client send a Email/set JMAP request to Server -> Server create a Jmap Email objects -> convert it into MIME format and save it to Database. 
When user client read a email (Email/get), I see server returns Jmap object (not Mime message).
 
My question is how does the JMAP client read that Jmap object as a email? Convert it into a MIME message or parsing it other way?

chibenwa:
> As i understand, JMAP email object based on RFC5322.

Correct

> When user client read a email (Email/get), I see server returns Jmap object (not Mime message).

Correct. 

To be more precise the server returns the properties the client asked for.

Some are high level, directly matches use case and requires no Mime knowledge eg: `subject` but some others are lower lever and the client can do advanced mime stuff not accounted for by the base spec.

Also a client might download the whole mime message if he wishes.

> My question is how does the JMAP client read that Jmap object as a email?

That's the point: a JMAP client might never end up dealing with any sort of MIME. He can work with the conveniant JMAP API without needing to understand MIME.

This makes it easy to write clients like INBOX or LTT.RS.

quantranhong1999:
I think we should separate In-Reply-To and References fields first. 
I think I gonna have some ideas related to In-Reply-To field.
I am thinking about dynamic algorithm for Guesing Thread Id problem. Child mail can based on succesor mail's Thread Id (through In-Reply-To) to decided its ThreadId. But I doubt in some cases maybe we can't find succesor mail based on In-Reply-To (maybe A and B discuss something for sometime already then reply a mail to C => C dont have succesor of that mail)
Doubting about this, but I still think we should separate In-Reply-To and References fields first.

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/ThreadIdGuessing.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.james.core.Username;
+import org.apache.james.mailbox.model.MessageId;
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MimeMessageId;
+import org.apache.james.mailbox.store.mail.model.Subject;
+
+public interface ThreadIdGuessing {
+    ThreadId naiveGuessThreadId(Username username, MessageId messageId, MimeMessageId thisMimeMessageId, MimeMessageId inReplyTo, List<MimeMessageId> references, Subject subject);
+
+    ThreadId guessThreadId(Username username, MimeMessageId thisMimeMessageId, MimeMessageId inReplyTo, List<MimeMessageId> references, Subject subject);
```
chibenwa, line 33: I disagree with two methods. We should have  methods, several implementation.

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/ThreadIdGuessing.java
```diff
@@ -0,0 +1,34 @@
...
+import java.util.List;
+
+import org.apache.james.core.Username;
+import org.apache.james.mailbox.model.MessageId;
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MimeMessageId;
+import org.apache.james.mailbox.store.mail.model.Subject;
+
+public interface ThreadIdGuessing {
+    ThreadId naiveGuessThreadId(Username username, MessageId messageId, MimeMessageId thisMimeMessageId, MimeMessageId inReplyTo, List<MimeMessageId> references, Subject subject);
```
chibenwa, line 31: The fact of being naive is an implementation detail that do not have its place in an interface.

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/Subject.java
```diff
@@ -0,0 +1,58 @@
...
+ * under the License.                                             *
+ ******************************************************************/
+
+package org.apache.james.mailbox.store.mail.model;
+
+import java.util.Objects;
+
+import com.google.common.base.MoreObjects;
+
+public class Subject {
```
chibenwa, line 26: Technically it should be `BaseSubject`: `After stripping automatically added prefixes such as “Fwd:”, “Re:”, “[List-Tag]”, etc., and ignoring white space, the subjects are the same. This avoids the situation where a person replies to an old message as a convenient way of finding the right recipient to send to but changes the subject and starts a new conversation.`

quantranhong1999: Why we should not let guessThreadId algorithm handle the stripping subject job also?

chibenwa: As you want. Your proposal makes sense.

quantranhong1999: We could write a stripping method in ThreadIdGuessing impl and use it one for all (for input' subject and other searched mails' subjects)

chibenwa: We already have code doing just that: `SearchUtil::getBaseSubject`... Though some more tests could be welcome, we can likely reuse it.

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/ThreadIdGuessing.java
```diff
@@ -28,7 +28,9 @@
 import org.apache.james.mailbox.store.mail.model.Subject;
 
 public interface ThreadIdGuessing {
```
chibenwa, line 30: I liked the ThreadIdGuessingAlgorithm terminology <3

chibenwa (approved):

chibenwa (approved):

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/NaiveThreadIdGuessingAlgorithmImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.james.core.Username;
+import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MimeMessageId;
+import org.apache.james.mailbox.store.mail.model.Subject;
+
+public class NaiveThreadIdGuessingAlgorithmImpl implements ThreadIdGuessingAlgorithm {
+    @Override
+    public ThreadId guessThreadId(Username username, MimeMessageId thisMimeMessageId, MimeMessageId inReplyTo, List<MimeMessageId> references, Subject subject) {
+        // To be implemented
+        return null;
```
Arsnael, line 33: Don't we throw a `NotImplementedException` in those cases usually?

chibenwa: We should implement the naive algorithm `new ThreadId(baseMessageId)` stuff....

quantranhong1999: This supposes to be the not naive implementation method without baseMessageId in params.
I don't understand Benoit's point yet..

chibenwa: Naive and real implementation should have the same API.

quantranhong1999: In the future when we already have the real implementation, can we delete the NaiveImpl and unnecessary basedMessageId param?

chibenwa: Yes I think so...

Arsnael (approved):

chibenwa (approved):

PR #484, opened 2021-06-08
JAMES-2886 Fix collection handling for PropertiesProvider

chibenwa:
Proof that we can now lauch 2 custom routes with webAdmin...

https://user-images.githubusercontent.com/6928740/121112398-8f229a00-c83a-11eb-9ecd-7f81ca145775.mp4

I discovered our property loading was actually flowed...

Review @Arsnael  please...

### DIFF
--- a/server/container/guice/configuration/src/main/java/org/apache/james/utils/DelegatedPropertiesConfiguration.java
+++ b/server/container/guice/configuration/src/main/java/org/apache/james/utils/DelegatedPropertiesConfiguration.java
@@ -275,15 +275,20 @@ public class DelegatedPropertiesConfiguration implements Configuration {
 
     @Override
     public String[] getStringArray(String key) {
-        return splitAndStripDoubleQuotes(configuration.getString(key))
+        return configuration.getList(key)
+            .stream()
+            .map(String.class::cast)
+            .flatMap(this::splitAndStripDoubleQuotes)
             .toArray(String[]::new);
     }
 
     @Override
     public List<Object> getList(String key) {
         try {
-            String rawList = configuration.get(String.class, key);
-            return splitAndStripDoubleQuotes(rawList)
+            return configuration.getList(key)
+                .stream()
+                .map(String.class::cast)
+                .flatMap(this::splitAndStripDoubleQuotes)
                 .collect(Guavate.toImmutableList());
         } catch (ConversionException e) {
             return configuration.getList(key);
@@ -372,9 +377,11 @@ public class DelegatedPropertiesConfiguration implements Configuration {
 
     private Stream<String> splitAndStripDoubleQuotes(String value) {
         return Optional.ofNullable(value)
+            .filter(s -> !StringUtils.isAllBlank(s))
             .map(notNullValue -> Stream
                 .of(StringUtils.strip(notNullValue, "\"").split(delimiter))
-                .map(String::trim))
+                .map(String::trim)
+                .filter(s -> !StringUtils.isAllBlank(s)))
             .orElseGet(Stream::empty);
     }
 }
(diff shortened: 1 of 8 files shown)


Arsnael (approved):

server/container/guice/configuration/src/main/java/org/apache/james/utils/DelegatedPropertiesConfiguration.java
```diff
@@ -372,6 +377,7 @@ public void unlock(LockMode mode) {
 
     private Stream<String> splitAndStripDoubleQuotes(String value) {
         return Optional.ofNullable(value)
+            .filter(s -> !StringUtils.isAllBlank(s))
```
jeantil, line 380: shouldn't filtering come after
```
.map(String::trim) ?
```
instead ? to avoid leaving empty values such as `" "` flow through and be converted to `""` ?

server/container/guice/configuration/src/main/java/org/apache/james/utils/PropertiesProvider.java
```diff
@@ -43,11 +43,20 @@
 import com.google.common.base.Strings;
 
 public class PropertiesProvider {
-
     private static final Logger LOGGER = LoggerFactory.getLogger("org.apache.james.CONFIGURATION");
     private static final char COMMA = ',';
     private static final String COMMA_STRING = ",";
 
+    public static Configuration getConfiguration(File propertiesFile) throws ConfigurationException {
```
jeantil, line 50: :thinking: move from private instance method to public static ... ? that's a huge scoping increase ... 

I'm not saying it's wrong but it feels really weird and it seems to be done exclusively for tests, isn't that a bit extreme ?

chibenwa: Maybe I could annotate it with `@VisibleForTesting` to make the intent clear?

chibenwa: > I'm not saying it's wrong but it feels really weird and it seems to be done exclusively for tests, isn't that a bit extreme ?

I'm trying to address a diverging beaviour between tests and 'production', initializing all the wiring to get a propertyProvider felt overwhelming...

server/container/guice/configuration/src/main/java/org/apache/james/utils/DelegatedPropertiesConfiguration.java
chibenwa: Right!

server/container/guice/configuration/src/main/java/org/apache/james/utils/PropertiesProvider.java
chibenwa: I found a better way...

server/container/guice/configuration/src/main/java/org/apache/james/utils/DelegatedPropertiesConfiguration.java
jeantil: but then the first one on line 380 becomes useless :p

server/container/guice/configuration/src/main/java/org/apache/james/utils/PropertiesProvider.java
```diff
@@ -48,6 +50,17 @@
     private static final char COMMA = ',';
     private static final String COMMA_STRING = ",";
 
+    @VisibleForTesting
+    public static PropertiesProvider forTesting() {
```
jeantil, line 54: :+1:

jeantil (approved):

chibenwa:
(Rebased to squash fixups prior merge)

Arsnael (approved):

PR #493, opened 2021-06-12
JAMES-3491 Do not send two JMAP events upon new messages

chibenwa:
Before we were firing two events on the JMAP event bus:

 - One for emails
 - One for mailbox

The following work aggregates both into a single notification
and thus reduces of around 30% RabbitMQ publish rate.

### DIFF
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxAndEmailChange.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxAndEmailChange.java
@@ -1,0 +1,220 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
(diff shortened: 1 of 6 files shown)


server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxAndEmailChange.java
```diff
@@ -0,0 +1,201 @@
+package org.apache.james.jmap.api.change;
```
Arsnael, line 1: Licence missing

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/MailboxChangesMethodContract.scala
```diff
@@ -889,7 +889,7 @@ trait MailboxChangesMethodContract {
             .addHeader(ACCEPT.toString, ACCEPT_RFC8621_VERSION_HEADER)
             .setBody(request)
             .build, new ResponseSpecBuilder().build)
-          .post
+          .post.prettyPeek()
```
Arsnael, line 892: debug spotted

Arsnael (approved):

Arsnael:
Rebased and squashed fixups

PR #497, opened 2021-06-14
JAMES-3599 Group execution: execute listeners together

chibenwa:
## Why

Every groups deserializes the event, which uses CPU and puts pressure on RabbitMQ.

rabbitmq-management

We can likely mutualize group executions to minimize deserialization, acks, and reduce pressure on RabbitMQ.

Note that we should try to preserve per-groups retries.

## How

design_before

design_after

## Results

Single queue with consumer rate equals publish rate

Screenshot from 2021-06-14 19-51-30

And perf definitl improved!

Screenshot from 2021-06-14 20-22-28


Arsnael (approved):

chibenwa:
## Before

 - RabbitMQ : 3,46% CPU - ACKs 1,43%
 - Play JSON : 3,13% CPU 

## After

 - RabbitMQ : 1,26% CPU - ACKs 0,15%
 - Play JSON : 2,33% CPU 
 - GC also notably decreased....

vttranlina (approved):

PR #499, opened 2021-06-16
Upgrade pdfbox to 2.0.24

Arsnael:
2 CVEs for versions anterior to 2.0.24 regarding pdfbox dependency have been reported:
- https://nvd.nist.gov/vuln/detail/CVE-2021-31812
- https://nvd.nist.gov/vuln/detail/CVE-2021-31811

This upgrade should solve them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2533,7 +2533,7 @@
             <dependency>
                 <groupId>org.apache.pdfbox</groupId>
                 <artifactId>pdfbox</artifactId>
-                <version>2.0.22</version>
+                <version>2.0.24</version>
                 <exclusions>
                     <exclusion>
                         <groupId>commons-logging</groupId>


vttranlina (approved):

chibenwa (approved):
To be honnest I would also welcome work to fully drop the dependency (I bet we use it just for tests?)

Arsnael:
Yes we use it only for tests it seems, for `PDFTextExtractor`. What would be the alternative to drop this?

chibenwa:
I would say using Tika text extractor in the tests relying on it...

Arsnael:
I tried and... well it's a bit more complicated than that. 

It would be good yes to get rid of a dependency, and from what I saw if you remove `PDFTextExtractor` you can ultimately remove `apache-james-mailbox-scanning-search` module, as the only class left would be `SimpleMessageSearchIndexTest` and I don't see why we could not move it to mailbox-store with `SimpleMessageSearchIndex`. 

However, two integration tests are using the `PDFTextExtractor`: 
* RabbitMQAwsS3GetMessageListMethodTest
* MemoryGetMessageListMethodTest

Except that... those tests are in junit4, and the only Tika extension is in junit5. So or we migrate those tests to junit5 possibly, or we create a rule with Tika. But it's not that trivial... 

Will try to dig a bit more on it later.

chibenwa:
Fine with the upgrade, let's not loose more time with this.

PR #504, opened 2021-06-21
JAMES-3516 Add threadId column to cassandra tables

quantranhong1999:
WIP: Added threadId column to messageIdTable and imapUidTable. 
I think I should add threadId to ComposedMessageIdWithMetaData class.

I have some questions:
- What is the meaning of these two tables' name? From my view, it looks like rather mailbox message table.
- What is IMAP Uid?
- I see table messageV2 storing message metadata. Should we add threadId to this table too?


Arsnael:
> I think I should add threadId to ComposedMessageIdWithMetaData class.

Hmmm I would say likely yes, or maybe even in `ComposedMessageId`? Should recheck all the usages... But probably not in this PR though?

> I have some questions:
> 
>     * What is the meaning of these two tables' name? From my view, it looks like rather mailbox message table.

Those are projections. If you look in `CassandraMessageModule` where they are defined, the description says it all:
* `messageIdTable`: Holds mailbox and flags for each message, lookup by mailbox ID + UID
* `imapUidTable`: Holds mailbox and flags for each message, lookup by message ID
 
>     * What is IMAP Uid?

https://datatracker.ietf.org/doc/html/rfc3501#section-2.3.1.1

To be simple here, IMAP UID is the ID message when using IMAP. And we use the message Id for JMAP
 
>     * I see table messageV2 storing message metadata. Should we add threadId to this table too?

messageV2 is not used anymore, it's messageV3 ;) But it would make sense to me to add a field here yes


Oh and don't forget to try adding the migration doc as well later!

quantranhong1999:
>Hmmm I would say likely yes, or maybe even in ComposedMessageId? Should recheck all the usages... But probably not in this PR though?

I see DAO layer parsing ComposeMessageIdWithMetaData to get needed metadata information. I guess I will add threadId to ComposeMessageIdWithMetaData then enable DAO layer parsing threadId. And I will do that in another PR while waiting for both you and Benoit' approvements for this PR.

src/site/xdoc/server/config-cassandra.xml
```diff
@@ -319,6 +319,12 @@
 
     </subsection>
 
+    <subsection name="From V10 to V11">
+
+      <p>Add threadId column to messageIdTable and imapUidTable in order to get a message's threadId.</p>
```
Arsnael, line 324: We should give the commands on Cassandra to perform for that I believe. But I consider that you probably don't know them.

I will try to support you on this part when I have a bit of time.

I think we need to modify the proper section in the new doc with Antora too :)

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/table/MessageIdToImapUid.java
```diff
@@ -29,6 +29,8 @@
 
     String MOD_SEQ = "modSeq";
 
-    String[] FIELDS = { MESSAGE_ID, MAILBOX_ID, IMAP_UID, MOD_SEQ,
+    String THREAD_ID = "threadId";
```
Arsnael, line 32: Maybe I would have added that constant in `CassandraMessageIds` and then import it here and in CassandraMessageIdTable instead

quantranhong1999: We already defined threadId in AppendResult(separately from ComposedMessageId) and MessageMetaData class. IMO it's not suitable if we put threadId into ComposedMessageId or CassandraMessageIds constrants also.

quantranhong1999:
Just a rebased on master to see if there is any failed test somewhere (build failed on CI but was green on local), nothing new added.

chibenwa (commented):
Adding a threadId collumn for adding a thread id collumn is IMO unsufficient. I would like that for Cassandra we read and write things to this newly added threadId collumn...

quantranhong1999:
Rebased to solve conflicts + add 2 commit 
- JAMES-3516 Enable reading and writing to threadId column
- JAMES-3516 Add threadId to ComposedMessageIdWithMetaData

chibenwa (commented):
Can we have an entry in `upgrade_instructions.md` please?

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdDAO.java
```diff
@@ -352,6 +357,7 @@ private PreparedStatement prepareSelectUidRangeLimited(Session session) {
                         MessageUid.of(row.getLong(IMAP_UID))))
                 .flags(FlagsExtractor.getFlags(row))
                 .modSeq(ModSeq.of(row.getLong(MOD_SEQ_LOWERCASE)))
+                .threadId(ThreadId.fromBaseMessageId(messageIdFactory.of(row.getUUID(THREAD_ID))))
```
chibenwa, line 360: Please handle the null case to allow a smooth upgrade

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdToImapUidDAO.java
```diff
@@ -248,6 +253,7 @@ private ComposedMessageIdWithMetaData toComposedMessageIdWithMetadata(Row row) {
                     MessageUid.of(row.getLong(IMAP_UID))))
                 .flags(FlagsExtractor.getFlags(row))
                 .modSeq(ModSeq.of(row.getLong(MOD_SEQ_LOWERCASE)))
+                .threadId(ThreadId.fromBaseMessageId(messageIdFactory.of(row.getUUID(THREAD_ID))))
```
chibenwa, line 256: Please handle the null case to allow a smooth upgrade

upgrade-instructions.md
```diff
@@ -19,7 +19,22 @@ Change list:
  - [Drop Cassandra schema version prior version 8](#drop-cassandra-schema-version-prior-version-8)
  - [Adopt UnboundID as a LDAP library](#adopt-unboundid-as-a-ldap-library)
  - [Review the architecture of the RabbitMQ event bus](#review-the-architecture-of-the-rabbitmq-event-bus)
+ - [Cassandra Schema update to V11](#cassandra-schema-update-to-v11)
  
+### Cassandra Schema update to V11
```
chibenwa, line 24:
```suggestion
### Adding threadId collumn to message metadata tables
```

There's no v11 as we did not define any upgrade task.

upgrade-instructions.md
```diff
@@ -19,7 +19,22 @@ Change list:
...
+
+Date 23/06/2021
+
+JIRA: https://issues.apache.org/jira/browse/JAMES-3516
+
+Concerned product: Distributed James
+
+Add threadId column to messageIdTable and imapUidTable in order to get a message's threadId.
+
+In order to benefit from this work, you need to [upgrade to the latest schema version](https://github.com/apache/james-project/blob/master/src/site/markdown/server/manage-webadmin.md#upgrading-to-the-latest-version).
```
chibenwa, line 34: Wrong. There's no upgrades to do here.

However we need to document how to create the threadId collumn on existing tables (ALTER TABLE ...)

upgrade-instructions.md
```diff
@@ -19,7 +19,22 @@ Change list:
...
+
+A James restart is advised after this migration in order to skip schema version reads.
```
chibenwa, line 36: Nope.

src/site/xdoc/server/config-cassandra.xml
```diff
@@ -319,6 +319,12 @@
 
     </subsection>
 
+    <subsection name="From V10 to V11">
```
chibenwa, line 322:
```suggestion
    <subsection name="Adding threadId collumn to message metadata table">
```

upgrade-instructions.md
Arsnael: It's more the other way around right? You should update the cassandra tables first before upgrading James?

upgrade-instructions.md
Arsnael: By using cqlsh: https://cassandra.apache.org/doc/latest/tools/cqlsh.html

Data manipulation with CQL: https://cassandra.apache.org/doc/latest/cql/dml.html# -> for finding your commands

upgrade-instructions.md
chibenwa: Yes.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdDAO.java
```diff
@@ -357,7 +358,14 @@ private PreparedStatement prepareSelectUidRangeLimited(Session session) {
                         MessageUid.of(row.getLong(IMAP_UID))))
                 .flags(FlagsExtractor.getFlags(row))
                 .modSeq(ModSeq.of(row.getLong(MOD_SEQ_LOWERCASE)))
-                .threadId(ThreadId.fromBaseMessageId(messageIdFactory.of(row.getUUID(THREAD_ID))))
+                .threadId(getThreadIdFromRow(row))
                 .build());
     }
+
+    private ThreadId getThreadIdFromRow(Row row) {
```
chibenwa, line 365: I would prefer the parameters to be `Row row, MessageId messageId` so that we do not read the messageId twice...

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdDAO.java
```diff
@@ -357,7 +358,14 @@ private PreparedStatement prepareSelectUidRangeLimited(Session session) {
...
+        if (row.getUUID(THREAD_ID_LOWERCASE) == null) {
```
chibenwa, line 366: I'd like we do not read THREAD_ID_LOWERCASE twice

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdToImapUidDAO.java
```diff
@@ -253,10 +254,17 @@ private ComposedMessageIdWithMetaData toComposedMessageIdWithMetadata(Row row) {
                     MessageUid.of(row.getLong(IMAP_UID))))
                 .flags(FlagsExtractor.getFlags(row))
                 .modSeq(ModSeq.of(row.getLong(MOD_SEQ_LOWERCASE)))
-                .threadId(ThreadId.fromBaseMessageId(messageIdFactory.of(row.getUUID(THREAD_ID))))
+                .threadId(getThreadIdFromRow(row))
                 .build();
     }
 
+    private ThreadId getThreadIdFromRow(Row row) {
```
chibenwa, line 261: idem

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdToImapUidDAO.java
```diff
@@ -253,10 +254,17 @@ private ComposedMessageIdWithMetaData toComposedMessageIdWithMetadata(Row row) {
...
+        if (row.getUUID(THREAD_ID_LOWERCASE) == null) {
```
chibenwa, line 262: idem

Arsnael (approved):

chibenwa (approved):

chibenwa:
Can you squash the fixups?

PR #509, opened 2021-06-24
JAMES-2366 Add missing tests for forwards to remote recipients

chibenwa:
@tlle14 reported issues (https://github.com/linagora/james-project/issues/4338)  regarding forwards when forwarding to remote recipients with a local copy.

The following test proves James code works and we could review settings of the involved platform.

Cc @Arsnael


chibenwa:
(passed both for master and 9f7425e )

Arsnael (approved):

PR #513, opened 2021-06-28
JAMES-3604 Use publish confirm

chibenwa:
CPU utilisation related to RabbitMQ as expected increased (1.30 % -> 2.6%) however latencies went from 67ms to 90+ ms at ~800 req/s.

RPC related to `Channel::confirmSelect` done repeatedly are rather expensive, and I notice some active creation of channels.

As such, I feel unconfident with the publish / confirm part of the work and would split it to another pull request...

https://www.rabbitmq.com/tutorials/tutorial-seven-java.html

```
Confirms should be enabled just once, not for every message published.
```

Not the case given my implementation.

For now I see several solutions:

  - rely on a sink to pass all messages as part of a single Flux. Thus init is done once. The done side is that the result is asynchronous to the caller (error handling would not be propagated)
  - enable publish confirms globally in ReactorRabbitMQChannelPool and re-implement send operation to skip confirmSelect activation...
  - Wrap channel in a super-class that activated `confirmSelect` only once (AtomicBoolean?)
  - ...
  
See https://github.com/apache/james-project/pull/508#issuecomment-869131638

### DIFF
--- a/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/ReactorRabbitMQChannelPool.java
+++ b/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/ReactorRabbitMQChannelPool.java
@@ -22,8 +22,12 @@ package org.apache.james.backends.rabbitmq;
 import java.io.IOException;
 import java.time.Duration;
 import java.util.Comparator;
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 
 import javax.annotation.PreDestroy;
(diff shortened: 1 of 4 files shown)


chibenwa:
## Before 

RabbitMQ uses 2.6% of CPU, we notice active channel creation and a lot of time is spent doing `confirmSelect`.

Screenshot from 2021-06-28 16-28-21

## After the last 2 commits

We no longer close the connection for nothing and publishConfirms are activated once, and RabbitMQ CPU is down to 2%.

chibenwa:
## Before (No publish confirms)

Screenshot from 2021-06-28 18-01-04


## After (First two commits)

Screenshot from 2021-06-28 18-01-38

## After (First two commit and performance fixes)

Screenshot from 2021-06-28 18-02-22

Arsnael (approved):

chibenwa:
```
james-server-queue-rabbitmq: There was a timeout or other error in the fork -> [Help 1]
```

vttranlina:
I read it

Arsnael:
> ```
> james-server-queue-rabbitmq: There was a timeout or other error in the fork -> [Help 1]
> ```

I've been seeing similar issues a couple of times lately, I'm not sure what to think...

chibenwa:
> I've been seeing similar issues a couple of times in builds lately, I'm not sure what to think...

On this build this was due to a failing test...

PR #515, opened 2021-06-29
JAMES-3516 Plug threadIdGuessingAlgorithm to MessageManager

quantranhong1999:
I have some questions:
- Am I supporting threadId for other protocols beside JMAP also? I am interacting with how James store threadId generally.
- Should we have CassandraThreadIdGuessingAlgorithm and InMemoryThreadIdGuessingAlgorithm... separately in the future?

### DIFF
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageStorer.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageStorer.java
@@ -40,8 +40,11 @@ import org.apache.james.mailbox.store.mail.AttachmentMapperFactory;
 import org.apache.james.mailbox.store.mail.MessageMapper;
 import org.apache.james.mailbox.store.mail.ThreadIdGuessingAlgorithm;
 import org.apache.james.mailbox.store.mail.model.MailboxMessage;
+import org.apache.james.mailbox.store.mail.model.MimeMessageId;
+import org.apache.james.mailbox.store.mail.model.Subject;
 import org.apache.james.mailbox.store.mail.model.impl.MessageParser;
 import org.apache.james.mailbox.store.mail.model.impl.PropertyBuilder;
+import org.apache.james.mailbox.store.mail.utils.MimeMessageHeadersUtil;
 import org.apache.james.mime4j.dom.Message;
 import org.apache.james.mime4j.message.HeaderImpl;
 import org.slf4j.Logger;
@@ -91,8 +94,11 @@ public interface MessageStorer {
         public Mono<Pair<MessageMetaData, Optional<List<MessageAttachmentMetadata>>>> appendMessageToStore(Mailbox mailbox, Date internalDate, int size, int bodyStartOctet, Content content, Flags flags, PropertyBuilder propertyBuilder, Optional<Message> maybeMessage, MailboxSession session, HeaderImpl headers) throws MailboxException {
             MessageMapper messageMapper = mapperFactory.getMessageMapper(session);
             MessageId messageId = messageIdFactory.generate();
-            // TODO get mime message header fields
-            ThreadId threadId = threadIdGuessingAlgorithm.guessThreadId(session.getUser(), messageId, null, null, null, null);
+            Optional<MimeMessageId> mimeMessageId = MimeMessageHeadersUtil.parseMimeMessageId(headers);
+            Optional<MimeMessageId> inReplyTo = MimeMessageHeadersUtil.parseInReplyTo(headers);
+            Optional<List<MimeMessageId>> references = MimeMessageHeadersUtil.parseReferences(headers);
+            Optional<Subject> subject = MimeMessageHeadersUtil.parseSubject(headers);
+            ThreadId threadId = threadIdGuessingAlgorithm.guessThreadId(session.getUser(), messageId, mimeMessageId, inReplyTo, references, subject);
 
             return mapperFactory.getMessageMapper(session)
                 .executeReactive(
@@ -151,8 +157,12 @@ public interface MessageStorer {
         public Mono<Pair<MessageMetaData, Optional<List<MessageAttachmentMetadata>>>> appendMessageToStore(Mailbox mailbox, Date internalDate, int size, int bodyStartOctet, Content content, Flags flags, PropertyBuilder propertyBuilder, Optional<Message> maybeMessage, MailboxSession session, HeaderImpl headers) throws MailboxException {
             MessageMapper messageMapper = mapperFactory.getMessageMapper(session);
             MessageId messageId = messageIdFactory.generate();
-            // TODO get mime message header fields
-            ThreadId threadId = threadIdGuessingAlgorithm.guessThreadId(session.getUser(), messageId, null, null, null, null);
+            Optional<MimeMessageId> mimeMessageId = MimeMessageHeadersUtil.parseMimeMessageId(headers);
+            Optional<MimeMessageId> inReplyTo = MimeMessageHeadersUtil.parseInReplyTo(headers);
+            Optional<List<MimeMessageId>> references = MimeMessageHeadersUtil.parseReferences(headers);
+            Optional<Subject> subject = MimeMessageHeadersUtil.parseSubject(headers);
+            ThreadId threadId = threadIdGuessingAlgorithm.guessThreadId(session.getUser(), messageId, mimeMessageId, inReplyTo, references, subject);
+
             MailboxMessage message = messageFactory.createMessage(messageId, threadId, mailbox, internalDate, size, bodyStartOctet, content, flags, propertyBuilder, ImmutableList.of());
 
             return mapperFactory.getMessageMapper(session)
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageStorer.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageStorer.java
@@ -38,10 +38,12 @@ import org.apache.james.mailbox.model.ParsedAttachment;
 import org.apache.james.mailbox.model.ThreadId;
 import org.apache.james.mailbox.store.mail.AttachmentMapperFactory;
 import org.apache.james.mailbox.store.mail.MessageMapper;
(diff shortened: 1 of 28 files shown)


mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageStorer.java
```diff
@@ -72,22 +73,25 @@
...
             this.messageParser = messageParser;
+            this.threadIdGuessingAlgorithm = threadIdGuessingAlgorithm;
         }
 
         @Override
         public Mono<Pair<MessageMetaData, Optional<List<MessageAttachmentMetadata>>>> appendMessageToStore(Mailbox mailbox, Date internalDate, int size, int bodyStartOctet, Content content, Flags flags, PropertyBuilder propertyBuilder, Optional<Message> maybeMessage, MailboxSession session) throws MailboxException {
             MessageMapper messageMapper = mapperFactory.getMessageMapper(session);
             MessageId messageId = messageIdFactory.generate();
-            ThreadId threadId = ThreadId.fromBaseMessageId(messageId);
+            // TODO get mime message header fields
```
chibenwa, line 93: I think you need a hint there...

MaximalBodyDescriptor do held a list of fields CF this debugging session:

Screenshot from 2021-06-29 16-17-11

We can contribute a `MaximalBodyDescriptor::getFiled(String name)` method in mime4J ?

Then we would likely be able to do threadId guessing easily in `StoreMessageManager::appendMessage` (the one with the big try catch...) Then we can pass the threadId as a parameter to other methods just like properties...

chibenwa: Likely even better (and less hacky) in arseProperties there is a header isntance we do not use...

```
private PropertyBuilder parseProperties(BodyOffsetInputStream bIn) throws IOException, MimeException {
        // Disable line length... This should be handled by the smtp server
        // component and not the parser itself
        // https://issues.apache.org/jira/browse/IMAP-122
        final MimeTokenStream parser = getParser(bIn);
        final HeaderImpl fields = readHeader(parser);
        
        // All we need for threadId guessing <3
        fields.getFields("Message-Id")
        
        final MaximalBodyDescriptor descriptor = (MaximalBodyDescriptor) parser.getBodyDescriptor();
        final MediaType mediaType = getMediaType(descriptor);
        final PropertyBuilder propertyBuilder = getPropertyBuilder(descriptor, mediaType.mediaType, mediaType.subType);
        setTextualLinesCount(parser, mediaType.mediaType, propertyBuilder);
        return propertyBuilder;
    }
 ```

Comment (quantranhong1999) [reply mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageStorer.java:93]: How about using this header instance to get headers fields needed for ThreadIdGuessingAlgorithm, and still place ThreadIdGuessingAlgorithm at MessageStorer::appendMessageToStore as I am doing now? Because I think threadId should go with messageId. Otherwise we should messageId generate out of MessageStorer::appendMessageToStore.

**Comment by quantranhong1999** [comment]
It's failing somewhere else related to JPA, still working on this.
Look like we are missing threadId at JPAMailboxMessage...

File: mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageStorer.java [hunk 2]
```diff
@@ -87,11 +91,14 @@ public WithAttachment(MailboxSessionMapperFactory mapperFactory, MessageId.Facto
         }
 
         @Override
-        public Mono<Pair<MessageMetaData, Optional<List<MessageAttachmentMetadata>>>> appendMessageToStore(Mailbox mailbox, Date internalDate, int size, int bodyStartOctet, Content content, Flags flags, PropertyBuilder propertyBuilder, Optional<Message> maybeMessage, MailboxSession session) throws MailboxException {
+        public Mono<Pair<MessageMetaData, Optional<List<MessageAttachmentMetadata>>>> appendMessageToStore(Mailbox mailbox, Date internalDate, int size, int bodyStartOctet, Content content, Flags flags, PropertyBuilder propertyBuilder, Optional<Message> maybeMessage, MailboxSession session, HeaderImpl headers) throws MailboxException {
             MessageMapper messageMapper = mapperFactory.getMessageMapper(session);
             MessageId messageId = messageIdFactory.generate();
-            // TODO get mime message header fields
-            ThreadId threadId = threadIdGuessingAlgorithm.guessThreadId(session.getUser(), messageId, null, null, null, null);
+            MimeMessageId mimeMessageId = MimeMessageHeadersUtil.parseMimeMessageId(headers);
```
chibenwa, line 97: Optional IMO

mailbox/api/src/test/java/org/apache/james/mailbox/MailboxManagerTest.java
```diff
@@ -144,6 +144,7 @@ void setUp() throws Exception {
         this.subscriptionManager = provideSubscriptionManager();
 
         this.message = Message.Builder.of()
+            .setMessageId("MimeMessageId")
```
chibenwa, line 147: We should IMO accept messages without MimeMessageId

mailbox/store/src/main/java/org/apache/james/mailbox/store/MessageStorer.java
```diff
@@ -147,11 +154,15 @@ public WithoutAttachment(MailboxSessionMapperFactory mapperFactory, MessageId.Fa
...
+            Optional<MimeMessageId> inReplyTo = MimeMessageHeadersUtil.parseInReplyTo(headers);
+            Optional<List<MimeMessageId>> references = MimeMessageHeadersUtil.parseReferences(headers);
+            Subject subject = MimeMessageHeadersUtil.parseSubject(headers);
```
chibenwa, line 163: Idem, what if for any reason a message do not have a subject?

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/utils/MimeMessageHeadersUtil.java
```diff
@@ -0,0 +1,58 @@
...
+import org.apache.james.mime4j.stream.Field;
+
+public class MimeMessageHeadersUtil {
+    public static MimeMessageId parseMimeMessageId(HeaderImpl headers) {
+        return new MimeMessageId(headers.getField("Message-ID").getBody());
+    }
+
+    public static Optional<MimeMessageId> parseInReplyTo(HeaderImpl headers) {
+        Field inReplyToField = headers.getField("In-Reply-To");
+        if (inReplyToField != null) {
```
chibenwa, line 38: Use Optional.ofNullable(...).map(...)

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/utils/MimeMessageHeadersUtil.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+        return Optional.empty();
+    }
+
+    public static Optional<List<MimeMessageId>> parseReferences(HeaderImpl headers) {
+        List<Field> mimeMessageIdFields = headers.getFields("References");
+        if (!mimeMessageIdFields.isEmpty()) {
+            List<MimeMessageId> mimeMessageIdList = mimeMessageIdFields.stream()
+                .map(mimeMessageIdField -> new MimeMessageId(mimeMessageIdField.getBody()))
+                .collect(Collectors.toList());
```
chibenwa, line 49: Guavate.toImmutableList()

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/utils/MimeMessageHeadersUtil.java
```diff
@@ -0,0 +1,58 @@
...
+            return Optional.of(mimeMessageIdList);
+        }
+        return Optional.empty();
+    }
+
+    public static Subject parseSubject(HeaderImpl headers) {
+        return new Subject(headers.getField("Subject").getBody());
```
chibenwa, line 56: `UnstructuredFieldImpl::getValue`

(We need to handle possible encoded characters IMO cf HeaderCollectionTest::getHeadersShouldDecodeValues)

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/utils/MimeMessageHeadersUtil.java
```diff
@@ -21,38 +21,39 @@
...
             return Optional.of(mimeMessageIdList);
         }
         return Optional.empty();
     }
 
-    public static Subject parseSubject(HeaderImpl headers) {
-        return new Subject(headers.getField("Subject").getBody());
+    public static Optional<Subject> parseSubject(HeaderImpl headers) {
+        return Optional.ofNullable(headers.getField("Subject"))
+            .map(field -> UnstructuredFieldImpl.PARSER.parse(field, DecodeMonitor.SILENT))
```
chibenwa, line 56: We do not need to parse the field if it is already a `UnstructuredFieldImpl`

chibenwa (approved):

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/utils/MimeMessageHeadersUtil.java
quantranhong1999: image

Actually we don't parse until we using .getValue, and in that if statement handle the situation you said.

chibenwa:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.22.2:test (default-test) on project apache-james-mailbox-cassandra: There are test failures.

java.lang.NullPointerException: threadId is required (252 occurence...)
```

quantranhong1999:
The related JPA tests fail is real. But these cassandra tests fail seems weird... they are green on local.

image

image

chibenwa:
Pro-tip: git rebase master & rerun ;-)

quantranhong1999:
Rebased on master + squash fixups

Arsnael (approved):

PR #523, opened 2021-07-02
JAMES-3606 Website quick-start.xml should mandate JRE 11

chibenwa:


src/site/xdoc/server/quick-start.xml
```diff
@@ -36,7 +36,7 @@
 Step 0: Requirements
 ####################
 
-  * JRE 1.8
+  * JRE 1.11
```
rouazana, line 39: As far as I know there is no more JRE after Java 10, so you should just name it `JDK 11`

chibenwa, line 39:
```suggestion
  * JDK 1.11
```

src/site/xdoc/server/quick-start.xml
```diff
@@ -36,7 +36,7 @@
 Step 0: Requirements
 ####################
 
-  * JRE 1.8
+  * JDK 1.11
```
chibenwa, line 39:
```suggestion
  * JDK 11
```

rouazana (approved):

vttranlina (approved):

PR #524, opened 2021-07-02
JAMES-3516 Create SearchGuessingAlgorithm and inject it into ThreadIdGuessingAlgorithm

quantranhong1999:
My idea is using MessageSearchIndex to get messageIds of a user, then use MessageIdMapper to find Mailbox Messages based on messageIds.

### DIFF
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/NaiveThreadIdGuessingAlgorithmImpl.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/NaiveThreadIdGuessingAlgorithmImpl.java
@@ -22,15 +22,38 @@ package org.apache.james.mailbox.store.mail;
 import java.util.List;
 import java.util.Optional;
 
+import javax.inject.Inject;
+
 import org.apache.james.core.Username;
+import org.apache.james.mailbox.MailboxSession;
+import org.apache.james.mailbox.exception.MailboxException;
 import org.apache.james.mailbox.model.MessageId;
+import org.apache.james.mailbox.model.MultimailboxesSearchQuery;
+import org.apache.james.mailbox.model.SearchQuery;
 import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MailboxMessage;
 import org.apache.james.mailbox.store.mail.model.MimeMessageId;
 import org.apache.james.mailbox.store.mail.model.Subject;
+import org.apache.james.mailbox.store.search.SearchGuessingAlgorithm;
+
+import reactor.core.publisher.Flux;
 
 public class NaiveThreadIdGuessingAlgorithmImpl implements ThreadIdGuessingAlgorithm {
+    private final SearchGuessingAlgorithm searchGuessingAlgorithm;
+
+    @Inject
+    public NaiveThreadIdGuessingAlgorithmImpl(SearchGuessingAlgorithm searchGuessingAlgorithm) {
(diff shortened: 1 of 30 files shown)


mailbox/store/src/main/java/org/apache/james/mailbox/store/search/SearchGuessingAlgorithmImpl.java
```diff
@@ -0,0 +1,108 @@
...
+        return searchMessageIds(expression, session, limit)
+            .flatMap(messageId -> messageIdMapper.findReactive(Collections.singletonList(messageId), fetchType));
+    }
+
+    @Override
+    public Flux<MessageId> searchMessageIds(MultimailboxesSearchQuery expression, MailboxSession session, long limit) throws MailboxException {
+        return getInMailboxIds(expression, session)
+            .filter(id -> !expression.getNotInMailboxes().contains(id))
+            .collect(Guavate.toImmutableSet())
+            .flatMapMany(Throwing.function(ids -> index.search(session, ids, expression.getSearchQuery(), limit)));
```
chibenwa, line 71: Use mailboxManager for the search...

mailbox/store/src/main/java/org/apache/james/mailbox/store/search/SearchGuessingAlgorithm.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.james.mailbox.MailboxSession;
+import org.apache.james.mailbox.exception.MailboxException;
+import org.apache.james.mailbox.model.MessageId;
+import org.apache.james.mailbox.model.MultimailboxesSearchQuery;
+import org.apache.james.mailbox.store.mail.MessageMapper;
+import org.apache.james.mailbox.store.mail.model.MailboxMessage;
+
+import reactor.core.publisher.Flux;
+
+public interface SearchGuessingAlgorithm {
```
chibenwa, line 31: I though implementing ThreadIdGuessingAlgorithm...

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/NaiveThreadIdGuessingAlgorithmImpl.java
```diff
@@ -22,15 +22,38 @@
...
+import org.apache.james.mailbox.model.SearchQuery;
 import org.apache.james.mailbox.model.ThreadId;
+import org.apache.james.mailbox.store.mail.model.MailboxMessage;
 import org.apache.james.mailbox.store.mail.model.MimeMessageId;
 import org.apache.james.mailbox.store.mail.model.Subject;
+import org.apache.james.mailbox.store.search.SearchGuessingAlgorithm;
+
+import reactor.core.publisher.Flux;
 
 public class NaiveThreadIdGuessingAlgorithmImpl implements ThreadIdGuessingAlgorithm {
```
chibenwa, line 41: No keep this class unchanged and implement another class for finding the threadId by doing searches...

mailbox/store/src/main/java/org/apache/james/mailbox/store/search/SearchGuessingAlgorithmImpl.java
quantranhong1999: That would be a loop: MailboxManager need ThreadIdGuessingAlgorithm and ThreadIdGuessingAlgorithm need MailboxManager

chibenwa: Injection systems like guice are designed to break these loops.

PR #534, opened 2021-07-09
JAMES-3516 Implement MailboxManager::getThread API and plug it into JMAP Thread/get

quantranhong1999:
WIP: Plug getThread API into JMAP Thread/get

### DIFF
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
@@ -34,6 +34,7 @@ import org.apache.james.mailbox.model.MailboxPath;
 import org.apache.james.mailbox.model.MessageId;
 import org.apache.james.mailbox.model.MessageRange;
 import org.apache.james.mailbox.model.MultimailboxesSearchQuery;
+import org.apache.james.mailbox.model.ThreadId;
 import org.apache.james.mailbox.model.search.MailboxQuery;
 import org.reactivestreams.Publisher;
 
@@ -323,6 +324,16 @@ public interface MailboxManager extends RequestAware, RightManager, MailboxAnnot
      *            the context for this call, not null
      */
     Publisher<MessageId> search(MultimailboxesSearchQuery expression, MailboxSession session, long limit) throws MailboxException;
+
+    /**
+     *
+     * @param threadId
+     *          target Thread
+     * @param session
+     *          the context for this call, not null
+     * @return  a list of MessageId of messages belong to that Thread
+     */
+    Publisher<MessageId> getThread(ThreadId threadId, MailboxSession session) throws MailboxException;
 
     /**
      * Does the given mailbox exist?
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
@@ -323,7 +323,7 @@ public interface MailboxManager extends RequestAware, RightManager, MailboxAnnot
      * @param session
      *            the context for this call, not null
      */
-    Publisher<MessageId> search(MultimailboxesSearchQuery expression, MailboxSession session, long limit) throws MailboxException;
+    Publisher<MessageId> search(MultimailboxesSearchQuery expression, MailboxSession session, long limit);
 
     /**
      * Returns the list of MessageId of messages belonging to that Thread
@@ -334,7 +334,7 @@ public interface MailboxManager extends RequestAware, RightManager, MailboxAnnot
      *          the context for this call, not null
      * @return  the list of MessageId of messages belonging to that Thread
      */
-    Publisher<MessageId> getThread(ThreadId threadId, MailboxSession session) throws MailboxException;
+    Publisher<MessageId> getThread(ThreadId threadId, MailboxSession session);
 
     /**
      * Does the given mailbox exist?
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
@@ -326,12 +326,13 @@ public interface MailboxManager extends RequestAware, RightManager, MailboxAnnot
     Publisher<MessageId> search(MultimailboxesSearchQuery expression, MailboxSession session, long limit) throws MailboxException;
 
     /**
+     * Returns the list of MessageId of messages belonging to that Thread
      *
      * @param threadId
      *          target Thread
      * @param session
      *          the context for this call, not null
-     * @return  a list of MessageId of messages belong to that Thread
+     * @return  the list of MessageId of messages belonging to that Thread
      */
     Publisher<MessageId> getThread(ThreadId threadId, MailboxSession session) throws MailboxException;
 
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/exception/ThreadNotFoundException.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/exception/ThreadNotFoundException.java
@@ -22,10 +22,6 @@ package org.apache.james.mailbox.exception;
 import org.apache.james.mailbox.model.ThreadId;
 
 public class ThreadNotFoundException extends MailboxException {
-    public ThreadNotFoundException(String message) {
-        super(message);
-    }
-
     public ThreadNotFoundException(ThreadId threadId) {
         super("Thread " + threadId.getBaseMessageId().serialize() + " can not be found");
     }
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/exception/ThreadNotFoundException.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/exception/ThreadNotFoundException.java
@@ -17,23 +17,16 @@
  * under the License.                                             *
  ******************************************************************/
 
-package org.apache.james.mailbox.store.mail;
+package org.apache.james.mailbox.exception;
 
-import java.util.List;
-import java.util.Optional;
-
-import org.apache.james.mailbox.MailboxSession;
-import org.apache.james.mailbox.exception.MailboxException;
-import org.apache.james.mailbox.model.MessageId;
 import org.apache.james.mailbox.model.ThreadId;
-import org.apache.james.mailbox.store.mail.model.MimeMessageId;
-import org.apache.james.mailbox.store.mail.model.Subject;
 
-import reactor.core.publisher.Flux;
-import reactor.core.publisher.Mono;
+public class ThreadNotFoundException extends MailboxException {
+    public ThreadNotFoundException(String message) {
+        super(message);
+    }
 
-public interface ThreadIdGuessingAlgorithm {
-    Mono<ThreadId> guessThreadIdReactive(MessageId messageId, Optional<MimeMessageId> thisMimeMessageId, Optional<MimeMessageId> inReplyTo, Optional<List<MimeMessageId>> references, Optional<Subject> subject, MailboxSession session) throws MailboxException;
-
-    Flux<MessageId> getMessageIdsInThread(ThreadId threadId, MailboxSession session) throws MailboxException;
+    public ThreadNotFoundException(ThreadId threadId) {
+        super("Thread " + threadId.getBaseMessageId().serialize() + " can not be found");
+    }
 }
--- a/mailbox/scanning-search/src/test/java/org/apache/james/mailbox/store/search/SearchThreadIdGuessingAlgorithmTest.java
+++ b/mailbox/scanning-search/src/test/java/org/apache/james/mailbox/store/search/SearchThreadIdGuessingAlgorithmTest.java
@@ -19,20 +19,30 @@
 
 package org.apache.james.mailbox.store.search;
 
+import org.apache.james.mailbox.MailboxSession;
 import org.apache.james.mailbox.inmemory.InMemoryCombinationManagerTestSystem;
+import org.apache.james.mailbox.inmemory.InMemoryMailboxManager;
 import org.apache.james.mailbox.inmemory.InMemoryMessageId;
+import org.apache.james.mailbox.inmemory.mail.InMemoryMapperProvider;
 import org.apache.james.mailbox.inmemory.manager.InMemoryIntegrationResources;
 import org.apache.james.mailbox.model.MessageId;
 import org.apache.james.mailbox.store.CombinationManagerTestSystem;
 import org.apache.james.mailbox.store.ThreadIdGuessingAlgorithmContract;
+import org.apache.james.mailbox.store.mail.MessageMapper;
 import org.apache.james.mailbox.store.mail.SearchThreadIdGuessingAlgorithm;
 import org.apache.james.mailbox.store.mail.ThreadIdGuessingAlgorithm;
+import org.apache.james.mailbox.store.mail.model.MapperProvider;
 
 public class SearchThreadIdGuessingAlgorithmTest extends ThreadIdGuessingAlgorithmContract {
+    InMemoryMailboxManager mailboxManager;
 
     @Override
     protected CombinationManagerTestSystem createTestingData() {
         InMemoryIntegrationResources resources = InMemoryIntegrationResources.defaultResources();
+
+        mailboxManager = resources.getMailboxManager();
+        eventBus = resources.getEventBus();
+        messageIdFactory = resources.getMessageIdFactory();
 
         return new InMemoryCombinationManagerTestSystem(
             resources.getMailboxManager(),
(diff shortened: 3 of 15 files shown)


mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
```diff
@@ -324,6 +325,16 @@ public final int hashCode() {
...
 
+    /**
+     *
+     * @param threadId
+     *          target Thread
+     * @param session
+     *          the context for this call, not null
+     * @return  a list of MessageId of messages belong to that Thread
+     */
+    Flux<MessageId> getThread(ThreadId threadId, MailboxSession session) throws MailboxException;
```
chibenwa, line 336: Use Publisher instead of Mono / Flux in TLD API

This allows subcomponents to not be dependant of the reactive technology...

chibenwa: No tests for this method?

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/SearchThreadIdGuessingAlgorithm.java
```diff
@@ -67,6 +67,18 @@ public SearchThreadIdGuessingAlgorithm(MailboxManager mailboxManager, MessageIdM
...
     }
 
+    @Override
+    public Flux<MessageId> getMessageIdsInThread(ThreadId threadId, MailboxSession session) throws MailboxException {
+        MultimailboxesSearchQuery expression = MultimailboxesSearchQuery.from(SearchQuery.matchAll()).build();
+
+        return Flux.from(mailboxManager.search(expression, session, Long.MAX_VALUE))
+            .collectList()
+            .flatMapMany(messageIds -> messageIdManager.getMessagesReactive(messageIds, FetchGroup.MINIMAL, session))
+            .filter(messageResult -> messageResult.getThreadId().equals(threadId))
```
chibenwa, line 77: TODO (Another PR)

Push this to the search engine.

How: 

`SearchQuery::withThreadId` and implement it for scanning search, lucene search as well as ElasticSearch.

mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
quantranhong1999: IMO we are testing ThreadIdGuessingAlgorithm in an inconvenient way. We depend on guice to resolve the loop of SearchThreadIdGuessingAlgorithm and MailboxManager, so we can test this from HTTP level. But we are having problems when testing ThreadIdGuessingAlgorithm APIs from the lower level (testing resource based on loop constructors). 

My proposal solutions:
- In the lower level tests: Using MessageMapper to save our custom messages with custom threadId (But we missed some actual guessing threadId)
OR
- SearchThreadIdGuessingAlgorithm depends on MailboxManger no more.

Personally, I would like to do something like inbox.appendMessage, and then getThread directly.
What do you think?

chibenwa: Or...

You can implement MailboxManager  by doing directly the search without calling the guessing algorithm?

quantranhong1999: At the moment, we just use searching for InMemory, we still need NaiveThreadIdGuessingAlgorithm for Cassandra and OpenJPA. And I would like to package related ThreadId APIs in ThreadIdGuessingAlgorithm.

What about recoding MailboxManager::search in SearchThreadIdGuessingAlgorithm? (a little duplicated though, but solve above problem)

chibenwa: No code duplication is not a solution.

mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
```diff
@@ -324,6 +325,16 @@ public final int hashCode() {
      */
     Publisher<MessageId> search(MultimailboxesSearchQuery expression, MailboxSession session, long limit) throws MailboxException;
 
+    /**
+     *
```
chibenwa, line 329: Returns the list of MessageId of messages belonging to that Thread

mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
```diff
@@ -324,6 +325,16 @@ public final int hashCode() {
...
+     * @param threadId
+     *          target Thread
+     * @param session
+     *          the context for this call, not null
+     * @return  a list of MessageId of messages belong to that Thread
```
chibenwa, line 334: * @return  the list of MessageId of messages belonging to that Thread

mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
```diff
@@ -333,7 +333,7 @@ public final int hashCode() {
      *          the context for this call, not null
      * @return  a list of MessageId of messages belong to that Thread
      */
-    Flux<MessageId> getThread(ThreadId threadId, MailboxSession session) throws MailboxException;
+    Publisher<MessageId> getThread(ThreadId threadId, MailboxSession session) throws MailboxException;
```
chibenwa, line 336: Methods with publishers should not throw.

They should return publisher that eventually completes with an error (`Flux.error`)

The error is returned when blocking the publisher not while generating the publisher.

mailbox/api/src/main/java/org/apache/james/mailbox/exception/ThreadNotFoundException.java
```diff
@@ -0,0 +1,32 @@
...
+ ******************************************************************/
+
+package org.apache.james.mailbox.exception;
+
+import org.apache.james.mailbox.model.ThreadId;
+
+public class ThreadNotFoundException extends MailboxException {
+    public ThreadNotFoundException(String message) {
+        super(message);
+    }
```
chibenwa, line 27: IMO that string method is likely not needed...

mailbox/scanning-search/src/test/java/org/apache/james/mailbox/store/search/SearchThreadIdGuessingAlgorithmTest.java
```diff
@@ -19,21 +19,29 @@
...
 import org.apache.james.mailbox.model.MessageId;
 import org.apache.james.mailbox.store.CombinationManagerTestSystem;
 import org.apache.james.mailbox.store.ThreadIdGuessingAlgorithmContract;
+import org.apache.james.mailbox.store.mail.MessageMapper;
 import org.apache.james.mailbox.store.mail.SearchThreadIdGuessingAlgorithm;
 import org.apache.james.mailbox.store.mail.ThreadIdGuessingAlgorithm;
+import org.apache.james.mailbox.store.mail.model.MapperProvider;
 
 public class SearchThreadIdGuessingAlgorithmTest extends ThreadIdGuessingAlgorithmContract {
+    InMemoryMailboxManager mailboxManager;
```
chibenwa, line 37: private?

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/SearchThreadIdGuessingAlgorithm.java
```diff
@@ -71,12 +72,12 @@ public SearchThreadIdGuessingAlgorithm(MailboxManager mailboxManager, MessageIdM
...
 
-        return Flux.from(mailboxManager.search(expression, session, Long.MAX_VALUE))
+        return Flux.from(mailboxManager.search(expression, session, Integer.MAX_VALUE))
             .collectList()
             .flatMapMany(messageIds -> messageIdManager.getMessagesReactive(messageIds, FetchGroup.MINIMAL, session))
             .filter(messageResult -> messageResult.getThreadId().equals(threadId))
-            .map(messageResult -> messageResult.getMessageId());
-
+            .map(messageResult -> messageResult.getMessageId())
+            .switchIfEmpty(Mono.error(new ThreadNotFoundException(threadId)));
```
chibenwa, line 80: .switchIfEmpty(Mono.error(() -> new ThreadNotFoundException(threadId)));

mailbox/store/src/test/java/org/apache/james/mailbox/store/ThreadIdGuessingAlgorithmContract.java
```diff
@@ -202,4 +231,41 @@ void givenOldMailWhenAddNonRelatedMailsThenGuessingThreadIdShouldBasedOnGenerate
...
+        MailboxMessage message3 = createMessage(mailboxId, ThreadId.fromBaseMessageId(newBasedMessageId));
+
+        Mailbox mailbox = new Mailbox(MailboxPath.inbox(USER), UID_VALIDITY, mailboxId);
+        messageMapper.add(mailbox, message1);
+        messageMapper.add(mailbox, message2);
+        messageMapper.add(mailbox, message3);
+
+        Flux<MessageId> messageIds = testee.getMessageIdsInThread(ThreadId.fromBaseMessageId(newBasedMessageId), mailboxSession);
+
+        assertThat(messageIds.collectList().block()).isEqualTo(List.of(message1.getMessageId(), message2.getMessageId(), message3.getMessageId()));
```
chibenwa, line 247: Indent

chibenwa: Use guava ImmutableList.of()

mailbox/store/src/test/java/org/apache/james/mailbox/store/ThreadIdGuessingAlgorithmContract.java
```diff
@@ -202,4 +231,41 @@ void givenOldMailWhenAddNonRelatedMailsThenGuessingThreadIdShouldBasedOnGenerate
...
+            threadId,
+            new Date(),
+            content.length(),
+            bodyStart,
+            new ByteContent(content.getBytes()),
+            new Flags(),
+            new PropertyBuilder().build(),
+            mailboxId);
+    }
 }
```
chibenwa, line 271: One mail in a tread?

mail 1, mail 3 in thread 1
mail 2 in thread 2

get thread 1 -> mail 1 + mail 3 (not related emails are not returned)

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/ThreadGetContract.scala
```diff
@@ -27,10 +29,27 @@ import org.apache.http.HttpStatus.SC_OK
...
+import org.apache.james.mailbox.MessageManager
+import org.apache.james.mailbox.model.MailboxPath
+import org.apache.james.mime4j.dom.Message
+import org.apache.james.mime4j.stream.RawField
+import org.apache.james.modules.MailboxProbeImpl
 import org.apache.james.utils.DataProbeImpl
 import org.junit.jupiter.api.{BeforeEach, Test}
 
+object ThreadGetContract {
+  private def createTestMessage: Message = Message.Builder
```
chibenwa, line 42: Unused :D

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/ThreadGetContract.scala
```diff
@@ -113,34 +140,57 @@ trait ThreadGetContract {
...
+  def givenMailsBelongToAThreadThenGetThatThreadShouldReturnExactThreadObject(server: GuiceJamesServer): Unit = {
+    val bobPath = MailboxPath.inbox(BOB)
+    server.getProbe(classOf[MailboxProbeImpl]).createMailbox(bobPath)
+
+    // given 2 mail with same thread
+    val message1: MessageManager.AppendResult = server.getProbe(classOf[MailboxProbeImpl])
+      .appendMessageAndGetAppendResult(BOB.asString(), bobPath,
+        MessageManager.AppendCommand.from(Message.Builder.of.setSubject("Test")
+        .setMessageId("Message-ID")
+          .setField(new RawField("In-Reply-To", "someInReplyTo"))
```
chibenwa, line 168: First message don't need the `In-Reply-To` field :D

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/ThreadGetContract.scala
```diff
@@ -160,16 +210,24 @@ trait ThreadGetContract {
...
+           |				"notFound": [
+           |
+           |				]
+           |			},
+           |			"c1"
+           |		]
+           |	]
+           |}""".stripMargin)
   }
 }
```
chibenwa, line 233: - In-Reply-To references messsage1 but different subject -> NOMATCH
 - REFERENCES -> match
  -Maybe 1-2 other fonctional tests.

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/ThreadGetMethod.scala
```diff
@@ -1,52 +1,80 @@
...
  * regarding copyright ownership.  The ASF licenses this file   *
  * to you under the Apache License, Version 2.0 (the            *
  * "License"); you may not use this file except in compliance   *
  * with the License.  You may obtain a copy of the License at   *
- *                                                              *
- *   http://www.apache.org/licenses/LICENSE-2.0                 *
- *                                                              *
+ * *
+ * http://www.apache.org/licenses/LICENSE-2.0                 *
+ * *
```
chibenwa, line 11: Oups you changed that license format

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/ThreadGetMethod.scala
```diff
@@ -1,52 +1,80 @@
...
+      accountId = accountId,
+      state = UuidState.INSTANCE,
+      list = threads.toList,
+      notFound = notFound)
+}
 
 class ThreadGetMethod @Inject()(val metricFactory: MetricFactory,
-                                          val sessionSupplier: SessionSupplier) extends MethodRequiringAccountId[ThreadGetRequest] {
+                                val sessionSupplier: SessionSupplier,
+                                val messageIdFactory: MessageId.Factory,
```
chibenwa, line 64: Please introduce a ThreadId.Factory class here (ThreadId.Factory would of course call the MessageId.Factory)

This reduces temporal coupling...

chibenwa:
@quantranhong1999 can you please rebase this work?

quantranhong1999:
> @quantranhong1999 can you please rebase this work?

I don't fix all reviews yet (some tests to be written). But ok I will rebase and fix other reviews later.
BTW, I think we should merge the search query first then we can plug it into this pr.

mailbox/api/src/main/java/org/apache/james/mailbox/MailboxManager.java
quantranhong1999: This method depends on MailboxManager::search which throws MailboxException. I guess there is nothing I can do with that.

Arsnael: I think those search methods were not reactive before and that throws stayed in the api interface. You can remove it I think. If you look at the current implementations of MailboxManager, they don't seem to throw it either

Arsnael: And yes there is other publisher methods still doing this too... we should clean up a bit :)

Arsnael (commented):
Read it, and added a comment in an existing thread

quantranhong1999:
We still did not handle **ids: null** case. But I guess we can come back later when we have an API get all Thread.

chibenwa:
The spec misses the mention `The ids argument may be null to fetch all at once.` for Thread as for `Email` so `ids = null ` is not to be suported.

When it is the case it is explicitly stated, as for `Mailbox` https://jmap.io/spec-mail.html#mailboxget

quantranhong1999:
Just rebase and drop https://github.com/apache/james-project/pull/539 NO REVIEW commit.
3 lastest commits: one solves Rene's comment, two write more tests.

Arsnael (approved):

chibenwa (approved):
Looks good to me.

chibenwa:
Please also squash the fixups...

PR #541, opened 2021-07-15
[PERFORMANCE] DropWizardMetricFactory: optimize wrapping monos 

chibenwa:
## Before

Screenshot from 2021-07-15 11-54-37

A good part of applicative CPU is used calling cancels on fluxes to get back a mono.

Zoom:

Screenshot from 2021-07-15 11-55-55

Here is the associated gatling result:

Screenshot from 2021-07-15 11-56-43

## After

Most of the cancels are gone (remains the ones of the authentication strategy)

Screenshot from 2021-07-15 11-57-45

Which translates with better gating perfs (~10% mean time and ~7% p99 decrease, ~0.5% throughput increase)

Screenshot from 2021-07-15 11-59-46


Arsnael (approved):

vttranlina (commented):
I read it

PR #546, opened 2021-07-22
JAMES-3544 Blob store should allow to list bucket

chibenwa:
This allow bucket based expiration to be in synchronisation with the underlying data store

### DIFF
--- a/server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/BucketNameResolverTest.java
+++ b/server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/BucketNameResolverTest.java
@@ -25,10 +25,25 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import org.apache.james.blob.api.BucketName;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 class BucketNameResolverTest {
     @Nested
     class EmptyPrefix {
+        @ParameterizedTest
+        @ValueSource(strings = {"namespace", "any", "bucketPrefix-aaa", "bucketPrefix-"})
+        void withShouldAddNewValuesInSet(String bucketNameString) {
+            BucketNameResolver resolver = BucketNameResolver.builder()
+                .noPrefix()
+                .namespace(BucketName.of("namespace"))
+                .build();
+
+            BucketName bucketName = BucketName.of(bucketNameString);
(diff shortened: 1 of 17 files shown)


vttranlina (approved):

server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/BucketNameResolverTest.java
```diff
@@ -50,10 +65,45 @@ void resolveShouldReturnValueWhenNamespace() {
...
+                .noPrefix()
+                .namespace(BucketName.of("namespace"))
+                .build();
+
+            assertThat(resolver.unresolve(BucketName.of("bucketName")))
+                .contains(BucketName.of("bucketName"));
+        }
+
+        @Test
+        void unresolveShouldReturnValueWhenNamespace() {
```
Arsnael, line 81: Isn't that test doing the same as one of the values in the parameterized test above `withShouldAddNewValuesInSet`?

server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/BucketNameResolverTest.java
```diff
@@ -50,10 +65,45 @@ void resolveShouldReturnValueWhenNamespace() {
             assertThat(resolver.resolve(BucketName.of("namespace")))
                 .isEqualTo(BucketName.of("namespace"));
         }
+
+        @Test
+        void unresolveShouldReturnPassedValue() {
```
Arsnael, line 70: Isn't that case checked above in the parameterized test `withShouldAddNewValuesInSet`?

Arsnael: In a more general manner, isn't this the same for tests in other nested classes?

chibenwa: > In a more general manner, isn't this the same for tests in other nested classes?

Yes but un top of different bucket resolver configuration and it (might or might not) yield different results.

Being very sure of the behaviour of that class is very important to me, I assume 100% having a very detailed test suite here.

chibenwa: Ok got your point

chibenwa: For the mathematical background https://en.wikipedia.org/wiki/Inverse_function <3

Arsnael (approved):

PR #553, opened 2021-07-24
JAMES-3618 Support LDAP for JPA guice based apps

chibenwa:
Includes:
 - JPA App
 - JPA-SMTP App

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1358,6 +1358,12 @@
             </dependency>
             <dependency>
                 <groupId>${james.groupId}</groupId>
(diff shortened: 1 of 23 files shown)


mbaechler (approved):
Looks good
I'm wondering if we should split data-api in several parts, it bundles very different concerns.

chibenwa:
I would tend to agree. 

At the very least for Sieve and MailRepository related classes. RRT, users and domains can in my opinion be considered some kind of identity components.

Arsnael (approved):

PR #555, opened 2021-07-27
JAMES-3614 The homepage should comply with the ASF release policy

chibenwa:
- Download page in the top links

Screenshot from 2021-07-27 16-35-41

 - Reference the getting started from downloads guide before the docker hands-on

Screenshot from 2021-07-27 16-35-48

 - We still keep a link to downloads inside the community block

Screenshot from 2021-07-27 16-36-00

### DIFF
--- a/src/homepage/index.html
+++ b/src/homepage/index.html
@@ -51,9 +51,8 @@ layout: default
         <li><a href="#roadmap">Roadmap</a></li>
         <li><a href="#posts">Last Posts</a></li>
         <li><a href="#second">Community</a></li>
-        <li><a href="#third">Contribute</a></li>
-        <li><a href="#use-cases">Use-cases</a></li>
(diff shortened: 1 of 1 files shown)


src/homepage/index.html
```diff
@@ -51,9 +51,8 @@ <h2>Emails at the heart of your business logic</h2>
         <li><a href="#roadmap">Roadmap</a></li>
         <li><a href="#posts">Last Posts</a></li>
         <li><a href="#second">Community</a></li>
-        <li><a href="#third">Contribute</a></li>
-        <li><a href="#use-cases">Use-cases</a></li>
```
Arsnael, line 55: Why getting rid of those two top links though?

chibenwa: To have ore space on the top link.

I find it uggly otherwise.

Both IMO are linked to community.

Arsnael (approved):

vttranlina (approved):

PR #563, opened 2021-07-30
JAMES-3620 Memory leak at org.apache.james.protocols.smtp.core.Abstra…

chibenwa:
…ctHookableCmdHandler


Arsnael (approved):

vttranlina (approved):
another way with the way that you wrote in Jira. but better (y)

PR #629, opened 2021-08-30
JAMES-3642 Fix Buggy behavior with Android Email mobile application

chibenwa:
### Actual behaviour

```
GIVEN Samsung Email mobile application
AND connected to an IMAP account located on an IMAP James server
AND load you emails (swap down if needed)
WHEN Refresh your emails (swap down a second time)
THEN the email account is rendered empty, only emails received since are displayed
```

Another refresh restores the original email list....

### Explanation

The second refresh uses QRESYNC SELECT command with the last known modification sequence modifier.

Thus James needs to compute the vanished responses (messages deletes since the last known modseq).

As James do not store expunges, we should interate the supplied range and identify the missing items, for which we will send vanished responses.

But James specifically filters out old messages from the liveness search, which results in all messages prior to the last known modification sequence to be considered deleted, effectively removing them from client cahe.

### The fix

Remove this buggy modseq condition: we should not answer that all old messages are deleted!

Also to avoid a full scan we can detect the deletions from the UID <-> MSN mapping of the mailbox we just selected (yeah).

### Alternative

We could add a projection table tracking expunged messages - along with the modseq.

However we should store all the history of deletions - otherwize some values as last known modseq would be rejected, and no client behaviour is specified in the spec if so.

Also storing even more data have a bitter-sweat taste to me...

### DIFF
--- a/protocols/imap/src/test/java/org/apache/james/imap/processor/SelectProcessorTest.java
+++ b/protocols/imap/src/test/java/org/apache/james/imap/processor/SelectProcessorTest.java
@@ -128,7 +128,7 @@ class SelectProcessorTest {
     }
 
     @Test
-    void vanishedResponsesShouldBeSentWheDeletes() throws Exception {
+    void vanishedResponsesShouldBeSentWhenDeletes() throws Exception {
         FakeImapSession session = new FakeImapSession();
         session.authenticated();
         session.setMailboxSession(mailboxSession);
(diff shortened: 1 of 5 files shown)


protocols/imap/src/test/java/org/apache/james/imap/processor/SelectProcessorTest.java
```diff
@@ -0,0 +1,173 @@
...
+                new DefaultImapEncoderFactory(new DefaultLocalizer(), true).buildImapEncoder(),
+                new ImapResponseComposerImpl(new OutputStreamImapResponseWriter(outputStream))),
+            session);
+
+        assertThat(new String(outputStream.toByteArray()))
+            .doesNotContain("* VANISHED (EARLIER) 1:4");
+    }
+
+    @Test
+    void vanishedResponsesShouldBeSentWheDeletes() throws Exception {
```
Arsnael, line 131:
```suggestion
    void vanishedResponsesShouldBeSentWhenDeletes() throws Exception {
```

Arsnael (approved):

PR #630, opened 2021-08-31
JAMES-3643 Local part resolution with virtual hosting

chibenwa:


server/data/data-cassandra/src/main/java/org/apache/james/user/cassandra/CassandraLocalPartLookupRepository.java
```diff
@@ -0,0 +1,69 @@
+package org.apache.james.user.cassandra;
```
Arsnael, line 1: Missing license

server/data/data-cassandra/src/main/java/org/apache/james/user/cassandra/LocalPartLookupModule.java
```diff
@@ -0,0 +1,21 @@
+package org.apache.james.user.cassandra;
```
Arsnael, line 1: Missing license

mailbox/store/src/main/java/org/apache/james/mailbox/store/SessionProviderImpl.java
```diff
@@ -59,16 +62,14 @@ public MailboxSession createSystemSession(Username userName) {
...
-        }
+        return validLogin(userid, passwd)
+            .map(Throwing.function(username -> createSession(username, MailboxSession.SessionType.User)))
+            .orElseThrow(BadCredentialsException::new);
     }
 
     @Override
     public MailboxSession loginAsOtherUser(Username adminUserid, String passwd, Username otherUserId) throws MailboxException {
-        if (! isValidLogin(adminUserid, passwd)) {
+        if (! validLogin(adminUserid, passwd).map(adminUserid::equals).orElse(false)) {
```
Arsnael, line 72:
```suggestion
        if (!validLogin(adminUserid, passwd).map(adminUserid::equals).orElse(false)) {
```

server/data/data-jpa/src/main/java/org/apache/james/user/jpa/JPAUsersDAO.java
```diff
@@ -103,6 +104,28 @@ public void init() {
...
+        EntityManager entityManager = entityManagerFactory.createEntityManager();
+
+        try {
+            List<JPAUser> results = entityManager
+                .createNamedQuery("findUserByLocalPart", JPAUser.class)
+                .setParameter("name", localPart.asString() + "@%")
+                .getResultList();
+            return results.stream()
+                .map(JPAUser::getUserName)
+                .collect(ImmutableList.toImmutableList());
```
Arsnael, line 118: Can't do it in one shot? Like:

```
            return entityManager
                .createNamedQuery("findUserByLocalPart", JPAUser.class)
                .setParameter("name", localPart.asString() + "@%")
                .getResultList()
                .stream()
                .map(JPAUser::getUserName)
```

chibenwa: Yes we could

Arsnael (approved):

chibenwa:
There seems to be little interest in the community, this work is out of date. Abort!

PR #648, opened 2021-09-08
[Documentation] Fix missing content-type in upgrade schema version co…

Arsnael:
…mmand

If the content type is not specified with curl, application/x-www-form-urlencoded is the default one, and then the request fails.

### DIFF
--- a/src/site/markdown/server/manage-webadmin.md
+++ b/src/site/markdown/server/manage-webadmin.md
@@ -2042,7 +2042,7 @@ Response codes:
 ### Upgrading to a specific version
 
 ```
-curl -XPOST http://ip:port/cassandra/version/upgrade -d '3'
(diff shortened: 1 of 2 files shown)


chibenwa (approved):

src/site/markdown/server/manage-webadmin.md
```diff
@@ -2042,7 +2042,7 @@ Response codes:
 ### Upgrading to a specific version
 
 ```
-curl -XPOST http://ip:port/cassandra/version/upgrade -d '3'
+curl -XPOST -H "Content-Type: application/json http://ip:port/cassandra/version/upgrade -d '3'
```
chibenwa, line 2045: Can you also  update it in /docs ?

Arsnael: Good catch, done!

chibenwa (approved):

PR #655, opened 2021-09-14
JAMES-3588 Mailet to propagate encoutered error

chibenwa:
Can be configured in the mailet container error processor to propagate errors to the calling context. Usefull for instance for error management with LMTP

EG:

```
        <processor state="error" enableJmx="true">
            <mailet match="All" class="MetricsMailet">
                <metricName>mailetContainerErrors</metricName>
            </mailet>
            <!-- If you want to notify the sender of the failure -->
            <mailet match="All" class="Bounce">
                <onMailetException>ignore</onMailetException>
            </mailet>
            <!-- Propagation of error to calling context - can be used to fail the LMTP transaction -->
            <mailet match="All" class="PropagateError">
                <onMailetException>propagate</onMailetException>
            </mailet>
        </processor>
```


Arsnael (commented):
No tests for this mailet?

Arsnael (approved):

PR #665, opened 2021-09-21
JAMES-3150 Better handle massive deletions as part of BloomFilterGCAl…

chibenwa:
…gorithm

Before we were deleting 128 objjects in parrallel.

This approach proved to be slow: we took 25 hours for 35M deletions...

With this proposed changeset we delete 1000 blobs per requests,
16 in parallel, using a S3 API to delete several objects at
once.


Arsnael (approved):

vttranlina (approved):
With AWS S3, they limit the request rate (throw an exception if too much request in time). 
Some other cloud vendors may be the same. 
So, I have a bit of thinking here

Arsnael:
> With AWS S3, they limit the request rate (throw an exception if too much request in time).
> Some other cloud vendors may be the same.
> So, I have a bit of thinking here

Not sure about how it works exactly but I would expect the S3 java client being used having a retry mechanism for such issue? If not we would have a lot of blobs disappearing I guess not only there?

PR #681, opened 2021-10-04
JAMES-3657 Generalise entity validation Upon creation

chibenwa:
Goal: apply entity validation when using CLI too:

```
GIVEN a group toto@domain.tld
WHEN I try to create the user toto@domain.tld using the CLI
THEN it fails

GIVEN an alias toto@domain.tld
WHEN I try to create the user toto@domain.tld using the CLI
THEN it fails
```

Move the validation logic to `data-library`.

Sorry for the Spring related hacks - you know how much I would love to remove Spring altogether!


server/container/spring/src/main/java/org/apache/james/container/spring/resource/SpringUserEntityValidator.java
```diff
@@ -0,0 +1,44 @@
+package org.apache.james.container.spring.resource;
```
Arsnael, line 1: license missing

Arsnael (approved):

Arsnael (approved):

PR #702, opened 2021-10-18
JAMES-3516 Small Scala fixes  to Thread/get

chibenwa:
Cc @quantranhong1999


quantranhong1999:
Read it. Approved if I could :+1:

PR #715, opened 2021-10-28
ability to assert mock emails without possible race condition

grault:
// reopening https://github.com/linagora/james-project/pull/4421 here as per @chibenwa 's request

We use the mock smtp to assert emails sent out from our test environments. The GET/DELETE api is very nice, but there's no guarantee that you don't get emails between these calls. It could be that by the time you call DELETE, you got an extra email that you haven't observed via GET.

One change in this PR is to amend DELETE to return the deleted emails, so you can be sure that you only delete emails that you can later observe. It's decided to be done this way to maximize backward compatibility for people already using DELETE.

We also poll number of emails and not necessarily want to retrieve all the emails every time and count them. For this, an extra GET endpoint is added which just returns the number of emails currently available in the repository.

For us, it would be nice to have these use cases covered by your upstream repo and then we wouldn't have to monitor for changes and rebase our patch on top of them. So in case you're not against covering these use cases, we're happy to contribute resolving changes in a different form/implementation if needed.

### DIFF
--- a/server/mailet/mock-smtp-server/src/main/java/org/apache/james/mock/smtp/server/HTTPConfigurationServer.java
+++ b/server/mailet/mock-smtp-server/src/main/java/org/apache/james/mock/smtp/server/HTTPConfigurationServer.java
@@ -244,7 +244,7 @@ public class HTTPConfigurationServer {
     }
 
     private Publisher<Void> getMailsCount(HttpServerRequest req, HttpServerResponse res) {
-        var count = receivedMailRepository.list().stream().count();
+        int count = receivedMailRepository.count();
 
         try {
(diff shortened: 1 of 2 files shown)


grault:
> I do think a comprehensive documentation of the HTTP endpoints would make it easier for people to re-use it. Just in case you are interested this could be a really appreciated second contribution to apache/james, if you have some spare time :-p

What about extending https://github.com/apache/james-project/tree/master/server/mailet/mock-smtp-server#readme with further one liners, using `curl`? `curl` speaks http and smtp. Showcasing the endpoints?

chibenwa:
Looks like a great idea to me!

PR #730, opened 2021-11-02
JAMES-3666 Fix DSNBounce exception when no Date header is present

ottoka:
When the DSNBounce mailet processes a mail without a Date header, it fails with the exception

`java.time.temporal.UnsupportedTemporalTypeException: Unsupported field: OffsetSeconds`

because the LocalDateTime used to construct the new date does not include timezone information required by the date formatter.

The fix is to construct the new date via ZonedDateTime instead.

There is actually a unit test for this, but the MimeMessageBuilder adds a Date header by default if not specified. The test fix is to explicitly remove the Date header after building.

### DIFF
--- a/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/DSNBounce.java
+++ b/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/DSNBounce.java
@@ -23,7 +23,6 @@ import static org.apache.james.transport.mailets.remote.delivery.Bouncer.DELIVER
 
 import java.net.InetAddress;
 import java.net.UnknownHostException;
-import java.time.LocalDateTime;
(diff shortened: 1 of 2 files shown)


chibenwa (approved):

Arsnael:
Nice contribution thanks!

Wondering why there is twice the exact same commit though?

chibenwa:
> Wondering why there is twice the exact same commit though?

My plan was to squash it all upon merge ;-)

> There is actually a unit test for this, but the MimeMessageBuilder adds a Date header by default if not specified. The test fix is to explicitly remove the Date header after building.

We might want to audit the code base, and MIME related tests regarding this defect. If it happened one time, we might have used a similar pattern in other places of the code and felt safe...

PR #743, opened 2021-11-10
JAMES-2557 Sieve should cleanup email after sending them

chibenwa:


Arsnael (approved):

Arsnael:
Hmmm... Is this correct? It seems you separated the work from https://github.com/apache/james-project/pull/742 but there is an extra fixup! on that commit in that PR that I don't see reflected here?

chibenwa:
Likely a mistake on #742

PR #744, opened 2021-11-11
JAMES-3668 Extra system properties from configuration file

ottoka:
Many aspects of JVM execution for James can be tweaked by system properties. However, a lot of them will clutter the command line and can become difficult to manage.

The solution is to read a properties file and set the contents as system properties. Since this can affect various class initializations, it should be done as soon as possible, e.g. immediately as first line of the server main() method.

This proposal adds a small utility class for this purpose. It will read a configuration file "conf/jvm.properties" if present, otherwise do nothing. The location of the file may be overridden by a system property "extra.props" for quick testing.

### DIFF
--- a/server/container/guice/common/src/main/java/org/apache/james/ExtraProperties.java
+++ b/server/container/guice/common/src/main/java/org/apache/james/ExtraProperties.java
@@ -1,0 +1,44 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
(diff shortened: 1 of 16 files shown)


chibenwa (commented):
Can we ship a sample configuration in `server/apps/**/sample-configuration`?

Also, we might want to switch the way JIB loads system properties (`server/apps/**/pom.xml` plugins)

Similary we might want to document this way of setting system properties in `server/apps/distributed-app/docs/???` and likely in src/site/xdoc/server/config-system???`

ottoka:
I am not sure what to put into the sample files, the stuff we have in there are highly project specific and likely won't help anybody else. I could put a mostly empty one with a comment saying "put any system properties here that you don't want on the command line". Any thoughts?

I don't think the JIB jvmFlags are a good fit, as they are a critical part of system setup that people should not override. 

I do agree that there should be documentation for this feature though, I will write some and deliver it in the next push.

ottoka:
Now with documentation (TM)

server/container/guice/common/src/main/java/org/apache/james/ExtraProperties.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.FileNotFoundException;
+import java.io.IOException;
+
+public class ExtraProperties {
+
+    public static final String OVERRIDE_PROPERTY = "extra.props";
+    public static final String DEFAULT_PATH = "conf/jvm.properties";
+
+    private ExtraProperties() {
+    }
```
chibenwa, line 32: We tend not to write such private constructor to prevent instanciation of the class.

chibenwa (approved):

Arsnael (approved):
Nice contribution, thanks a lot :)

server/container/guice/common/src/main/java/org/apache/james/ExtraProperties.java
ottoka: OK, some do some don't. I removed it to keep things James style.

chibenwa (approved):

PR #746, opened 2021-11-12
JAMES-3669 Delay on authentication failure

ottoka:
For standalone James installations, there should be some basic protection against people/bots abusing James as a password oracle for brute-force/dictionary attacks. This needs to be enforced in a central location, so it affects all of the various protocols supported by James.

This proposal adds an option verifyFailureDelay to usersrepository.xml, which delays the response if someone tries to authenticate with a non-existing user or wrong password. There is intentionally no distinction between these two cases, so it also covers username guessing attacks.

Introducing this feature should not affect existing James installations, so the default is 0 delay/disabled.

T-Shirt size S.

### DIFF
--- a/server/apps/distributed-app/docs/modules/ROOT/pages/configure/usersrepository.adoc
+++ b/server/apps/distributed-app/docs/modules/ROOT/pages/configure/usersrepository.adoc
@@ -36,6 +36,9 @@ recipient local part than as login for different protocols.
 |user's name. Allow a user to access to the https://tools.ietf.org/html/rfc4616#section-2[impersonation command],
 acting on the behalf of any user.
 
+| verifyFailureDelay
+| Delay after a failed authentication attempt with an invalid user name or password. Duration string defaulting to seconds, e.g. `2`, `2s`, `2000ms`. Default `0s` (disabled).
+
 |===
 
 == Configuring a LDAP
--- a/server/data/data-library/src/main/java/org/apache/james/user/lib/UsersRepositoryImpl.java
+++ b/server/data/data-library/src/main/java/org/apache/james/user/lib/UsersRepositoryImpl.java
@@ -19,6 +19,7 @@
 
 package org.apache.james.user.lib;
 
+import java.time.temporal.ChronoUnit;
 import java.util.Iterator;
 import java.util.Optional;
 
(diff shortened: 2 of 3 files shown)


server/apps/distributed-app/docs/modules/ROOT/pages/configure/usersrepository.adoc
```diff
@@ -36,6 +36,9 @@ recipient local part than as login for different protocols.
 |user's name. Allow a user to access to the https://tools.ietf.org/html/rfc4616#section-2[impersonation command],
 acting on the behalf of any user.
 
+| verifyFailureDelay
+| Delay after a failed authentication attempt with an invalid user name or password. Duration string defaulting to seconds. No delay if 0 or omitted.
```
chibenwa, line 40: Please indicate units being used, which unit is used if none is specified, and also what is the default value (`0`).

chibenwa: Also add this property within `src/site/xdoc/server/config-usersrepository.xml`...

ottoka: Uh, it does say all these things...? But maybe some examples would be more helpful. i.e.: 2, 2s, 2000ms
And of course I always forget about the other documentation location...

ottoka: Pushed, hope this is better?

server/apps/distributed-app/docs/modules/ROOT/pages/configure/usersrepository.adoc
```diff
@@ -36,6 +36,9 @@ recipient local part than as login for different protocols.
 |user's name. Allow a user to access to the https://tools.ietf.org/html/rfc4616#section-2[impersonation command],
 acting on the behalf of any user.
 
+| verifyFailureDelay
+| Delay after a failed authentication attempt with an invalid user name or password. Duration string defaulting to seconds, e.g. `2`, `2s`, `2000ms`. Default `0s`.
```
chibenwa, line 40:
```suggestion
| Delay after a failed authentication attempt with an invalid user name or password. Duration string defaulting to seconds, e.g. `2`, `2s`, `2000ms`. Default `0s`, which disables the feature.
```

ottoka: Makes sense. I will put (disabled) there, and also in the xdoc.

chibenwa (approved):

Arsnael (approved):

PR #747, opened 2021-11-15
JAMES-3534 Cassandra implementation for CustomIdentityDAO

quantranhong1999:
resolve https://github.com/linagora/james-project/issues/4424

### DIFF
--- a/server/data/data-jmap-cassandra/src/main/scala/org/apache/james/jmap/cassandra/identity/CassandraCustomIdentityModule.scala
+++ b/server/data/data-jmap-cassandra/src/main/scala/org/apache/james/jmap/cassandra/identity/CassandraCustomIdentityModule.scala
@@ -1,0 +1,48 @@
+/** **************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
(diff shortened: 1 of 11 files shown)


server/data/data-jmap-cassandra/src/main/scala/org/apache/james/jmap/cassandra/identity/CassandraCustomIdentityModule.scala
```diff
@@ -0,0 +1,48 @@
...
+
+object CassandraCustomIdentityModule {
+  val MODULE: CassandraModule = CassandraModule.builder()
+    .`type`(EMAIL_ADDRESS)
+    .statement(statement => statement
+      .addColumn(EmailAddress.NAME, text())
+      .addColumn(EmailAddress.EMAIL, text()))
+
+    .table(CassandraCustomIdentityTable.TABLE_NAME)
+    .comment("Hold user custom identities data")
```
chibenwa, line 36: Mention Jmap rfc-8621 in this comment imo...

chibenwa (approved):
Looks overall ok to me.

Some work could have been done to enable commutativiry of updates on distinct fields (in a concurrent environment) but this is likely overkill for identities and would complexify the code a lot. So better written like this.

Commutativity: A x B = B x A

Here A -> update replyTo and B -> update bcc

Concurrent scenario exemple

Read A 
Read B
Update A
Update B

With the current code A update of replyTo would be lost. By only updating the fields we need we could imagine keeping both updates (but it likely is more complicated!)

Arsnael (approved):

PR #748, opened 2021-11-15
JAMES-3670 : Configurable restore location for deleted messages

ottoka:
James should have an option to specify which mailbox receives any restored messages from the Deleted Messages Vault, i.e. something other than the default of 'Restored-Messages'.

This is necessary to use the Deleted Messages Vault in combination with POP3, since POP3 users cannot access any other mailboxes than 'INBOX'.

The solution is to add an extra configuration option 'restoreLocation' within deletedMessagesVault.properties, which can be injected into the RestoreService for this purpose.

T-Shirt size S.

### DIFF
--- a/server/container/guice/mailbox-plugin-deleted-messages-vault/src/main/java/org/apache/james/modules/vault/DeletedMessageVaultModule.java
+++ b/server/container/guice/mailbox-plugin-deleted-messages-vault/src/main/java/org/apache/james/modules/vault/DeletedMessageVaultModule.java
@@ -29,7 +29,7 @@ import com.google.inject.Scopes;
 public class DeletedMessageVaultModule extends AbstractModule {
     @Override
     protected void configure() {
-        install(new DeletedMessageVaultRetentionModule());
+        install(new DeletedMessageVaultConfigurationModule());
         install(new VaultTaskSerializationModule());
 
         bind(BucketNameGenerator.class).in(Scopes.SINGLETON);
--- a/server/protocols/webadmin/webadmin-mailbox-deleted-message-vault/src/test/java/org/apache/james/webadmin/vault/routes/DeletedMessagesVaultRoutesTest.java
+++ b/server/protocols/webadmin/webadmin-mailbox-deleted-message-vault/src/test/java/org/apache/james/webadmin/vault/routes/DeletedMessagesVaultRoutesTest.java
@@ -111,7 +111,7 @@ import org.apache.james.utils.UpdatableTickingClock;
 import org.apache.james.vault.DeletedMessage;
 import org.apache.james.vault.DeletedMessageVault;
 import org.apache.james.vault.DeletedMessageZipper;
-import org.apache.james.vault.RetentionConfiguration;
+import org.apache.james.vault.VaultConfiguration;
 import org.apache.james.vault.blob.BlobStoreDeletedMessageVault;
 import org.apache.james.vault.blob.BlobStoreVaultGarbageCollectionTaskAdditionalInformationDTO;
 import org.apache.james.vault.blob.BucketNameGenerator;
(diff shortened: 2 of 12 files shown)


server/container/guice/mailbox-plugin-deleted-messages-vault/src/main/java/org/apache/james/modules/vault/DeletedMessagesVaultRestoreModule.java
```diff
@@ -0,0 +1,54 @@
...
+import com.google.inject.AbstractModule;
+import com.google.inject.ConfigurationException;
+import com.google.inject.Provides;
+
+public class DeletedMessagesVaultRestoreModule extends AbstractModule {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DeletedMessagesVaultRestoreModule.class);
+
+    @Override
+    protected void configure() {
+    }
```
chibenwa, line 41: We likely do not need this empty configure method...

server/container/guice/mailbox-plugin-deleted-messages-vault/src/main/java/org/apache/james/modules/vault/DeletedMessageVaultModule.java
```diff
@@ -30,6 +30,7 @@
     @Override
     protected void configure() {
         install(new DeletedMessageVaultRetentionModule());
+        install(new DeletedMessagesVaultRestoreModule());
```
chibenwa, line 33: Why creating a seperate module instead of putting the binding here?

server/protocols/webadmin/webadmin-mailbox-deleted-message-vault/src/test/java/org/apache/james/webadmin/vault/routes/DeletedMessagesVaultRoutesTest.java
```diff
@@ -198,7 +199,7 @@ void beforeEach() throws Exception {
         taskManager = new MemoryTaskManager(new Hostname("foo"));
         JsonTransformer jsonTransformer = new JsonTransformer();
 
-        RestoreService vaultRestore = new RestoreService(vault, mailboxManager);
+        RestoreService vaultRestore = new RestoreService(vault, mailboxManager, RestoreLocationConfiguration.DEFAULT);
```
chibenwa, line 202: Shouldn't we merge `RestoreLocation` and `RetentionConfiguration` pojos into a single DeletedMessageVaultConfiguration pojo with 2 fields?

chibenwa:
(Note that i am globally fine with the code here, just proposing tiny enhancements to improve it...)

server/protocols/webadmin/webadmin-mailbox-deleted-message-vault/src/test/java/org/apache/james/webadmin/vault/routes/DeletedMessagesVaultRoutesTest.java
ottoka: I was keeping them separate since they are used in different locations in the code base. And initially did not want to introduce more change impact than necessary, while it all was a private patch. But of course I can merge it all together now, which will also eliminate the extra module you commented on above.

server/container/guice/mailbox-plugin-deleted-messages-vault/src/main/java/org/apache/james/modules/vault/DeletedMessagesVaultRestoreModule.java
ottoka: Huh, missed that one, will remove it.

server/protocols/webadmin/webadmin-mailbox-deleted-message-vault/src/test/java/org/apache/james/webadmin/vault/routes/DeletedMessagesVaultRoutesTest.java
chibenwa: I am globally ok with the patch as off today but would prefer a single pojo ;-)

ottoka: Done, see latest push.

chibenwa (approved):

Arsnael (approved):
Nice thanks for this :)

PR #749, opened 2021-11-15
JAMES-3671 : Glowroot instrumentation for POP3 protocol

ottoka:
For better analysis capability in POP3 related issues, James should have glowroot support for this protocol as well.

This is considerably easier with a minor refactoring, making all POP3 command handlers extend an AbstractPOP3CommandHandler which unifies their onCommand methods.

On this basis we can provide a simple glowroot plugin (pop3.json) for each James app variant.

T-Shirt size S.

### DIFF
--- a/protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3CommandHandler.java
+++ b/protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3CommandHandler.java
@@ -1,0 +1,31 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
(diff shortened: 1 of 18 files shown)


protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3CommandHandler.java
```diff
@@ -0,0 +1,30 @@
...
+
+package org.apache.james.protocols.pop3.core;
+
+import org.apache.james.protocols.api.Request;
+import org.apache.james.protocols.api.Response;
+import org.apache.james.protocols.pop3.POP3Session;
+
+public abstract class AbstractPOP3CommandHandler {
+    
+    public abstract Response onCommand(POP3Session session, Request request);
```
chibenwa, line 28: For readers not familiar with this, gloroot struggles implementing interfaces, hence the abstract class...

(Correct me if I am wrong)

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3CommandHandler.java
```diff
@@ -0,0 +1,30 @@
...
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.protocols.pop3.core;
+
+import org.apache.james.protocols.api.Request;
+import org.apache.james.protocols.api.Response;
+import org.apache.james.protocols.pop3.POP3Session;
+
+public abstract class AbstractPOP3CommandHandler {
```
chibenwa, line 26: Can't we have this extends CommandHandler<T> ?

That would avoid repeating this info in every pop3 class definition...

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3CommandHandler.java
ottoka: Yes excatly.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3CommandHandler.java
ottoka: Agreed, this is somewhat redundant now, will fix it.

chibenwa (approved):

Arsnael (approved):

PR #750, opened 2021-11-16
JAMES-3672 : Support TLS authentication via client certificate

ottoka:
In order to limit access to trusted partners/users only, James should support TLS with certificate-based client authentication.

For this purpose, TLS configuration is extended with the desired authentication mode (none, optional, required), and the associated trust store to validate any received client certificates. Example:

```
<tls socketTLS="true" startTLS="false">
...
  <clientAuth required="true">
    <truststore>file://conf/truststore</truststore>
    <truststoreType>JKS</truststoreType>
    <truststoreSecret>yoursecret</truststoreSecret>
  </clientAuth>
</tls>
```

This is implemented mostly in `AbstractConfigurableAsyncServer` and associated Netty infrastructure.

T-Shirt size M.


ottoka:
BTW: When I created the unit test for this, I noticed that apparently there is no code that uses the <algorithm> and <provider> settings. Did I miss something? Are the documentation and samples outdated? Did the relevant code get lost at some point? I am confused. Could someone more experienced with the James code base have a look please?

protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractSSLAwareChannelPipelineFactory.java
```diff
@@ -64,6 +66,12 @@ public ChannelPipeline getPipeline() throws Exception {
             if (enabledCipherSuites != null && enabledCipherSuites.length > 0) {
                 engine.setEnabledCipherSuites(enabledCipherSuites);
             }
+            if (Boolean.TRUE.equals(clientAuth)) {
+                engine.setNeedClientAuth(true);
+            }
+            if (Boolean.FALSE.equals(clientAuth)) {
+                engine.setWantClientAuth(true);
+            }
```
chibenwa, line 74:
```suggestion
            if (clientAuth) {
                engine.setNeedClientAuth(true);
            } else {
                engine.setWantClientAuth(true);
            }
```

Looks simpler?

protocols/netty/src/main/java/org/apache/james/protocols/netty/BasicChannelUpstreamHandler.java
```diff
@@ -215,6 +215,14 @@ protected ProtocolSession createSession(ChannelHandlerContext ctx) throws Except
             if (enabledCipherSuites != null && enabledCipherSuites.length > 0) {
                 engine.setEnabledCipherSuites(enabledCipherSuites);
             }
+            Boolean clientAuth = secure.getClientAuth();
+            if (Boolean.TRUE.equals(clientAuth)) {
+                engine.setNeedClientAuth(true);
+            }
+            if (Boolean.FALSE.equals(clientAuth)) {
+                engine.setWantClientAuth(true);
+            }
```
chibenwa, line 224: Idem

```suggestion
            if (clientAuth) {
                engine.setNeedClientAuth(true);
            } else {
                engine.setWantClientAuth(true);
            }
```

?

server/apps/distributed-app/docs/modules/ROOT/pages/configure/ssl.adoc
```diff
@@ -84,6 +84,29 @@ Please note `JKS` keystore format is also supported (default value if no keystor
 </tls>
 ....
 
+When you enable TLS, you may also configure the server to request a client certificate for authentication:
```
chibenwa, line 87:
```suggestion
=== Client authentication via certificates

When you enable TLS, you may also configure the server to request a client certificate for authentication:
```

We might like to have a dedicated (sub) section dedicated to client authentication...

server/protocols/protocols-library/src/test/resources/testServerPlain.xml
```diff
@@ -0,0 +1,9 @@
+<testerver enabled="true">
+    <jmxName>testserver-custom</jmxName>
+    <bind>0.0.0.0:25</bind>
```
chibenwa, line 3:
```suggestion
    <bind>0.0.0.0:0</bind>
```

Bonding a non fixed port is cleaner as it ensures the build can run several time concurrently on the same computer.

Port can then be retrieved programmatically in the tests.

chibenwa (commented):
Question: Is client authentication also applied upon STARTTLS or is it omly applies to plain TLS?

src/site/xdoc/server/config-ssl-tls.xml
```diff
@@ -109,6 +109,29 @@
 &lt;/tls&gt;
 </source>
 
+      <p>When you enable TLS, you may also configure the server to request a client certificate for authentication:</p>
```
chibenwa, line 112: Idem a distinct sub-section could be nice.

protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractSSLAwareChannelPipelineFactory.java
ottoka: Actually this is a tri-state logic: null=none=no client auth, false=want=optional client auth, true=need=mandatory client auth. So clientAuth can be null here, and in that case we must not call either of the engine.setXXXClientAuth(true).

Admittedly this is a somewhat ugly API in the JDK, there is no good way around it. I can rename clientAuth to clientAuthRequired to maybe make things a bit clearer.

protocols/netty/src/main/java/org/apache/james/protocols/netty/BasicChannelUpstreamHandler.java
ottoka: see above re tri-state logic

server/protocols/protocols-library/src/test/resources/testServerPlain.xml
ottoka: OK. This is supposed to be a configuration test only, so the address is never used anyway. I would have liked to have something like 1.2.3.4:5 here, but since the bind address is unfortunately not testable, I might as well keep it all zeroes.

server/apps/distributed-app/docs/modules/ROOT/pages/configure/ssl.adoc
ottoka: Sure, I will update the documentation accordingly.

src/site/xdoc/server/config-ssl-tls.xml
ottoka: sure thing.

ottoka:
> Question: Is client authentication also applied upon STARTTLS or is it omly applies to plain TLS?

The point is that the TLS negotiation will fail and prevent the connection if the client provides an unknown/invalid certificate, or none at all (in need mode, required=true). So you could use it with StartTLS as well if that is what you want to happen. Admittedly it makes more sense in a private mail network where you can use straight SMTPS on port 465 from the start.

protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractSSLAwareChannelPipelineFactory.java
ottoka: See latest push.

chibenwa:
`Maybe a future extension could somehow make the client certificate chain available to matchers and mailets, so they can e.g. allow relaying only on connections with trusted client certificates.`

Yes sure, carrying information from the protocols to the mailet context can easily be done via mail attributes if needed.

ottoka:
@chibenwa  Do you have any insights on the  `algorithm` and `provider` properties I mentioned above?

chibenwa:
> @chibenwa Do you have any insights on the algorithm and provider properties I mentioned above?

Yes and I forgot to reply on this.

This sounds like legacy to me, and I am in favor of removing the two options from the documentation.

ottoka:
OK, I will remove the comment from the unit test.
What about the documentation fix, seems to me like this should be a separate issue/PR? Should I create one, or will you do this?

server/protocols/protocols-library/src/test/resources/testServerDefaults.xml
```diff
@@ -0,0 +1,2 @@
+<testerver enabled="true">
```
Arsnael, line 1:
```suggestion
<testserver enabled="true">
```

Idem for all the other xml files :)

Arsnael (commented):
Just some minor typo comments, but I'm globally ok with this work thanks :)

server/protocols/protocols-library/src/test/resources/testServerDefaults.xml
ottoka: Sigh, copy and paste strikes again. Thanks for spotting this.

Arsnael (approved):

protocols/api/src/main/java/org/apache/james/protocols/api/Encryption.java
```diff
@@ -29,15 +29,17 @@
     private final SSLContext context;
     private final boolean starttls;
     private final String[] enabledCipherSuites;
+    private Boolean clientAuth;
 
-    private Encryption(SSLContext context, boolean starttls, String[] enabledCipherSuites) {
+    private Encryption(SSLContext context, boolean starttls, String[] enabledCipherSuites, Boolean clientAuth) {
```
chibenwa, line 34: Maybe an enum would be better than a Boolean.

Especially an enum would enable to explicit the usage of `null` that is unclear to me.

server/apps/distributed-app/docs/modules/ROOT/pages/configure/ssl.adoc
```diff
@@ -84,6 +84,32 @@ Please note `JKS` keystore format is also supported (default value if no keystor
...
+    <truststore>file://conf/truststore</truststore>
+    <truststoreType>JKS</truststoreType>
+    <truststoreSecret>yoursecret</truststoreSecret>
+  </clientAuth>
+</tls>
+....
+
+With `required="true"`, the server refuses any client connections that do not include a client certificate.
+
+With `required="false"`, the servers requests an optional client certificate, but also accepts connections without one.
```
chibenwa, line 108: What is the benefit of `required=false` ?

chibenwa:
Sorry I am back from 01 week of vacation so I am having a new, deeper look at this work ;-)

server/apps/distributed-app/docs/modules/ROOT/pages/configure/ssl.adoc
ottoka: It is one of the options you have in the TLS handshake protocol. The idea is that servers could allow different levels of access whether they decided to authenticate or not.
Right now this does not make much sense in James. But someone could implement a way to access the client certificate (if any) from the protocol session, e.g. to implement SASL EXTERNAL authentication mode. Or even from mailets to do whatever.
Now that I think about it, it is probably easier to just remove the option for now; if someone needs it later they can easily add it again.

protocols/api/src/main/java/org/apache/james/protocols/api/Encryption.java
ottoka: Agreed, enums kind of are the Java way to express tri-state or multi-state logic. But as discussed below, I will remove the unused optional authentication feature for now, so this will become a simple `boolean`.

chibenwa: Even with too states enums brings in strong typing thus better express the intent and should rather be used.

(Refs: Clean code book, chapter I don't remember but clearly boolean arguments should be discouraged.)

ottoka: Fixed in latest push.

server/apps/distributed-app/docs/modules/ROOT/pages/configure/ssl.adoc
ottoka: Fixed in latest push. I decided to keep the three states in an enum because that's how TLS works, but removed the option for WANT from the configuration. Now its either NEED with `<clientAuth>...</clientAuth>` ot NONE without this XML element.

ottoka:
No problem, I hope you enjoyed your vacation :-)

protocols/api/src/main/java/org/apache/james/protocols/api/Encryption.java
```diff
@@ -46,13 +48,15 @@ public static Encryption createTls(SSLContext context) {
...
      */
-    public static Encryption createTls(SSLContext context, String[] enabledCipherSuites) {
-        return new Encryption(context, false, enabledCipherSuites);
+    public static Encryption createTls(SSLContext context, String[] enabledCipherSuites, ClientAuth clientAuth) {
+        return new Encryption(context, false, enabledCipherSuites, clientAuth);
     }
 
     public static Encryption createStartTls(SSLContext context) {
-        return createStartTls(context, null);
+        return createStartTls(context, null, null);
```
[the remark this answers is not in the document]
ottoka: Oops, missed this one.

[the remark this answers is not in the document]
ottoka: Aaand fixed.

chibenwa (commented):
As documented here, as a user I would understand the use of client certificate would span accross all mail protocols (SMTP, POP3, ManageSieve, IMAP)

However the changeset only enable client certificate verification when creating the SSL engine for SMTP.

A quick IDE inspection on `Encryption.getContext` shows usages in `IMAPServer` and `ManageSieveServer`. I deduce that those call paths are unaffected by the change.

Going a bot deeper on the code, it looks like the `SSLContext`  returned by the `Encryption` class are not ready to work with. After calling `encryption.getContext`, we then need to modify (if required) the cypher suites used by the SSL context, and with your change we also need to specify the mode of client auth. This is an issue asit is unclear to the caller of `Encryption.getContext` what to do with the context, and settings secified while building the Encryption class MIGHT only be applied if the caller follow a very specific set of actions, that ends up duplicated. This anti-pattern is called temporal coupling. This subtilty is easy to miss, and we 

My recommendation is to have the `Encryption.getContext` method applying the cypher suite and client auth extra configuration before returning it to the caller, thus the caller can safely be ignorant to these details. Ideally, getting read of the getters on cyphers and client auth should also be possible. (I would highly welcome such a refactoring)

chibenwa:
> No problem, I hope you enjoyed your vacation :-)

A lot!

(Sorry still picky on this one, Netty layer is a bit legacy, thus changes there are not the easiest one to conduct out!)

ottoka:
> Sorry still picky on this one, Netty layer is a bit legacy, thus changes there are not the easiest one to conduct out!
That's fine, anything that improves code quality is a good thing. And cooperation on such issues is why we do all this in the first place, right? :-)

protocols/netty/src/main/java/org/apache/james/protocols/netty/BasicChannelUpstreamHandler.java
```diff
@@ -215,6 +216,14 @@ protected ProtocolSession createSession(ChannelHandlerContext ctx) throws Except
...
                 engine.setEnabledCipherSuites(enabledCipherSuites);
             }
+            ClientAuth clientAuth = secure.getClientAuth();
+            if (ClientAuth.NEED.equals(clientAuth)) {
+                engine.setNeedClientAuth(true);
+            }
+            if (ClientAuth.WANT.equals(clientAuth)) {
+                engine.setWantClientAuth(true);
+            }
+            LOGGER.info("start SMTP with TLS client auth need {} want {}", engine.getNeedClientAuth(), engine.getWantClientAuth());
```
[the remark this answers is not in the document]
ottoka: Right, and it was supposed to be debug.

protocols/netty/src/main/java/org/apache/james/protocols/netty/NettyServer.java
```diff
@@ -122,6 +122,7 @@ protected ChannelPipelineFactory createPipelineFactory(ChannelGroup group) {
             maxCurConnectionsPerIP,
             group,
             secure != null ? secure.getEnabledCipherSuites() : null,
+            secure != null ? secure.getClientAuth() : null,
```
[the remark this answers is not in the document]
ottoka: Right, just did it this way to mimic the line above.

ottoka:
> My recommendation is to have the Encryption.getContext method applying the cypher suite and client auth extra configuration before returning it to the caller, thus the caller can safely be ignorant to these details. Ideally, getting read of the getters on cyphers and client auth should also be possible. (I would highly welcome such a refactoring)

Seems like a sensible solution to me. But should I do the refactoring in this PR, or is it too far beyond its scope and should be a separate PR?

ottoka:
In the latest push I went with the second option. `Encryption` now has a method `createSSLEngine()` which applies the settings, and the object is passed through the protocol/implementation stack to wherever an engine needs to be created. This avoids the temporal coupling anti-pattern you mentioned. As a nice benefit it also should make things easier in case someone needs more settings to apply to the resulting SSLEngine.

chibenwa (approved):

Arsnael (approved):

PR #758, opened 2021-11-23
JAMES-3225 Force a dependency cleanup on the build phase in the Jenki…

Arsnael:
We have more and more corrupted maven packages in the local repo on some workers it seems lately, crashing builds very quickly. The goal here is to force the update of maven packages to avoid corruptions.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,7 +86,7 @@ pipeline {
         stage('Build') {
             steps {
                 echo 'Building'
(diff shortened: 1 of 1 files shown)


Arsnael:
That does not seem to work on the concerned node H27 that this PR is using atm: https://github.com/apache/james-project/pull/752#pullrequestreview-812239498, I still have the same error build

PR #759, opened 2021-11-24
JAMES-3675 : Generalize vacation handling

ottoka:
Vacation handling in James is tightly bound to JMAP, where it is a mandatory part of the specification. However, this feature is actually of general use. It should be possible to manage vacation settings through other means as well, such as the WebAdmin interface. This also enables easier integration of James with third party services, e.g. a web admin GUI.

Since this is a rather large refactoring, I have split it into three commits:

**Storage Layer**

- Move `NotificationRegistry`, `RecipientId`, `Vacation`, `VacationPatch`,`VacationRepository` and associated tests from _james-server-data-jmap_ to _james-server-data-api_. 
- Copy `AccountId` to _james-server-data-api_, since it is used heavily in JMAP, but needed to preserve type safety in the generalized storage layer. Add a trivial helper `AccountIdUtil` to convert a JMAP `AccountId` to its general equivalent.
- Move `MemoryNotificationRegistry`, `MemoryVacationRepository` and associated tests from _james-server-data-jmap_ to _james-server-data-memory_.
- Move the respective bindings from `MemoryDataJmapModule`  to `MemoryDataModule` in _james-server-guice-memory_.
- Move `CassandraNotificationRegistry`, `CassandraNotificationRegistryDAO`, `CassandraNotificationRegistryModule`, `CassandraVacationDAO`, `CassandraVacationModule`, `CassandraVacationRepository`, `CassandraNotificationTable`, `CassandraVacationTable` and associated tests from _james-server-data-jmap-cassandra_ to _james-server-data-cassandra_.
- Move the respective bindings from `CassandraJmapModule` to a new `CassandraVacationModule` in `james-server-guice-cassandra`, and bind the latter in all servers that use cassandra.

**Vacation Mailet**

- Move `VacationMailet`, `VacationReply`, `MimeMessageBodyGenerator` and associated tests from _james-server-jmap-draft_ to _james-server-mailets_.
- Add upgrade instructions to adjust the package name of `VacationMailet` in `mailetconainer.xml`

**WebAdmin Routes**

- Create `VacationRoutes`, `VacationDTO` and associated tests in _james-server-webadmin-data_.
- Bind the routes within `DataRoutesModule` in _james-server-guice-webadmin-data_.
- Document `VacationRoutes` usage via WebAdmin interface.

Everything else in this PR is mainly adjustments to the package name changes.


server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/VacationRoutes.java
```diff
@@ -0,0 +1,199 @@
...
+            VacationPatch vacationPatch = VacationPatch.builder()
+                .subject(updateTo(vacationDto.getSubject()))
+                .textBody(updateTo(vacationDto.getTextBody()))
+                .htmlBody(updateTo(vacationDto.getHtmlBody()))
+                .fromDate(updateTo(vacationDto.getFromDate()))
+                .toDate(updateTo(vacationDto.getToDate()))
+                .isEnabled(updateTo(vacationDto.getEnabled()))
+                .build();
+            vacationRepository.modifyVacation(accountId, vacationPatch)
+                .then(notificationRegistry.flush(accountId)).block();
```
chibenwa, line 122: A nice refactoring would be to have the `vacationRepository` flushing by itself the `notificationRegistry` (or an abstraction on top of them) so that we don't duplicate that statement...

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/VacationRoutes.java
```diff
@@ -0,0 +1,199 @@
...
+            return Responses.returnNoContent(response);
+        } catch (JsonExtractException e) {
+            throw ErrorResponder.builder()
+                .statusCode(HttpStatus.BAD_REQUEST_400)
+                .type(ErrorResponder.ErrorType.INVALID_ARGUMENT)
+                .message("Malformed JSON input")
+                .cause(e)
+                .haltError();
+        }
```
chibenwa, line 131: All the handling of `JsonExtractException` should already be done by the default error handler of the WebAdminServer. We likely can remove such statements here.

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/VacationRoutes.java
```diff
@@ -0,0 +1,199 @@
...
```
chibenwa, line 157: idem

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/VacationRoutes.java
```diff
@@ -0,0 +1,199 @@
...
+        return Responses.returnNoContent(response);
+    }
+
+    private void testUserExists(Request request) {
+        boolean exists;
+        try {
+            exists = usersRepository.contains(Username.of(request.params(USER_NAME)));
+        } catch (UsersRepositoryException e) {
+            exists = false;
+        }
```
chibenwa, line 182: Mutability spotted ;-)

Here this could easily be solved with one more method extraction.

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    @Test
+    void getVacationFails() {
+        vacationRepository.modifyVacation(AccountId.fromString(BOB), VacationPatch.builderFrom(VACATION).build()).block();
+
+        when()
+            .get(VacationRoutes.VACATION + SEPARATOR + CAROL)
+        .then()
+            .statusCode(HttpStatus.BAD_REQUEST_400);
```
chibenwa, line 127: This fails because the user CAROL do not exist, correct?

Shouldn't it be a 404?

Also please assert the fields in the JSON body of the response (that should contain very useful information.)

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    @Test
+    void postVacationFails() {
+        given()
+            .body("{\"enabled\":true,\"subject\":\"On vacation again\"}")
+        .when()
+            .post(VacationRoutes.VACATION + SEPARATOR + CAROL)
+        .then()
+            .statusCode(HttpStatus.BAD_REQUEST_400);
```
chibenwa, line 194: Idem 404?

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
```diff
@@ -0,0 +1,297 @@
...
+            .statusCode(HttpStatus.NO_CONTENT_204);
+
+        Vacation vacation = vacationRepository.retrieveVacation(bob).block();
+        assertThat(vacation).isNotNull();
+        assertThat(vacation.isEnabled()).isTrue();
+        assertThat(vacation.getFromDate()).isEmpty();
+        assertThat(vacation.getToDate()).isEmpty();
+        assertThat(vacation.getSubject()).isEqualTo(Optional.of("More vacation"));
+        assertThat(vacation.getTextBody()).isEmpty();
+        assertThat(vacation.getHtmlBody()).isEmpty();
```
chibenwa, line 248: In `patchVacationModifies`

I would expect only the subject to be changed, but the other properties to be untouched.

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
```diff
@@ -0,0 +1,297 @@
...
+            .statusCode(HttpStatus.NO_CONTENT_204);
+
+        Vacation vacation = vacationRepository.retrieveVacation(bob).block();
+        assertThat(vacation).isNotNull();
+        assertThat(vacation.isEnabled()).isFalse();
+        assertThat(vacation.getFromDate()).isEmpty();
+        assertThat(vacation.getToDate()).isEmpty();
+        assertThat(vacation.getSubject()).isEmpty();
+        assertThat(vacation.getTextBody()).isEmpty();
+        assertThat(vacation.getHtmlBody()).isEmpty();
```
chibenwa, line 284: When we have many assertions we generally wrap them in a soft assertion:

```
SoftAssertions.assertThatSoftly(softly -> {
    softly.assertThat(vacation).isNotNull();
    softly.assertThat(vacation.isEnabled()).isFalse();
    // ...
});
```

In case of failures it executes all the statements and do a global reports instead of failing on the first one.

This enables fixing all the issues at once.

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
```diff
@@ -0,0 +1,297 @@
...
+        Boolean registered = notificationRegistry.isRegistered(bob, alice).block();
+        assertThat(registered).isFalse();
+    }
+
+    @Test
+    void deleteVacationFails() {
+        when()
+            .delete(VacationRoutes.VACATION + SEPARATOR + CAROL)
+        .then()
+            .statusCode(HttpStatus.BAD_REQUEST_400);
```
chibenwa, line 295: 404?

chibenwa (commented):
`org.apache.james.jmap.draft.model.VacationResponse` can give inspiration on how to handle partial updates for PATCH.

If simpler we could keep going with just the POST and drop the PATCH, if acceptable to you...

server/container/guice/memory/src/main/java/org/apache/james/modules/data/MemoryDataModule.java
```diff
@@ -79,6 +85,15 @@ protected void configure() {
...
         bind(DLPConfigurationStore.class).to(EventSourcingDLPConfigurationStore.class);
 
+        bind(MemoryVacationRepository.class).in(Scopes.SINGLETON);
+        bind(VacationRepository.class).to(MemoryVacationRepository.class);
+
+        bind(MemoryNotificationRegistry.class).in(Scopes.SINGLETON);
+        bind(NotificationRegistry.class).to(MemoryNotificationRegistry.class);
+
+        bind(DefaultZonedDateTimeProvider.class).in(Scopes.SINGLETON);
+        bind(ZonedDateTimeProvider.class).to(DefaultZonedDateTimeProvider.class);
```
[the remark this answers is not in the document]
ottoka: I remember actually seeing a ClockModule somewhere. But as you say, it is a different issue for a different ticket.

server/protocols/jmap-draft/src/main/java/org/apache/james/jmap/draft/utils/AccountIdUtil.java
```diff
@@ -0,0 +1,28 @@
...
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.jmap.draft.utils;
+
+import org.apache.james.jmap.api.model.AccountId;
+
+public class AccountIdUtil {
+    public static org.apache.james.vacation.api.AccountId converted(AccountId accountId) {
```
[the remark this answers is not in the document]
ottoka: Sure, if you prefer it this way. I was not sure what the James style is for such things.

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/VacationResponseSetMethod.scala
```diff
@@ -37,6 +36,8 @@ import org.apache.james.jmap.routes.SessionSupplier
 import org.apache.james.jmap.vacation.{VacationResponseSetError, VacationResponseSetRequest, VacationResponseSetResponse, VacationResponseUpdateResponse}
 import org.apache.james.mailbox.MailboxSession
 import org.apache.james.metrics.api.MetricFactory
+import org.apache.james.vacation.api.{VacationPatch, VacationRepository}
+import org.apache.james.vacation.api.{AccountId => ConvertedAccountId}
```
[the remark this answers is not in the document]
ottoka: Makes sense, and matches the Java use.

mpt/impl/smtp/cassandra-rabbitmq-object-storage/src/test/resources/mailetcontainer.xml
```diff
@@ -60,7 +60,7 @@
             <mailet match="All" class="RemoveMimeHeader">
                 <name>bcc</name>
             </mailet>
-            <mailet match="RecipientIsLocal" class="org.apache.james.jmap.mailet.VacationMailet">
+            <mailet match="RecipientIsLocal" class="org.apache.james.transport.mailets.VacationMailet">
```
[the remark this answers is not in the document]
ottoka: Heh, now that you mention it, sure. Another benefit of making it general purpose.

server/apps/distributed-app/docs/modules/ROOT/pages/operate/webadmin.adoc
```diff
@@ -551,6 +551,78 @@ Response codes:
...
+
+Response codes:
+
+* 204: The vacation settings were successfully updated
+* 400: The user name or the payload is invalid
+
+=== Patch vacation settings
+
+....
+curl -XPATCH http://ip:port/vacation/usernameToBeUsed
```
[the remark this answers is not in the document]
ottoka: > -> null reset to null
-> absent ignores.

I Would prefer it this way, but unfortunately the JSON deserializer cannot distinguish these cases. It results in null or Optional.empty regardless, whether I specify the property or omit it. This is the reason why I added two methods to handle null/absent properties, one to update (POST) and one to reset (PATCH).

But overall, I believe the update (POST) is the more useful method, and there is a workaround to DELETE and then POST a partial object again if someone actually needs the reset functionality. So I will drop PATCH, since having only one method is probably less confusing to the end user.

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/VacationRoutes.java
```diff
@@ -0,0 +1,199 @@
...
+    public VacationDTO getVacation(Request request, Response response) {
+        testUserExists(request);
+        try {
+            AccountId accountId = AccountId.fromString(request.params(USER_NAME));
+            Vacation vacation = vacationRepository.retrieveVacation(accountId).block();
+            return VacationDTO.from(vacation);
+        } catch (IllegalStateException e) {
+            LOGGER.info("Invalid get on user vacation", e);
+            throw ErrorResponder.builder()
+                .statusCode(HttpStatus.NOT_FOUND_404)
```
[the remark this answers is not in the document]
ottoka: Probably a leftover from an earlier development iteration. I will remove it.

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/VacationRoutes.java
ottoka: Just checked, it actually does return 400 on bad JSON. I can remove the catch clause and just declare the method `throws JsonExtractException`. Good to know :-)

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
ottoka: Sure, I will change it accordingly. And good to know, I sow this elsewhere and was wondering what it was for.

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
ottoka: Agreed, 404 is more specific in case someone needs to analyze the result.

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/VacationRoutes.java
ottoka: Sigh, ok ok.

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
ottoka: PATCH was supposed to reset any properties missing/null in the request JSON. But since I removed PATCH, this is a moot point :-)

server/protocols/webadmin/webadmin-data/src/main/java/org/apache/james/webadmin/routes/VacationRoutes.java
ottoka: I actually missed that one in my first implementation, so I wholeheartedly agree. If I am refactoring all this anyway, might as well do it right. Will create a fourth commit for that though to keep things separate.

ottoka:
Latest push includes a fourth commit, which introduces a `VacationService` interface and its `DefaultVacationService` implementation. This wraps access to `VacationRepository` and `NotificationRegistry` in such a way as to ensure a user's registry is flushed whenever the user's vacation settings get modified.

ottoka:
Just to be sure - there is no JPA implementation of VacationRepository and NotificationRegistry, is there? Or did I miss it?

chibenwa:
> Just to be sure - there is no JPA implementation of VacationRepository and NotificationRegistry, is there? Or did I miss it?

Yes, no JPA as we did not port JMAP protocol over JPA.

chibenwa (approved):
Looks good to me

ottoka:
> Yes, no JPA as we did not port JMAP protocol over JPA.

With this refactoring, it would make sense to implement JPA vacation handling independently of JMAP. Unfortunately, I have never used JPA, so I do not feel comfortable enough with its database to implement this feature in a production quality manner :-(

chibenwa:
The fact that we can now implement JPA vacation support do not mean that we necessarily should.

IMO we can let that to others, if somebody benefits of it.

Arsnael:
Hello! There is a conflict to solve here :)

ottoka:
As long as it is just in the upgrade instruction documentation, that's an easy fix!

ottoka:
Should I worry about the failed tests on the Jenkins build? IMAPServerTest runs fine on my local build...

Arsnael:
I don't think it was relevant, I restarted a build :)

chibenwa:
`This branch cannot be rebased due to conflicts` Can you have a look @ottoka ?

Overwize looks great to me ;-)

ottoka:
I originally fixed the documentation conflict in the GitHub UI, which produced a merge commit. Probably that caused the issue. I re-did the fix manually and re-based everything on the current master. So hopefully the latest push will do better.

server/apps/distributed-app/docs/modules/ROOT/pages/operate/webadmin.adoc
```diff
@@ -551,6 +551,64 @@ Response codes:
...
+== Administrating vacation settings
+
+=== Get vacation settings
+
+....
+curl -XGET http://ip:port/vacation/usernameToBeUsed
+....
+
+Resource name usernameToBeUsed representing valid users, hence it should
+match the criteria at xref:confgure/usersrepository.adoc[User Repositories documentation]
```
Arsnael, line 563:
```suggestion
match the criteria at xref:configure/usersrepository.adoc[User Repositories documentation]
```

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
```diff
@@ -0,0 +1,257 @@
...
+            .body("message", is("The user 'carol@example.org' does not exist"));
+    }
+
+    @Test
+    void postVacationCreates() throws Exception {
+        AccountId bob = AccountId.fromString(BOB);
+        RecipientId alice = RecipientId.fromMailAddress(new MailAddress(ALICE));
+
+        vacationRepository.modifyVacation(bob, VacationPatch.builderFrom(VACATION).build())
+                .then(notificationRegistry.register(bob, alice, Optional.empty())).block();
```
Arsnael: With this isn't this test technically an update, like the below test? Shouldn't we remove this instead to really test a creation from nothing?

Arsnael (commented):
Sorry to review this work so late. Very good work overall thank you :) but I have 2 small comments

server/apps/distributed-app/docs/modules/ROOT/pages/operate/webadmin.adoc
ottoka: Oops, fixed in latest push.

server/protocols/webadmin/webadmin-data/src/test/java/org/apache/james/webadmin/routes/VacationRoutesTest.java
ottoka: It is a misnomer here, I actually wanted to test that all fields can be updated. But you are right, creation from scratch is also a good test case, so I added an extra one. There now is Create, UpdatesAll, UpdatesPartial.

Arsnael (approved):

PR #760, opened 2021-11-24
JAMES-3534 Identity/destroy should not fail for identities whose alias is deleted

quantranhong1999:
After writing the test and having a look again at the delete API logic, I realize that we don't need to change any logic to the destroy case.
Still the test for sure.

### DIFF
--- a/server/container/guice/common/src/main/java/org/apache/james/utils/DataProbeImpl.java
+++ b/server/container/guice/common/src/main/java/org/apache/james/utils/DataProbeImpl.java
@@ -119,6 +119,12 @@ public class DataProbeImpl implements GuiceProbe, DataProbe {
     }
 
     @Override
+    public void removeUserAliasMapping(String fromUser, String fromDomain, String toAddress) throws Exception {
+        MappingSource source = MappingSource.fromUser(fromUser, fromDomain);
+        recipientRewriteTable.removeAliasMapping(source, toAddress);
+    }
+
+    @Override
     public void addDomainAliasMapping(String aliasDomain, String deliveryDomain) throws Exception {
         recipientRewriteTable.addDomainAliasMapping(MappingSource.fromDomain(Domain.of(aliasDomain)), Domain.of(deliveryDomain));
     }
--- a/server/data/data-api/src/main/java/org/apache/james/probe/DataProbe.java
+++ b/server/data/data-api/src/main/java/org/apache/james/probe/DataProbe.java
@@ -75,6 +75,8 @@ public interface DataProbe {
 
     void addUserAliasMapping(String fromUser, String fromDomain, String toAddress) throws Exception;
 
+    void removeUserAliasMapping(String fromUser, String fromDomain, String toAddress) throws Exception;
+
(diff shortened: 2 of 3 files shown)


chibenwa (changes requested):
Ok for Identity/set destroy.

Please add the following tests:

```
GIVEN bob-alias@domain.tld
AND I create a new custom identity with bob-alias@domain.tld
WHEN an admin delete bob-alias@domain.tld
THEN Identity/get no longer returns the identity
```

```
GIVEN bob-alias@domain.tld
AND I create a new custom identity with bob-alias@domain.tld
WHEN an admin delete bob-alias@domain.tld
THEN Identity/set update of that identity fails
```

```
GIVEN bob-alias@domain.tld
AND I update the default identity of bob-alias@domain.tld
WHEN an admin delete bob-alias@domain.tld
THEN Identity/get no longer returns the identity
```

```
GIVEN bob-alias@domain.tld
AND I update the default identity of bob-alias@domain.tld
WHEN an admin delete bob-alias@domain.tld
THEN Identity/set update of that identity fails
```

Arsnael:
Read it

quantranhong1999:
Added commit https://github.com/apache/james-project/pull/760/commits/19731a2d3adf322abb3820e77153f2e6ca69632f following 4 test cases Benoit suggested.
I think this is more likely to belong to Tung work (fix bug for get and update case).
@vttranlina you can cherry-pick this commit. And I will remove this commit after you take it.

chibenwa (approved):

Arsnael (approved):

Arsnael:
Needs a rebase

Arsnael:
I don't get why it keeps failing with the same errors... It makes no sense, the code in backends has not been modified. And also that code is used in https://github.com/apache/james-project/pull/761 that is green.....

Do we have troubles with a worker again?

quantranhong1999:
> Do we have troubles with a worker again?
Maybe. The NoClassDefFoundError seems like the CI can not load properly the maven dependency.

Arsnael:
Ok I'm retrying with adding an extra cleanup in the CI build, see if we get a better result

Arsnael:
Still the same... I don't think it's maven deps or coming from us. But maybe something directly wrong with the worker. You have issues like this:
```
java.lang.ExceptionInInitializerError
Caused by: org.testcontainers.containers.ContainerLaunchException: Container startup failed
Caused by: org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageName=<resolving>, imagePullPolicy=DefaultPullPolicy())
Caused by: com.github.dockerjava.api.exception.DockerClientException: Failed to parse docker configuration file
Caused by: java.io.IOException: Failed to parse docker config.json
Caused by: org.testcontainers.shaded.com.fasterxml.jackson.databind.JsonMappingException: 
No content to map due to end-of-input
 at [Source: /home/jenkins/.docker/config.json; line: 1, column: 0]
```
No idea what the hell is that `/home/jenkins/.docker/config.json` file... I will try to open a duplicated PR, to force the build on an other worker. I don't see better solution.

PR #771, opened 2021-12-03
JAMES-2435 Docker run's port param should be synchronized with webadmin.properties and curl examples's port

quantranhong1999:


examples/custom-webadmin-route/README.md
```diff
@@ -67,7 +67,7 @@ $ docker run -d \
    -v $PWD/webadmin.properties:/root/conf/webadmin.properties \
    -v $PWD/exts:/root/extensions-jars \
    -v $PWD/keystore:/root/conf/keystore \
-   -p 25:25 \
+   -p 8000:8000 \
```
Arsnael, line 70: Nothing wrong here actually... Look at a few lines below. Not exposing the port to the host in Docker does not mean you can't access it at all.

Your container has an IP, that you can see with a `docker inspect` command on it. If you call your IP directly with the port number, you can access webadmin, like in the example below: `http://172.17.0.2:8000`

quantranhong1999: So what does 25:25 mean here?

chibenwa: Binding SMTP port. It likely is not required.

Arsnael: Port 25 is SMTP...

PR #772, opened 2021-12-03
JAMES-3674 Provide PBKDF2 for password encryption

chibenwa:


Arsnael (approved):
Not an encryption expert but I understand it becomes necessary to adopt a default standard password encryption, as stated in the related ticket.

The code looks good to me :)

PR #803, opened 2021-12-20
JAMES-3680 Fix JMAP authentication strategies configuration

Arsnael:


vttranlina (approved):

server/container/guice/protocols/jmap/src/main/java/org/apache/james/jmap/draft/JMAPModule.java
```diff
@@ -207,7 +207,7 @@ JMAPDraftConfiguration provideDraftConfiguration(PropertiesProvider propertiesPr
                 .keystoreType(configuration.getString("tls.keystoreType", null))
                 .secret(configuration.getString("tls.secret", null))
                 .jwtPublicKeyPem(loadPublicKey(fileSystem, ImmutableList.copyOf(configuration.getStringArray("jwt.publickeypem.url"))))
-                .authenticationStrategies(Optional.ofNullable(configuration.getList(String.class, "authentication.strategy.draft", null)))
+                .authenticationStrategies(Optional.ofNullable(ImmutableList.copyOf(configuration.getStringArray("authentication.strategy.draft"))))
```
chibenwa, line 210: The Optional.ofNullable is not needed...

Arsnael: Optional.of then?

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/core/JmapRfc8621Configuration.scala
```diff
@@ -60,7 +61,7 @@ object JmapRfc8621Configuration {
         .getOrElse(UPLOAD_LIMIT_DEFAULT),
       maxTimeoutSeconds = Optional.ofNullable(configuration.getInteger(WEB_PUSH_MAX_TIMEOUT_SECONDS_PROPERTY, null)).map(Integer2int).toScala,
       maxConnections = Optional.ofNullable(configuration.getInteger(WEB_PUSH_MAX_CONNECTIONS_PROPERTY, null)).map(Integer2int).toScala,
-      authenticationStrategies = Optional.ofNullable(configuration.getList(classOf[String], AUTHENTICATION_STRATEGIES, null)).toScala)
+      authenticationStrategies = Optional.ofNullable(ImmutableList.copyOf(configuration.getStringArray(AUTHENTICATION_STRATEGIES))).toScala)
```
chibenwa, line 64: Idem 

The Optional.ofNullable is not needed ? Or the null case throws a NPE...

server/container/guice/protocols/jmap/src/main/java/org/apache/james/jmap/draft/JMAPModule.java
chibenwa: Maybe not, as it handled the not present case.

I think:

```
Optional.ofNullable(configuration.getStringArray("authentication.strategy.draft")).map(ImmutableList::copyOf)
```

is better as it preserves nullity.

Also this shows we can benefit from unit tests on this conf parsing. You can get a look at what was done for `WebAdminConfiguration`.

Arsnael: Yeah will do later

Arsnael: > is better as it preserves nullity.

Well the failures in the tests (and when debugging them) shows that it doesn't... will find an other way

Arsnael: We always end up with a Some([]) with this... while the initial code actually is ok on the nullity. I tried as well to add the auth strategies in jmap.properties in integration tests for jmap rfc8621 and ran AuthenticationContract and it's green. 

Maybe it's what i did with Helm, or I messed up somewhere else... Will test deeper when I have time

Arsnael:
Just some weird things happening with K8S if with an env var. But nothing wrong with that base code. Closing it.

PR #813, opened 2021-12-28
solve the problem when sending messages with Chinese subjects by some clients, subject displayed are as garbled codes

hanxianlong:
When I send an email, Chinese subject, by wps excel, the subject displayed as "?????", which are garbled codes. So I changed the two java files here, and the problem is solved.

image

### DIFF
--- a/server/container/core/src/main/java/org/apache/james/server/core/MailHeaders.java
+++ b/server/container/core/src/main/java/org/apache/james/server/core/MailHeaders.java
@@ -19,16 +19,15 @@
 
 package org.apache.james.server.core;
 
+import org.apache.mailet.base.RFC2822Headers;
+
+import javax.mail.MessagingException;
+import javax.mail.internet.InternetHeaders;
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.Enumeration;
-
-import javax.mail.MessagingException;
-import javax.mail.internet.InternetHeaders;
-
-import org.apache.mailet.base.RFC2822Headers;
 
 /**
  * This interface defines a container for mail headers. Each header must use
@@ -67,7 +66,7 @@ public class MailHeaders extends InternetHeaders implements Serializable, Clonea
      */
     public MailHeaders(InputStream in) throws MessagingException {
         super();
-        load(in);
+        load(in, true);
     }
 
     /**
--- a/server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SendMailHandler.java
+++ b/server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SendMailHandler.java
@@ -19,10 +19,6 @@
 
 package org.apache.james.smtpserver;
 
-import java.io.IOException;
-
-import javax.inject.Inject;
-
 import org.apache.commons.configuration2.Configuration;
 import org.apache.james.protocols.smtp.SMTPSession;
 import org.apache.james.protocols.smtp.dsn.DSNStatus;
@@ -33,6 +29,10 @@ import org.apache.james.queue.api.MailQueueFactory;
 import org.apache.mailet.Mail;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+import javax.mail.Message;
+import java.io.IOException;
 
 /**
  * Queue the message
@@ -70,6 +70,8 @@ public class SendMailHandler implements JamesMessageHook {
         LOGGER.debug("sending mail");
 
         try {
+            Message message = mail.getMessage();
+            message.setSubject(mail.getMessage().getSubject());
             queue.enQueue(mail);
             LOGGER.info("Successfully spooled mail {} from {} on {} for {}", mail.getName(), mail.getMaybeSender(), session.getRemoteAddress().getAddress(), mail.getRecipients());
         } catch (Exception me) {


server/container/core/src/main/java/org/apache/james/server/core/MailHeaders.java
```diff
@@ -67,7 +66,7 @@ public MailHeaders() {
      */
     public MailHeaders(InputStream in) throws MessagingException {
         super();
-        load(in);
+        load(in, true);
```
chibenwa, line 69: I think we would benefit from having a `private static boolean ALLOW_UTF_8 = true;` constant to explain what this nagic value means.

server/container/core/src/main/java/org/apache/james/server/core/MailHeaders.java
```diff
@@ -19,17 +19,16 @@
...
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Serializable;
 import java.util.Enumeration;
 
-import javax.mail.MessagingException;
-import javax.mail.internet.InternetHeaders;
-
-import org.apache.mailet.base.RFC2822Headers;
-
```
chibenwa, line 32: This import reorder is likely not checkstyle friendly... Do you succeed to compile using maven after your changes?

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SendMailHandler.java
```diff
@@ -70,6 +70,8 @@ public HookResult onMessage(SMTPSession session, Mail mail) {
         LOGGER.debug("sending mail");
 
         try {
+            Message message = mail.getMessage();
+            message.setSubject(mail.getMessage().getSubject());
```
chibenwa, line 74: Given the change to load headers using UTF-8 is this necessary?

chibenwa (commented):
Hello,

Thanks for your time and interest on this topic. Indeed this is a topic to address though I suppose this arise for non-standars encodings.

You work would benefit from a test reproducing the issue, that your code changes will eventually solve.

Could you also share the EML (sources) of the incriminated emails?

Best regards,

Benoit

chibenwa:
```
ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (check-style) on project james-server-core: You have 2 Checkstyle violations. -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (check-style) on project james-server-core: You have 2 Checkstyle violations.
```

server/container/core/src/main/java/org/apache/james/server/core/MailHeaders.java
hanxianlong: Hi @chibenwa here are two other mails' EML I sent days before, please check it.  One is the error mail whose subject is error code "?", the other normal subject "gong zi tiaoa封技术测试邮件，如有打扰请忽略。“ was sent by the same email client after I fixed the issue.
[https://files.cnblogs.com/files/hanxianlong/error_mails_eml.zip](url)

server/container/core/src/main/java/org/apache/james/server/core/MailHeaders.java
hanxianlong: Actually, when I cloned the master branch, I cannot compiled it succeed because of the checkstyle. So I just comment the checkstyle and then compiled succeeded.....

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SendMailHandler.java
hanxianlong: I don't thinks it IS necessary in my experience, because it looks like the codes do nothing but just re-set the subject of the message...........
However, If I remove this two lines, the error code of subject issue will not be solved. I don't know why, do you have any other suggestions?

server/container/core/src/main/java/org/apache/james/server/core/MailHeaders.java
hanxianlong: @chibenwa Could you provide me a mail account which is established by JAMES and you can access the email server to check the log, I can send a email by WPS-excel client again to re-procude the problem for you.

chibenwa:
Hello,

Given the example you provided I wrote quickly an integration test reprodicng the issue https://github.com/chibenwa/james-project/tree/pr-813

So far I can not demonstrate Apache James SMTP server, at least with its default configuration, alters received messages.

Are we speaking of outgoing emails?

Regards,

Benoit TELLIER

hanxianlong:
@chibenwa Yes, we are speaking outgoing emails. I'll buy a temprory VPS on Aliyun and then setup a temporary Apache James server with the default configuration, to re-produce the issue for you.
I'll contact you later when I set it up, TELLIER.

chibenwa:
> I'll contact you later when I set it up, TELLIER.

No please. To get me help you keep the conversation public.

> Yes, we are speaking outgoing emails. 

Ok. I did not succeed to write a test reproducing this.

Maybe you could share your mailetcontainer.xml file?

chibenwa:
Hello,

I added an integration test using the message you provided and could not get RemoteDelivery to alter the message. CF https://github.com/chibenwa/james-project/tree/pr-813 latest commit.

Which version of James are you using?

chibenwa:
Hello,

We can't reproduce the bug you describe.

Please share detailed configuration, email of the message, steps to reproduce and sequence of operation.

Regards,

Benoit

chibenwa:

chibenwa:
No answer, closing.

Reopen if relevant.

PR #820, opened 2022-01-04
Get docker config json

chibenwa:


chibenwa:
Proved the file is empty, which causes test containers to fail (reproduced within a local environment)

PR #823, opened 2022-01-05
JAMES-3432 Prevent 'Port 8000 not mapped' exception in Remote Deliver…

chibenwa:
…y tests


chibenwa:
Without that I get failures locally.

I do not know why the CI did not complain.

Encountered while diagnosing https://github.com/apache/james-project/pull/813

Arsnael (approved):

PR #831, opened 2022-01-07
JAMES-3687 Demonstrate issues with deletes of delayed mails

chibenwa:
- The Pulsar mail queue currently do not delete delayed emails

 Explanation: the sequenceId of the mail is reset when
 moved from the scheduled topic to the out topic, thus
 bypassing deletes emitted while it was sitting in the
 scheduled queue.

 (this mechanism exist to prevent deletions to apply on
 future emails)

 We would likely need the sequence to be backed by some
 applicative metadata - like a timestamp?

  - Second problem: filter removal based on sequence number

 Goal: Prevent uncontrolled growth by removing
 meaningless deletion filters

 Assumptions: emails are ordered

 Pitfall: delayed messages are out of order.

 Please note that the JMS mailqueue is buggy regarding
 delayed messages being deleted, I used the memory implementation as a reference.

### DIFF
--- a/server/queue/queue-api/src/test/java/org/apache/james/queue/api/DelayedManageableMailQueueContract.java
+++ b/server/queue/queue-api/src/test/java/org/apache/james/queue/api/DelayedManageableMailQueueContract.java
@@ -23,6 +23,7 @@ import static org.apache.james.queue.api.Mails.defaultMail;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.time.Duration;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
@@ -33,6 +34,7 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 import reactor.core.publisher.Flux;
+import reactor.core.scheduler.Schedulers;
 
 @ExtendWith(ExecutorExtension.class)
 public interface DelayedManageableMailQueueContract extends DelayedMailQueueContract, ManageableMailQueueContract {
@@ -68,6 +70,61 @@ public interface DelayedManageableMailQueueContract extends DelayedMailQueueCont
         getManageableMailQueue().clear();
 
         assertThat(getManageableMailQueue().getSize()).isEqualTo(0L);
+    }
+
+    @Test
+    default void delayedEmailsShouldBeDeleted() throws Exception {
+        getManageableMailQueue().enQueue(defaultMail()
+                .name("abc")
+                .build(),
+            5L,
+            TimeUnit.SECONDS);
+        // The queue being FIFO a second email can serve as a wait condition
+        getManageableMailQueue().enQueue(defaultMail()
+                .name("def")
+                .build(),
+            5L,
+            TimeUnit.SECONDS);
+
+        getManageableMailQueue().remove(ManageableMailQueue.Type.Name, "abc");
+
+        ArrayList<String> names = new ArrayList<>();
+        Flux.from(getManageableMailQueue().deQueue())
+            .subscribeOn(Schedulers.elastic())
+            .subscribe(item -> names.add(item.getMail().getName()));
+
+       Awaitility.await().untilAsserted(() -> assertThat(names).contains("def"));
+        assertThat(names).containsExactly("def");
+    }
+
+    @Test
+    default void delayedEmailsShouldBeDeletedWhenMixedWithOtherEmails() throws Exception {
+        getManageableMailQueue().enQueue(defaultMail()
+                .name("abc")
+                .build(),
+            5L,
+            TimeUnit.SECONDS);
+
+        getManageableMailQueue().remove(ManageableMailQueue.Type.Name, "abc");
+
+        // The newer email
+        getManageableMailQueue().enQueue(defaultMail()
+                .name("def")
+                .build());
+        // The queue being FIFO a third email can serve as a wait condition
+        getManageableMailQueue().enQueue(defaultMail()
+                .name("ghi")
+                .build(),
+            5L,
+            TimeUnit.SECONDS);
+
+        ArrayList<String> names = new ArrayList<>();
+        Flux.from(getManageableMailQueue().deQueue())
+            .subscribeOn(Schedulers.elastic())
+            .subscribe(item -> names.add(item.getMail().getName()));
+
+       Awaitility.await().untilAsserted(() -> assertThat(names).contains("ghi"));
+        assertThat(names).containsExactly("def", "ghi");
     }
 
     @Test
--- a/server/queue/queue-api/src/test/java/org/apache/james/queue/api/DelayedManageableMailQueueContract.java
+++ b/server/queue/queue-api/src/test/java/org/apache/james/queue/api/DelayedManageableMailQueueContract.java
@@ -93,7 +93,8 @@ public interface DelayedManageableMailQueueContract extends DelayedMailQueueCont
             .subscribeOn(Schedulers.elastic())
             .subscribe(item -> names.add(item.getMail().getName()));
 
-       Awaitility.await().untilAsserted(() -> assertThat(names).contains("def"));
+       Awaitility.await()
+           .untilAsserted(() -> assertThat(names).contains("def"));
         assertThat(names).containsExactly("def");
     }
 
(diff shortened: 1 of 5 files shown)


chibenwa:
CC @jeantil  @mbaechler

jeantil:
I suggest not starting a third discussion on the same topic. as I [wrote in the first discussion](https://github.com/apache/james-project/pull/808#discussion_r780170243) the behavior for this case was not specified in the contracts so we implemented the simplest that passed the contracts :)

With that said, implementing the requested behavior is not so difficut. I propose two options depending on the degree of accuracy we require : 
- attach a monotonously increasing identifier such as a timestamp (or an ObjectId for increased precision) to all enqueued messages and use that in the filter. This is sensitive to clock drift which could lead to some missed messages (not sure if it's an issue)
- query the scheduled topic offset on filter creation and attach that to the filter in addition to the out topic offset. filter scheduled messages as they are consumed and before they are requeued in the out topic. a filter would then expire when both the out topic and the scheduled topic's offset have been exceeded. (a simple way to implement this with the current setup is to duplicate the filter  actor and the filter command queue and reuse the existing logic for the scheduled queue)

chibenwa:
> I suggest not starting a third discussion on the same topic.

Fair. But failing tests have added value and shows fonctionnal expectations I have. 

It also makes the whole discussion less abstract (and show I care enough to dedicate a bit of time to the topic).

I do not feel like this just is valueless conversation duplication.

BTW it proved we have two distinct problems while the design review only allowed to catch one!

> so we implemented the simplest that passed the contracts :)

I have limited faight in these contracts, delays is not an area covered by the previous RabbitMQ implementation and we did devote limited amount of time testing edge cases, interactions with other MailQueue features. It should not be taken as "written in marble".

> attach a monotonously increasing identifier such as a timestamp

That's what I meant by "We would likely need the sequence to be backed by some applicative metadata - like a timestamp?". Of course it brings additional problems like accounting for imperfect clock synchronisation with a tolerated clock skew.

> query the scheduled topic offset on filter creation and attach that to the filter in addition to the out topic offset. filter scheduled messages as they are consumed and before they are requeued in the out topic. a filter would then expire when both the out topic and the scheduled topic's offset have been exceeded. (a simple way to implement this with the current setup is to duplicate the filter actor and the filter command queue and reuse the existing logic for the scheduled queue)

:+1:

jeantil:
Before we undertake aligning the behavior with the updated contract, can you confirm the following :

> Removal should apply on all items enqueued before remove is called regardless of whether they are scheduled in the future or not. Removal should not apply on any items that are enqueued after the call to remove.

in other words, removal applies on enqueue order and not on dequeue order. 

Assuming we implement the 2nd option I proposed : duplicate the filter logic for items consumed from the scheduled topic before they are injected in the out topic using the schedule topic offset at the time of remove as a reference for filtering, do you still see 

> Second problem: filter removal based on sequence number

as a fundamental problem that needs to be solved ? I can't think of anything but I could be missing something.

chibenwa:
> in other words, removal applies on enqueue order and not on dequeue order.

Yes.

Imagine for instance a given sender uses my SaaS instance to send SPAM. I spot the issue, dis-activate his account and purge the queue of his messages, both ready for sending and scheduled.

Also, similarily if a given recipient start bouncing, we wmight have a lot of messages in the queue, both without delays and with delays (retries). If the situation is hopeless, I might want to remove all emails to that very recipient, both scheduled and not scheduled.

> Assuming we implement the 2nd option I proposed : duplicate the filter logic for items consumed from the scheduled topic before they are injected in the out topic using the schedule topic offset at the time of remove as a reference for filtering, do you still see

:+1: 


>> Second problem: filter removal based on sequence number

> as a fundamental problem that needs to be solved ? I can't think of anything but I could be missing something.

I do not understand your reply

If the reference to apply deletes is the time of enqueue, as we have delays, these would thus not be ordered at the time of dequeue. Thus we could not craft an heuristic relying on an ordered sequence to simplify the deletion problem (performance versus correctness).

I do not see other solutions than keeping the filters forever (as we do not know if a message for which filter applies is still in the queue)

 - We could define an upper bound to delays, thus allowing keeing the same strategy 
 
  - A smarter solution could be to : 
      - Upon delete browse the queue to find matching messages
      - Build a distributed filter using the enqueueId (unique idenfier for this email transit) + expected time of dequeue = delay + time of enqueue
      - We could cleanup filters (that here would apply for individual emails) after a threshold after the expected time of dequeue is reached. Like 1 day after the time of the dequeue.

Or a double queuing system could be used:
 - one topic for emails marked as deleted (builds the filter)
 - one topic for emails effectively deleted (cleans the filter)
      
 - Or we could document the limitation and keep it for later. Thus keeping the filters forever.

chibenwa:
As these are just tests I think we can merge them to the contract prior to the fix. 

I opened a separate issue https://issues.apache.org/jira/browse/JAMES-3698

jeantil:
you will need to @Disable them in the PulsarMailqueueTest, we will enable them again when we merge (adding the @Disabled will ensure we get a conflict when rebasing :)

server/queue/queue-pulsar/src/test/java/org/apache/james/queue/pulsar/PulsarMailQueueTest.java
```diff
@@ -238,4 +238,18 @@ private void assertThatStoreIsEmpty() {
...
+    public void delayedEmailsShouldBeDeleted() {
+
+    }
+
+    @Test
+    @Override
+    @Disabled("JAMES-3687 Delayed deletes are buggy")
+    public void delayedEmailsShouldBeDeletedWhenMixedWithOtherEmails() {
+
+    }
```
chibenwa, line 254: > you will need to @disable them in the PulsarMailqueueTest, we will enable them again when we merge (adding the 
> @disabled will ensure we get a conflict when rebasing :)

Isn't this done here?

jeantil:
My bad. I shouldn't have tried to do the  review  while managing the kids.

Le mar. 11 janv. 2022 à 16:24, Benoit TELLIER ***@***.***> a
écrit :

server/queue/queue-api/src/test/java/org/apache/james/queue/api/DelayedManageableMailQueueContract.java
```diff
@@ -70,6 +72,61 @@ default void delayedMessagesShouldBeCleared() throws Exception {
...
+            TimeUnit.SECONDS);
+
+        getManageableMailQueue().remove(ManageableMailQueue.Type.Name, "abc");
+
+        ArrayList<String> names = new ArrayList<>();
+        Flux.from(getManageableMailQueue().deQueue())
+            .subscribeOn(Schedulers.elastic())
+            .subscribe(item -> names.add(item.getMail().getName()));
+
+       Awaitility.await().untilAsserted(() -> assertThat(names).contains("def"));
```
Arsnael, line 96: indent

server/queue/queue-api/src/test/java/org/apache/james/queue/api/DelayedManageableMailQueueContract.java
```diff
@@ -70,6 +72,61 @@ default void delayedMessagesShouldBeCleared() throws Exception {
...
+                .build(),
+            5L,
+            TimeUnit.SECONDS);
+
+        ArrayList<String> names = new ArrayList<>();
+        Flux.from(getManageableMailQueue().deQueue())
+            .subscribeOn(Schedulers.elastic())
+            .subscribe(item -> names.add(item.getMail().getName()));
+
+       Awaitility.await().untilAsserted(() -> assertThat(names).contains("ghi"));
```
Arsnael, line 126: indent

server/queue/queue-api/src/test/java/org/apache/james/queue/api/DelayedManageableMailQueueContract.java
chibenwa, line 96:
```suggestion
       Awaitility.await()
           .untilAsserted(() -> assertThat(names).contains("def"));
```

server/queue/queue-api/src/test/java/org/apache/james/queue/api/DelayedManageableMailQueueContract.java
chibenwa, line 126:
```suggestion
       Awaitility.await()
           .untilAsserted(() -> assertThat(names).contains("ghi"));
```

Arsnael (approved):

chibenwa:
@jeantil do you think I can merge this?

jeantil (approved):

jeantil:
yes

PR #836, opened 2022-01-11
JAMES-3700 Define basic expectations for Pulsar MailQueue dead letter…

chibenwa:
… policy

Cc @mbaechler @jeantil 

https://issues.apache.org/jira/browse/JAMES-3700

Dead lettering was off my radar...


Arsnael (approved):

jeantil:
hi benoit, 

It seems reasonable we will see how it goes trying to put it in place.

chibenwa:
From what I could see by having JSON schemas defined on the underlying topics, we could get away with a "type safe" approach thus avoiding such edge cases.

PR #838, opened 2022-01-12
Transiant dependencies upgrades

chibenwa:


quantranhong1999:
Read it. Nothing to add.
Q: How can you know if there is a vulnerable issue with a library?

Arsnael (approved):

chibenwa:
> Q: How can you know if there is a vulnerable issue with a library?

A: You need to follow security vulnerabilities. I am subscribed to Apache Software foundation announce mailing list so I can follow the vulnerabilities and releases of Apache softwares. https://www.apache.org/foundation/mailinglists.html#foundation-announce

Then... You need some tools! There is for instance dependabot that can be activated on Github and analysis your dependencies. You get warnings by mail or when pushing on the repo. EG: https://github.com/linagora/james-project/security/dependabot . While easy to install dependabot is limited.

Hence another option is https://app.snyk.io import your projects in there and you are good to go!
 
 - On the huge +1 it is able to do a deep scan of docker images, including OS libraries and the raw list of Java libraries. I use docker image scan on the Linagora docker images to know that we are having vulnerable dependencies.
 - On the huge -1 Snyk is bad at maven dependencies and do not handle dependency clashes well. It would report both the old and the new while only the new dependency might be applied. So it has a lot of false positives. Yet, given the use of docker image scan it present in depth analysis of the dependency chain bringing the dependency.
 
 Now on the dependency front, we can see that we are mostly up to date except:
   - The cassandra driver, unmaintained for 2 years, an upgrade to the version 4.x would eventually be needed here.
   - The ElasticSearch driver. We are maintaining compatibility with 7.10 whose driver is (a bit) outdated. Hopefully an upgrade to version 8.something would solve that.
   - Netty 3 stuff would eventually benefit from a Netty 4 upgrade. Warning, big work ahead!
   - And the Spark HTTP dependency brings an old Netty version. I did not try yet an override there...
   
Also there are some complementary static analysis tools like Sonar that could be mobilised to similar tasks. https://www.sonarqube.org/ . 
  
Does this answers your question?

quantranhong1999:
> Does this answers your question?

It does. Thank you for your dedicated answer.

PR #860, opened 2022-01-25
JAMES-3708 Invalid mail address: Demonstrate NPE in RemoteDelivery

chibenwa:
I will likely work on solving this by providing stricter parsing in the `MailAddress` POJO...

Maybe @jeantil @mbaechler feedback would be useful to me.

### DIFF
--- a/core/src/test/java/org/apache/james/core/MailAddressTest.java
+++ b/core/src/test/java/org/apache/james/core/MailAddressTest.java
@@ -28,7 +28,6 @@ import javax.mail.internet.AddressException;
 import javax.mail.internet.InternetAddress;
 
 import org.assertj.core.api.Assertions;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
@@ -48,13 +47,9 @@ class MailAddressTest {
                 GOOD_ADDRESS,
                 GOOD_QUOTED_LOCAL_PART,
                 "server-dev@james-apache.org",
-                "a..b@domain.com",
                 "server-dev@[127.0.0.1]",
-                "server-dev@#123",
-                "server-dev@#123.apache.org",
                 "server.dev@james.apache.org",
-                "\\.server-dev@james.apache.org",
-                "server-dev\\.@james.apache.org")
+                "\\.server-dev@james.apache.org")
             .map(Arguments::of);
     }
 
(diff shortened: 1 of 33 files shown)


Arsnael (approved):

core/src/test/java/org/apache/james/core/MailAddressTest.java
```diff
@@ -47,6 +48,7 @@
                 GOOD_ADDRESS,
                 GOOD_QUOTED_LOCAL_PART,
                 "server-dev@james-apache.org",
+                "a..b@domain.com",
```
jeantil, line 51: hmm I don't know according to wikipedia  that's not a valid address because the dot occurs twice in a row which seems to match [RFC5322](https://datatracker.ietf.org/doc/html/rfc5322#section-3.2.3) this would need to be quoted to be a valid local part since it is not a dot-atom

For reference examples from wikipedia : 
good addresses: 
- Abc@example.com
- Abc@10.42.0.1
- Abc.123@example.com
- user+mailbox/department=shipping@example.com
- !#$%&'*+-/=?^_`.{|}~@example.com
- "Abc@def"@example.com
- "Fred Bloggs"@example.com
- "Joe.\\Blow"@example.com
- Loïc.Accentué@voilà.fr8
- pelé@exemple.com
- δοκιμή@παράδειγμα.δοκιμή
- 我買@屋企.香港
- 二ノ宮@黒川.日本
- медведь@с-балалайкой.рф
- संपर्क@डाटामेल.भारत


bad addresses
- Abc.example.com
- Abc.@example.com
- Abc..123@example.com

jeantil:
Apart from the fact that the PR seems to introduce an invalid address in the "goodAddresses" list (but maybe that's on purpose to demonstrate the issue ?), I agree that ignoring exceptions by replacing a value by null is not a good idea :) especially if it triggers errors/DOS/... downstream. 

I am not familiar enough with the parsing parts of james but I feel that this should be rejected early when parsing the SMTP message...

core/src/test/java/org/apache/james/core/MailAddressTest.java
chibenwa: Yes I know. Sorry it is indeed misleading.

> (but maybe that's on purpose to demonstrate the issue ?)

Yes it is, as my fix suggests to move it to bad address/

chibenwa:
>I am not familiar enough with the parsing parts of james but I feel that this should be rejected early when parsing the SMTP message...

Actually it is not even the SMTP message - James cares little about it and if we do - eg in JMAP we are lenient

We are speaking here of SMTP enveloppe which is even worse akka `RCPT TO: <a..b@domain.tld>`.

I'm converting the PR as a draft to denote I plan further work on this and that indeed I do not wich to include a bad address in the list of good addresses...

chibenwa:
This is more or less what I have in mind on this topic.

As I changed some rules and as I changed some exception messages I am expecting a bunch of CI failures hence I will keep it as a draft for the time being, until I have a :green_apple:  build.

chibenwa:
UN-WIP as this is :green_apple:

Arsnael (approved):

PR #861, opened 2022-01-25
JAMES-3709 Refactoring POP3 command handlers to share common code

ottoka:
Make TopCmdHandler re-use common code from RetrCmdHandler as protected respondWithMessage method.
Keep only specific argument parsing in retr and top handler methods respectively.
Add unit tests for everything, but in particular getMessageContent throwing or yielding null, which was missing in TopCmdHandler before.

I did not include the handling of IndexOutOfBoundsException / NoSuchElementException from TopCmdHandler, could not see when this would happen. I assume maybe Mailbox.getMessage() did that at some point, but not anymore. That would also explain the missing null check.


chibenwa (approved):

Arsnael (approved):

ottoka:
This latest refactoring introduces a base class AbstractPOP3MessageCommandHandler for all message related commands, not only TOP and RETR. Access to deleted messages throws IOException instead of returning null. Further expands tests.
Also includes and fixes https://github.com/apache/james-project/pull/868
Bonus: Fixes wrong numbering in LIST command after deleting some messages.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3MessageCommandHandler.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.james.protocols.api.ProtocolSession;
+import org.apache.james.protocols.api.Request;
+import org.apache.james.protocols.api.Response;
+import org.apache.james.protocols.pop3.POP3Response;
+import org.apache.james.protocols.pop3.POP3Session;
+import org.apache.james.protocols.pop3.mailbox.MessageMetaData;
+
+import com.google.common.collect.ImmutableList;
+
+public abstract class AbstractPOP3MessageCommandHandler extends AbstractPOP3CommandHandler {
```
chibenwa, line 36: An abstact class of an abstract class  seems a bit far fetched to me.

We used to / still have similar code in other places and this had never ended up  turning out great.

I am an advocate of composition compared to inheritence when it comes up to code sharing.

For instance we could have a cool little POJO helper being `MessageCommand`

```
// POJO
class MessageCommand { 
   static Optional<MessageCommand> fromLine(String line) { ... here goes the parsing logic not linked to the processor stuff...  } 
   
   private final int num;
}
```

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/MessageMetaDataUtils.java
```diff
@@ -28,14 +29,14 @@
 public class MessageMetaDataUtils {
 
     /**
-     * Returns the {@link MessageMetaData} for the given message number or <code>null</code> if it can not be 
-     * found.
+     * Returns the {@link MessageMetaData} for the given message number.
+     * @throws IOException if it can not be found.
      */
-    public static MessageMetaData getMetaData(POP3Session session, int number) {
+    public static MessageMetaData getMetaData(POP3Session session, int number) throws IOException {
```
chibenwa, line 35: I would promote this to the `POP3Session` rather than adding an artificial "utils" class

`POP3Session::getMetadation(int number)`

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3MessageCommandHandler.java
chibenwa: This would make it easier to write unit test!

server/protocols/protocols-pop3/src/test/java/org/apache/james/pop3server/POP3ServerTest.java
```diff
@@ -544,6 +549,49 @@ void testKnownUserInboxWithMessages() throws Exception {
...
+            Flux.from(mailboxManager.getMailbox(mailboxPath, session).listMessagesMetadata(MessageRange.all(), session))
+            .map(i -> i.getComposedMessageId().getUid())
+            .collectList().block(),
+            session);
+
+        Reader r = pop3Client.retrieveMessageTop(entries[0].number, 0);
+
+        assertThat(r).isNull();
+
+        // Fails: The NPE is not handled and causes the connection to abort...
```
chibenwa, line 591: Still true? Or shall we remove this outdated comment?

ottoka: Oops, forgot to remove it.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/MessageMetaDataUtils.java
ottoka: This utility class was already there, but it should not be hard to move its content to where it is actually needed.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/AbstractPOP3MessageCommandHandler.java
ottoka: AbstractPOP3CommandHandler was only introduced to get a grip on POP3 with Glowroot, and it does not really have any code in it. Of course the AbstractPOP3MessageCommandHandler could be a utility class instead, that accepts one or two lambdas instead of having protected methods. I am not quite sure how your MessageCommand code example fits in here though.

chibenwa: > I am not quite sure how your MessageCommand code example fits in here though.

To be fair it would just wrap the line parsing part.

ottoka: In the second commit I switched from inheritance to delegation. Is this closer to what you envisioned?

Arsnael:
Little checkstyle issue during the build, easy to fix :+1: 
```
03:14:34,002 [INFO] There is 1 error reported by Checkstyle 8.29 with checkstyle.xml ruleset.

03:14:34,009 [ERROR] src/main/java/org/apache/james/protocols/pop3/core/DeleCmdHandler.java:[22,8] (imports) UnusedImports: Unused import - java.io.IOException.
```

ottoka:
How ever did this get through...? Fixed in latest push.

Arsnael (approved):

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandDelegate.java
```diff
@@ -0,0 +1,101 @@
...
+                    return handleMessageExists(session, data, args);
+                } else {
+                    return new POP3Response(POP3Response.ERR_RESPONSE, "Message (" + num + ") already deleted.");
+                }
+
+            } catch (IOException ioe) {
+                return new POP3Response(POP3Response.ERR_RESPONSE, "Message (" + num + ") does not exist.");
+            }
+        } else {
+            return POP3Response.ERR;
```
chibenwa, line 84: This is a 40+ lines of codes method, we should split it in smaller chunks.

Here we can distinguish the following responsibilities:
 - Parsing (we could introduce a small POJO to combine num and args in the return type)
 - Processing

ottoka: OK, now I see what you meant with the parsing POJO. Done in latest commit.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandArguments.java
```diff
@@ -0,0 +1,77 @@
...
+    public static POP3MessageCommandArguments fromRequest(Request request) {
+        String parameters = request.getArgument();
+        if (parameters == null) {
+            return INVALID;
+        }
+
+        try {
+            List<Integer> args = Arrays.stream(parameters.trim().split(" "))
+                .map(String::trim).filter(arg -> !arg.isEmpty()).map(Integer::parseInt)
+                .collect(Collectors.toUnmodifiableList());
```
chibenwa, line 46:
```suggestion
            List<Integer> args = Splitter.on(' ')
            .omitEmptyStrings()
            .trimResults()
            .splitToList(parameters)
            .stream()
            .map(Integer::parseInt)
                .collect(ImmutableList.toImmutableList());
```

Rationnals:
 - `String::split` hides a regex compilation and evaluation that I would better avoid on each POP3 command
 - Splitter is a common, readable pattern we use in many other places
 - Unmodifiable list is a less strong version of immutable list, we should rather use the later.
 
 (Sorry for being picky)

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/mailbox/MessageMetaData.java
```diff
@@ -59,4 +59,15 @@ public long getSize() {
...
+    /**
+     * Check whether POP3 UID is compatible with RFC1939
+     */
+    private static boolean isRFC1939Compatible(String uid) {
+        if (uid == null) {
+            return false;
+        }
+
+        return IntStream.range(0, uid.length())
+            .allMatch(i -> uid.charAt(i) >= 0x21 && uid.charAt(i) <= 0x7E);
```
chibenwa, line 71: Guava comes with some handly `CharMatcher` with a nice allMatch method that for sure will perform better than its string based counterpart

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandArguments.java
ottoka: Now that you mention it, I remember we talked about this before. Committed.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandArguments.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class POP3MessageCommandArguments {
+    
+    private static POP3MessageCommandArguments INVALID = new POP3MessageCommandArguments(false, 0, Optional.empty());
+    
+    private final boolean valid;
+    private final int messageNumber;
+    private final Optional<Integer> lineCount;
+
+    public static POP3MessageCommandArguments fromRequest(Request request) {
```
chibenwa, line 38:
```suggestion
    public static Optional<POP3MessageCommandArguments> fromRequest(Request request) {
```

`private final boolean valid` is clearly an anti-pattern as we purpously build some invalid objects. Let's just avoid return them in the first place!

(actually the magic constants ture/false and 0 catched my eyes, I was about to comment about them but then looking deeper it seems there's more.)

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandArguments.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+
+        try {
+            List<Integer> args = Splitter.on(' ')
+            .omitEmptyStrings()
+            .trimResults()
+            .splitToList(parameters)
+            .stream()
+            .map(Integer::parseInt)
+                .collect(ImmutableList.toImmutableList());
```
chibenwa, line 51: Indent fail, sorry!

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandArguments.java
ottoka: ... needs imports as well. Fixed in latest push.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandArguments.java
ottoka: Easily fixed.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/mailbox/MessageMetaData.java
ottoka: OK, I just moved the existing code here, but might as well clean it up.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandArguments.java
ottoka: I originally thought about Optional, but it is a bit awkward here since I would need something like .orElseReturn(...) which obviously does not exist.
Anyway, it is safer to use with Optional, so I changed it in the latest push. Bonus, no more magic numbers/constants.

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandDelegate.java
```diff
@@ -46,24 +47,25 @@ public POP3MessageCommandDelegate(Collection<String> commands) {
...
         if (session.getHandlerState() == POP3Session.TRANSACTION) {
 
-            POP3MessageCommandArguments args = POP3MessageCommandArguments.fromRequest(request);
-            if (!args.isValid()) {
+            Optional<POP3MessageCommandArguments> args = POP3MessageCommandArguments.fromRequest(request);
+            if (args.isEmpty()) {
                 return handleSyntaxError();
             }
-
+            int num = args.get().getMessageNumber();
```
chibenwa, line 54: Calling `.get` on Optionals is cheatting.

You are supposed to transform it, and unbox it in a functional style. Method extraction helps a lot. Eg:

```
return args.map(this::handleValidArgs).orElseGet(this::handleSyntaxError);
```

chibenwa (approved):
Thanks for carring about my comments.

I'm making more comments but that is for the record / sport...

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/mailbox/MessageMetaData.java
chibenwa: Yes that's a trap when moving old code: it shows up in the reviews and people complains! :-)

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandDelegate.java
ottoka: Right, looks a lot cleaner this way. Fixed in latest push.

ottoka:
BTW I actually prefer it when people comment on my code. It improves code quality, catches the occasional bug, and I get to learn the preferred James style. So thank you!

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandDelegate.java
```diff
@@ -0,0 +1,88 @@
...
+import com.google.common.collect.ImmutableList;
+
+public abstract class POP3MessageCommandDelegate {
+
+    private final String command;
+    
+    public POP3MessageCommandDelegate(Collection<String> commands) {
+        if (commands.isEmpty()) {
+            throw new IllegalArgumentException("must provide at least one command keyword");
+        }
```
chibenwa, line 42:
```suggestion
    Preconditions.checkArguments(!commands.isEmpty(), "You must provide at least one command keyword");
```

Would be the prefered way to do this.

chibenwa (approved):
Looks great to me!

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/core/POP3MessageCommandDelegate.java
ottoka: I forgot about Preconditions. Fixed in push.

ottoka:
Latest push squashed commits and rebased on current master.

chibenwa (approved):

PR #867, opened 2022-01-26
Jwt-revoker container update for oidc example

Arsnael:
The rpc client from Krakend don't seem to really handle reconnections. It makes more sense as well to open a connection and close it when we finish using it.

### DIFF
--- a/examples/oidc/jwt-revoker/main.go
+++ b/examples/oidc/jwt-revoker/main.go
@@ -43,39 +43,24 @@ func getEnvironmentVariable(envVariable string, defaultVariable string) string {
 var key = getEnvironmentVariable("JWT_CLAIM", "sid")
 
 func main() {
-	krakendHost := getEnvironmentVariable("KRAKEND_HOST", "krakend")
-	krakendPort := getEnvironmentVariable("KRAKEND_PORT", "1234")
-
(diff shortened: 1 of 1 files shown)


chibenwa:
So, for every request we do open a new connection, right?

Performance wise, this do not look wise! This might be acceptable as "logouts" are expected to be rare, but I would prefer us to just renew connections if they are broken, not all the time.

Arsnael:
> So, for every request we do open a new connection, right?
> 
> Performance wise, this do not look wise! This might be acceptable as "logouts" are expected to be rare, but I would prefer us to just renew connections if they are broken, not all the time.

Honesly the krakend rpc client looks very simple to me, without such mechanisms... but yeah I will try to check... That might imply connection detection, retries, etc... and I'm not proficient in golang just saying

Arsnael:
Ideal would be to have one persistent connection between jwt-revoker and krakend for resources and performances concerns. But that would imply a considerable effort to implement necessary things like a pool of connections, connection loss detection, retry mechanisms, ... in golang

Because logout is considered not happening often, it's totally acceptable for now to only create a connection when adding a token to krakend and then closing it down (which is what the above PR proposes).

chibenwa (approved):
I think one connection per logout is acceptable, especially in an example folder.

PR #868, opened 2022-01-26
JAMES-3709 Concurrent message deletion can cause issue in POP3

chibenwa:
cc @ottoka

### DIFF
--- a/server/protocols/protocols-pop3/src/test/java/org/apache/james/pop3server/POP3ServerTest.java
+++ b/server/protocols/protocols-pop3/src/test/java/org/apache/james/pop3server/POP3ServerTest.java
@@ -19,6 +19,7 @@
 package org.apache.james.pop3server;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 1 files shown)


server/protocols/protocols-pop3/src/test/java/org/apache/james/pop3server/POP3ServerTest.java
```diff
@@ -544,6 +549,55 @@ void testKnownUserInboxWithMessages() throws Exception {
...
+            Flux.from(mailboxManager.getMailbox(mailboxPath, session).listMessagesMetadata(MessageRange.all(), session))
+            .map(i -> i.getComposedMessageId().getUid())
+            .collectList().block(),
+            session);
+
+        Reader r = pop3Client.retrieveMessageTop(entries[0].number, 0);
+
+        assertThat(r).isNotNull();
+        assertThat(IOUtils.toString(r)).isEqualTo("-ERR\r\n");
+
```
ottoka, line 596: In order for the test to pass, after the fix, the two lines above should be:
assertThat(r).isNull();
which pop3client returns in case of an error. But this is what we want, since a fixed handler should respond with "-ERR message is already deleted".

(The two lines do work without the fix, since initially you get "+OK message follows" but with a null content stream, so the subsequent read via IOUtils.toString(r) cause the NPE. Then the "content" read is the resulting "-ERR".)

ottoka:
I can also confirm that this test indeed reproduces the issue. I can also confirm that my fixes from https://github.com/apache/james-project/pull/861 lets it succeed, given the change from my comment above. How to proceed from here?

chibenwa:
> I can also confirm that this test indeed reproduces the issue. I can also confirm that my fixes from #861 lets it succeed, given the change from my comment above. How to proceed from here?

I would suggest you cherry-pick and adapt my commit, remove the @Disabled in your PR.

Also this would be a good opportunity IMO to handle Mailbox::getMessage nullity issue.

server/protocols/protocols-pop3/src/test/java/org/apache/james/pop3server/POP3ServerTest.java
chibenwa: Subttle bug!

chibenwa:
Fixed as part of https://github.com/apache/james-project/pull/861

PR #874, opened 2022-02-06
[SONARLINT] Fixing a bunch of sonarlint wanrings across the project

chibenwa:
- Bug fixes
 - Performance enhancements
 - Code style refactorings

### DIFF
--- a/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/jsieve/FileIntoAction.java
+++ b/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/jsieve/FileIntoAction.java
@@ -73,26 +73,17 @@ public class FileIntoAction implements MailAction {
     public void execute(ActionFileInto anAction, Mail aMail, final ActionContext context) throws MessagingException {
         String destinationMailbox = anAction.getDestination();
         MailAddress recipient;
-        boolean delivered = false;
-        try {
-            recipient = ActionUtils.getSoleRecipient(aMail);
-            
-            if (!(destinationMailbox.length() > 0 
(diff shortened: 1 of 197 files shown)


server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/jsieve/FileIntoAction.java
```diff
@@ -73,26 +73,17 @@ public void execute(Action action, Mail mail, ActionContext context) throws Mess
...
-                destinationMailbox =  HIERARCHY_DELIMITER + destinationMailbox;
-            }
-            
-            final String mailbox = destinationMailbox.replace(HIERARCHY_DELIMITER, '/');
-            final String url = "mailbox://" + recipient.asString() + mailbox;
-            //TODO: copying this message so many times seems a waste
-            context.post(url, aMail);
-            delivered = true;
-        } catch (MessagingException ex) {
-            LOGGER.debug("Error while storing mail into {}.", destinationMailbox, ex);
```
Arsnael, line 91: So we removing here the debug error log statement so we can get rid of the try catch block right? Is that ok?

Arsnael (approved):
Nice little fixes :)

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/jsieve/FileIntoAction.java
chibenwa: It's OK as the exception is re-thrown.

Log then throw is a classic anti-pattern upon exception handling.

Arsnael:
Rebase to solve conflicts

PR #888, opened 2022-02-18
JAMES-3150 Fix RabbitMQWebAdminServerBlobGCIntegrationTest

chibenwa:
An input stream was reused, thus empty on the second try. We
should fix this input mistake and correct the wrong assertions
based on it.


Arsnael (approved):

quantranhong1999 (approved):

Arsnael:
Green on the CI, don't know why a step is still yellow here

PR #890, opened 2022-02-20
JAMES-3150 Improvments regarding Blob deduplication and garbage collection

chibenwa:
- Apply duplicating/deduplicating module choosing for Cassandra guice product too

 -> This is what the doc says
 -> This limits behavior divergence between the Cassandra and
 Distributed products

 Because there's nothing worse that choosing the wrong BlobStore
 we need to apply this configuration.

 AES encryption ships as a bonus...

 - GC BloomFilter default false positive rate 0.8 -> 0.01

0.8 as a false positive rate do not make sense.

We furthermore slightly better document bloom filter sizing
properties.

 - Allow configuration of generation duration and family


Arsnael:
Likely related :) 

```
java.lang.AssertionError: 
1 expectation failed.
JSON path additionalInformation.bloomFilterAssociatedProbability doesn't match.
Expected: is <0.8F>
  Actual: <0.01F>

	at org.apache.james.webadmin.integration.rabbitmq.RabbitMQWebAdminServerTaskSerializationIntegrationTest.blobGCTaskShouldComplete(RabbitMQWebAdminServerTaskSerializationIntegrationTest.java:7
```

Arsnael (approved):
Overall I approve

Arsnael (approved):

PR #898, opened 2022-02-26
JAMES-3715 Graceful shutdown for Netty servers

chibenwa:
Waits ongoing tasks being executed on Netty servers for a few seconds
before releasing underlying resources (like DB connections).

Especially on top of Cassandra components, this is a good first step
to minimize risks of shutdowns to cause inconsistencies.

The Netty4 migration enabled us to easily write such resilient
behaviour while it was not implemented for the previously existing
Netty 3 code. CF https://github.com/apache/james-project/pull/886

### DIFF
--- a/mpt/impl/smtp/cassandra-pulsar/src/test/resources/smtpserver.xml
+++ b/mpt/impl/smtp/cassandra-pulsar/src/test/resources/smtpserver.xml
@@ -43,6 +43,7 @@
             <handler class="org.apache.james.smtpserver.fastfail.ValidRcptHandler"/>
             <handler class="org.apache.james.smtpserver.CoreCmdHandlerLoader"/>
         </handlerchain>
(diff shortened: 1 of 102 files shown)


chibenwa:
This code need to be reworked: the graceful shutdown process takes ~2 seconds and we don't want to wait that long in tests.

Solution -> make it configurable

chibenwa:
Reminder to self: I still need to turn off graceful shutdown in mailet integration tests...

glennosss:
I think you'll find `unbindNow()` is a cleaner way of doing this - see: https://github.com/glennosss/james-project/blob/netty4-combo/protocols/api/src/main/java/org/apache/james/protocols/api/ProtocolServer.java#L67

chibenwa:
Alreanative PR welcomed ;-)

PR #899, opened 2022-02-27
JAMES-3715 Small cleanups following Netty 4 migration

chibenwa:
Cc @ff-wl I'm  interested by your feedback...

Playing around with the code resulting from the Netty 4 migration ( https://github.com/apache/james-project/pull/886 ) I realized:

  - That the threading model was further simplified than I thought (which is good!) and that we could get rid of a few configuration options and of a few extra classes (yeah!)
  - That some ssl related options were leftover
  - And that the ioWorker option was no longer applied while we easily could.
  
@ff-wl feel free to cherry-pick the commits.


ff-wl:
> Cc @ff-wl I'm interested by your feedback...
> 
> Playing around with the code resulting from the Netty 4 migration ( #886 ) I realized:
> 
> * That the threading model was further simplified than I thought (which is good!) and that we could get rid of a few configuration options and of a few extra classes (yeah!)
> * That some ssl related options were leftover
> * And that the ioWorker option was no longer applied while we easily could.
> 
> @ff-wl feel free to cherry-pick the commits.

looks good for me. I've cherry-picked all commits and will push them to the #886 branch.

chibenwa:
My bad https://github.com/apache/james-project/pull/899/commits/f7f0c182ee1260d330ad2ed974bd07f9f1e07436 should be removed from here: it is a mistake... Sorry...

PR #902, opened 2022-03-03
JAMES-3719 Reactive textual content extraction with Apache Tika

chibenwa:
Tika was called from reactive code and was doing blocking HTTP calls from within
the MIME parsing code.

This generate:
 - An unneeded thread consumption as we have some threads waiting for Tika
   response
 - Potentially dangerous blocking calls: for instance the InVM event bus was
  doing such calls on the parallel thread pool (where it is critical NOT to
  block)...
 - Also the connection was opened on a per-call basis, not being reused.

 We introduce the following changes:
  - Reactification of the TextExtractor API
  - We re-implement the HTTP calls done by TikaTextExtractor with reactor-netty
  which allows us to pool HTTP connections and do this in a non-blocking
  reactive fashion.
  - We provide a reactive cache using the caffeine caching library - Guava
  caches are blocking thus not an option...
  - We uncouple the text extraction from the MIME parsing phase by introducing
  an intermediate POJO. Doing so requires us to do a post-parsing copy of
  content.

 Only do the copy if necessary. We don't want to copy large attachments for whom no text is going to be extracted...

  - Finally we reactify index content generation for ElasticSearch code.

### DIFF
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/extractor/TextExtractor.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/extractor/TextExtractor.java
@@ -24,6 +24,7 @@ import java.io.InputStream;
 import org.apache.james.mailbox.model.ContentType;
 
 import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
 
 public interface TextExtractor {
     default boolean applicable(ContentType contentType) {
@@ -33,7 +34,8 @@ public interface TextExtractor {
     ParsedContent extractContent(InputStream inputStream, ContentType contentType) throws Exception;
 
     default Mono<ParsedContent> extractContentReactive(InputStream inputStream, ContentType contentType) {
-        return Mono.fromCallable(() -> extractContent(inputStream, contentType));
+        return Mono.fromCallable(() -> extractContent(inputStream, contentType))
+            .subscribeOn(Schedulers.elastic());
     }
 
 }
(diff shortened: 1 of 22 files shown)


mailbox/api/src/main/java/org/apache/james/mailbox/extractor/TextExtractor.java
```diff
@@ -23,8 +23,14 @@
...
 import org.apache.james.mailbox.model.ContentType;
 
+import reactor.core.publisher.Mono;
+
 public interface TextExtractor {
 
     ParsedContent extractContent(InputStream inputStream, ContentType contentType) throws Exception;
 
+    default Mono<ParsedContent> extractContentReactive(InputStream inputStream, ContentType contentType) {
+        return Mono.fromCallable(() -> extractContent(inputStream, contentType));
```
mbaechler, line 36: given that extractContent is blocking, would it makes sense to enforce this to be executed in an elastic scheduler?

mbaechler (approved):
it looks good

mailbox/api/src/main/java/org/apache/james/mailbox/extractor/TextExtractor.java
chibenwa: Good question. In practice this is, with current implementation, a byte array input stream. But yes I will put that subscribeOn for the sake of genericity or we might regret it later.

Good catch!

Arsnael (approved):

PR #908, opened 2022-03-07
JAMES-3715 Execute core handlers outside of the event loop

chibenwa:
## Why?

 - Core handlers are blocking and one should never block on the event loop.
 - Order of outgoing messages is not preserved for items sent from threads within and out of the event loop
 
## Current caveats

https://gitter.im/netty/netty?at=62257c72d1b64840db5467af

Currently I encounter issues with SMTP pipelining: the client sends all the SMTP requests into a single network hop.

```
        Socket client = new Socket(bindedAddress.getAddress().getHostAddress(), bindedAddress.getPort());
        buf.append("HELO TEST");
        buf.append("\r\n");
        buf.append("MAIL FROM: <test@localhost>");
        buf.append("\r\n");
        buf.append("RCPT TO: <test2@localhost>");
        buf.append("\r\n");
        buf.append("DATA");
        buf.append("\r\n");
        buf.append("Subject: test");
        buf.append("\r\n");
        buf.append("\r\n");
        buf.append("content");
        buf.append("\r\n");
        buf.append(".");
        buf.append("\r\n");
        buf.append("quit");
        buf.append("\r\n");

        OutputStream out = client.getOutputStream();
        out.write(buf.toString().getBytes());
        out.flush();
```

CF https://github.com/apache/james-project/blob/634be36800422259e40121248fb0d07475024a9c/server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SMTPServerTest.java#L1887

The way `DATA` is handled is that it adds a handler prior the core SMTP handler: https://github.com/chibenwa/james-project/blob/ffc0d4a8b22508b8f5b58594d14041d1f6bc3acf/protocols/netty/src/main/java/org/apache/james/protocols/netty/NettyProtocolTransport.java#L162

```
        channel.pipeline().addBefore(eventExecutors, HandlerConstants.CORE_HANDLER, "lineHandler" + lineHandlerCount, new LineHandlerUpstreamHandler(session, overrideCommandHandler));
```

Everything works fine if the core handler is running on the event loop, however once we switch it to a distinct executor, the pipeline modification is no longer applied and the subsequent message content is interpreted as SMTP commands as it the line handler was not there.

CF https://github.com/chibenwa/james-project/blob/ffc0d4a8b22508b8f5b58594d14041d1f6bc3acf/protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractChannelPipelineFactory.java#L78

```
        // Fails
        pipeline.addLast(eventExecutorGroup, HandlerConstants.CORE_HANDLER, createHandler());
        // Succeed
        pipeline.addLast(HandlerConstants.CORE_HANDLER, createHandler());
```

Is there any way to modify the pipeline 'synchronously' from outside the event loop?


Arsnael (commented):
Read it

chibenwa:
Todo rebase on master ?

Arsnael (approved):

chibenwa:
To be continued at https://github.com/apache/james-project/pull/924

PR #903, opened 2022-03-04
JAMES-3715 Second round of small refactorings following netty 4 migration

chibenwa:
This is likely not addressing issues reported so far in https://github.com/apache/james-project/pull/886 but it could be beneficial...

PR #926, opened 2022-03-22
JAMES-3715 Small round of Netty4 polishing

chibenwa:
Rebased version of https://github.com/apache/james-project/pull/903

TODO quich performance checks.


mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/GhostMailbox.java
```diff
@@ -35,6 +35,14 @@
     public static final String TYPE = "type";
     public static final String MAILBOX_ID = "mailboxId";
 
+    public static boolean isDebugEnabled() {
+        return LOGGER.isDebugEnabled();
+    }
+
+    public static boolean isInfoEnabled() {
+        return LOGGER.isDebugEnabled();
```
Arsnael, line 43: You return here if debug is enabled, nothing to do with info...

I don't think the method `isInfoEnabled` would be used anyway, and it's not in your code for now. We can remove it?

chibenwa:
Forced pushed to remove the last commit which was unrelated.. Was :green_apple:

Arsnael (approved):

chibenwa:
Forced pushed to solve a rebase conflict...

PR #930, opened 2022-03-23
JAMES-3738 Ease overrides of Encryption

chibenwa:


server/protocols/protocols-library/src/main/java/org/apache/james/protocols/lib/SslConfig.java
```diff
@@ -0,0 +1,126 @@
+package org.apache.james.protocols.lib;
```
chibenwa, line 1: Missing license

chibenwa:
Build was :green_apple: , I force pushed to reword the commits messages so that they include the right ticket numbers

Arsnael (approved):

chibenwa:
Build was :green_apple: , I force pushed to solve a rebase conflict

chibenwa:
Rebased and force pushed to solve conflict

PR #932, opened 2022-03-24
JAMES-3732 Turn Netty leak detection to paranoid in protocols unit test

chibenwa:


Arsnael:
As I commented in the respective JIRA ticket... it's cool but... adding just paranoid level for leak detection in netty in tests... just for the pleasure of it.... if we don't find a way to fail the test when such a leak is detected I hardly see the point... Except for making our logs potentially bigger and our builds longer.

Well except if someone accepts to checks the full log of each build for leak detections?

chibenwa:
Yes I know, currently build don't fail upon leak.

But IMO turning on leak detection is better than not turning it on at all hence this PR.

Arsnael:
I just feel this is not really gonna be exploited... like cool we enable it but nobody looks at it. Then I'm waiting to see how potentially longer the build will become... I'm half convinced TBH but sure let's try

chibenwa:
> Then I'm waiting to see how potentially longer the build will become...

Enabled only in protocols tests, not a concern to be honnest

chibenwa:
> I just feel this is not really gonna be exploited... like cool we enable it but nobody looks at it. Then I'm waiting to see how potentially longer the build will become... I'm half convinced TBH but sure let's try

You are free to configure the build to fail ^^

Arsnael (approved):

PR #935, opened 2022-03-24
JAMES-3736 IMAP use a Map instead of a chain

chibenwa:
Do a map lookup instead of calling processors in the chain.

This leads to very messy stack, which complicates analysis of IMAP workloads.


Arsnael (approved):

chibenwa:
(Build was :green_apple: , I just reworded the commit with the right JIRA ticket)

PR #952, opened 2022-04-05
JAMES-3700 Add dead letter policy for Pulsar MailQueue

haanhvu:
see https://issues.apache.org/jira/browse/JAMES-3700

@chibenwa can you check if I'm going the right way or not?


chibenwa (commented):
It's definitly going in the right direction.

Is there a mechanism to know, for a message in dead-letter, which topic it originate from?

Also, in case of failed processing, it would be very handy to get an applicative log (in James log). Is it the case today?

Also, I know you are interested in GSOC, can you formally propose to participate to it in https://issues.apache.org/jira/browse/JAMES-3695 ? It would be important for us to do the mentoring requests upfront, I don't know how we stand regarding the agenda but we might need to act timely on this topic...

chibenwa:
Apparently, scala compilation fails...

```
[ERROR] Failed to execute goal net.alchim31.maven:scala-maven-plugin:4.5.6:compile (scala-compile-first) on project james-server-queue-pulsar: scala compilation failed -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecu
```

Can you check?

jeantil (changes requested):

chibenwa:
Hello @haanhvu 

Any follow up on this work?

haanhvu:
> Hello @haanhvu
> 
> Any follow up on this work?

I'm at the end of my school term and having some stuff to complete. Will come back to this PR this weekend.

chibenwa:

haanhvu:
> Is there any plan to carry on this work?

No sorry. Let me close this now.

PR #957, opened 2022-04-08
JAMES-3744 remove redundant hashmaps which lead to OOM 

chibenwa:
Screenshot from 2022-04-08 21-01-17

CF https://github.com/reactor/reactor-netty/commit/fe6653fac44df16664bdd0628ad56cd142175bca

By caching the params and matchers this class creates a laaarge amount of memory never cleaned up.

This only grow, and this memory leak could be leveraged to conduct DOS attacks by memory exhaustion.


server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/EventSourceRoutes.scala
```diff
@@ -165,7 +165,7 @@ class EventSourceRoutes@Inject() (@Named(InjectionKeys.RFC_8621) val authenticat
       .action(this.handleSSE)
       .corsHeaders,
     JMAPRoute.builder
-      .endpoint(new Endpoint(HttpMethod.OPTIONS, EVENT_SOURCE))
+      .endpoint(new Endpoint(HttpMethod.GET, EVENT_SOURCE))
```
quantranhong1999, line 168:
```suggestion
      .endpoint(Endpoint.ofFixedPath(HttpMethod.GET, EVENT_SOURCE))
```

Arsnael (approved):

server/protocols/jmap/src/main/java/org/apache/james/jmap/Endpoint.java
```diff
@@ -45,11 +55,11 @@ public String getPath() {
 
     public boolean matches(HttpServerRequest request) {
         return method.equals(request.method())
-            && uriPathTemplate.matches(request.uri());
+            && uriMatcher.matches(request.uri());
     }
 
-    UriPathTemplate getUriPathTemplate() {
-        return uriPathTemplate;
+    UriMatcher getUriPathTemplate() {
```
quantranhong1999, line 61:
```suggestion
    UriMatcher getUriMatcher() {
```

quantranhong1999 (approved):

PR #958, opened 2022-04-08
JAMES-3745 Use FastByteArrayOutputStream as a short lived object

chibenwa:
From connection scoped to request scoped.

I started studying production heap dump in an effort to reduce the per user memory footprint. In conjunction with issues discussed in JAMES-3740 I found that in FastByteArrayOutputStream, upon writing a long line this buffer is sized up and never sized down.

We can end up with 60KB+ memory allocated per user... Forever.

While the impact is not as huge as the one described in JAMES-3740 yet for James deployment with a high connection density this could be non neglictible (100s MB of RAM).

Furthermore given the long lived lifecycle of this buffer (connection) it is very likely to be promoted to the old gen, which could build up to eventually get long GC pauses.

Suggestion: create a new IMAP composer for each request, 2 -> 64KB allocation is anyway neglictibe, likely cheaper than the lookups to get them back! We will thus allocate more memory, yes, but only for short lived durations, wich JVM GC is really good at.

IMO it is so easy to avoid this unwanted behaviour that it would be sad not to do it.


chibenwa:
fastByteArray

quantranhong1999 (approved):

PR #965, opened 2022-04-15
JAMES-3750 Add an upsert method in ElasticsearchIndexer

Arsnael:


quantranhong1999 (approved):

chibenwa:
Forgive me if I'm wrong but won't we get the same behaviour than "upsert"by doing a regular call to 'index' ?

Arsnael:
> Forgive me if I'm wrong but won't we get the same behaviour than "upsert"by doing a regular call to 'index' ?

Does it? If you index something that exists it updates it, that's not an error? I will check, if it is then my bad!

Arsnael:
Ok it does... Nevermind this :)

PR #976, opened 2022-04-26
[PERF] Some enhancement for IMAP

chibenwa:
We recently worked on a more meaningfull IMAP performance test that highlighted many tiny and not so tiny area for performance enhancements that this work try to address.


Arsnael (approved):

chibenwa:
Rebase + CapabilityTest should match bean contract

PR #977, opened 2022-04-26
Revert "JAMES-3747 Reactive implementation for RabbitMQ channelPool"

chibenwa:
This reverts commit 8564a2ecf72e0a7ef404fe491799fb8140dea1f6.

### DIFF
--- a/backends-common/rabbitmq/pom.xml
+++ b/backends-common/rabbitmq/pom.xml
@@ -83,10 +83,6 @@
             <artifactId>feign-slf4j</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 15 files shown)


Arsnael (approved):
What was the issue exactly?

chibenwa:
To be faily honnest I do not know... Some RabbitMQ MPT tests starts failing saying they can't open channels so it seems related. I reviewed the parameters of these tests but could not come up with a fix - though I did not put much efforts on it...

Thus it's better to revert for now.

quantranhong1999 (approved):

PR #980, opened 2022-05-03
JAMES-1489 Fix decoding for complex UTF-8 searches

chibenwa:
- Reader index was  not properly reset for
multiple subsequent continuations.
 - Charset was not carried over for all the
 terms of the search.

### DIFF
--- a/protocols/imap/src/main/java/org/apache/james/imap/decode/parser/SearchCommandParser.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/decode/parser/SearchCommandParser.java
@@ -24,6 +24,7 @@ import java.nio.charset.UnsupportedCharsetException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 
 import org.apache.james.imap.api.ImapConstants;
 import org.apache.james.imap.api.ImapMessage;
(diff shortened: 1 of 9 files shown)


Arsnael (approved):

Arsnael:
Hmm lots of null pointer exceptions in tests... like this one:
```
org.apache.james.imap.decode.parser.SearchCommandParserCharsetTest.testASCIICharset
java.lang.NullPointerException
	at org.apache.james.imap.decode.parser.SearchCommandParserCharsetTest.checkValid(SearchCommandParserCharsetTest.java:169)
	at org.apache.james.imap.decode.parser.SearchCommandParserCharsetTest.testASCIICharset(SearchCommandParserCharsetTest.java:142)
```

chibenwa:
Yes, tests with mocks!

Arsnael (approved):

protocols/imap/src/main/java/org/apache/james/imap/decode/parser/SearchCommandParser.java
```diff
@@ -49,6 +50,18 @@
...
+        private Optional<Charset> previousCharset = Optional.empty();
+
+        public void setCharset(Charset charset) {
+            previousCharset = Optional.of(charset);
+        }
+
+        public Charset getCharset() {
+            return previousCharset.orElse(null);
+        }
+    }
```
quantranhong1999, line 63: Q: Why do we use Optional when directly returning null?

quantranhong1999 (approved):

protocols/imap/src/main/java/org/apache/james/imap/decode/parser/SearchCommandParser.java
chibenwa: A: because I wanted to minimize the changes in this complex call, it would have had required to change 20+ method signature.

PR #981, opened 2022-05-04
[UPGRADE] Security related upgrades

chibenwa:


Arsnael (approved):

quantranhong1999 (approved):
`dgrijalva/jwt-go` is no longer maintained. We may need to move to `golang-jwt/jwt` at some point.

chibenwa:
> dgrijalva/jwt-go is no longer maintained. We may need to move to golang-jwt/jwt at some point.

Ticket? :-)

PR #985, opened 2022-05-04
JAMES-3761 Adds configurable bucketname to MimeMessageStore

mbaechler:
Co-Authored-By: Matthieu Baechler <matthieu@apache.org>

### DIFF
--- a/server/blob/blob-common/src/main/java/org/apache/james/blob/api/Store.java
+++ b/server/blob/blob-common/src/main/java/org/apache/james/blob/api/Store.java
@@ -77,14 +77,15 @@ public interface Store<T, I> {
         private final Encoder<T> encoder;
         private final Decoder<T> decoder;
         private final BlobStore blobStore;
+        private final BucketName bucketName;
 
-        public Impl(BlobPartsId.Factory<I> idFactory, Encoder<T> encoder, Decoder<T> decoder, BlobStore blobStore) {
+        public Impl(BlobPartsId.Factory<I> idFactory, Encoder<T> encoder, Decoder<T> decoder, BlobStore blobStore, BucketName bucketName) {
             this.idFactory = idFactory;
             this.encoder = encoder;
             this.decoder = decoder;
             this.blobStore = blobStore;
+            this.bucketName = bucketName;
         }
-
         @Override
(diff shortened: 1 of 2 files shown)


chibenwa (approved):
No objections. As it stands it seems to benefit custom code only?

Arsnael (approved):

Arsnael:
```
22:14:18,258 [INFO] There are 2 errors reported by Checkstyle 8.29 with checkstyle.xml ruleset.

22:14:18,263 [ERROR] src/main/java/org/apache/james/blob/api/Store.java:[89] (whitespace) EmptyLineSeparator: 'METHOD_DEF' should be separated from previous statement.

22:14:18,264 [ERROR] src/main/java/org/apache/james/blob/api/Store.java:[96] (whitespace) EmptyLineSeparator: 'METHOD_DEF' should be separated from previous statement.
```

jeantil:
> No objections. As it stands it seems to benefit custom code only?

It is used in the BlobStore backed MailRepository of #987 but we felt that the change made sense on its own as there can be multiple MimeMessageStores in a single assembly even before the BlobStore backed Mailrepository. 
For example pulsar-cassandra-smtp-relay uses both CassandraMailRepository and the pulsar mailqueue both of which use a MimeMessageStore. As it stands neither adds custom qualifiers while wiring the message store. Unless I am mistaken this means that they will both store their respective blobs in the same bucket. While there are not direct consequences this is doesn't seem to be a very desirable situation.

chibenwa:
```
07:25:20,278 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (check-style) on project james-server-mail-store: You have 1 Checkstyle violation. -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.0:check (check-style) on project james-server-mail-store: You have 1 Checkstyle violation.
```

Still....

jeantil:
I can't thank checkstyle enough for this unhelpful message that nicely omits the place where the violation occurs, my workday is starting I will continuer searching what's wrong later

mbaechler:
test this please

PR #988, opened 2022-05-05
JAMES-2080 Use flattened type for indexing headers

chibenwa:
## TODO

 - [ ] Measure index size benefits
 - [ ] Measure indexation time benefits
 - [ ] Once validated write detailed upgrade instructions


chibenwa:
Let's be serious, pushing tokenisation on James side is not a good idea...

PR #991, opened 2022-05-06
JAMES-3744 Fixed URI matcher should be more flexible

quantranhong1999:
This allows matching routes with query param.


chibenwa:
Can we have a unit test in JMAP tooling reproducing what you try to fix?

chibenwa:
IMO Fixed UtiMatcher could be made smarter to ignore query parameters, IE if the matched sting contains `?` we instead took the substring until `?`. Would it solves your problem too?

A unit test for Fixed Uri matcher would then be enough to me to back this changeset.

chibenwa (approved):

Arsnael (approved):

vttranlina (approved):

PR #992, opened 2022-05-06
JAMES-3534 Avoid blocking calls in Identity/set

chibenwa:


vttranlina (approved):

quantranhong1999 (approved):
It's weird that CI did not spot this and maybe our local also. My local ITs seem only failing when I debug the .toStream recently.

PR #996, opened 2022-05-11
[BUILD] Increase ConcurrentTestRunnerTest await time

chibenwa:
Locally 50ms was borderline with random failures, so
I expect the default value of 100 ms to be too low for
being reliable on a busy CI server where context switch
might be made expensive with expensive concurrent builds.

I hereby propose a safer, higher value of 2 seconds.


chibenwa:
cf https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-902/2/testReport/junit/org.apache.james.util.concurrency/ConcurrentTestRunnerTest/runShouldPerformAllOperationsEvenOnExceptions/

Arsnael (approved):

PR #997, opened 2022-05-11
JAMES-3765: Improve some IMAP commands (STORE, COPY, MOVE) performance

quantranhong1999:


Arsnael (approved):

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraIndexTableHandler.java
```diff
@@ -129,8 +129,7 @@ public Mono<Void> updateIndexOnAdd(MailboxMessage message, CassandraId mailboxId
             .then();
     }
 
-    public Mono<Void> updateIndexOnAdd(Collection<MailboxMessage> messages, CassandraId mailboxId) {
-        int lowConcurrency = 2;
+    public Mono<Void> updateIndexOnAdd(Collection<MailboxMessage> messages, CassandraId mailboxId, int reactorConcurrency) {
```
chibenwa, line 132: Please use `fairConcurrency` variable like below and do not leak this detail outside.

'Index' projections are not doing LWT thus enabling concurrency here will not result in CAS contension,

chibenwa, line 132:
```suggestion
    public Mono<Void> updateIndexOnAdd(Collection<MailboxMessage> messages, CassandraId mailboxId) {
        int fairConcurrency = 4;
        // ...
```

mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/MessageMapper.java
```diff
@@ -207,7 +207,7 @@ default Mono<List<UpdatedFlags>> resetRecentReactive(Mailbox mailbox) {
     MessageMetaData copy(Mailbox mailbox, MailboxMessage original) throws MailboxException;
 
     default List<MessageMetaData> copy(Mailbox mailbox, List<MailboxMessage> original) throws MailboxException {
-        return original.stream()
+        return original.parallelStream()
```
chibenwa, line 210: parrallelStream is a dangerous abstraction, we should not use it. Doing one range at the time is likely enough BTW.

If you really want to do stuff parrallely migrate this wall chunk of code to reactor.

chibenwa:
Read of the night: https://www.batey.info/cassandra-anti-pattern-misuse-of.html

TLDR batch can improve perf when acting on a single partition.

Prime candidate: index table handler: recents deleted unseen.

Then maybe also messageIdTable if it do not impact things too much.

We could see a 50% enhancement from this but perf gains needs to be tested. Also batches could likely benefit is from a given threshold. Eg above 8 entries use batch...

chibenwa:
Quick local bench for batch on top of CassandraDeletedMessageDAO

DAO modifications:

```
    public Mono<Void> addDeleted(CassandraId cassandraId, Flux<MessageUid> uids) {
        return uids.reduce(new BatchStatement(BatchStatement.Type.UNLOGGED), (batch, uid) ->
            batch.add(addStatement.bind()
                .setUUID(MAILBOX_ID, cassandraId.asUuid())
                .setLong(UID, uid.asLong())))
            .flatMap(cassandraAsyncExecutor::executeVoid);
    }
```

Original bench:

```
    @Test
    void bench() {
        Flux.range(0, 100000)
            .doOnNext(i -> {
                if (i % 1000 == 0) {
                    System.out.println(i);
                }
            })
            .flatMap(i -> testee.addDeleted(MAILBOX_ID, MessageUid.of(i)), 4)
            .blockLast();
    }
```

Executes in 20s locally. Without specifying concurrency on the flux this is still 9.2 seconds for an unfair Cassandra query budget consumption.

With batch:

```
    @Test
    void bench2() {
        Flux.range(0, 100000)
            .map(MessageUid::of)
            .window(1000)
            .flatMap(uids -> testee.addDeleted(MAILBOX_ID, uids), 2)
            .blockLast();
    }
```

Executes in 2.4 seconds.

This tends to prove the benefit of unlogged batch as a performance improvments when performing multiple inserts to the same primary key as mentionned by the link above.

We still need gatling confirmation...

quantranhong1999:
I did put an effort to try to optimize messageIdTable and imapUidTable delete with the batch statement here https://github.com/quantranhong1999/james-project/commit/b1fa4063d2910899d7f3afc0b5d950fd2a9f1fb1 (some tests are failed though). You can come and have a look if I am going the exact way.

Also, we should have a dedicated ticket to improve our Cassandra queries using the Batch statement where appropriate.

I will try a custom image to see how this impacts next.

chibenwa:
Don't use batches on imapUidTable !

We should not mix primary keys in batches OR we will have a big performance hit!

quantranhong1999:
> We should not mix primary keys in batches OR we will have a big performance hit!

Partition key more exactly? BTW I did some group-by to avoid messing up the partition key together.

chibenwa:
Please re-read https://www.batey.info/cassandra-anti-pattern-misuse-of.html

quantranhong1999:
> Please re-read https://www.batey.info/cassandra-anti-pattern-misuse-of.html

`Most of the time you don't want to use unlogged batches with Cassandra. The time you should consider it is when you have multiple inserts/updates for the same partition key.`

chibenwa:
Entries have different primary keys on imapUidDAO so we should not use batches there. Ever.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraDeletedMessageDAO.java
```diff
@@ -123,12 +124,34 @@ public Mono<Void> addDeleted(CassandraId cassandraId, MessageUid uid) {
                 .setLong(UID, uid.asLong()));
     }
 
+    public Mono<Void> addDeleted(CassandraId cassandraId, Flux<MessageUid> uids) {
+        return uids
+            .window(65535)
```
vttranlina, line 129: just curious, why the number is 65535?

quantranhong1999: But maybe I should reduce this quite a bit here.

chibenwa: Yes IMO 1024 was cool on local benches.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraDeletedMessageDAO.java
```diff
@@ -123,12 +124,34 @@ public Mono<Void> addDeleted(CassandraId cassandraId, MessageUid uid) {
                 .setLong(UID, uid.asLong()));
     }
 
+    public Mono<Void> addDeleted(CassandraId cassandraId, Flux<MessageUid> uids) {
```
chibenwa, line 127: We have a simpler reactive pipeline if we have a collection here. We should likely consider it.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraDeletedMessageDAO.java
```diff
@@ -123,12 +124,34 @@ public Mono<Void> addDeleted(CassandraId cassandraId, MessageUid uid) {
...
 
+    public Mono<Void> removeDeleted(CassandraId cassandraId, Flux<MessageUid> uids) {
+        return uids
+            .window(65535)
+            .flatMap(partialUidsFlux -> partialUidsFlux.reduce(new BatchStatement(BatchStatement.Type.UNLOGGED), (batch, uid) ->
+                batch.add(deleteStatement.bind()
+                    .setUUID(MAILBOX_ID, cassandraId.asUuid())
+                    .setLong(UID, uid.asLong()))))
+            .flatMap(cassandraAsyncExecutor::executeVoid)
+            .then();
```
chibenwa, line 152: If we use a collection we could see if the size is worth it to use a batch and batch things from (say) 2+items...

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraDeletedMessageDAO.java
Arsnael: Then maybe setting up a static constant to make this more clear? :)

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraDeletedMessageDAO.java
```diff
@@ -123,12 +127,47 @@ public Mono<Void> addDeleted(CassandraId cassandraId, MessageUid uid) {
...
+                addStatement.bind()
+                    .setUUID(MAILBOX_ID, cassandraId.asUuid())
+                    .setLong(UID, uids.iterator().next().asLong()));
+        } else {
+            return Flux.fromIterable(uids)
+                .window(BATCH_STATEMENT_WINDOW)
+                .flatMap(partialUidsFlux -> partialUidsFlux.reduce(new BatchStatement(BatchStatement.Type.UNLOGGED), (batch, uid) ->
+                    batch.add(addStatement.bind()
+                        .setUUID(MAILBOX_ID, cassandraId.asUuid())
+                        .setLong(UID, uid.asLong()))))
```
chibenwa, line 142: We IMO do not need a Flux to build the batch statements.

```
Stream<BatchStatement> batches = Lists.partition(uids, 1024)
    .stream()
    .map(uidBatch -> {
        BatchStatement batch = new BatchStatement(BatchStatement.Type.UNLOGGED);
        uidBatch.forEach(uid -> batch.add(addStatement.bind()
                        .setUUID(MAILBOX_ID, cassandraId.asUuid())
                        .setLong(UID, uid.asLong()))));
        return batch;
    });
return Flux.fromStream(batches)
    .concatMap(cassandraAsyncExecutor::executeVoid)
    .then();
```

Rationals: being asynchronous is not needed when building the batches... `window` operator likely is expensive.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraDeletedMessageDAO.java
```diff
@@ -123,12 +127,47 @@ public Mono<Void> addDeleted(CassandraId cassandraId, MessageUid uid) {
...
+                .flatMap(cassandraAsyncExecutor::executeVoid)
```
chibenwa, line 143: Executing 128 batches in parallel sounds like a bad idea. concatMap maybe? or flatMap with a really low concurrency?

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreMailboxManager.java
```diff
@@ -590,7 +590,7 @@ private List<MailboxRenamedResult> doRenameMailbox(Mailbox mailbox, MailboxPath
                 .asUserBound();
             locker.executeWithLock(from, (LockAwareExecution<Void>) () -> {
                 block(mapper.findMailboxWithPathLike(query)
-                    .flatMap(sub -> {
+                    .concatMap(sub -> {
```
chibenwa, line 593: I think we can agree this concatMap is not really beneficial...

Should we roll it back? What do you think?

quantranhong1999: Me I think RENAME is not a common command to be called much in the real world and even its edge case with so many sub mailboxes, therefore I would agree on choosing the performance here.

quantranhong1999:
## Audit massive operation performance

### 12k messages, 50 batch size, no change (master branch)


### 12k messages, 50 batch size, flatMap change only (+ concatMap for rename)


### 12k messages, 50 batch size, flatMap and BatchStatement change (+ concatMap for rename)


### 12k messages, 20000 batch size, flatMap and BatchStatement change (+ concatMap for rename)



As we can see from the results, this PR effort leads to a significant boost in IMAP SELECT, STORE, COPY, and MOVE commands.

A degradation on RENAME command (concatMap) =>  rollbacked to flatMap for RENAME (agreed above).

chibenwa (approved):

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraIndexTableHandler.java
```diff
@@ -172,19 +177,21 @@ private Mono<Void> manageApplicableFlagsOnFlagsUpdate(CassandraId mailboxId, Lis
...
+        List<MessageUid> addDeletedUids = updatedFlags
+            .stream()
+            .filter(flag -> flag.isModifiedToSet(Flags.Flag.DELETED))
+            .map(UpdatedFlags::getUid)
+            .collect(Collectors.toList());
+        List<MessageUid> removeDeletedUids = updatedFlags
+            .stream()
+            .filter(flag -> flag.isModifiedToUnset(Flags.Flag.DELETED))
+            .map(UpdatedFlags::getUid)
+            .collect(Collectors.toList());
```
vttranlina, line 189:
```suggestion
        ImmutableList.Builder<MessageUid> addDeletedUidsBuilder = ImmutableList.builder();
        ImmutableList.Builder<MessageUid> removeDeletedUidsBuilder = ImmutableList.builder();

        updatedFlags
            .forEach(flag -> {
                if (flag.isModifiedToSet(Flags.Flag.DELETED)) {
                    addDeletedUidsBuilder.add(flag.getUid());
                } else if (flag.isModifiedToUnset(Flags.Flag.DELETED)) {
                    removeDeletedUidsBuilder.add(flag.getUid());
                }
            });
```
it will avoid re-stream, re-filter?

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraIndexTableHandler.java
```diff
@@ -262,19 +275,21 @@ private Mono<Void> manageUnseenMessageCountsOnFlagsUpdate(CassandraId mailboxId,
     }
 
     private Mono<Void> manageRecentOnFlagsUpdate(CassandraId mailboxId, List<UpdatedFlags> updatedFlags) {
-        return Flux.fromIterable(updatedFlags)
-            .concatMap(flags -> manageRecentOnFlagsUpdate(mailboxId, flags))
-            .then();
-    }
+        List<MessageUid> addRecentUids = updatedFlags
```
vttranlina, line 278: similar

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraIndexTableHandler.java
```diff
@@ -292,26 +313,46 @@ private Mono<Void> checkDeletedOnAdd(CassandraId mailboxId, Flags flags, Message
...
-        }
-        return Mono.empty();
+    private Mono<Void> updateFirstUnseenOnDelete(CassandraId mailboxId, Collection<ComposedMessageIdWithMetaData> composedMessageIdWithMetaData) {
+        return firstUnseenDAO.removeUnread(mailboxId, composedMessageIdWithMetaData.stream().filter(composeId -> !composeId.getFlags().contains(Flags.Flag.SEEN))
+            .map(composeId -> composeId.getComposedMessageId().getUid())
+            .collect(Collectors.toList()));
+    }
+
+    private Mono<Void> updateFirstUnseenOnFlagsUpdate(CassandraId mailboxId, List<UpdatedFlags> updatedFlags) {
+        List<MessageUid> addUnreadUids = updatedFlags
```
vttranlina, line 342: similar

vttranlina (approved):

PR #1002, opened 2022-05-17
JAMES-3431 Stricter validation for DSN ENVID

chibenwa:
Setting an invalid EndId leads to emails not being
transmissible to third party systems. Stricter validation
fixes this.


mailet/api/src/test/java/org/apache/mailet/DsnParametersTest.java
```diff
@@ -128,6 +123,12 @@ void ofShouldThrowOnNullValue() {
                 .isInstanceOf(NullPointerException.class);
         }
 
+        @Test
+        void ofShouldThrowOnBadValue() {
```
Arsnael, line 127: Shouldn't we test for `=` and `+` cases?

Arsnael (approved):

PR #1008, opened 2022-05-19
[JAMES-3767] rename pulsar artifact

jeantil:
Hello,

I renamed the artifact as discussed by email, I also added an bit of readme.

Contrary to what we wrote in the mail, I decided to remain on an SQL store for domains, recipient rewrite tables and users. The reason is that while relying on files would be enough for our specific use case, it would also seriously limit the `scaling` part of the artifact.


Arsnael (approved):
Looks alright to me

server/apps/scaling-pulsar-smtp/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+Cassandra is pretty expensive to operate and only makes sense for really large scale deployments. On the other hand, there are cost-efficient SAAS deployments of pulsar. For small scale deployments destined to grow this app will instead rely on 
+- A blobstore (for mail repository) 
+- A PostgresSQL database for reference data (domain list, recipient rewrite table and users)
+
+The idea being that it will be easy enough to operate this assembly at a "reasonable" cost in the cloud until the scale is large enough to warrant switching to the fully distributed app. 
+
+## Building
+
+In order to build both the ZIP and docker packaging, run:
```
chibenwa, line 15: As far as I am aware of there is no ZIP packaging in place for this

chibenwa (approved):

jeantil:
My guess is flaky test ...
```
Condition with lambda expression in org.apache.james.transport.mailets.MailReprocessingIntegrationTest was not fulfilled within 1 minutes.
```

server/apps/scaling-pulsar-smtp/README.md
jeantil: fixedup

PR #1015, opened 2022-05-25
JAMES-3756 DelegationStore API + Memory + Contract

chibenwa:


server/data/data-api/src/main/java/org/apache/james/user/api/DelegationStore.java
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * @return Lists of the users authorized to impersonnate to baseUser.
+     */
+    Publisher<Username> authorizedUsers(Username baseUser);
+
+    Publisher<Void> clear(Username baseUser);
+
+    Publisher<Void> addAuthorizedUser(Username baseUser, Username userWithAccess);
+
+    Publisher<Void> removeAuthorizedUser(Username baseUser, Username userWithAccess);
```
quantranhong1999, line 35: Maybe we could have a little java doc for other APIs?

quantranhong1999: BTW should we have an API to check if a user is authorized to impersonate to baseUser?

quantranhong1999 (approved):

server/data/data-api/src/main/java/org/apache/james/user/api/DelegationStore.java
chibenwa: > BTW should we have an API to check if a user is authorized to impersonate to baseUser?

That's what the `Authorizator` API is doing ;-)

chibenwa: > Maybe we could have a little java doc for other APIs?

Boiler plate. I don't like java doc. I added that one so that we do not mix the delegated user and the base user. It seemed uneeded for other methods. Even this javadoc addition is questionable IMO.

Arsnael (approved):

PR #1018, opened 2022-05-25
Upgrade ES to 8.2

Arsnael:
Still a draft... I have to focus on something else but I post the beginning anyways... Does not compile, lots of things to change actually with the new java api rest client


Arsnael:
Change of strategy, I duplicated the backend es module and do the v8 modification on it first, so the rest can still compile. When it's stable I will remove v7 and wire it correctly...

Multiple issues though, it seems the tests checking the connection with ES are not even working, timeout. Need to check if something changed in docker image setup, conf, etc...

backends-common/elasticsearch-v8/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+
+        <!-- Prevents https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-28491 -->
+        <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-cbor</artifactId>
+            <version>2.13.1</version>
+        </dependency>
```
chibenwa, line 65: Still needed?

backends-common/elasticsearch-v8/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+            <artifactId>javax.annotation-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-configuration2</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-to-slf4j</artifactId>
+            <version>2.17.1</version>
```
chibenwa, line 90: Don't we have a generic version for this?

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/search/ScrolledSearch.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private Mono<ScrollResponse<ObjectNode>> buildRequest(Optional<String> scrollId) {
+        return scrollId.map(id ->
+            client.scroll(new ScrollRequest.Builder()
+                    .scrollId(scrollId.get())
+                    .scroll(TIMEOUT)
+                    .build()))
+            //.orElseGet(() -> client.search(searchRequest));
+            .get(); // TODO: scroll and search return different responses now, thus this trick not working...
+                    // not sure of the best way to solve that yet
```
chibenwa, line 98: Likely the level of indirection needs to be escalated...

backends-common/elasticsearch-v8/pom.xml
Arsnael: Just a copy/paste of the v7 module. Means it was not cleaned up

backends-common/elasticsearch-v8/pom.xml
Arsnael: Not sure... likely not

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/search/ScrolledSearch.java
Arsnael: Tried quickly without success... will dig into it after I manage to make the rest work

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/ElasticSearchIndexer.java
```diff
@@ -84,7 +85,7 @@ public Mono<BulkResponse> update(List<UpdatedRepresentation> updatedDocumentPart
                 .index(aliasName.getValue())
                 .id(updatedDocumentPart.getId().asString())
                 .action(document -> document
-                    .withJson(new StringReader(updatedDocumentPart.getUpdatedDocumentPart())))
+                    .doc(new RawValue(updatedDocumentPart.getUpdatedDocumentPart())))
```
Arsnael, line 88: I've been looking all afternoon for that... Rational being `withJson` is not working on bulkRequests atm => https://discuss.elastic.co/t/java-8-1-bulk-request/302423

...

Arsnael:
I reworked the git history, so it's easier to review... I separated the copy of the v7 backends es module and the actual changes for v8. So just review what's after the copy commit :)

Most of the tests are green now. What's left:
- `ClientProviderImplConnectionAuthESIgnoreSSLCheckTest` and `ClientProviderImplConnectionAuthESOverrideTrustStoreTest`. Something is wrong with the nginx and es communication... Take into account that an other task (not here) would be to get rid of nginx, as ES seems to integrate xpack in v8 and integrates realms, certificates, ssl, ... security in general.
- `ScrolledSearch` still need to fix that one, and reenable its tests.

Then... we will be good for wiring it to the rest :)

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/search/ScrolledSearch.java
```diff
@@ -93,9 +93,14 @@ private Mono<ScrollResponse<ObjectNode>> buildRequest(Optional<String> scrollId)
                     .scrollId(scrollId.get())
                     .scroll(TIMEOUT)
                     .build()))
-            //.orElseGet(() -> client.search(searchRequest));
-            .get(); // TODO: scroll and search return different responses now, thus this trick not working...
-                    // not sure of the best way to solve that yet
+            .orElseGet(() -> client.search(searchRequest)
```
Arsnael, line 96: So actually here I kept the scroll for now and just adapted the code to quickly migrate...

However I believe though we can optimize this greatly in an other task/PR. 

https://www.elastic.co/guide/en/elasticsearch/reference/8.2/paginate-search-results.html#scroll-search-results => stating scrolling is not recommended anymore for deep pagination.

Instead they advice to use the search_after param in the search method: https://www.elastic.co/guide/en/elasticsearch/reference/8.2/paginate-search-results.html#scroll-search-results

That would require a little refactoring.

chibenwa: Maybe we could create a ticket "evaluate PIT + scroll_after" ?

Arsnael: > Regarding search_after, requesting a point in time looks expensive and likely is a blocker. Before adoption I'd like to have a performance evaluation...

I'm not sure it's more expensive than requesting a scrollId... Felt a bit the same mechanism to me, you have it for the time of your scroll and then you clear it. 

Agreed that it should be subject to a performance tests of course before adoption :)

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/ElasticSearchIndexer.java
```diff
@@ -0,0 +1,140 @@
...
+                return Mono.empty();
+            });
+    }
+
+    public Mono<BulkResponse> delete(List<DocumentId> ids, RoutingKey routingKey) {
+        BulkRequest.Builder bulkBuilder = new BulkRequest.Builder();
+
+        if (ids.isEmpty()) {
+            return Mono.empty();
+        }
```
vttranlina, line 104:
```suggestion
        if (ids.isEmpty()) {
            return Mono.empty();
        }
        
        BulkRequest.Builder bulkBuilder = new BulkRequest.Builder();
```

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/IndexCreationFactory.java
```diff
@@ -0,0 +1,376 @@
...
+                FinalStage indexName(IndexName indexName);
+            }
+
+            public static class FinalStage {
+                private final int nbShards;
+                private final int nbReplica;
+                private final int waitForActiveShards;
+                private final IndexName indexName;
+                private final ImmutableList.Builder<AliasName> aliases;
+                private ImmutableList.Builder<IndexCreationCustomElement> customAnalyzers;
```
vttranlina, line 119: We can use `final`?

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/IndexCreationFactory.java
```diff
@@ -0,0 +1,376 @@
...
+                    .sneakyThrow());
+            } catch (IOException e) {
+                LOGGER.error("Error while creating index : ", e);
+            }
+            return client;
+        }
+
+        private void createAliasIfNeeded(ReactorElasticSearchClient client, IndexName indexName, AliasName aliasName) throws IOException {
+            if (!aliasExist(client, aliasName)) {
+                client.updateAliases(
```
vttranlina, line 211: Missing `.block()` ?

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/IndexCreationFactory.java
```diff
@@ -0,0 +1,376 @@
...
+                            .add(new AddAction.Builder()
+                                .index(indexName.getValue())
+                                .alias(aliasName.getValue())
+                                .build())
+                            .build())
+                        .build());
+            }
+        }
+
+        private boolean aliasExist(ReactorElasticSearchClient client, AliasName aliasName) {
```
vttranlina, line 223: Should we merge `aliasExist` and `updateAliases` in a single pipeline?

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/IndexCreationFactory.java
```diff
@@ -0,0 +1,376 @@
...
+            } catch (ElasticsearchException exception) {
+                if (exception.getMessage().contains(INDEX_ALREADY_EXISTS_EXCEPTION_MESSAGE)) {
+                    LOGGER.info("Index [{}] already exists", indexName.getValue());
+                } else {
+                    throw exception;
+                }
+            }
+        }
+
+        private boolean indexExists(ReactorElasticSearchClient client, IndexName indexName) {
```
vttranlina, line 253: Should we merge `indexExists` and `createIndex` in a single pipeline?

backends-common/elasticsearch-v8/src/test/java/org/apache/james/backends/es/v8/ElasticSearchIndexerTest.java
```diff
@@ -0,0 +1,259 @@
...
+        awaitForElasticSearch(new MatchQuery.Builder()
+            .field("message")
+            .query("trying")
+            .build()
+            ._toQuery(), 1L);
+    }
+    
+    @Test
+    void indexMessageShouldThrowWhenJsonIsNull() {
+        assertThatThrownBy(() -> testee.index(DOCUMENT_ID, null, ROUTING))
```
vttranlina, line 97:
```suggestion
        assertThatThrownBy(() -> testee.index(DOCUMENT_ID, null, ROUTING).block())
```

Arsnael:
Mailbox es is green, quota search es is green... it seems being only about finishing wiring up things correctly, cleaning up, rebase, ...

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/ElasticSearchIndexer.java
```diff
@@ -113,7 +113,7 @@ public Mono<BulkResponse> delete(List<DocumentId> ids, RoutingKey routingKey) {
         return client.bulk(bulkBuilder.build())
             .onErrorResume(exception -> {
                 LOGGER.warn("Error while deleting index", exception);
-                return Mono.empty();
+                return Mono.error(exception);
             });
```
chibenwa, line 117: Log then rethrow is an anti-pattern...

Here we likely should not do anything I guess.

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/ElasticSearchIndexer.java
```diff
@@ -92,7 +92,7 @@ public Mono<BulkResponse> update(List<UpdatedRepresentation> updatedDocumentPart
         return client.bulk(bulkBuilder.build())
             .onErrorResume(exception -> {
                 LOGGER.warn("Error while updating index", exception);
-                return Mono.empty();
+                return Mono.error(exception);
```
chibenwa, line 95: Log then rethrow is an anti-pattern...

Here we likely should not do anything I guess.

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/ElasticSearchIndexer.java
Arsnael: There is tests checking that an exception is propagated correctly to the top... if I leave Mono.empty() well the test just hangs forever... which sounds logical as we don't propagate the exception... I don't get how that was passing before though?

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/ElasticSearchIndexer.java
Arsnael: Same remark

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/ElasticSearchIndexer.java
Arsnael: alright get rid of the onErrorResume and propagate the exception  to the caller, got it

chibenwa: I mean:
 - Remove the log as we are o longer managing the exception here
 - Then the onErrorResume blocks becomes more or less useless and can be removed.

chibenwa:
@vttranlina we should refrain from too many changes t once, this work is already large enough.

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/IndexCreationFactory.java
chibenwa: Let's avid too much refactorings IMO

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/IndexCreationFactory.java
chibenwa: Let's avid too much refactorings IMO

backends-common/elasticsearch-v8/pom.xml
Arsnael: We don't. Only used here in this module... Want to do sth about it?

backends-common/elasticsearch-v8/src/main/java/org/apache/james/backends/es/v8/IndexCreationFactory.java
Arsnael: Could be a refactoring in an other PR, too much changes already as stated by @chibenwa

server/apps/memory-app/sample-configuration/elasticsearch.properties
```diff
@@ -20,12 +20,4 @@
...
-# Configuration file for ElasticSearch
-
-# Reports for metrics into ElasticSearch
-# Defaults to elasticsearch.masterHost : on which server to publish metrics
-elasticsearch.http.host=elasticsearch
-elasticsearch.http.port=9200
-elasticsearch.metrics.reports.enabled=true
-elasticsearch.metrics.reports.period=30
-elasticsearch.metrics.reports.index=james-metrics
+# Configuration file for ElasticSearch
```
chibenwa, line 23: This file itself no longer make sense, right?

server/apps/distributed-app/docs/modules/ROOT/pages/operate/guide.adoc
```diff
@@ -137,7 +137,7 @@ Currently, an administrator can monitor indexation failures through
 `ERROR` log review. You can as well
 xref:operate/webadmin.adoc#_listing_failed_events[list failed events] by
 looking with the group called
-`org.apache.james.mailbox.elasticsearch.v7.events.ElasticSearchListeningMessageSearchIndex$ElasticSearchListeningMessageSearchIndexGroup`.
+`org.apache.james.mailbox.elasticsearch.v8.events.ElasticSearchListeningMessageSearchIndex$ElasticSearchListeningMessageSearchIndexGroup`.
```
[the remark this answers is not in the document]
chibenwa: We may need to add manual unbindings in RabbitMQ to avoid issues in upgrade instructions.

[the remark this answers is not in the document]
Arsnael: > Will this cause issue?

like every time we change a group listener in dead letter... that design based on the FQDN isn't good IMO, it just breaks if you replay dead letters and you still had events in there with the old one after an upgrade.

> We may need to add manual unbindings in RabbitMQ to avoid issues in upgrade instructions.

You mean like replay your dead letters, make sure it's empty of events of that group before upgrading James?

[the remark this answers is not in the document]
chibenwa: I mean manually deletes the bindings from the event queues in RabbitMQ

upgrade-instructions.md
```diff
@@ -17,6 +17,21 @@ Changes to apply between 3.7.x and 3.8.x will be reported here.
...
+### Migration to ElasticSearch 8
+
+Date: 09/06/2022
+
+Concerned products: Distributed James, Cassandra James
+
+ElasticSearch implementation has been upgraded to support the version 8 of it.
+
+You should migrate your ElasticSearch servers from version 7 to 8 before upgrading James. We recommend to use
+at least the version 8.2 of ElasticSearch.
```
chibenwa, line 31: Add: Users can either perform a migration of their data from version 7 to 8 or might start from an empty version 8 ElasticSearch cluster and [link] reindex data.

mailbox/elasticsearch-v8/src/main/java/org/apache/james/mailbox/elasticsearch/v8/MailboxMappingFactory.java
```diff
@@ -19,272 +19,241 @@
...
+            .withJson(new StringReader(generateMappingContent()))
+            .build();
+    }
+    
+    public static String generateMappingContent() {
+        return "{" +
+            "  \"dynamic\": \"strict\"," +
+            "  \"" + ROUTING + "\": {" +
+            "    \"" + REQUIRED + "\": true" +
+            "  }," +
```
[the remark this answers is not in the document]
Arsnael: Probably with the builder, but it's painful enough honestly, that mapping is huge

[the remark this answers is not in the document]
chibenwa: Get something working first, then other contributors could iterate to improve overall code quality if needed ;-)

We already did invest a lot of time on this, let's not focus on details.

Arsnael:
Squashed fixups, removed the commits when we were removing metrics es reporter module as well

mbaechler:
I'm late to the party, but are we removing the support for the last OSI-compliant version of elasticsearch?

chibenwa:
> I'm late to the party, but are we removing the support for the last OSI-compliant version of elasticsearch?

What is the alternative?

Unmaintained ElasticSearch version?

Amazon alternatives?

An expensive migration to SOLR?

Bitter sweat choices...

mbaechler:
> > I'm late to the party, but are we removing the support for the last OSI-compliant version of elasticsearch?
> 
> What is the alternative?
> 
> Unmaintained ElasticSearch version?
> 
> Amazon alternatives?
> 
> An expensive migration to SOLR?
> 
> Bitter sweat choices...

As far as I understand, OpenSearch is a fork of ES 7.10. Thus, the existing code for ES 7 should work out of the box with opensearch and benefits from support from the OpenSearch "community".

I don't have any problem with adding an implementation for ES 8 but we don't have to remove the ES 7 code, right?

chibenwa:
(for the record internally we rely on OpenDistro and maintain a separate test suite of mailbox-es7 on top of OpenDistro)

The feeling I get is that thanks to decision making at Eladstic one need to choose his side and we won't get both.

PR #1020, opened 2022-05-25
JAMES-3772 Fixing a blocking thread issue with reactive RabbitMQ when releasing a channel in the pool

Arsnael:


chibenwa (changes requested):
Won't change anything to be honnest. 

Even if you execute the reactor logic on an elastic thread you are still blocking on an elastic thread.

chibenwa:
CF https://github.com/apache/james-project/pull/1021 for a solution.

Alternatives: revert to a blocking channel pool, or switch all rabbitMQ consumer to an elastic thread.

Arsnael:
> Won't change anything to be honnest.
> 
> Even if you execute the reactor logic on an elastic thread you are still blocking on an elastic thread.

Yes I thought about it on my way home yesterday, it does not change anything indeed

PR #1023, opened 2022-05-26
JAMES-3773 get rid of Reactor Elastic scheduler

chibenwa:
-> Convert blocking non-reactor calls to boundedElastic
 -> Convert calls that wraps reactor.block call in a dedicated scheduler to avoid starvation
 -> Remove unneeded subscribeOn calls
 -> Reactify things slightly better where possible.

## TODO

 - [x] Performance tests, IMAP, JMAP
 - [x] jstack : we expect more control on the count of threads
 - [x] Run distributed integration tests
 - [x] Open a JIRA & reword this.
 
 ## Conclusion
 
 Equivalent performance for IMAP, JMAP draft, JMAP RFC-8621 and SMTP.
 
 Uncarefull .block calls leads to dead locks, we need to be carefull on this.

### DIFF
--- a/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraUidProvider.java
+++ b/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraUidProvider.java
@@ -77,7 +77,7 @@ public class CassandraUidProvider implements UidProvider {
         this.insertStatement = prepareInsert(session);
         Duration firstBackoff = Duration.ofMillis(10);
         this.retrySpec = Retry.backoff(cassandraConfiguration.getUidMaxRetry(), firstBackoff)
-            .scheduler(Schedulers.elastic());
+            .scheduler(Schedulers.parallel());
     }
 
     private PreparedStatement prepareSelect(Session session) {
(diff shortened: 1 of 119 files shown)


chibenwa:
I confirm the staged approach solves the dead locks cf last fixup.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraUidProvider.java
```diff
@@ -77,7 +77,7 @@ public CassandraUidProvider(Session session, CassandraConfiguration cassandraCon
         this.insertStatement = prepareInsert(session);
         Duration firstBackoff = Duration.ofMillis(10);
         this.retrySpec = Retry.backoff(cassandraConfiguration.getUidMaxRetry(), firstBackoff)
-            .scheduler(Schedulers.elastic());
+            .scheduler(Schedulers.parallel());
```
quantranhong1999, line 80: Q: Why are there somewhere you switch `.elastic` to `.parallel` and somewhere else to `.boundedElastic`?
BTW I thought that we should not do a potential blocking job (here use elastic before) on a parallel thread?

chibenwa: A: When it's just about delays elastic scheduler is overkill. We are not dealing with a blocking task hence parallel.

Arsnael:
```
java.lang.NullPointerException
	at org.apache.james.events.RabbitMQEventBusTest.tearDown(RabbitMQEventBusTest.java:139)
```
For all tests in RabbitMQEventBusTest

Arsnael:
Rebase needed :)

PR #1030, opened 2022-05-30
[PERF] Improve SMTP performance

chibenwa:
This change set proposes a 38% performance improvement for the
following workload (reduced CPU usage):

```
smtp-source -A -C100 -r 1 -l 500000 -m 5000 -s 20 -d -c -f any@other.com -M other.com -t firstname14.surname14@upn.integration-open-paas.org localhost:25
```

Rationals:

 - Submitted emails are read line by line and a couple of session
 scoped values where retrieved from session attachment for each line:
 mail size, header reached, etc. This causes not cheap HashMap reads
 for each email line. This can easily be replaced by
 getters/setters on the SMTP session to save the map read/writes.
 - The data was accessed with the following transformation chain:
 ByteBuf -> NIO byte buffer -> byte[]. But expensive reference
 counting is triggered by Netty under the hood. Doing directly
 ByteBuf -> byte[] prevents this.
 - SMTP commands metric was incremented on each body line (!) which
 also incurs a performance hit.
 - Immutable handlers where retrieved on each request while we could
 save them to fields to avoid re-computation.
 - Avoid re-parsing charset.
 - Use Deque::peekFirst instead of Iterables::getFirst when looking
 up for line overrides (same than IMAP).
 - Compute debug logs only if needed.
 - Fasten slightly parsing SMTP parameters
 
 This can be seen in the follwoing flame graph:
 
[smtp-cpu.zip](https://github.com/apache/james-project/files/8795220/smtp-cpu.zip)

Screenshot from 2022-05-30 10-31-49

## After

Screenshot from 2022-05-30 13-15-04

[ssmcpu.zip](https://github.com/apache/james-project/files/8795917/ssmcpu.zip)


protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/DataLineMessageHookHandler.java
```diff
@@ -58,16 +57,15 @@ public class DataLineMessageHookHandler implements DataLineFilter, ExtensibleHan
...
             .orElseThrow(() -> new RuntimeException("'" + DataCmdHandler.MAILENV.asString() + "' has not been filled."));
 
         OutputStream out = getMessageOutputStream(env);
         try {
             // 46 is "."
-            // Stream terminated            
-            int c = line.get();
-            if (line.remaining() == 2 && c == 46) {
+            // Stream terminated
+            if (line.length == 3 && line[0] == 46) {
```
quantranhong1999, line 68:
```suggestion
            if (line.length == 2 && line[0] == 46) {
```

quantranhong1999 (approved):

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1030/5/testReport/?cloudbees-analytics-link=scm-reporting%2Ftests%2Ffailed

I think those failed tests are related to this PR, please check

Arsnael (approved):

quantranhong1999 (approved):

PR #1046, opened 2022-06-10
JAMES-3763 BlobMailRepository: No isolation across mail repositories

chibenwa:
This is an issue as, for instance, when getting rid of the SPAMs that I
received I might deletes email in ERROR that actually would like to
redeliver causing thus data loss.

Cc @mbaechler @jeantil

### DIFF
--- a/server/mailrepository/mailrepository-blob/src/main/scala/org/apache/james/mailrepository/blob/BlobMailRepository.scala
+++ b/server/mailrepository/mailrepository-blob/src/main/scala/org/apache/james/mailrepository/blob/BlobMailRepository.scala
@@ -23,28 +23,25 @@ import java.util
 import java.util.Date
 import java.util.stream.Stream
 
+import com.google.common.collect.ImmutableMap
+import javax.inject.Inject
 import javax.mail.MessagingException
 import javax.mail.internet.MimeMessage
-
-import scala.jdk.CollectionConverters.IterableHasAsJava
-
 import org.apache.commons.lang3.tuple.Pair
 import org.apache.james.blob.api.BlobStore.StoragePolicy.SIZE_BASED
 import org.apache.james.blob.api.Store.Impl
 import org.apache.james.blob.api.{BlobId, BlobPartsId, BlobStore, BlobStoreDAO, BlobType, BucketName, Store}
 import org.apache.james.blob.mail.MimeMessagePartsId
-import org.apache.james.mailrepository.api.{MailKey, MailRepository}
-import org.apache.mailet.{Attribute, AttributeName, AttributeValue, Mail, PerRecipientHeaders}
-
-import com.google.common.collect.ImmutableMap
-
-import play.api.libs.json.{Format, Json}
-import reactor.core.publisher.{Flux, Mono}
-import scala.jdk.StreamConverters._
-
 import org.apache.james.core.{MailAddress, MaybeSender}
+import org.apache.james.mailrepository.api.{MailKey, MailRepository, MailRepositoryUrl}
 import org.apache.james.server.blob.deduplication.BlobStoreFactory
 import org.apache.james.server.core.MailImpl
+import org.apache.mailet.{Attribute, AttributeName, AttributeValue, Mail, PerRecipientHeaders}
+import play.api.libs.json.{Format, Json}
+import reactor.core.publisher.{Flux, Mono}
+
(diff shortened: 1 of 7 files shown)


server/mailrepository/mailrepository-blob/src/test/java/org/apache/james/mailrepository/blob/BlobMailRepositoryTest.java
```diff
@@ -56,4 +58,15 @@ void setup() {
     public MailRepository retrieveRepository() {
         return blobMailRepository;
     }
+
+    @Override
+    public MailRepository retrieveRepository(MailRepositoryPath path) {
+        return blobMailRepository;
```
jeantil, line 64: I don't actually agree with the test setup. We intend to instantiate a mailrepository per mailrepository url and use the url to define the bucket to use. as such the retrieveRepository(path) would return a different repository backed by a different bucket for a different path

jeantil: Note that this is exactly what you do in the casandra implementatio of the test, you return a different repository for a different path ..

chibenwa: Current code injecting mailRepository passes the URL to implementations via either config or constructor, and shoud have the underlying mail repository adapt its storage to the URL.

I'll adapt the implem.

Regards

server/mailrepository/mailrepository-blob/src/main/scala/org/apache/james/mailrepository/blob/BlobMailRepository.scala
```diff
@@ -138,12 +135,12 @@ object BlobMailRepository {
   }
 }
 
-class BlobMailRepository(val blobStore: BlobStoreDAO,
+class BlobMailRepository @Inject() (val blobStore: BlobStoreDAO,
```
jeantil, line 138: Why add @Inject here ?

server/mailrepository/mailrepository-blob/src/main/scala/org/apache/james/mailrepository/blob/BlobMailRepository.scala
```diff
@@ -138,12 +135,12 @@ object BlobMailRepository {
...
                          val blobIdFactory: BlobId.Factory,
-                         val mimeMessageStore: Store[MimeMessage, MimeMessagePartsId]
-                        ) extends MailRepository {
-  private val bucketName: BucketName = BucketName.of("mailMetadata")
-  private val mimeMessageBucketName: BucketName = BucketName.of("mimeMessageData")
+                         val mimeMessageStore: Store[MimeMessage, MimeMessagePartsId],
+                         val url: MailRepositoryUrl) extends MailRepository {
```
jeantil, line 141: as I said in my email, this is something we were going to be working on in our next session but if you need it faster than that, then sure :D

server/mailrepository/mailrepository-blob/src/main/scala/org/apache/james/mailrepository/blob/BlobMailRepository.scala
chibenwa: Cause MemoryMailRepositoryStore would need it.

server/mailrepository/mailrepository-blob/src/main/scala/org/apache/james/mailrepository/blob/BlobMailRepository.scala
chibenwa: Well it is more that without this this implem is trivially broken...

That's also a good opportunity to harden the mail repository contract.

jeantil: Improving the contract does sound good.

We will resolve the conflicts and finalize the fix when we get there as this won't be enough since we probably also want to get a different mimeMessageStore...

chibenwa:
@jeantil do you think we can merge this?

chibenwa:
See https://github.com/apache/james-project/pull/1053

PR #1047, opened 2022-06-10
JAMES-3778 Demonstrate a Graphite integration

chibenwa:
Screenshot from 2022-06-10 18-01-45

More complex that I though.... I needed an extension mechanism to start the collector.

# TODO

  - [x] README for the example
  - [x] Mention the example in the doc
  - [x] Document startable registration
  
 CC @jeantil


server/container/guice/utils/src/main/java/org/apache/james/utils/ExtensionConfiguration.java
```diff
@@ -27,22 +27,31 @@
 import com.google.common.collect.ImmutableList;
 
 public class ExtensionConfiguration {
-    public static final ExtensionConfiguration DEFAULT = new ExtensionConfiguration(ImmutableList.of());
+    public static final ExtensionConfiguration DEFAULT = new ExtensionConfiguration(ImmutableList.of(), ImmutableList.of());
 
     public static ExtensionConfiguration from(Configuration configuration) {
         return new ExtensionConfiguration(
             Arrays.stream(configuration.getStringArray("guice.extension.module"))
```
jeantil, line 34: I had no idea james could alread do that ...

server/container/guice/utils/src/main/java/org/apache/james/utils/UserStartables.java
```diff
@@ -0,0 +1,43 @@
...
+import com.github.fge.lambdas.Throwing;
+import com.google.inject.AbstractModule;
+import com.google.inject.Singleton;
+import com.google.inject.multibindings.ProvidesIntoSet;
+
+public class UserStartables implements Startable {
+    public static class Module extends AbstractModule {
+        @ProvidesIntoSet
+        @Singleton
+        InitializationOperation graphiteExporter(UserStartables startables) {
```
jeantil, line 18:
```suggestion
        InitializationOperation initializationOperations(UserStartables startables) {
```

maybe ?

server/container/guice/utils/src/main/java/org/apache/james/utils/ExtensionConfiguration.java
```diff
@@ -27,22 +27,31 @@
...
+            Arrays.stream(configuration.getStringArray("guice.extension.startable"))
                 .map(ClassName::new)
                 .collect(ImmutableList.toImmutableList()));
     }
 
     private final List<ClassName> additionalGuiceModulesForExtensions;
+    private final List<ClassName> startables;
 
-    public ExtensionConfiguration(List<ClassName> additionalGuiceModulesForExtensions) {
+    public ExtensionConfiguration(List<ClassName> additionalGuiceModulesForExtensions, List<ClassName> startables) {
```
jeantil, line 45: :thinking: it's hard to tell since I never used this mechanism before but looking @ org.apache.james.modules.server.ElasticSearchMetricReporterModule it looks like I can already declare 
```
guice.extension.module=org.apache.james.modules.server.ElasticSearchMetricReporterModule
```
or 
```
guice.extension.module=org.apache.james.modules.server.GraphiteMetricReporterModule
```
and that would be installed and the `start` would be called during server startup when InitializationOperation are collected. 

Therefore all the UserStartable extension is unneeded or am I missing something ?

jeantil: I was confused because in the existing apps the ElasticSearchMetricReporterModule is always explicitely installed in the main, is there a difference I am missing ?

jeantil:
#mindBlown the extension mechanism should definitely be used more often  none of the existing assemblies seem to use it ! instead everything is hardcoded in the main

chibenwa:
> #mindBlown the extension mechanism should definitely be used more often none of the existing assemblies seem to use it ! 

I do use it with my customers ;-)

https://www.youtube.com/watch?v=zr8qpNkL6U4&ab_channel=TheApacheFoundation if you want to listen to my broken english.

This allows injecting your stuff across the wall extension set: routes, listeners mailet...

Also the way it composes means that you can't overrides james bindings or integrate into unplanned places within James, bringing in an additional safety net ;-)

> instead everything is hardcoded in the main

Huge +1

 - https://github.com/apache/james-project/tree/master/server/mailet/rate-limiter and https://github.com/apache/james-project/tree/master/server/mailet/rate-limiter-redis also work with this.
 - I proposed RSpamD extension set to also use these mechanisms https://issues.apache.org/jira/browse/JAMES-3775
 - I would be happy to move some of the hard wired stuff as extensions.
 
 Current limitation includes not being able to submit tasks on the distributed task manager (because

server/container/guice/utils/src/main/java/org/apache/james/utils/ExtensionConfiguration.java
chibenwa: > won't it be picked up when loading all the InitializationOperation ?

No. 

Because InitializationOperation is used by core James stuff thus not open for use.

That also coupled with the way child injectors work in guice.

jeantil: alright thanks for the clarification

chibenwa:
## TODO

 - [ ] JIRA
 - [ ] Unit tests for User defined Startables
 - [ ] Documentation for User defined Startables

chibenwa:
## TODO

 - [x] JIRA
 - [x] Unit tests for User defined Startables
 - [x] Documentation for User defined Startables
 
 ## In another PR
 
 - [ ] Extract the ElasticSearch metric reporter as an extension
 - [ ] Document the ElasticSearch metric reporter new set up.

PR #1050, opened 2022-06-13
JAMES-3780 ExpireMailboxTask should support large number of users

ottoka:
Current implementation produces
`
java.lang.IllegalStateException: Iterating over a toIterable() / toStream() is blocking, which is not supported in thread parallel-2
`
because the user list can be a BlockingIterable, e.g. in Cassandra.

The solution is to run user listing in the elastic scheduler which supports blocking.

### DIFF
--- a/server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/service/ExpireMailboxService.java
+++ b/server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/service/ExpireMailboxService.java
@@ -187,7 +187,8 @@ public class ExpireMailboxService {
                         SearchQuery.headerDateBefore("Expires", now, DateResolution.Second)
                     )
             );
-            return Iterators.toFlux(usersRepository.list())
+            // Note: user list may be a blocking iterable, must run on a scheduler that supports this.
+            return Iterators.toFlux(usersRepository.list()).subscribeOn(ReactorUtils.BLOCKING_CALL_WRAPPER)
                 .transform(ReactorUtils.<Username, Task.Result>throttle()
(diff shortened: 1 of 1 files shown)


server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/service/ExpireMailboxService.java
```diff
@@ -187,7 +188,8 @@ public Mono<Result> expireMailboxes(Context context, RunningOptions runningOptio
                         SearchQuery.headerDateBefore("Expires", now, DateResolution.Second)
                     )
             );
-            return Iterators.toFlux(usersRepository.list())
+            // Note: user list may be a blocking iterable, must run on a scheduler that supports this.
+            return Iterators.toFlux(usersRepository.list()).subscribeOn(Schedulers.elastic())
```
chibenwa, line 192:
```suggestion
            return Iterators.toFlux(usersRepository.list()).subscribeOn(ReactorUtils.BLOCKING_CALL_WRAPPER)
```

 -> elastic is deprecated in favor of bounded elastic
 -> nested calls to bounded elastic causes dead locks hence when we know we would end up calling .block or equivalent we need to use a distinct scheduler.
 
Having a reactive version of userRepository.list would be handy!

ottoka: Thanks for the feedback, I am still tripping over the fine details of reactor, sigh. Fixed in latest push.

chibenwa (approved):

quantranhong1999 (approved):

Arsnael (approved):

PR #1052, opened 2022-06-16
[REACTOR] Bring back `.subscribeOn(ReactorUtils.BLOCKING_CALL_WRAPPER)` for EmailSubmissionSetMethod

quantranhong1999:
Underlying call to CanSendFrom could be blocked which is not allowed in non-blocking thread
`"block()/blockFirst()/blockLast() are blocking, which is not supported in thread reactor-http-epoll-8"`


chibenwa:
I really wonder if this is legit as we encounter this for a custom component.

Maybe we should rather fix our current component, I proposed n idea of fix going this way...

chibenwa (changes requested):
See above

PR #1053, opened 2022-06-16
 JAMES-3763 BlobMailRepository: No isolation across mail repositories

jeantil:
I think this should replace #1046 

The implementation in #1046 doesn't account for the mime message store bucketname which needs to be instanciated based on the url too so that both computations are synced. 
Also the way the bucketname is built in #1046 would result in bucket names such as
`mailMetadata-var/mail/error`. In this MR we propose to translate to `var/mail/error/mailMetadata` instead since that maps better with the way most s3 blobstore handle hierarchical namespacing. 

This PR fixes the existing implementation to pass the contract tests while keeping the MailRepository unaware of its MailRepositoryUrl. The injection part will be handled separately in a future PR


server/mailrepository/mailrepository-blob/src/main/scala/org/apache/james/mailrepository/blob/BlobMailRepository.scala
```diff
@@ -6,9 +6,9 @@
  * to you under the Apache License, Version 2.0 (the            *
  * "License"); you may not use this file except in compliance   *
  * with the License.  You may obtain a copy of the License at   *
- *                                                              *
- *   http://www.apache.org/licenses/LICENSE-2.0                 *
- *                                                              *
+ * *
+ * http://www.apache.org/licenses/LICENSE-2.0                 *
+ * *
```
chibenwa, line 11: Oups

chibenwa (approved):

jeantil:
: thinking: I did run man validate before pushing
I'll fix that around lunch

Le ven. 17 juin 2022 à 04:17, Benoit TELLIER ***@***.***> a
écrit :

Arsnael (approved):

server/mailrepository/mailrepository-blob/src/main/scala/org/apache/james/mailrepository/blob/BlobMailRepository.scala
jeantil: I fixed this (I force pushed the amended commit instead of pushing a fixup)

chibenwa:
It was green before the rebase right?

Then yes ;-)

jeantil:
Funny  : new build, new seemingly unrelated cassadra errors : 
```
com.datastax.oss.driver.api.core.servererrors.InvalidQueryException: unconfigured table annotation
	at org.apache.james.mailbox.cassandra.CassandraSubscriptionManagerTest.setUp(CassandraSubscriptionManagerTest.java:95)
```
maybe related to https://github.com/apache/james-project/pull/1066 ? 
1 green build out of 11 ... :cry:

chibenwa:
> 1 green build out of 11 ... :cry:  

Yes, sorry.

I'm dedicated to improve that...

Also the Cassandra driver update (3.11 -> 4) did lower the request timeout from 2s to 5s which on the ASF infra is apparently noticeable.

BTW I think the risk associated with that merge request is low so please go ahead ;-)

jeantil:
>I'm dedicated to improve that...

I had not noticed it was this bad till recently, i'll think about it

>BTW I think the risk associated with that merge request is low so please go ahead ;-)
thanks, I'll rebase and merge then

PR #1056, opened 2022-06-17
[REACTOR] Reactify CanSendFrom::userCanSendFrom

quantranhong1999:
This change should improve the reactive chain where using CanSendFrom::userCanSendFrom (reactive).
One place left using this is `SenderAuthIdentifyVerificationRcptHook` (SMTP) which we could not migrate to reactive soon for now, therefore we need to keep CanSendFrom::userCanSendFrom (non-reactive).

### DIFF
--- a/server/data/data-api/src/main/java/org/apache/james/rrt/api/CanSendFrom.java
+++ b/server/data/data-api/src/main/java/org/apache/james/rrt/api/CanSendFrom.java
@@ -24,8 +24,6 @@ import org.apache.james.core.MailAddress;
 import org.apache.james.core.Username;
 import org.reactivestreams.Publisher;
 
-import reactor.core.publisher.Mono;
-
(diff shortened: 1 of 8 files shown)


server/data/data-api/src/main/java/org/apache/james/rrt/api/CanSendFrom.java
```diff
@@ -30,6 +33,10 @@ public interface CanSendFrom {
      */
     boolean userCanSendFrom(Username connectedUser, Username fromUser);
 
+    default Publisher<Boolean> userCanSendFromReactive(Username connectedUser, Username fromUser) {
```
chibenwa, line 36: CanSendFromImpl don't need to wrap it's blocking calls? Shouldn't it implement this default method?

Arsnael (approved):

PR #1067, opened 2022-06-27
[IMPROVEMENT] Prevent RMI from doing System.gc every hour

chibenwa:
Screenshot from 2022-06-27 15-40-46

Suspicious! GC every hour with JMX activated... I'll shoot down the usual suspect.

### DIFF
--- a/server/apps/cassandra-app/sample-configuration/jvm.properties
+++ b/server/apps/cassandra-app/sample-configuration/jvm.properties
@@ -42,3 +42,7 @@ config.file=conf/cassandra-driver.conf
 
 # Fails explicitly on missing configuration file rather that taking implicit values. Defautls to false.
 # james.fail.on.missing.configuration=true
+
+# JMX, when enable causes RMI to plan System.gc every hour. Set this instead to once every 1000h.
+sun.rmi.dgc.server.gcInterval=3600000000
+sun.rmi.dgc.client.gcInterval=3600000000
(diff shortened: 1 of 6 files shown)


Arsnael (approved):

chibenwa:
> :exploding_head: 

Not surprised that's a big classic in Devoxx JVM talks. I listen to that when washing dishes....

jeantil:
i do sometimes too but clearly not enough. I would be very curious to understand the rationale that drove to scheduling a gc every hour ...

chibenwa:
Likely some long living connections  with some churns.

Long living connections cause the old gen to fill up.

And the churn turns those old gen objects into garbadges.

I bet the intent is to try to achieve reasonnable pause time by forcing periodic GCs rather than let a huge GC pause build up. THough debatable this 'could' make sense for JMX intensive apps.

Now in james context we only use JMX for the CLI and in the heap dump I worked with I never even come to read 'rmi' or 'jmx' thus I believe that for James this GC pause is highly irrelevant.

Also I came across some comments in Cassandra source code explicitly disabling this : https://github.com/apache/cassandra/blob/39e89fd636ee4343eb2201820da87881cbc749e2/src/java/org/apache/cassandra/utils/JMXServerUtils.java#L110

jeantil:
>THough debatable this 'could' make sense for JMX intensive apps
sounds like premature optimization built-in right into the JDK :p

PR #1073, opened 2022-06-30
Add a vhost configuration option to RabbitMQ

Arsnael:
Some queues (the james work queues) don't get created if we use multiple virtual hosts and they have been created on an other one by an other James already, as we list all queues before creating if not existing.

Adding here a vhost option to just check the queues on that vhost, which should allow the creation of the missing queues on the targeted vhost if they got created somewhere else already.

### DIFF
--- a/upgrade-instructions.md
+++ b/upgrade-instructions.md
@@ -18,26 +18,6 @@ Change list:
 
 - [Adding authorized_users column to user table](#adding-authorized_users-column-to-user-table)
 - [Migration to Cassandra driver 4](#migration-to-cassandra-driver-4)
-- [Adding vhost option to Rabbitmq configuration](#adding-vhost-option-to-rabbitmq-configuration)
-
-### Adding vhost option to Rabbitmq configuration
-
-Date: 30/06/2022
-
-Concerned product: Distributed James
-
-It is possible now to specify a virtual host to be able to create exchanges and queues on a different one
-than the default "/". You just need to add in `rabbitmq.properties` file:
-
-```
-vhost=[VHOST]
-```
-
-Also don't forget to add it to the uri, like that:
-
-```
-amqp://[USERNAME]:[PASSWORD]@[HOST]:[PORT]/[VHOST]
-```
 
 ### Migration to Cassandra driver 4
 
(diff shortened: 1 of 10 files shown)


upgrade-instructions.md
```diff
@@ -18,6 +18,26 @@ Change list:
...
+
+```
+vhost=[VHOST]
+```
+
+Also don't forget to add it to the uri, like that:
+
+```
+amqp://[USERNAME]:[PASSWORD]@[HOST]:[PORT]/[VHOST]
+```
```
chibenwa, line 40: I don't think we should add an upgrade instruction. There is no breaking changes.

IMO remove it.

chibenwa (commented):
Looks good to me!

quantranhong1999 (commented):
Read it.

quantranhong1999 (approved):

chibenwa (approved):

vttranlina (approved):

mogui:
I've tryied to use this configuration but queues get created on the same default rabbit vhost /
There is something I'am missing?

Arsnael:
@mogui which version of James do you use? If it's a released version (like 3.7.0) and not compiled from the master branch, it's normal then as it didn't get released yet (should with the 3.8.0)

Arsnael:
I might have answered a bit fast on this topic. After rechecking it seems obvious in the code some queues are created by using this `vhost` parameter and others by parsing it out of the `uri ` parameter. 
So don't hesitate to add the vhost on the uri as well for now.

It's not convenient and should be unified though. We will take a look at it. Thanks for reporting it @mogui

mogui:
I am using 3.7.1 somehow it works with vhosts in the uri but not for all queue `JamesMailQueue` still gets create on default vhost `/`. It's annoying havingtwo james using the same rabbitmq.
Should I upgrade to 3.8 right?

Arsnael:
3.8 has not been released yet. So you need to compile from master yourself if you want the vhost for the mailqueues on RabbitMQ I'm afraid yes

mogui:
@Arsnael there is an expected timeline for the release?

PR #1061, opened 2022-06-22
Migrate ES-v7 to OpenSearch 2.0

Arsnael:
A much more reworked cleaner version of https://github.com/apache/james-project/pull/1051

Still a few things remain:
- [ ] the ssl tests on the backends to debug and re-enable
- [ ] the sort issue with null fields... there was a workaround in the POC (that I personally don't like but I don't think we can do better)... I was hoping that the PR to fix on the opensearch client would be merged with a bit of change but.... nope (disabled the test for now, might put back the workaround though)
- [ ] debugging what's left

PR #1076, opened 2022-07-04
[DUPLICATE] JAMES-3771 Migrate ES-v7 to OpenSearch 2.0

quantranhong1999:
Just a duplication of https://github.com/apache/james-project/pull/1061.
Seeking a green build (could experiment a bit with the build here).


backends-common/opensearch/src/test/java/org/apache/james/backends/opensearch/DockerElasticSearchExtension.java
```diff
@@ -60,4 +61,9 @@ public void awaitForElasticSearch() {
     public DockerElasticSearch getDockerElasticSearch() {
         return elasticSearch;
     }
+
+    @Override
+    public void afterAll(ExtensionContext context) {
+        elasticSearch.stop();
```
Arsnael, line 68: I'm pretty sure this is worse... did you test it locally? The docker container is usually reused in the module by all test classes... If you stop it after all tests in a class, it's a disaster for the ones after no? But I could be wrong...

backends-common/opensearch/src/test/java/org/apache/james/backends/opensearch/DockerElasticSearchExtension.java
```diff
@@ -37,6 +38,7 @@ public void afterEach(ExtensionContext context) {
 
     @Override
     public void beforeEach(ExtensionContext extensionContext) {
+        elasticSearch.start();
```
Arsnael, line 41: Ok didn't refresh well just saw this... Still not convinced though... it's in a beforeEach as well and you shut it down in afterAll... BeforeAll?

quantranhong1999: I had an error before when putting it into BeforeAll. Not sure why but BTW this still does not work and likely not the problem...

PR #1084, opened 2022-07-11
JAMES-3775 Docker container for testing RSpamD

Arsnael:


third-party/rspamd/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.james</groupId>
+        <artifactId>third-party</artifactId>
+        <version>3.8.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>apache-james-rspamd</artifactId>
+    <name>Apache James :: Third Party :: RSpamD</name>
+    <description>RSpamD Java client and testing utilities</description>
```
chibenwa, line 33: Add a mention about the use of http

third-party/rspamd/src/test/java/org/apache/james/rspamd/DockerRSpamD.java
```diff
@@ -0,0 +1,68 @@
...
+    private static final String DEFAULT_TAG = "3.2-r2-alpine3.16.0-r0";
+    private static final int DEFAULT_PORT = 11334;
+
+    private final DockerRedis dockerRedis;
+    private final GenericContainer<?> container;
+    private Network network;
+
+    public DockerRSpamD() {
+        this.network = Network.newNetwork();
+        this.dockerRedis = new DockerRedis(network);
```
chibenwa, line 38: Redis?

Arsnael: Doesn't start without it... Actually on all the images I tried (remember, there is no ifficial rspamd image) that's one I found that only required redis... others required other stuff

It seems that many modules of rspamd use redis, and that's probably why custom rspamd images all xpect a redis to start (from what I could test): 

We will use Redis as well anyway in our deployments, and we have already a docker redis image in james tests, so I thought was ok... If no then please find me one then, cause I spent a day last week swooping custom images

chibenwa: That s ok

third-party/rspamd/pom.xml
Arsnael: Gonna fix that but want to add that the build was green though in case :)

third-party/rspamd/pom.xml
```diff
@@ -30,7 +30,7 @@
 
     <artifactId>apache-james-rspamd</artifactId>
     <name>Apache James :: Third Party :: RSpamD</name>
-    <description>RSpamD Java client and testing utilities</description>
+    <description>RSpamD Java and HTTP client and testing utilities</description>
```
chibenwa, line 33: RSpamD Java client (HTTP) and testing utilities

quantranhong1999 (approved):

vttranlina (approved):

chibenwa (approved):

PR #1085, opened 2022-07-13
JAMES-3791 Remote Delivery uses a pool of SMTP sessions

ottoka:
Remote Delivery now uses pools of SMTP and SMTPS sessions. This prevents a race condition in setting per-delivery properties.

The pools replace the former shared session objects. During tryDeliveryToHost(), the selectSession() method now borrows a session from the appropriate pool. Its properties can be changed as necessary for the current delivery. Once delivery is done, the new method releaseSession() finally returns this session to its pool. The pool resets any delivery specific properties to prevent information leaks.

The session pools are unbounded, so they can scale to whatever maximum thread concurrency is required. Otherwise, the default pool configuration should be fine.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2527,6 +2527,11 @@
             </dependency>
             <dependency>
                 <groupId>org.apache.commons</groupId>
+                <artifactId>commons-pool2</artifactId>
+                <version>2.11.1</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.commons</groupId>
                 <artifactId>commons-text</artifactId>
                 <version>1.9</version>
             </dependency>
--- a/server/mailet/mailets/pom.xml
+++ b/server/mailet/mailets/pom.xml
@@ -216,6 +216,10 @@
             <artifactId>commons-lang3</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-pool2</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.apache.james</groupId>
             <artifactId>apache-mime4j-dom</artifactId>
         </dependency>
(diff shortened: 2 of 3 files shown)


jeantil:
I'm unable to connect to GitHub today and checked the diff from my phone.
Sorry if my questions are naive because of the reduced context.

- As it, releasing the session rethrows any Exception by wrapping it in a
MessagingException. Won't this end in the delivery being marked as failed
internally even if the mail was actually accepted ?  I would have expected
an implementation similar to that of `closeTransport` which logs an error
but doesn't rethrows to preserve the delivery status.
- I haven't traced the flow completely (not trivial on mobile GitHub:) but
what happens if closeTransport throws ? (Could that end up in sessions that
are never released back to the pool ?)

Le mer. 13 juil. 2022 à 14:56, Karsten Otto ***@***.***> a
écrit :

ottoka:
Right, it makes sense to ignore exceptions on releaseSession() and just log a warning. My understanding of commons pool is that abandoned objects (not returned to the pool) should be mostly harmless if they are comparatively cheap (like our sessions) and the pool is unbounded. So this holds as well for your second case, e.g. if closeTransport() unexpectedly throws some RuntimeException.

Arsnael (approved):

jeantil:
hmm I may have missed something but I think the wrong commit was merged to master : [ad05e79](https://github.com/apache/james-project/commit/ad05e797507bbab2bbb3709b402c3627262874c3) instead of [6c89b75](https://github.com/apache/james-project/pull/1085/commits/6c89b7588637a5dbcd21a46d24aa282b195e99ec)

 As a result that same commit was cherry-picked to be proposed for backport to 3.7.x 

was that on purpose ?

ottoka:
Strange. I just cloned a fresh James repo, and its master branch has the version of MailDelivrerToHost where releaseSession() just logs a warning. So I think we are good?

PR #1087, opened 2022-07-15
JAMES-3791 Remote Delivery uses a pool of SMTP sessions (3.7.x)

ottoka:
Straight cherry-pick from master. As a critical bug fix, this should be in 3.7.x too.


chibenwa (approved):

ottoka:
Having clarified that the version of MailDelivrerToHost on master actually is what it is supposed to be, I guess we can try this again. Note that here too, releaseSession() just logs a warning, so any exception will not affect the outcome of a successful remote delivery.

Arsnael (approved):

PR #1094, opened 2022-07-27
Jenkinsfile cleanup changes

Arsnael:


Arsnael:
Few things here... I switched to cleanWs() instead of deleteDir(). Rationals => https://stackoverflow.com/questions/54019121/what-is-the-difference-between-deletedir-cleanws-and-wscleanup-in-jenkins-pi

Some comments saying deleteDir() is bad with a docker context. Like it gets confused because the mounted directory went away... Was thinking could that explain why sometimes we get 404 container not found in the build? Workers seems to have the capacity to hold 2 builds at the same time... I don't know if deleteDir() could create those issues by when starting a 2nd build on the worker, it fucks up docker context with the first build running... Anyways it's just a try

On an other note I saw the Cassandra build using it as well in their Jenkinsfile as pre clean step. They don't seem to have a post clean... well if you clean your workspace everytime before building I think it's not necessary to keep it too (specially only in success case)

As well I added `skipDefaultCheckout(true)` in options because they advice it in the plugin doc: https://plugins.jenkins.io/ws-cleanup/ 
And I think it makes sense... What we were doing until now was having that automatic checkout done, then cleanup the workspace (so wiping the thing we did before out) and then redoing a checkout in the next step... Then the automatic checkout is useless

chibenwa (approved):

PR #1106, opened 2022-08-05
JAMES-3775 RSpamD mailet should support rewrite subject

quantranhong1999:


third-party/rspamd/src/main/java/org/apache/james/rspamd/RSpamDScanner.java
```diff
@@ -36,25 +36,36 @@
...
     @Override
     public void service(Mail mail) throws MessagingException {
         AnalysisResult rSpamDResult = rSpamDHttpClient.checkV2(new MimeMessageInputStream(mail.getMessage())).block();
 
         mail.getRecipients()
             .forEach(recipient -> appendRSpamDResultHeader(mail, recipient, rSpamDResult));
+
+        rSpamDResult.getDesiredRewriteSubject()
+            .filter(any -> rewriteSubject)
+            .ifPresent(Throwing.consumer(desiredRewriteSubject -> mail.getMessage().setSubject(desiredRewriteSubject)));
```
chibenwa, line 68:
```
if (rewriteSubject) {
            rSpamDResult.getDesiredRewriteSubject()
                .ifPresent(Throwing.consumer(desiredRewriteSubject -> mail.getMessage().setSubject(desiredRewriteSubject)));
}
```

?

chibenwa (approved):

third-party/rspamd/src/main/java/org/apache/james/rspamd/RSpamDScanner.java
quantranhong1999: That was exactly the solution that came up to me first, but then I switch to the current code which looks more functional.
Anyway as you wish.

Arsnael (approved):

chibenwa (approved):

chibenwa:
Solve the conflict please :-)

PR #1107, opened 2022-08-05
JAMES-3798 fix bounce exception when no date header is present

ouvtam:


chibenwa (commented):
Thanks a lot for the contribution!

Having a unit test for MailModifier would be awesome, if you have 5 minutes to devote to it!

chibenwa:
I propose the following test.

Cheers, and thanks for the contribution.

Arsnael (approved):
Looks good to me. Thank you for the kind contribution :)

quantranhong1999 (approved):

ouvtam:
@chibenwa thanks for adding the test. I was just about to start writing the unit test.

PR #1112, opened 2022-08-09
JAMES-3775 Increase RSpamD scoring stability

quantranhong1999:
RSpamD scoring delegates to many async modules and then aggregate the result. The implicit default `task_timeout` is 8 seconds which is sometime not long enough to return the full result.
Increasing timeout to a bigger value (say 20s) could let RSpamD enough time to fully score a mail.

### DIFF
--- a/third-party/rspamd/src/test/java/org/apache/james/rspamd/DockerRSpamDExtensionTest.java
+++ b/third-party/rspamd/src/test/java/org/apache/james/rspamd/DockerRSpamDExtensionTest.java
@@ -23,9 +23,11 @@ import static org.apache.james.rspamd.DockerRSpamD.PASSWORD;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.hamcrest.core.Is.is;
 
+import org.apache.james.junit.categories.Unstable;
 import org.apache.james.util.Port;
 import org.apache.james.webadmin.WebAdminUtils;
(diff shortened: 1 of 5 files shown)


third-party/rspamd/src/test/resources/rspamd-config/actions.conf
```diff
@@ -1,29 +0,0 @@
-# Actions settings
```
vttranlina, line 1: this file is unnecessary?
Don't forget we haven't any spam.eml sample file with scores greater than 15

Arsnael (commented):
Read it

vttranlina:
On my PC
image

chibenwa:
Screenshot from 2022-08-10 22-39-50

chibenwa:
Can we please run rspamd tests as unstable tests while you investigate this issue / relax assertions to not include scoring?

quantranhong1999:
I investigated the issue and likely it is a bug regarding to RSpamD (reported) and no more we can do at least ATM.

> Can we please run rspamd tests as unstable tests while you investigate this issue / relax assertions to not include scoring?

The only place we assert the score (on master) is on a checkHam test which is not observed an unstable score issue.
But sure, relaxing the score there also is better, just in case.

chibenwa:
Please rebase @quantranhong1999

PR #1119, opened 2022-08-11
JAMES-3802 Webadmin route to clean events of a dead letter group

quantranhong1999:


event-bus/api/src/test/java/org/apache/james/events/EventDeadLettersContract.java
```diff
@@ -197,6 +197,14 @@ default void removeShouldThrowWhenGroupIsNull() {
                 .isInstanceOf(IllegalArgumentException.class);
         }
 
+        @Test
+        default void removeAllEventsOfAGroupShouldThrowWhenGroupIsNull() {
```
chibenwa, line 201: What is the result of calling `remove(GROUPA)` then `groupsWithFailedEvents()` ?

vttranlina (approved):

event-bus/api/src/test/java/org/apache/james/events/EventDeadLettersContract.java
quantranhong1999: It should remove the group (in Cassandra group table) (y)

chibenwa (approved):

Arsnael (approved):

PR #1121, opened 2022-08-11
JAMES-3803 RemoteDelivery uses different scheduler for dequeuing

AdBuch:
This is necessary as else in high load a delivery-task can queue in the same thread as the never ending dequeuing-task and stay there forever


chibenwa (approved):
I am working at reproducing the issue described and hope to come up with a test for it!

chibenwa:
Reproduced, I confirm the fix works as intended. This error was specific for ActiveMQ.

chibenwa:
Should we do the same in JamesMailSpooler ?

AdBuch:
JamesMailSpooler should not be affected as it still uses Schedulers.elastic() which has no bounderies and no internal queues

Arsnael (approved):

quantranhong1999 (approved):

vttranlina (approved):

PR #1141, opened 2022-08-22
JAMES-3808 Tag PulsarMailQueueTest::clearShouldNotFailWhenBrowsingIte…

chibenwa:
…rating as unstable

This failed 6 times out of 50

Cc @ jeantil


jeantil:
That one is different from the other two and looks like a pulsar bug (5xx error returned by the pulsar API)

jeantil (approved):

PR #1143, opened 2022-08-23
[BUILD] Enforce the use of a local m2 repo

chibenwa:
This provides otherwise missing build isolation


Arsnael (approved):
Ah nice one! :)

PR #1156, opened 2022-08-25
JAMES-3775 Rspamd Scanner should distinguish reject from add_header

chibenwa:
Now positions headers on `add header` action (which would end up moving the email in user spam box) and distinguish reject action (email not delivered).

Scores threshold are configurable on RspamD side.

### DIFF
--- a/third-party/rspamd/README.md
+++ b/third-party/rspamd/README.md
@@ -69,7 +69,7 @@ might not even be desirable.
 </processor>
 
 <!--Store rejected spam emails (with a very high score) -->
-<processor state="virus" enableJmx="false">
+<processor state="spam" enableJmx="false">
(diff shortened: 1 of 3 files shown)


third-party/rspamd/README.md
```diff
@@ -58,6 +67,13 @@ you can specify how James process mail virus. We provide a sample Rspamd mailet
         <subjectPrefix>[VIRUS]</subjectPrefix>
     </mailet>
 </processor>
+
+<!--Store rejected spam emails (with a very high score) -->
+<processor state="virus" enableJmx="false">
```
Arsnael, line 72:
```suggestion
<processor state="spam" enableJmx="false">
```

Arsnael (approved):

PR #1159, opened 2022-08-26
[BUILD] Provision Cassandra tables in parallel

chibenwa:
Localy table creation went from 42s to 6s.

### DIFF
--- a/backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/CassandraTableManager.java
+++ b/backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/CassandraTableManager.java
@@ -51,11 +51,11 @@ public class CassandraTableManager {
     public InitializationStatus initializeTables(CassandraTypesProvider typesProvider) {
         KeyspaceMetadata keyspaceMetadata = session.getMetadata().getKeyspaces().get(session.getKeyspace().get());
 
-        return module.moduleTables()
-                .parallelStream()
-                .map(table -> table.initialize(keyspaceMetadata, session, typesProvider))
-                .reduce(InitializationStatus::reduce)
-                .orElse(InitializationStatus.ALREADY_DONE);
+        return Flux.fromIterable(module.moduleTables())
+            .flatMap(table -> table.initialize(keyspaceMetadata, session, typesProvider))
+            .reduce(InitializationStatus::reduce)
+            .switchIfEmpty(Mono.just(InitializationStatus.ALREADY_DONE))
+            .block();
     }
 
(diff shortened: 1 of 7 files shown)


vttranlina (approved):

quantranhong1999 (approved):

backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/CassandraTableManager.java
```diff
@@ -52,9 +52,9 @@ public InitializationStatus initializeTables(CassandraTypesProvider typesProvide
         KeyspaceMetadata keyspaceMetadata = session.getMetadata().getKeyspaces().get(session.getKeyspace().get());
 
         return module.moduleTables()
-                .stream()
+                .parallelStream()
```
mbaechler, line 55: it will use the fork/join executor to execute IO base tasks, you should switch to reactor + elastic executor

chibenwa: Got lazy, sorry...

Arsnael (approved):

backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/CassandraTypesCreator.java
```diff
@@ -39,9 +42,10 @@ public CassandraTypesCreator(CassandraModule module, CqlSession session) {
     public InitializationStatus initializeTypes() {
         KeyspaceMetadata keyspaceMetadata = session.getMetadata().getKeyspaces().get(session.getKeyspace().get());
 
-        return types.stream()
-                .map(type -> type.initialize(keyspaceMetadata, session))
-                .reduce(InitializationStatus::reduce)
-                .orElse(InitializationStatus.ALREADY_DONE);
+        return Flux.fromIterable(types)
+            .flatMap(type -> type.initialize(keyspaceMetadata, session))
+            .reduce(InitializationStatus::reduce)
```
mbaechler, line 47: nitpick: I would rather use the reduce with initial state for clarity

mbaechler (approved):

backends-common/cassandra/src/main/java/org/apache/james/backends/cassandra/init/CassandraTypesCreator.java
chibenwa: The semantic is actually different and is not straightforward with the syntax you propose. I will keep this as is.

PR #1168, opened 2022-08-30
JAMES-3810: Backport of fixes to 3.7.x

ottoka:
This is mostly a cherry pick of the respective commits from JAMES-3810. The only "backport" is the use of Schedulers.elastic() instead of ReaktorUtils.BLOCKING_CALL_WRAPPER since the latter does not exist yet in 3.7.x.

Note: Merging this should probably wait until the pending 3.7.1 release is decided.


chibenwa:
No we can merge it now, it would just not be included in 3.7.1 (but potentially in a future 3.7.2)

Arsnael (approved):

PR #1170, opened 2022-08-31
JAMES-3732 Fix a typo in poms regarding the leak detector

chibenwa:
Disable leak detector in tests as we do not achieve good resource management in
those tests. Enable it selectively in a few relevant projects.


Arsnael (approved):

jeantil:
I guess it has the merit to clarify the situation

jeantil (approved):

chibenwa:
server-protocols-smtp depends on the memory mailqueue which clones mail and do not dispose of them... Hence the fixup.

PR #1172, opened 2022-08-31
JAMES-3755 Add a classifiedAsSpam parameter to Rspamd routes

chibenwa:
This allows for instance reporting messages manually reported by the user to the
Spam mailbox with re-reporting messages already classified as Spam by Rspamd,
avoiding creating a positive retro-action loop, and only reporting relevant messages.

Similar for Ham reporting.


Arsnael:
Can be rebased :)

Arsnael (commented):
The code looks good but isn't the documentation for that extra param on rspamd routes missing?

Arsnael (approved):

PR #1175, opened 2022-08-31
JAMES-3812 Provide a ToPlainText mailet

chibenwa:


chibenwa:
🍏 build

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/ToPlainText.java
```diff
@@ -0,0 +1,144 @@
...
+                mutated = convertRemainingHtmlToPlainText(multipart.getBodyPart(i));
+            }
+            return mutated;
+        }
+        return false;
+    }
+
+    // true if the message content is mutated
+    private boolean convertRemainingHtmlToPlainText(BodyPart bodyPart) throws IOException, MessagingException {
+        if (bodyPart.getContent() instanceof Multipart) {
```
Arsnael, line 125: This if clause and the one in line 112 are exactly the same code inside. Extracting it into a private method maybe would be better?

Arsnael (approved):

PR #1181, opened 2022-09-05
WIP IMAP throttler

chibenwa:
## Prior the "reactive IMAP migration"

IMAP request execution is blocking, there is a thread pool of max "ioThreads" (default 20) taking care of 1 request each.

 -> Bad resource utilisation as IO threads get's sleeping most of the time => this was the motivation to migrate to a "reactive style of code".
 -> QOS: at any given time there is at most IO thread requests processed at once.

While the "resource utilisation" of this model might be detrimental, the QOS part of this model might actually be beneficial.

## After the "reactive IMAP migration"

The model is now fully asynchronous with few threads processing potentially thousands of requests.

 -> Good resource utilisation
 -> No QOS

## Diagnostic

On one of our prod instances we notice IMAP request activity spikes that actually overwhelm the servers:

 -> Exhaust the http concurrency on the S3 driver
 -> SERIAL timeouts on Cassandra side.
 -> Eventually build up as bad as to get an OOM

I believe that this situation was prevented by the QOS bringed in by the "blocking thread pool" execution model.

I believe bringing back in that QOS as an option might be beneficial.

## TODO

QOS (concurrentIMAPRequests option in imapserver.xml) defaults to no QOS (unbounded)

Find a way to implement this with Reactor / Netty

Experiment to see if this stabilizes OpenPaaS.linagora.com. If validated contribute it upstream.

### DIFF
--- a/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ReactiveThrottler.java
+++ b/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ReactiveThrottler.java
@@ -1,0 +1,94 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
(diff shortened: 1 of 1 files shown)


quantranhong1999:
An effort contributing unit tests for the throttler : https://github.com/quantranhong1999/james-project/commit/81d80229f642acfc60dc517a51b51e029d57b8fd#r83411951

chibenwa:
See https://github.com/apache/james-project/pull/1185

PR #1182, opened 2022-09-07
JAMES-3815 Handle possible null values in imapUidTable and messageIdTable

quantranhong1999:
This NPE likely comes from badly applied [Rework message denormalization migration](https://github.com/apache/james-project/blob/master/upgrade-instructions.md#rework-message-denormalization) which creates in `imapUidTable` and `messageIdTable` a few rows with null `internalDate`, `bodyStartOctet`, `fullContentOctets` and `headerContent`. Now the Cassandra driver 4 code change just triggers that NPE (can not convert a null to Date).

Checked: `bodyStartOctet`, `size`, `headerContent` should be already good from NPE.

### DIFF
--- a/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdDAO.java
+++ b/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdDAO.java
@@ -524,10 +524,10 @@ public class CassandraMessageIdDAO {
                 .modSeq(ModSeq.of(row.getLong(MOD_SEQ_LOWERCASE)))
                 .threadId(getThreadIdFromRow(row, messageId))
                 .build())
-            .bodyStartOctet(row.getInt(BODY_START_OCTET_LOWERCASE))
+            .bodyStartOctet(row.get(BODY_START_OCTET_LOWERCASE, Integer.class))
             .internalDate(Optional.ofNullable(row.getInstant(INTERNAL_DATE_LOWERCASE))
                 .map(Date::from))
-            .size(row.getLong(FULL_CONTENT_OCTETS_LOWERCASE))
+            .size(row.get(FULL_CONTENT_OCTETS_LOWERCASE, Long.class))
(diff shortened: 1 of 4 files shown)


vttranlina:
image
Should we fix the NPE of other classes in this pr?

quantranhong1999:
> Should we fix the NPE of other classes in this pr?

IMO it depends on which tables and do we allow to write null into that. I will check tmr.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdDAO.java
```diff
@@ -538,4 +540,36 @@ private ThreadId getThreadIdFromRow(Row row, MessageId messageId) {
         }
         return ThreadId.fromBaseMessageId(CassandraMessageId.Factory.of(threadIdUUID));
     }
+
+    @VisibleForTesting
+    Mono<Void> insertNullInternalDateAndHeaderContent(CassandraMessageMetadata metadata) {
```
chibenwa, line 545: Imo overkill

chibenwa (commented):
What if size is null? HeaderContent? bodyStartOctet ?
Finally it is not a pb of badly applied migration but a regression (you) introduced during cassandra driver 4.0 upgrade.

Arsnael:
> Finally it is not a pb of badly applied migration but a regression (you) introduced during cassandra driver 4.0 upgrade.

You mean "we"...?

Arsnael:
> What if size is null? HeaderContent? bodyStartOctet ? 

Answer in his first comment of the issue: 

```
Checked: bodyStartOctet, size, headerContent should be already good from NPE.
```

Arsnael (commented):
Read it looks ok to me so far

quantranhong1999:
> What if size is null? HeaderContent? bodyStartOctet ?

We already wrap `Optional.ofNullable` when reading them. 
Even `bodyStartOctet` and `size` are special cases when the driver does not return the `null` value but defaults 0 instead. May be return actual `null` here will be more exact.

Arsnael (approved):

chibenwa (approved):

PR #1185, opened 2022-09-09
JAMES-3816 Limit IMAP concurrent requests

chibenwa:


Arsnael (approved):

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ReactiveThrottler.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    private void onRequestDone() {
+        concurrentRequests.decrementAndGet();
+        Publisher<Void> throttled = queue.poll();
+        if (throttled != null) {
+            Mono.from(throttled)
+                .doFinally(any -> onRequestDone())
+                .subscribeOn(Schedulers.parallel())
```
quantranhong1999, line 87: I have a concern here about the `Schedulers.parallel()`: the deeper layer of our code may contain some blocking code that is not friendly with the parallel scheduler.

chibenwa: It is the responsibilities to wrap such calls and execute it on an elastic scheduler.

Blocking calls on the netty event loops are critical too and should be avoided at all cost, too.

This changes nothing...

quantranhong1999 (approved):

Arsnael (approved):

PR #1203, opened 2022-09-21
[STABILITY] Limit number of threads in parallel when creating cassand…

Arsnael:
…ra tables

I encountered some issues when trying to run James with a Cassandra not initialized yet. I could observe sometimes James starting to create the tables on Cassandra and then stopping on a driver timeout exception. 

We likely try to create too many tables in parallel at the same time. Limiting the concurrency seems to help greatly. It might help with the build stability too.


chibenwa (approved):
CassandraTypeProvider too maybe? (can be a separate PR)

quantranhong1999 (approved):

PR #1206, opened 2022-09-23
[PULSAR] Disable randomly failing test

chibenwa:


jeantil:
I don't know, I have been running it until failure and it seems to remain green  (30 iterations for now, I'll let it run while I grab lunch)

jeantil:
I reviewed the test and it relies on a 100ms wait to ensure the filter has been registered. Since the filter registration round-trips through pulsar even in a single instance process, I guess it can sometimes take longer that 100ms on the CI. 

I think I already proposed introducing a dilation/scaling factor for waits on the CI platform.. I'll discuss this more with @mbaechler tomorrow to see if we can come up with a more deterministic approach for tests.
image

chibenwa:
Ok closing, but I'll re-open if the CI fails again here

jeantil:
for reference https://issues.apache.org/jira/browse/JAMES-3696

PR #1207, opened 2022-09-25
JAMES-3820 implement DNSRBLHandler as a MailHook instead of RcptHook …

ouvtam:
…to call it as soon as possible after SMTP authentication.

At the moment the DNSRBL handler is implemented as a RcptHook. Thus, for every RCPT TO call this handler will be called and a blocklist lookup will be issued.

One can argue It makes sense to implement the handler as a ConnectHandler, so the blocklist check is done as early as possible. However, if SMTP AUTH is successful then we should allow the connecting client anyway.

Therefore it makes sense to implement the DNSRBL handler at MAIL FROM stage that is MailHook. One caveat: According to [RFC 4954](https://datatracker.ietf.org/doc/html/rfc4954#section-5), authentication information can optionally provided as ESMTP AUTH parameter with a single value in the 'MAIL FROM:' command.

### DIFF
--- a/examples/custom-smtp-command/README.md
+++ b/examples/custom-smtp-command/README.md
@@ -67,7 +67,7 @@ public class MyCmdHandlerLoader implements HandlersPackage {
             MailSizeEsmtpExtension.class,
             UsersRepositoryAuthHook.class,
             AuthRequiredToRelayRcptHook.class,
-            SenderAuthIdentifyVerificationRcptHook.class,
+            SenderAuthIdentifyVerificationHook.class,
             PostmasterAbuseRcptHook.class,
             ReceivedDataLineFilter.class,
             DataLineJamesMessageHookHandler.class,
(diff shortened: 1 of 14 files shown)


chibenwa (approved):

ouvtam:
As discussed in https://issues.apache.org/jira/browse/JAMES-3820 I cleaned up additional RcptHooks that should rather be MailHooks.

chibenwa (approved):

chibenwa:
`Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.1.2:check (check-style) on project james-server-protocols-smtp: You have 2 Checkstyle violations.`

Arsnael (approved):

ouvtam:
sorry, I missed that one. I've updated the PR.

chibenwa:
```
06:21:05,972 [ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project custom-smtp-command: Compilation failure
```

More compile errors :-)

ouvtam:
Some tests are still failing. I am going to fix it.

ouvtam:
I need to reintroduce the RcptHooks (see discussion in https://issues.apache.org/jira/browse/JAMES-3820) because of different behaviour when running James as MSA (mail submission agent) or MTA (mail transfer agent).

So this PR needs some rework.

chibenwa:
> So this PR needs some rework.

I turn the PR into a draft to make this clear.

Once necessary fixes are in place, we can un-draft this.

chibenwa (approved):
Looks good to me.

chibenwa:
`SmtpIdentityVerificationTest` are failing...

ouvtam:
Some tests are failing but they are not caused by my changes.

Arsnael (approved):

PR #1211, opened 2022-09-27
[JAMES-3696] makes PulsarMailQueue tests more stable

jeantil:
Previously to this patch, filter registration was going through a pulsar delivery even for non-distributed James. This meant tests that verify the filtering behavior had to wait for the message to be consumed back from the queue.
    
This lead us to introduce an artificial wait to ensure that the filter was correctly registered. Unfortunately, the delay was not always enough, especially on the CI which lead to flaky tests.
    
I propose to add a small optmization which registers the filter with the receiving instance synchronously before publishing it to pulsar.
This way, the wait time to filter  registration can be much smaller.

Because the filter stage uses a set, adding the same filter multiple times should have a negligible impact on performances.

There was already a test in place that verifies that filters are distributed through the cluster `ensureThatDeletionApplyOnBrowsingBothInstancesWithTwoInstancesOfMailQueue`

relates to https://github.com/apache/james-project/pull/1206


chibenwa (approved):

Arsnael (approved):

jeantil:
:green_apple: build

quantranhong1999 (approved):

PR #1212, opened 2022-09-28
JAMES-3827 Improve indexing of email address domains

chibenwa:
Given an email with `bob@domain.tld`

-> `bob@domain.tld` matches
-> `bob` (localpart) matches
-> `domain.tld` (domain) matches
-> `domain` (part of the domain) do not match. (and it should).

I think we could easily propose a better 'search experience' by improving indexing and searches for domains.

Traps to avoid:

Given `bob@domain.tld` search we should not return results of other addresses of `domain.tld` domain.

Given `domain.tld` search we should not return unrelated search results from the `tld` top-domain.

migration strategy will be aturaly covered upon the OpenSearch migration.

### DIFF
--- a/mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/json/EMailer.java
+++ b/mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/json/EMailer.java
@@ -30,10 +30,19 @@ public class EMailer implements SerializableMessage {
 
     private final Optional<String> name;
     private final String address;
+    private final String domain;
 
-    public EMailer(Optional<String> name, String address) {
+    public EMailer(Optional<String> name, String address, String domain) {
         this.name = name;
         this.address = address;
+        this.domain = la(domain);
+    }
+
+    String la(String s) {
+        if (s.contains(".")) {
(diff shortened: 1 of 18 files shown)


Arsnael (approved):

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/json/EMailer.java
```diff
@@ -30,10 +30,19 @@ public class EMailer implements SerializableMessage {
...
+    private final String domain;
 
-    public EMailer(Optional<String> name, String address) {
+    public EMailer(Optional<String> name, String address, String domain) {
         this.name = name;
         this.address = address;
+        this.domain = la(domain);
+    }
+
+    String la(String s) {
```
quantranhong1999, line 41: Can we have a better name for this method?

quantranhong1999 (commented):
And a bit in `upgrade-instructions.md`?

chibenwa:
> And a bit in upgrade-instructions.md?

Not needed for James. People going from 3.7.x to 3.8.0 will reindex their data in OpenSearch...

PR #1220, opened 2022-09-30
JAMES-3826 Introduce additional healthcheck mechanism

quantranhong1999:


server/container/guice/common/src/main/java/org/apache/james/PeriodicalHealthChecks.java
```diff
@@ -46,7 +47,7 @@ public class PeriodicalHealthChecks implements Startable {
     private Disposable disposable;
 
     @Inject
-    PeriodicalHealthChecks(Set<HealthCheck> healthChecks, PeriodicalHealthChecksConfiguration configuration) {
+    PeriodicalHealthChecks(@Named("jamesHealthCheckWithAdditionalOnes") Set<HealthCheck> healthChecks, PeriodicalHealthChecksConfiguration configuration) {
```
chibenwa, line 50: Maybe we can think of a better name for this injection.

How about... `@Named("resolved-checks") Set<HealthCheck>`

chibenwa (commented):
The exemple is a very good initiative IMO. 

We need a mention in `exemples/README.md`. I see no changes there...

We need a test in `server/apps/memory-app` where we pass via a Guice ovveride a `PeriodicalHealthChecksConfiguration` that defines an additional check (check being on the test class path) and verify that this check is well expose via webadmin to ensure bindings are well done (think avoid regressions)

server/container/guice/common/src/main/java/org/apache/james/PeriodicalHealthChecks.java
quantranhong1999: Sure. BTW I have already written the integration test: `HealthCheckExtensionTest`.

chibenwa (approved):

Arsnael (approved):

PR #1225, opened 2022-10-03
JAMES-3816 Manage IMAP throttling errors

chibenwa:
Previously on rejected IMAP computation the connexion was closed which may be a bit too harsh

### DIFF
--- a/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ReactiveThrottler.java
+++ b/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ReactiveThrottler.java
@@ -23,6 +23,7 @@ import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.james.imap.api.ImapMessage;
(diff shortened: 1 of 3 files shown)


server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ReactiveThrottler.java
```diff
@@ -73,7 +81,7 @@ public Mono<Void> throttle(Publisher<Void> task) {
                 String.format(
                     "The IMAP server has reached its maximum capacity "
                         + "(concurrent requests: %d, queue size: %d)",
-                    maxConcurrentRequests, maxQueueSize)));
+                    maxConcurrentRequests, maxQueueSize), imapMessage));
```
quantranhong1999, line 84: Missing `%s` for imapMessage

quantranhong1999 (approved):

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ReactiveThrottler.java
chibenwa: No it's not, it's a parameter to the exception, not the string

Arsnael (approved):

PR #1228, opened 2022-10-04
JAMES-3831 Specification for JMAP Mail RFC-3831 Identity sortOrder ad…

chibenwa:
…ditional property


quantranhong1999 (approved):
The commit message should say RFC-8621 instead of RFC-3831.

Arsnael (approved):

PR #1232, opened 2022-10-05
JAMES-3832  Support TLS host name verification

ottoka:
Using host names instead of IP addresses for RemoteDelivery allows enabling of TLS host name verification in security sensitive deployments, via Java Mail property mail.smtp(s).ssl.checkserveridentity in the mailet configuration.

I do not see any disadvantages with this change, but the intricacies of server lookup are complex, and I may not see the full picture. Any reasons why this would be a problem?

### DIFF
--- a/server/apps/distributed-app/docs/modules/ROOT/partials/RemoteDelivery.adoc
+++ b/server/apps/distributed-app/docs/modules/ROOT/partials/RemoteDelivery.adoc
@@ -46,6 +46,7 @@ to switch the connection to a TLS-protected connection before issuing any login 
 * *sslEnable* (optional) - a Boolean (true/false) indicating whether to use SSL to connect and use the SSL port unless
 explicitly overridden. Default is false. Setting up to true will result in delivery attempts in SMTPS on port 465 with a fallback
 to SMTP on port 25. The trust-store if needed can be customized by *-Djavax.net.ssl.trustStore=/root/conf/keystore*.
+* *verifyServerIdentity* (optional) - a Boolean (true/false) indicating whether to match the remote server name against its certificate on TLS connections. Default is true. Disabling this runs the risk of someone spoofing a legitimate server and intercepting mails, but may be necessary to contact servers that have strange certificates, no DNS entries, are reachable by IP address only, and similar edge cases.
 * *gateway* (optional) - a String containing a comma separated list of patterns defining the gateway servers to be used to
 deliver mail regardless of the recipient address. If multiple gateway servers are defined, each will be tried in definition order
 until delivery is successful. If none are successful, the mail is bounced. The pattern is *host[:port]* where:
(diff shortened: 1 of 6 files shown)


chibenwa:
Nice finding!

Shouldn't mail.smtp(s).ssl.checkserveridentity be set to true by default when unspecified by the user?

This would be much more secure, i can already see me forgetting to set this in halth of my environments!

ottoka:
Not sure about enabling verification by default. It requires people to correctly set up their certificates with the right FQDN, possibly using subjectAltName extensions, wildcards etc. In my experience, this is usually the case with commercial certificates, but not necessarily if people roll their own self-signed ones. And it is also a question of valuing connectivity over security or vice versa.

I would keep verification disabled by default to keep the current behavior for people just upgrading James, but add it to the sample configurations (set to true?) so people don't forget it on new deployments.

For that it would also make sense to have our own mailet config option "verifyHostname" or so, which sets the Java Mail properties internally and avoid the inevitable confusion about the correct usage of mail.smtp or mail.smtp**S** variants.

chibenwa:
`mail.smtp(s).ssl.checkserveridentity` should at the very least be seriously documented, that's for sure!

> I would keep verification disabled by default to keep the current behavior for people just upgrading James, but add it to the sample configurations (set to true?) so people don't forget it on new deployments.

Agree this would be the minimum.

According to me, remote delivery have 2 use cases:

 - Gatewaying: you did set up an other MX on your site to send emails to others domains on behalf of James. James could maintain a persistent SMTP connection to this gateway. The IP / hostname of the gateway is hard coded in James. All is in local network, so SSL is only secondary line of defense, less crucial. SMTPS can be setted up.
 - MX mode: DNS resolution and hands over the mail to the remote domain. Third party is unknown, and changes for every mail. No persistent connection can be attempted. Port 25 SMTP is used, STARTTLS can opportunistically encrypt the exchange, but there's a dilemma between delivery/security. Here we deal with third party on the internet.
 
 I think if a certificate is handed to us in MX mode, it needs to be seriously validated to avoid Man In The Middle attacks. I could see people opening CVEs on this just like they did for Netty.
 
 Ref: https://security.snyk.io/vuln/SNYK-JAVA-IONETTY-1042268
 
 We should expect sites on the internet to have correctly configured certificates. Failure to do so would be a good reason NOT to deliver the mail.
 
 Also, hobbyists that just want something to work should be able to:
   - Get quickly related errors
   - And turn of related configuration

ottoka:
OK, maybe I am too conservative regarding changes in software behavior. The way I understand you, people generally want server verification, maybe even expected it to be there all along. And know enough to at least create certificates with CN=myhost.mydomain.org. Running without server verification would be a rare exception.

So I will
- add a new mailet property "verifyServerIdentity" which defaults to TRUE
- extend the mailet documentation accordingly, warning people to know the risk of disabling it
- add a note to the upgrade instructions about the change and how to disable if necessary.

Regarding getting related errors, Java Mail produces IOException("Can't verify identity of server: " + server) on connection, which will lead to a temporary mail failure (with subsequent retries). Seems clear enough as a problem indication to me, and an appropriate reaction by James: According to the premise above, such a situation is most likely a temporary issue, e.g. after changing certificates at the remote end. We can expect it to be fixed in a timely manner, so the next delivery attempt will succeed.

chibenwa:
That would be perfect. We should mention this in `upgrade-instructions.md` too IMO.

chibenwa:
Be careful also that the LOG is at least at WARN level... Remote delivery log levels are rather low as I can recall...

chibenwa (approved):
Dont forget to document this also in RemoteDelivery javadoc

Arsnael (approved):

ottoka:
> Dont forget to document this also in RemoteDelivery javadoc

Always something to forget, sigh. But a good thing too, I noticed the Security section was somewhat outdated and needed a rewrite. The latest push fixes it.

chibenwa (approved):

ottoka:
Sigh... this breaks several integration tests that use fake domain/mailhost names in combination with localhost IP address. Before this worked, but now the fake names get passed to Java Mail instead, which cannot resolve them obviously.
I could change the code to only connect by hostname is verifyServerIdentity is enabled, and fall back to the old IP-based method otherwise. But then I would have to
1. disable verifyServerIdentity in test configurations and builders all over the codebase; good chance I will miss some
2. -OR- change the default of verifyServerIdentity to FALSE, and rely on upgrade notice and sample configurations to let users know they might want to enable it.
Any suggestions?

chibenwa:
1. Seems better to me to be honnest

ottoka:
In the latest push, I only use hostnames when when it is needed, i.e. both verification and either SMTPS or StartTLS is enabled. This should keep the issue out of the integration test, which generally do not use TLS.

ottoka:
Latest push was just a rebase on master, so the CI would not skip any tests. I am glad it passed this time, so I am confident my solution works without causing problems and is ready for merging. I will do another rebase and fix the conflict in upgrade-instructions.md that popped up in the meantime.

PR #1233, opened 2022-10-06
JAMES-3604 Allow quorum queues only on mail queues

Arsnael:
Performance tests showed that there was some serious performance issues when having quorum queues for event notifications as well with IMAP.

### DIFF
--- a/event-bus/distributed/src/main/java/org/apache/james/events/KeyRegistrationHandler.java
+++ b/event-bus/distributed/src/main/java/org/apache/james/events/KeyRegistrationHandler.java
@@ -19,6 +19,7 @@
 
 package org.apache.james.events;
 
+import static org.apache.james.backends.rabbitmq.Constants.ALLOW_QUORUM;
 import static org.apache.james.backends.rabbitmq.Constants.AUTO_DELETE;
 import static org.apache.james.backends.rabbitmq.Constants.EXCLUSIVE;
 import static org.apache.james.events.RabbitMQEventBus.EVENT_BUS_ID;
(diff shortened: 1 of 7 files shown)


chibenwa (approved):

event-bus/distributed/src/main/java/org/apache/james/events/KeyRegistrationHandler.java
```diff
@@ -126,7 +127,7 @@ private void declareQueue(Sender sender) {
                 .durable(configuration.isEventBusNotificationDurabilityEnabled())
                 .exclusive(!EXCLUSIVE)
                 .autoDelete(AUTO_DELETE)
-                .arguments(configuration.workQueueArgumentsBuilder(AUTO_DELETE).build()))
+                .arguments(configuration.workQueueArgumentsBuilder(!ALLOW_QUORUM).build()))
```
quantranhong1999, line 130: I know it could work e.g `ALLOW_QUORUM` is true -> so we understand that we don't want to use the quorum queue for e.g event bus queue.

What if the `ALLOW_QUORUM` constant is false? Then I must understand that we want to use the quorum queue for event bus queue.

Should we make this more clear? Say always put `false` for other queues than mailQueue?

Arsnael: Don't get what's not clear... It's the same principle with other constants  defined here (AUTO_DELETE for example). They are set to true to make it more understandable when you read the code than just true or false.

`ALLOW_QUORUM` and `!ALLOW_QUORUM`... It speaks well enough for itself. I don't get what is confusing you here?

The `ALLOW_QUORUM` is only for mail queues already, the event notifications queue have `!ALLOW_QUORUM`

quantranhong1999: Ok if we will not change the `ALLOW_QUORUM` value then I am fine ^^

Arsnael: It's a constant :) so shouldn't change^^

quantranhong1999 (approved):

vttranlina (approved):

PR #1235, opened 2022-10-06
JAMES-3831 Implement urn:apache:james:params:jmap:mail:identity:sortorder extension

chibenwa:


server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/IdentitySetContract.scala
```diff
@@ -926,6 +1027,115 @@ trait IdentitySetContract {
...
+           |                        "sortOrder": 125
+           |                    }
+           |                ]
+           |            },
+           |            "c2"
+           |        ]
+           |    ]
+           |}""".stripMargin)
+  }
+
```
Arsnael, line 1138: Can we add a test to check we can't modify the sortOrder without using the custom capability?

chibenwa: Well it would be possible, however it do not hurt inter-operability (to want to modify the sort order you need to be aware it exists) => This corner case do not hurt.

But feel free to add a test if you believe it is necessary.

Arsnael (approved):

chibenwa:
Rebased to solve a merge conflict.

PR #1236, opened 2022-10-07
[BUILD] Tag Pulsar related work as unstable

chibenwa:
Failed in 4 builds out of 7.

```
com.github.fge.lambdas.ThrownByLambdaException: org.apache.pulsar.client.admin.PulsarAdminException$NotFoundException: Tenant does not exist
Caused by: org.apache.pulsar.client.admin.PulsarAdminException$NotFoundException: Tenant does not exist
Caused by: org.apache.pulsar.shade.javax.ws.rs.NotFoundException: HTTP 404 Tenant does not exist
```


jeantil:
I won't be able to look into this quickly not sure Matthieu has anymore
time.
So go ahead.
Current hypothesis is that the namespace creation is async, we don't wait
for it , or at least not enough, and pulsar suffers from slowdowns on the
CI.
I don't remember having this issue on builds before the version upgrade a
couple weeks ago. Nor have we seen it locally during our session last
Tuesday.

Le ven. 7 oct. 2022 à 04:17, Benoit TELLIER ***@***.***> a
écrit :

chibenwa:
Thanks for the quick reply.

I propose to merge this then reverts once the builds proved stable.

PR #1264, opened 2022-10-21
JAMES-3775 Handle possible empty `rspamdTimeoutInSeconds` when deserializing task additional information

quantranhong1999:
Make this field more robust again possible null cases.

### DIFF
--- a/third-party/rspamd/src/main/java/org/apache/james/rspamd/task/RunningOptions.java
+++ b/third-party/rspamd/src/main/java/org/apache/james/rspamd/task/RunningOptions.java
@@ -20,6 +20,7 @@
 package org.apache.james.rspamd.task;
 
 import java.time.Duration;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Predicate;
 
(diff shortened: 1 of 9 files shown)


Arsnael (approved):

third-party/rspamd/src/main/java/org/apache/james/rspamd/task/RunningOptions.java
```diff
@@ -66,27 +66,19 @@ public boolean test(MessageResult messageResult) {
...
-    }
-
+    private final Optional<Duration> rspamdTimeout;
 
     public RunningOptions(@JsonProperty("periodInSecond") Optional<Long> periodInSecond,
                           @JsonProperty("messagesPerSecond") int messagesPerSecond,
                           @JsonProperty("samplingProbability") double samplingProbability,
                           @JsonProperty("classifiedAsSpam") Optional<Boolean> classifiedAsSpam,
-                          @JsonProperty("rspamdTimeoutInSeconds") Duration rspamdTimeout) {
+                          @JsonProperty("rspamdTimeoutInSeconds") Optional<Duration> rspamdTimeout) {
```
chibenwa, line 81: I do not think we do need to have this optionalin the running Options.

In the end, we will always choose a timeout, thus it needs to be explicit here.

quantranhong1999: Reworked.

Arsnael (approved):

PR #1273, opened 2022-10-25
Fix release year for 3.7.0

marcelstoer:
For 3.7.0 it should read "Released March 202**2**" rather than "Released March 2021".


chibenwa:
Good catch. Thanks for the PR

Arsnael (approved):

PR #1274, opened 2022-10-26
JAMES-3842 Access SSLSession from Hooks and Handlers

ottoka:
James supports SMTPS and StartTLS, but currently has no way to programmatically access the resulting SSLSession from SMTP Hooks, POP3 Command Handlers etc. I propose to make the SSLSession accessible through the ProtocolSession.

This way, hooks/handlers can examine the SSL/TLS parameters negotiated with the currently connected client. They can check e.g. TLS version and selected cipher suite to evaluate the security level of the connection. With certificate based client authentication enabled, they can get the provided client certificate, e.g. to check against a permission list. Or a JamesMessageHook could attach it to the Mail object, so this information becomes even available to Mailets after spooling.

### DIFF
--- a/protocols/api/src/test/java/org/apache/james/protocols/api/AbstractProtocolTransportTest.java
+++ b/protocols/api/src/test/java/org/apache/james/protocols/api/AbstractProtocolTransportTest.java
@@ -27,9 +27,12 @@ import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.stream.IntStream;
+
+import javax.net.ssl.SSLSession;
 
 import org.apache.james.protocols.api.handler.LineHandler;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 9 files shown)


protocols/api/src/test/java/org/apache/james/protocols/api/AbstractProtocolTransportTest.java
```diff
@@ -81,6 +84,11 @@ public boolean isStartTLSSupported() {
                 throw new UnsupportedOperationException();
             }
             
+            @Override
+            public Optional<SSLSession> getSSLSession() {
+                throw new UnsupportedOperationException();
```
chibenwa, line 89: Why not return Optional.empty ?

chibenwa (commented):
Similar changes in ImapSession could be nice.

protocols/api/src/test/java/org/apache/james/protocols/api/AbstractProtocolTransportTest.java
ottoka: I am following the pattern here, since the method is not used in the tests. Throwing "unsupported" is different from returning Optional.empty() which would make it appear as a regular non-SSL transport.

ottoka:
Latest push includes ImapSession.

chibenwa (approved):

Arsnael (approved):

PR #1278, opened 2022-10-27
JAMES-3843 VacationMailet should use null sender

ottoka:
When the VacationMailet sends a vacation message, it uses the original recipient as envelope sender. However, RFC 5230 states
"5.1. SMTP MAIL FROM Address
The SMTP MAIL FROM address of the message envelope SHOULD be set to <>."
While we are technically OK because of the SHOULD, I do not see any real requirement to deviate from the recommendation. In fact, using the null sender can help prevent mail loops. So I propose we change the VacationMailet accordingly.

### DIFF
--- a/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/VacationMailet.java
+++ b/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/VacationMailet.java
@@ -46,12 +46,15 @@ import reactor.core.publisher.Mono;
 
 public class VacationMailet extends GenericMailet {
 
+    public static final String EXPLICIT_SENDER_PROPERTY = "james.vacation.sender.explicit";
+
     private static final Logger LOGGER = LoggerFactory.getLogger(VacationMailet.class);
 
     private final VacationService vacationService;
     private final ZonedDateTimeProvider zonedDateTimeProvider;
     private final AutomaticallySentMailDetector automaticallySentMailDetector;
     private final MimeMessageBodyGenerator mimeMessageBodyGenerator;
+    private final boolean explicitSender;
 
     @Inject
     public VacationMailet(VacationService vacationService, ZonedDateTimeProvider zonedDateTimeProvider,
@@ -61,6 +64,7 @@ public class VacationMailet extends GenericMailet {
         this.zonedDateTimeProvider = zonedDateTimeProvider;
         this.automaticallySentMailDetector = automaticallySentMailDetector;
         this.mimeMessageBodyGenerator = mimeMessageBodyGenerator;
+        this.explicitSender = Boolean.parseBoolean(System.getProperty(EXPLICIT_SENDER_PROPERTY, "false"));
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


chibenwa (approved):
Good idea!

Arsnael (approved):

Arsnael:
```
org.apache.james.jmap.memory.MemoryVacationRelayIntegrationTest.forwardingAnEmailShouldWork

Error Message

Assertion condition defined as a lambda expression in org.apache.james.jmap.VacationRelayIntegrationTest that uses java.lang.String null within 1 minutes.

Stacktrace

org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a lambda expression in org.apache.james.jmap.VacationRelayIntegrationTest that uses java.lang.String null within 1 minutes.
Caused by: java.util.concurrent.TimeoutException
```
and
```
org.apache.james.jmap.rabbitmq.RabbitMQAwsS3VacationRelayIntegrationTest.forwardingAnEmailShouldWork

Error Message

Assertion condition defined as a lambda expression in org.apache.james.jmap.VacationRelayIntegrationTest that uses java.lang.String null within 1 minutes.

Stacktrace

org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a lambda expression in org.apache.james.jmap.VacationRelayIntegrationTest that uses java.lang.String null within 1 minutes.
Caused by: java.util.concurrent.TimeoutException
```
Possibly related?

ottoka:
Seems I missed the integration tests, I will take a look.

ottoka:
The integration tests fail since the FakeSMTP server does not handle an empty sender correctly:
`MAIL FROM:<>`
`250 2.1.0 Ok`
`RCPT TO:<ray@yopmail.com>`
`503 5.5.1 Error: need MAIL command`
Digging further I find https://github.com/andris9/simplesmtp/blob/master/lib/server.js#L403 , which we hit since since Javascript treats undefined/null/empty string all as false. Not sure if this is a feature or a bug in simplesmtp.

In either case, I don't have the time to fix a third party project. I could
1. do nothing, drop this PR, close the Jira ticket as won't fix. After all the RFC only has a SHOULD requirement here. - _sad_
2. disable the vacation integration tests - _bad_
3. make the sender conditional on a system property or mailet configuration, e.g. "explicitSender=true" would send the address as before, "explicitSender=false" would use the null sender. Default false. - _annoying_

Any suggestions?

chibenwa:
If going for the annoying a comment next to the option would be critical

chibenwa:
Other options: frop FakeSmtp in this test in favor of https://github.com/apache/james-project/tree/master/server/mailet/mock-smtp-server

CF https://github.com/apache/james-project/tree/master/server/mailet/remote-delivery-integration-testing

ottoka:
FakeSmtp is used in quite a lot of tests, I don't really want to touch it. So I went with the annoying option instead, using a boolean system property "james.vacation.sender.explicit" (default false). Its main purpose is as a workaround for the integration tests, I don't see any real word use case. Should I still add documentation and/or sample configuration for it?

chibenwa:
> FakeSmtp is used in quite a lot of tests, I don't really want to touch it. So I went with the annoying option instead, using a boolean system property "james.vacation.sender.explicit" (default false). Its main purpose is as a workaround for the integration tests, I don't see any real word use case. Should I still add documentation and/or sample configuration for it?

Let me have a try at the mock server today...

chibenwa:
https://github.com/apache/james-project/pull/1287 as promised.

ottoka:
Nice work! Let's close this PR and continue the topic with https://github.com/apache/james-project/pull/1287.

PR #1287, opened 2022-11-01
JAMES-3843 VacationMailet should use null sender

chibenwa:
https://github.com/apache/james-project/pull/1278 without the custom config hack to turn off the nullSender, and without turning off the integration tests.

CC @ottoka


Arsnael (approved):

ottoka:
Nice work, much better solution than my annoying workaround! I will close https://github.com/apache/james-project/pull/1278 and let you take over with this one.

ottoka (approved):

PR #1301, opened 2022-11-09
JAMES-3788 Fix PROXY protocol support with TLS

chibenwa:
Given James is responsible of TLS termination, PROXY protocol should be handled first, before SSL.

Add examples for TLS + PROXY support.

Cc @ouvtam


Arsnael (approved):

ouvtam:
Looks good to me.

Any issues with STARTTLS since the example only covers TLS?

chibenwa:
> Any issues with STARTTLS since the example only covers TLS?

No because connection is already establish when SSL negotiations begins.

PR #1302, opened 2022-11-10
JAMES-3830 Do not limit Quota/get to user scope

chibenwa:


chibenwa:
CF bug reported by Manh in https://github.com/linagora/tmail-flutter/issues/33

Arsnael (approved):

PR #1304, opened 2022-11-10
JAMES-3851 TLS host name verification should handle trailing dot

ottoka:
I noticed that sometimes RemoteDelivery opens a connection using a fully qualified hostname that ends with a trailing dot, like "mail.example.org." I believe James may get that from MX resolving, since afaik DNS servers may do this to indicate an absolute FQDN vs. a relative one. This is not an issue when establishing a connection, but will break TLS hostname verification, since the CN and SubjectAltNames in server certificates never use trailing dots.

Consequently, RemoteDelivery should strip a trailing dot from the hostname before connecting.

Note that this is the minimal fix for the verification issue; log messages related to remote delivery will still show the hostname with a trailing dot. I am not sure if this is intended behavior, i.e. to see exactly what James got from DNS resolution. Otherwise, an alternative would be to fix this early in MXHostAddressIterator. Let me know if you prefer the latter.


chibenwa:
Looks good to me.

chibenwa (approved):

Arsnael (approved):

PR #1305, opened 2022-11-11
[FIX] Refresh SpamAssassin testing DockerFile

chibenwa:


chibenwa:
Merged because it turned the build into a nightmare.

I'm happy to fix any remarks may it be needed.

PR #1357, opened 2022-12-12
[FIX] Avoid clearing the schemaVersion

chibenwa:
Apart from the first test that did create the table, follow up tests have no way to know the actual version ad thus choose the lowest. Avoiding clearing this table avoids this.


vttranlina (commented):
read it

PR #1364, opened 2022-12-28
JAMES-3870 Group IMAP response line within TCP packets

chibenwa:
Today each Imap response line is transmitted in a
distinct TCP packet.

This causes a lot of network overhead as
=> TCP headers are added for each response line.
   To give an idea a LIST response line is 35 bytes
   long but result in a 101 bytes TCP frame so a 188%
   overcost....
=> TCP ack are conducted for each line independently.
   An ACK is 66 bytes

This is especially problematic for LIST, FETCH commands that actually result in many (100, 1000, maybe millions) response lines.

master-list-responses-with-flush

We should try to limit the calls to "flush" with Netty, and force the flush only once per IMAP command (at the end of processing). Netty is free to transmit some data earlier if it's buffer states requires it.

The entire mailbox list take 1 packet to transmit (1420 bytes total for 41 mailboxes so ~35 bytes per mailbox) and a single ACK (66 bytes).

pr-list-responses-without-flush

### DIFF
--- a/protocols/imap/src/main/java/org/apache/james/imap/api/process/ImapSession.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/api/process/ImapSession.java
@@ -263,4 +263,8 @@ public interface ImapSession extends CommandDetectionSession {
     }
 
     void schedule(Runnable runnable, Duration waitDelay);
+
+    default void flush() {
+
+    }
 }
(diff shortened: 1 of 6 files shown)


vttranlina (approved):

chibenwa:
Performance tests: 

 - Gatling performances are equivalent
 - Flame graph show a 100% performance enhancement of SSL wrapping and TCP socket flushing for IMAP. This result in a 2-3% CPU enhancement overall. Wrapping and sending the same payload is of course faster if performed on bigger TCP frames...
 
 The performance impact hereby mesured is tiny, but consistent with the above exposed facts.

PR #1377, opened 2023-01-06
[UPGRADE] January 2023 dependency upgrades

chibenwa:


vttranlina (approved):

quantranhong1999:
```java
java.lang.IllegalStateException: block()/blockFirst()/blockLast() are blocking, which is not supported in thread wait-2
```
seems related

PR #1381, opened 2023-01-09
[PERF] Handle range in UnseenSearchOverride

chibenwa:


chibenwa:
Perf tests shows a 3x speedup despite onging degraded performance on OVH side

(mean 49ms p99 236 ms while before this was mean 286 ms p99 742 ms)

quantranhong1999 (approved):

vttranlina (approved):

PR #1388, opened 2023-01-11
JAMES-3756 SessionTranslator should only ask 1 delegation source

quantranhong1999:
Instead of two sources with the same truth (https://github.com/linagora/james-project/issues/4673).
(`sessionProvider.authenticate(session.getUser).as(targetUser)` called the 2nd source `authorized users`).

### DIFF
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/SessionProvider.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/SessionProvider.java
@@ -110,6 +110,8 @@ public interface SessionProvider {
         return otherUserId -> loginAsOtherUser(givenUserid, otherUserId);
     }
 
+    MailboxSession loginAsOtherUserWithoutCheckingDelegation(Username givenUserid, Username otherUserId);
+
     /**
      * <p>
      * Logs the session out, freeing any resources. Clients who open session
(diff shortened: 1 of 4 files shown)


chibenwa:
I have two major issues wit this design:

 - 1. the naming suck (): too long, not precise
  - 2. with this change SessionProvider exposes the APIs to bypass its own role pushing validation logics to the applicative layers, which is unsafe and breaks encapsulation.
  - 3. Too much methods does (almost) the same things.
  
3 big code smells ...
  
Especially on authorization I want to be able to review code and state "that's fine, it only relies on the session provider, so I know authorization logic won't be messed up". With this change I cannot make such assumptions.

As an alternative I would try to refine the SessionOrovider API  to make it more composable and expressive while keeping it concise and safe.

Please see https://github.com/chibenwa/james-project/tree/poc-refacto-delegation as a possible alternative.

chibenwa:
TODO rebase https://github.com/chibenwa/james-project/tree/poc-refacto-delegation

(I will do)

quantranhong1999:
> (I will do)

I can do it, no worry.

chibenwa:
Ok, just I would need it for some other topics.

quantranhong1999:
> Ok, just I would need it for some other topics.

For you: https://github.com/quantranhong1999/james-project/tree/rebase-poc-session-translator

PR #1390, opened 2023-01-12
[REFACTOING] Improve SessionProvider API

chibenwa:
- Decrease method cardinality for login / delegation methods
 - Remove uneeded methods 
    - getPathDelimiter() - can be carried by the MailboxSession instead. No need to enforce the delimiter choice for a whole SessionProvider impl.
    - logout: isOpen is never used, this bit of logic can be strip away.


quantranhong1999 (approved):
`SessionProvider` APIs look more beautiful and clear to me <3

vttranlina (approved):

PR #1392, opened 2023-01-16
JAMES-3867 Allow IMAP extensions configuration

chibenwa:
One can define custom, per-IMAP server
configuration values.


vttranlina (approved):

quantranhong1999 (approved):
Could we add a bit doc in `config-imap4.xml`,... ?

vttranlina (approved):

PR #1400, opened 2023-01-18
[boyscout] logs bound port when webmin server starts

jeantil:
Logging bound port at startup helps make sure the ops setup is correct. The idea was lifted from `AbstractConfigurableAsyncServer` which is used for the SMTP server.


chibenwa (approved):
<3

chibenwa:
Yes it looks like it...

Arsnael (approved):

jeantil:
build was green

PR #1402, opened 2023-01-19
JAMES-3875 JPAMailRepository fix

AdBuch:
This fixes the JAMES-3875, cur of to long error messages who lead to exception, and replacing setObject(null) with type to avoid type missmatch


chibenwa (approved):

chibenwa:
I think you can also open this PR against 3.7.x if you want to benefit from a potential 3.7.4.

PR #1405, opened 2023-01-19
JAMES-3756 DelegationStoreAuthorizator should not fail on bad admin v…

chibenwa:
…irtualHosting

It is preferable to fallback to forbidden.


vttranlina (commented):
good catch

vttranlina (approved):

Arsnael (approved):

PR #1406, opened 2023-01-20
JAMES-3878 - Remove icu4j dependency

woj-tek:
Remove icu4j dependency and replace required functionality with other dependencies

https://issues.apache.org/jira/projects/JAMES/issues/JAMES-3878

### DIFF
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/model/MailboxAnnotationKey.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/model/MailboxAnnotationKey.java
@@ -25,21 +25,18 @@ import org.apache.commons.lang3.StringUtils;
 
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
-import com.ibm.icu.text.UnicodeSet;
+import com.google.re2j.Pattern;
 
 public class MailboxAnnotationKey {
 
     public static final String SLASH_CHARACTER = "/";
     public static final String TWO_SLASH_CHARACTER = "//";
 
-    private static final UnicodeSet NAME_ANNOTATION_PATTERN = new UnicodeSet("[[a-z][A-Z][0-9]]")
-            .add(SLASH_CHARACTER)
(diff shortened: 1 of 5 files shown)


mailbox/api/src/main/java/org/apache/james/mailbox/model/MailboxAnnotationKey.java
```diff
@@ -25,21 +25,18 @@
...
 
 public class MailboxAnnotationKey {
 
     public static final String SLASH_CHARACTER = "/";
     public static final String TWO_SLASH_CHARACTER = "//";
 
-    private static final UnicodeSet NAME_ANNOTATION_PATTERN = new UnicodeSet("[[a-z][A-Z][0-9]]")
-            .add(SLASH_CHARACTER)
-            .freeze();
+    private static final Pattern NAME_ANNOTATION_PATTERN = Pattern.compile("^([\\w\\d/]*)$");
```
chibenwa, line 35: My preference goes to relying on Guava CharMatcher and avoid REGEX alltogether if that works.

chibenwa: For instance:

```
    private static final CharMatcher NAME_ANNOTATION_PATTERN = CharMatcher.inRange('a', 'z')
        .or(CharMatcher.inRange('A', 'Z'))
        .or(CharMatcher.inRange('0', '9'))
        .or(CharMatcher.is('/'));
```

Passes the unit tests

Arsnael (commented):
Read it, nothing to add :)

mailbox/api/src/main/java/org/apache/james/mailbox/model/MailboxAnnotationKey.java
woj-tek: Switched to CharMatcher

chibenwa (approved):

PR #1409, opened 2023-01-26
Adds minimal pulsar authentication support to pulsar configuration

jeantil:
The initial implementation only supported no auth, which doesn't work well in a prod environment. 

This PR adds support for 
- explicit no auth
- basic user/password auth
- token auth

### DIFF
--- a/backends-common/pulsar/src/main/scala/org/apache/james/backends/pulsar/PulsarClients.scala
+++ b/backends-common/pulsar/src/main/scala/org/apache/james/backends/pulsar/PulsarClients.scala
@@ -1,0 +1,68 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.backends.pulsar
+
+import com.sksamuel.pulsar4s.{PulsarAsyncClient, PulsarClient, PulsarClientConfig}
+import org.apache.commons.io.IOUtils
+import org.apache.pulsar.client.admin.PulsarAdmin
+import org.apache.pulsar.client.impl.auth.{AuthenticationBasic, AuthenticationDisabled, AuthenticationToken}
+
+import java.io.Closeable
+import javax.annotation.PreDestroy
+import scala.jdk.CollectionConverters.MapHasAsJava
+
+object PulsarClients {
+  def create(configuration: PulsarConfiguration): PulsarClients = {
+    val pulsarAuth = configuration.auth match {
+      case Auth.NoAuth => new AuthenticationDisabled()
+      case Auth.Token(value) => new AuthenticationToken(value)
+      case Auth.Basic(userId, password) =>
+        val basic = new AuthenticationBasic()
+        basic.configure(Map("userId" -> userId, "password" -> password).asJava)
+        basic
+    }
+
+    val adminClient: PulsarAdmin =
+      PulsarAdmin.builder()
+        .serviceHttpUrl(configuration.adminUri)
+        .authentication(pulsarAuth)
+        .build()
+
+    val asyncClient: PulsarAsyncClient = {
+      PulsarClient(
+        PulsarClientConfig(
+          serviceUrl = configuration.brokerUri,
+          authentication = Some(pulsarAuth)
+        )
+      )
+    }
+
+    new PulsarClients(adminClient, asyncClient)
+  }
+}
+
+
+class PulsarClients(val adminClient: PulsarAdmin, val asyncClient: PulsarAsyncClient) {
+  @PreDestroy
+  def stop(): Unit = {
+    val closeableAsyncClient: Closeable = () => asyncClient.close()
+    IOUtils.closeQuietly(adminClient, closeableAsyncClient)
+  }
+}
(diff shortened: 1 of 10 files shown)


backends-common/pulsar/src/main/scala/org/apache/james/backends/pulsar/PulsarConfiguration.scala
```diff
@@ -1,42 +1,76 @@
-/****************************************************************
+/** **************************************************************
```
chibenwa, line 1: Let's avoid changes in the license formatting

server/queue/queue-pulsar/src/main/scala/org/apache/james/queue/pulsar/PulsarMailQueue.scala
```diff
@@ -110,11 +108,8 @@ class PulsarMailQueue(
   private implicit val implicitSystem: ActorSystem = system
   private implicit val ec: ExecutionContextExecutor = system.dispatcher
   private implicit val implicitBlobIdFactory: BlobId.Factory = blobIdFactory
-  private implicit val client: PulsarAsyncClient = PulsarClient(config.pulsar.brokerUri)
-  private val admin = {
-    val builder = PulsarAdmin.builder()
-    builder.serviceHttpUrl(config.pulsar.adminUri).build()
-  }
+  private implicit val client: PulsarAsyncClient = config.pulsar.asyncClient()
+  private val admin = config.pulsar.adminClient()
```
chibenwa, line 113: Shouldn't the pulsar client be a global entity rather than a "per queue" entity?

 -> Sounds like expensive resources to maintain
 -> the chances of the leaks increases

My opinion is that it would better fit as a backend-pulsar component injected down the stack.

chibenwa (commented):
Otherwise looks great to me.

backends-common/pulsar/src/test/java/org/apache/james/backends/pulsar/PulsarConfigurationTest.java
```diff
@@ -155,8 +158,106 @@ void fromShouldReturnTheConfigurationWhenRequiredParametersAreGiven() {
...
+
+        String namespace = "namespace";
+
+        configuration.addProperty("namespace", namespace);
+        assertThatThrownBy(() -> PulsarConfiguration.from(configuration))
+                .isInstanceOf(IllegalStateException.class)
+                .hasMessage("You need to specify a non-empty value for " + PulsarConfiguration.AUTHENTICATION_TOKEN_PROPERTY_NAME());
+    }
+    @Test
+    void fromShouldReturnTheConfigurationWithTokenAuthenticationWhenRequiredParametersAreGiven() {
```
Arsnael, line 182: Confused with the name of the test, as it seems it's basic auth and not token auth that is being tested here 

Also a blank line between each test would help visibility :)

backends-common/pulsar/src/test/java/org/apache/james/backends/pulsar/PulsarConfigurationTest.java
```diff
@@ -155,8 +158,106 @@ void fromShouldReturnTheConfigurationWhenRequiredParametersAreGiven() {
 
         configuration.addProperty("namespace", namespace);
         assertThat(PulsarConfiguration.from(configuration))
-                .isEqualTo(new PulsarConfiguration(brokerUri, adminUri, new Namespace(namespace)));
+                .isEqualTo(new PulsarConfiguration(brokerUri, adminUri, new Namespace(namespace), Auth.noAuth()));
     }
+    @Test
+    void fromShouldThrowWithTokenAuthenticationWhenTokenIsMissing() {
```
Arsnael, line 164: I don't see the test that it returns successfully the conf if token auth params are valid

server/queue/queue-pulsar/src/main/scala/org/apache/james/queue/pulsar/PulsarMailQueue.scala
jeantil: We will have to check all the implications but this sounds reasonable to me. Would you accept a `lazy val` on `pulsarConfiguration` to replace the `def` for `asyncClient` and `adminClient` respectivey ?

chibenwa: Shouldnt a configuration be a dumb pojo?

jeantil: we reworked the code to match your suggestion

backends-common/pulsar/src/main/scala/org/apache/james/backends/pulsar/PulsarClients.scala
```diff
@@ -0,0 +1,49 @@
+package org.apache.james.backends.pulsar
```
chibenwa, line 1: Missing license

server/container/guice/queue/pulsar/src/main/java/org/apache/james/queue/pulsar/module/PulsarQueueModule.java
```diff
@@ -67,10 +68,16 @@ private Configuration rawConfiguration(PropertiesProvider propertiesProvider) th
     private PulsarConfiguration pulsarConfiguration(@Named(PULSAR_CONFIGURATION_NAME) Configuration configuration) {
         return PulsarConfiguration.from(configuration);
     }
+    @Provides
```
chibenwa, line 71:
```suggestion
    }
    
    @Provides
```

chibenwa (approved):

backends-common/pulsar/src/test/java/org/apache/james/backends/pulsar/PulsarConfigurationTest.java
jeantil: I updated the tests, added the blank lines and added the missing tests

backends-common/pulsar/src/test/java/org/apache/james/backends/pulsar/PulsarConfigurationTest.java
jeantil: fixedup

backends-common/pulsar/src/main/scala/org/apache/james/backends/pulsar/PulsarClients.scala
jeantil: fixedup

backends-common/pulsar/src/main/scala/org/apache/james/backends/pulsar/PulsarClients.scala
```diff
@@ -1,3 +1,22 @@
...
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ * *
+ * http://www.apache.org/licenses/LICENSE-2.0                 *
+ * *
```
Arsnael, line 11: Just need to fix quickly the format of the license, sorry :)

jeantil: fixed

PR #1416, opened 2023-01-30
[PERF] Use UnsynchronizedBufferedInputStream

chibenwa:
Synchronized methods becomes ore expensive with higher JRE versions (loom) making calling
`BufferedInputStream::read()` method for each byte prohibitive.

See https://issues.apache.org/jira/projects/IO/issues/IO-786

Screenshot from 2023-01-30 09-06-03


chibenwa:
> EDIT: and are we sure this does not make some tests unstable? I see the first two runs failing on the same test class MemoryTaskManagerTest

Not sure how it relates :-p

quantranhong1999 (approved):

PR #1420, opened 2023-02-02
Fix a leak in TextExtractor

mbaechler:
I propose a new contract to make sure `TextExtractor` implementations close the provided input streams.

I only made it to work with Default implementation but I'm sure it can be extended to other implementations.


mailbox/api/src/test/java/org/apache/james/mailbox/extractor/TextExtractorContract.java
```diff
@@ -0,0 +1,59 @@
+package org.apache.james.mailbox.extractor;
```
chibenwa, line 1: Missing license

chibenwa (approved):
Good initiative!

Arsnael (commented):
Read it nothing to add, cheers for that :)

Arsnael (approved):

quantranhong1999 (approved):

PR #1424, opened 2023-02-06
[FIX] JMAP urn:ietf:params:jmap:submission submissionExtensions shoul…

chibenwa:
…d be a map

Default JSON formatting renders the map as an iterable, so as a JSON collection instead of the desired JSON object.


vttranlina (approved):

Arsnael (approved):

Arsnael:
```
org.apache.james.jmap.rfc8621.memory.MemoryCustomMethodTest.getShouldReturnCorrectSession


JSON documents are different:
Different value found in node "accounts.29883977c13473ae7cb7678ef767cbfbaffc8a44a6e463d971d23a65c1dc4af6.accountCapabilities.urn:ietf:params:jmap:submission.submissionExtensions", expected: <[]> but was: <{}>.
Different value found in node "capabilities.urn:ietf:params:jmap:submission.submissionExtensions", expected: <[]> but was: <{}>.
```
Related

quantranhong1999 (approved):

PR #1432, opened 2023-02-10
JAMES-3870 Prevent partial flush in IMAP

chibenwa:
## Actual behaviour

Screenshot from 2023-02-09 22-29-33

```
Error fetching folders: status: expecting ')' or attribute
```

Here are IMAP debug logs:

```
* STATUS "Test email box" (MESSAGES'
[imapx:A] I/O: ' 0 UIDNEXT 1 UIDVALIDITY 748930662 UNSEEN 0)
```

Notice the IMAP STATUS response being split into 2 IMAP messages.

## Reason

Evolution do not support an IMAP message to be split into 2 TCP frames.

Avoid pushing partial TCP frames.

Suspect: https://github.com/apache/james-project/pull/1364

## Resolution test

 - [ ] QA to validate #1364 impact: revert and test live...
 - [ ] Find a way to disable partial sends with the netty stack
 - [ ] Write a test reproducing the partial send behaviour

### DIFF
--- a/server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IMAPServerTest.java
+++ b/server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IMAPServerTest.java
@@ -977,7 +977,7 @@ class IMAPServerTest {
                 .doesNotThrowAnyException();
         }
 
-        @Test
+        @RepeatedTest(100)
         void authenticatePlainShouldSucceed() {
             assertThatCode(() ->
                 testIMAPClient.connect("127.0.0.1", port)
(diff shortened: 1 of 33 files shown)


Arsnael (approved):

server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IMAPServerTest.java
```diff
@@ -977,7 +977,7 @@ void loginShouldSucceed() {
                 .doesNotThrowAnyException();
         }
 
-        @Test
+        @RepeatedTest(100)
```
Arsnael, line 980: debug?

Arsnael:
```
logoutCommandShouldWork{GuiceJamesServer} – org.apache.james.IMAPIntegrationTest

Error
Connection closed without indication.
Stacktrace
java.io.EOFException: Connection closed without indication.
	at org.apache.james.IMAPIntegrationTest.logoutCommandShouldWork(IMAPIntegrationTest.java:67)
```

server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IMAPServerTest.java
chibenwa: No that's intended to catch the race condition

PR #1436, opened 2023-02-10
JAMES-3882 Move messages into DeletedMessageVault asynchronously for …

chibenwa:
…Cassandra


Arsnael (approved):

server/container/guice/cassandra/src/main/java/org/apache/james/modules/mailbox/CassandraMailboxModule.java
```diff
@@ -229,6 +229,7 @@ protected void configure() {
             .addBinding().to(MailboxAnnotationListener.class);
         Multibinder.newSetBinder(binder(), EventListener.ReactiveGroupEventListener.class)
             .addBinding().to(DeleteMessageListener.class);
+        Multibinder.newSetBinder(binder(), DeleteMessageListener.DeletionCallback.class);
```
quantranhong1999, line 232: Q: what is implementation this binded to? Maybe this is duplicated with the binding declared in `CassandraDeletedMessageVaultModule` ?

chibenwa: None. It just creates an empty set for injections. Without this injections fails without the deletedMessageVault module.

quantranhong1999 (approved):

PR #1438, opened 2023-02-13
JAMES-3867 Missing ACLResponseEncoder

chibenwa:


Arsnael (approved):

quantranhong1999 (approved):
I wonder if we should have a smarter way to list those classes automatically to avoid manual mistakes.
e.g: scan all `ImapProcessor` classes in package `org.apache.james.imap.processor`
scan all `ImapCommandParser` classes in `org.apache.james.imap.decode.parser`
scan all `ImapResponseEncoder` classes in `org.apache.james.imap.encode`.

PR #1440, opened 2023-02-13
[Boyscout] Some tiny improvements

quantranhong1999:


protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/esmtp/AuthCmdHandler.java
```diff
@@ -356,7 +356,7 @@ protected Response doDelegation(SMTPSession session, Username username) {
                 if (SMTPRetCode.AUTH_FAILED.equals(res.getRetCode())) {
                     LOGGER.warn("{} was not authorized to connect as {}", session.getUsername(), username);
                 } else if (SMTPRetCode.AUTH_OK.equals(res.getRetCode())) {
-                    LOGGER.info("{} was authorized to connect as {}", session.getUsername(), username);
+                    LOGGER.warn("{} was authorized to connect as {}", session.getUsername(), username);
```
chibenwa, line 359: Why? It's a success, no problem?

chibenwa (approved):

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/esmtp/AuthCmdHandler.java
```diff
@@ -382,7 +382,7 @@ protected Response doAuthTest(SMTPSession session, Username username, String pas
 
                 if (res != null) {
                     if (SMTPRetCode.AUTH_FAILED.equals(res.getRetCode())) {
-                        LOGGER.info("AUTH method {} failed", authType);
+                        LOGGER.warn("AUTH method {} failed", authType);
                     } else if (SMTPRetCode.AUTH_OK.equals(res.getRetCode())) {
```
jeantil, line 386: I'm curious, why do you think this is worth raising to `warn`? 
At my current company, warns and errors result in sentry notifications to the corresponding engineering team, we tend to avoid notifications for client errors.

quantranhong1999: If a client's action is harmful, does it deserve a warning?

Say someone's account is being brute-force attack.

chibenwa: I tend to agree with Jean.

Maybe we can discriminate the by having a logger dedicated to failed logging attemps? EG

`org.apache.james.protocols.smtp.AUTHENTICATION` ?

 -> We could put success logs at debug
 -> INFO logs for failures
 
 Then this could easily be instrumented if needed...

jeantil: > If a client's action is harmful, does it deserve a warning?
> Say someone's account is being brute-force attack.

What we do at my current employer is that we still log it as info, but we create monitors in our observability tool to get alerts when unusual patterns occur. This way we are only notified from actual attacks and not from casual user errors.

using a dedicated logger would indeed help to create the monitor more easily

chibenwa (approved):

Arsnael (approved):

jeantil:
***@***.*** *approved this pull request. ;)

Le mar. 14 févr. 2023 à 08:04, Rene Cordier ***@***.***> a
écrit :

chibenwa:
Screenshot from 2023-02-23 15-10-14

jeantil (approved):

tungtv202 (approved):

PR #1443, opened 2023-02-16
JAMES-3885 Changing username

chibenwa:
This is a POC on defining a pluggable mecanism to change of username (divorce, wedding)

## TODO

 - [x] Unit tests for `ForwardUsernameChangeTaskStep` (memory based)
 - [x] Integration tests: route exposed, task accepted and executed, forward from old to new user created, old user forwards migrated.
 - [x] Documentation
 - [x] Add an option "From stage" to resume failed username changes...

### DIFF
--- a/server/data/data-api/src/main/java/org/apache/james/rrt/ForwardUsernameChangeTaskStep.java
+++ b/server/data/data-api/src/main/java/org/apache/james/rrt/ForwardUsernameChangeTaskStep.java
@@ -1,0 +1,86 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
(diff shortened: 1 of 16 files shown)


server/data/data-api/src/main/java/org/apache/james/rrt/ForwardUsernameChangeTaskStep.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Inject
+    public ForwardUsernameChangeTaskStep(RecipientRewriteTable rrt) {
+        this.rrt = rrt;
+    }
+
+    @Override
+    public Publisher<Void> changeUsername(Username oldUsername, Username newUsername) {
+        MappingSource oldSource = MappingSource.fromUser(oldUsername);
+         MappingSource newSource = MappingSource.fromUser(newUsername);
```
Arsnael, line 48: indent

server/protocols/webadmin-integration-test/memory-webadmin-integration-test/src/test/java/org/apache/james/webadmin/integration/memory/MemoryUsernameChangeIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.james.webadmin.WebAdminUtils;
+import org.apache.james.webadmin.routes.TasksRoutes;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import io.restassured.RestAssured;
+import io.restassured.specification.RequestSpecification;
+
+class MemoryUsernameChangeIntegrationTest {
```
Arsnael, line 58: Only memory integration tests? Is that enough?

chibenwa: I believe it is :-)

Arsnael (approved):

PR #1447, opened 2023-02-20
JAMES-3440 Email/query (EmailQueryView usage case) should filter share mailboxes

quantranhong1999:


quantranhong1999:
> We should open a similar fix on JMAP draft if needed. Can you check?

Sure, I will check and open another PR id needed.

chibenwa:
> Sure, I will check and open another PR if needed.

Let's create an issue on this right now, so as not to forget, nheeee

PR #1453, opened 2023-02-21
JAMES-2601 Implement a LDAP healthCheck

HoussemNasri:
As the title says, added a new health checker to check LDAP server's health and tested it.

Note: checkShouldReturnHealthyIfLdapIsRunning failed when I ran it locally. Apparently, there is an exception thrown by the ldap user repository that says "Unable check user existence from ldap".

I'm sure that the test container has started because I can view it from Docker Desktop, but for some reason, the exception is still thrown.

image

### DIFF
--- a/server/container/guice/data-ldap/src/main/java/org/apache/james/data/LdapUsersRepositoryModule.java
+++ b/server/container/guice/data-ldap/src/main/java/org/apache/james/data/LdapUsersRepositoryModule.java
@@ -45,7 +45,7 @@ public class LdapUsersRepositoryModule extends AbstractModule {
         bind(UsersRepository.class).to(ReadOnlyUsersLDAPRepository.class);
         bind(DelegationStore.class).to(NaiveDelegationStore.class);
         bind(Authorizator.class).to(UserRepositoryAuthorizator.class);
-        bind(HealthCheck.class).to(LdapHealthCheck.class);
+        Multibinder.newSetBinder(binder(), HealthCheck.class).addBinding().to(LdapHealthCheck.class);
     }
 
     @Provides
--- a/server/container/guice/data-ldap/src/main/java/org/apache/james/data/LdapUsersRepositoryModule.java
+++ b/server/container/guice/data-ldap/src/main/java/org/apache/james/data/LdapUsersRepositoryModule.java
@@ -18,7 +18,6 @@
  ****************************************************************/
 package org.apache.james.data;
 
-import com.google.inject.multibindings.Multibinder;
 import org.apache.commons.configuration2.ex.ConfigurationException;
 import org.apache.james.adapter.mailbox.UserRepositoryAuthorizator;
 import org.apache.james.core.healthcheck.HealthCheck;
@@ -37,6 +36,7 @@ import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
 import com.google.inject.Scopes;
 import com.google.inject.Singleton;
+import com.google.inject.multibindings.Multibinder;
 import com.google.inject.multibindings.ProvidesIntoSet;
 
 public class LdapUsersRepositoryModule extends AbstractModule {
--- a/server/container/guice/data-ldap/src/main/java/org/apache/james/data/LdapUsersRepositoryModule.java
+++ b/server/container/guice/data-ldap/src/main/java/org/apache/james/data/LdapUsersRepositoryModule.java
@@ -18,6 +18,7 @@
  ****************************************************************/
 package org.apache.james.data;
 
+import com.google.inject.multibindings.Multibinder;
 import org.apache.commons.configuration2.ex.ConfigurationException;
 import org.apache.james.adapter.mailbox.UserRepositoryAuthorizator;
 import org.apache.james.core.healthcheck.HealthCheck;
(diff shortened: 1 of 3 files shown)


server/data/data-ldap/src/main/java/org/apache/james/user/ldap/LdapHealthCheck.java
```diff
@@ -0,0 +1,38 @@
+package org.apache.james.user.ldap;
```
chibenwa, line 1: You are missing the compulsary ASF V2 license header

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/LdapHealthCheck.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    @Override
+    public ComponentName componentName() {
+        return COMPONENT_NAME;
+    }
+
+    @Override
+    public Mono<Result> check() {
+        return Mono.fromCallable(() -> ldapUserRepository.getUserByName(LDAP_TEST_USER))
```
chibenwa, line 33: Please handle the case "user not found" and return healthy in this case: we should not require the ldat-test user to be preprovisionned.

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/LdapHealthCheck.java
```diff
@@ -0,0 +1,38 @@
...
+            .map(user -> Result.healthy(COMPONENT_NAME))
+            .onErrorResume(e -> Mono.just(Result.unhealthy(COMPONENT_NAME, "Error checking LDAP server!", e)))
+            .doOnError(e -> LOGGER.error("Error in LDAP server", e));
```
chibenwa, line 36: LDAP connections are blocking, we should schedule them on the `bundedElastic` scheduler....

server/data/data-ldap/src/test/java/org/apache/james/user/ldap/LdapHealthCheckTest.java
```diff
@@ -0,0 +1,41 @@
+package org.apache.james.user.ldap;
```
chibenwa, line 1: Missing license

server/container/guice/data-ldap/src/main/java/org/apache/james/data/LdapUsersRepositoryModule.java
```diff
@@ -43,6 +45,7 @@ public void configure() {
         bind(UsersRepository.class).to(ReadOnlyUsersLDAPRepository.class);
         bind(DelegationStore.class).to(NaiveDelegationStore.class);
         bind(Authorizator.class).to(UserRepositoryAuthorizator.class);
+        bind(HealthCheck.class).to(LdapHealthCheck.class);
```
chibenwa, line 48: No this needs to be done through the multi-binder:

```suggestion
        Multibinder.newSetBinder(binder(), HealthCheck.class).addBinding().to(LdapHealthCheck.class);
```

server/data/data-ldap/src/test/java/org/apache/james/user/ldap/LdapHealthCheckTest.java
```diff
@@ -0,0 +1,41 @@
...
+        ldapContainer = DockerLdapSingleton.ldapContainer;
+    }
+
+    @Test
+    void checkShouldReturnUnhealthyIfLdapIsDown() {
+        ldapContainer.pause();
+
+        Result checkResult = ldapHealthCheck.check().block();
+        assertNotNull(checkResult);
+        assertTrue(checkResult.isUnHealthy());
```
chibenwa, line 30: Use assertj library:

```suggestion
        assertThat(checkResult.isUnHealthy()).isTrue();
```

This is a common project practice and it tends to be more readable than Junit assertions.

server/data/data-ldap/src/test/java/org/apache/james/user/ldap/LdapHealthCheckTest.java
```diff
@@ -0,0 +1,41 @@
...
+
+        ldapContainer.unpause();
+    }
+
+    @Test
+    void checkShouldReturnHealthyIfLdapIsRunning() {
+        Result checkResult = ldapHealthCheck.check().block();
+        assertNotNull(checkResult);
+        assertTrue(checkResult.isHealthy());
```
chibenwa, line 39:
```suggestion
        assertThat(checkResult.isHealthy()).isTrue();
```

Arsnael (commented):
Reviewed it, nothing to add. Thanks for the kind contribution :)

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/LdapHealthCheck.java
HoussemNasri: > Please handle the case "user not found" and return healthy in this case

I think it's already handled. Since `getUserByName` returns null if the user is not found, so the final result should be healthy as long the repository method doesn't throw an exception. This is the first time I use the reactor library, so there might be a flaw in my logic.

server/container/guice/data-ldap/src/main/java/org/apache/james/data/LdapUsersRepositoryModule.java
HoussemNasri: Ah right! because we need to inject the list of health check implementations somewhere else in the code.

Arsnael (approved):

quantranhong1999 (approved):

chibenwa:
@HoussemNasri do you agree with the fixes? Can we merge this?

HoussemNasri:
Yes, please go ahead

PR #1468, opened 2023-02-27
JAMES-3890 Allow parallel execution of safe tasks (3.7.x backport)

ottoka:
This is a backport of  #1452 to the 3.7.x branch. In order to keep it minimal, there are some notable changes compared to the master version:
- Task.runAsync() uses Schedulers.elastic() since ReactorUtils.BLOCKING_CALL_WRAPPER has not been introduced yet.
- SerialTaskManagerWorker.close() does not cancel and wait for tasks to finish. Seems that was introduced later an never backported. While I could add this as a potential stability improvement, I am not sure what the impact would be.
- There is no ExpireMailboxTask yet to use as a showcase, though I did a smoke test with a private extension that was its spiritual predecessor.

### DIFF
--- a/server/task/task-api/pom.xml
+++ b/server/task/task-api/pom.xml
@@ -48,6 +48,10 @@
             <artifactId>guava</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.projectreactor</groupId>
+            <artifactId>reactor-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.awaitility</groupId>
             <artifactId>awaitility</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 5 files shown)


Arsnael:
Is it really necessary to backport this to 3.7.x though?

ottoka:
I originally thought it would make sense, but now that you mention it... No, it is not strictly necessary right now, it would only be a preparation for the future, in case we need to backport something else that relies on async task management.
But if you would prefer to keep changes on the 3.7.x branch to a minimum, I am OK with dropping this backport and just keep the repo for future reference just in case, until there is a 3.8.0 release.

chibenwa (approved):
If somebody needs it, and contributes it then let's happily include it.

ottoka:
Well, I can use it to improve the operations of my predecessor to ExpireMailboxTask, so I would like to include it.
@Arsnael : Unless you have a strong objection?

Arsnael (approved):
No strong objections, go for it!

PR #1472, opened 2023-03-01
JAMES-2643 Optimize listing mapping by destination on top of JPA

HoussemNasri:
Fixes https://issues.apache.org/jira/browse/JAMES-2643


Arsnael (approved):
Hi @HoussemNasri 

Thanks for the contribution, it looks good to me!

PR #1475, opened 2023-03-10
JAMES-3895 Automatically provision default (IMAP) mailboxes

woj-tek:
I opted to create default mailboxes when `INBOX` is created without explicitly checking if they exist to lower the impact on the repositories as the check is made on each logging in.


woj-tek:
Wrong branch, mea culpa.

PR #1476, opened 2023-03-10
JAMES-3895 Automatically provision default (IMAP) mailboxes

woj-tek:
I opted to create default mailboxes when `INBOX` is created without explicitly checking if they exist to lower the impact on the repositories as the check is made on each logging in.

### DIFF
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/DefaultMailboxes.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/DefaultMailboxes.java
@@ -38,7 +38,7 @@ public interface DefaultMailboxes {
     String TEMPLATES = "Templates";
     String RESTORED_MESSAGES = "Restored-Messages";
 
-    List<String> DEFAULT_MAILBOXES = ImmutableList.of(INBOX, OUTBOX, SENT, TRASH, DRAFTS, SPAM);
+    List<String> DEFAULT_MAILBOXES = ImmutableList.of(INBOX, OUTBOX, SENT, TRASH, DRAFTS, ARCHIVE, SPAM);
 
     static List<MailboxPath> defaultMailboxesAsPath(Username username) {
         return DEFAULT_MAILBOXES.stream()
--- a/protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
@@ -159,6 +159,10 @@ public abstract class AbstractAuthProcessor<R extends ImapRequest> extends Abstr
     private void provisionMailbox(String mailbox, ImapSession session, MailboxManager mailboxManager,
                                   MailboxSession mailboxSession) throws MailboxException {
         var mailboxPath = PathConverter.forSession(session).buildFullPath(mailbox);
+        if (Mono.from(mailboxManager.mailboxExists(mailboxPath, mailboxSession)).block()) {
+            LOGGER.debug("{} exists. No need to create it.", mailbox);
+            return;
+        }
         try {
             mailboxManager.createMailbox(mailboxPath, MailboxManager.CreateOption.CREATE_SUBSCRIPTION, mailboxSession)
                     .ifPresentOrElse(id -> LOGGER.info("Provisioning mailbox {}. {} created.", mailbox, id),
--- a/protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
@@ -21,11 +21,13 @@ package org.apache.james.imap.processor;
 import java.util.Optional;
 
 import org.apache.james.core.Username;
+import org.apache.james.imap.api.ImapConfiguration;
 import org.apache.james.imap.api.display.HumanReadableText;
 import org.apache.james.imap.api.message.request.ImapRequest;
 import org.apache.james.imap.api.message.response.StatusResponseFactory;
 import org.apache.james.imap.api.process.ImapSession;
 import org.apache.james.imap.main.PathConverter;
+import org.apache.james.mailbox.DefaultMailboxes;
 import org.apache.james.mailbox.MailboxManager;
 import org.apache.james.mailbox.MailboxSession;
 import org.apache.james.mailbox.exception.BadCredentialsException;
@@ -50,6 +52,7 @@ public abstract class AbstractAuthProcessor<R extends ImapRequest> extends Abstr
 
     // TODO: this should be configurable
     private static final int MAX_FAILURES = 3;
+    private ImapConfiguration imapConfiguration;
 
     @FunctionalInterface
     protected interface MailboxSessionAuthWithDelegationSupplier {
@@ -59,6 +62,13 @@ public abstract class AbstractAuthProcessor<R extends ImapRequest> extends Abstr
     public AbstractAuthProcessor(Class<R> acceptableClass, MailboxManager mailboxManager, StatusResponseFactory factory,
                                  MetricFactory metricFactory) {
         super(acceptableClass, mailboxManager, factory, metricFactory);
+    }
+
+    @Override
+    public void configure(ImapConfiguration imapConfiguration) {
+        super.configure(imapConfiguration);
+
+        this.imapConfiguration = imapConfiguration;
     }
 
     protected void doAuth(AuthenticationAttempt authenticationAttempt, ImapSession session, ImapRequest request, Responder responder, HumanReadableText failed) {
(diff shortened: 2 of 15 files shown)


mailbox/api/src/main/java/org/apache/james/mailbox/DefaultMailboxes.java
```diff
@@ -38,7 +38,7 @@ public interface DefaultMailboxes {
     String TEMPLATES = "Templates";
     String RESTORED_MESSAGES = "Restored-Messages";
 
-    List<String> DEFAULT_MAILBOXES = ImmutableList.of(INBOX, OUTBOX, SENT, TRASH, DRAFTS, SPAM);
+    List<String> DEFAULT_MAILBOXES = ImmutableList.of(INBOX, OUTBOX, SENT, TRASH, DRAFTS, ARCHIVE, SPAM);
```
chibenwa, line 41: Pretty sure this will cause some tests to fail (JMAP mailbox provisionning).

Those tests would need to be fixed, too.

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
```diff
@@ -137,17 +147,29 @@ protected void provisionInbox(ImapSession session, MailboxManager mailboxManager
...
+        var mailboxPath = PathConverter.forSession(session).buildFullPath(mailbox);
+        try {
+            mailboxManager.createMailbox(mailboxPath, MailboxManager.CreateOption.CREATE_SUBSCRIPTION, mailboxSession)
+                    .ifPresentOrElse(id -> LOGGER.info("Provisioning mailbox {}. {} created.", mailbox, id),
+                                     () -> LOGGER.warn(
+                                             "Provisioning mailbox {} successful. But no MailboxId have been returned.",
+                                             mailbox));
+        } catch (MailboxExistsException e) {
+            LOGGER.warn("Mailbox {} created by concurrent call. Safe to ignore this exception.", mailbox);
+        }
```
chibenwa, line 170: I bet that if the mailbox already exist prior to calling this method:
 
 - That a `MailboxExistException` will be thrown
 -and  That we will have a misleading warning log.
 
 My advice: call `mailboxExists` prior to creation to avoid both the exception and the log...

mailbox/api/src/main/java/org/apache/james/mailbox/DefaultMailboxes.java
woj-tek: Unfortunately I only run test of the modules I modified, mostly due to all suite with all integration test taking hours (probably https://issues.apache.org/jira/browse/JAMES-3879 would help with "test-ability" here :-) ).

What's the desired way forward - keeping ARCHIVE as default (I would say it's quite elementary folder) and identifying all those test or removing it from the list of defaults?

As for testing goes, I checked the build (https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1476/1/testReport/) and there are a lot of failures, but from the cursory look most of them are due to configuration not being set during tests causing nasty NPE so most likely a missing class/injection in the tests...

woj-tek:
There is a problem with lots of tests failing (which I mentioned in the in-line comment) -  I checked the sources and if I get them right, it would seem that in order to fix those tests virtually all tests in `org.apache.james.mpt.imapmailbox` would have to be updated as they all rely on logging in. Alternatively I could set `imapConfiguration` in `ImapConfiguration` to `null` and do the null check in `org.apache.james.imap.processor.AbstractAuthProcessor#provisionInbox` but this would be just an ugly fix (workaround).

Btw. is `apache-james-mpt-imapmailbox-core` package used anywhere or is it only for tests? It looks like testing code but it's in `main` instead of `test` directory.

chibenwa:
org.apache.james.jmap.rabbitmq.RabbitMQAwsS3GetMailboxesMethodTest.getMailboxesShouldReturnAllMailboxesWhenIdsIsNull

```
JSON path [0][1].list doesn't match.
Expected: a collection with size <8>
```

We need to add the Archive in the assertions of this test?

woj-tek:
Most likely, though I first wanted to address main issue that caused ~1400 tests fail, with that done (https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1476/3/) I'll go over remaining tests and see what the issue.

woj-tek:
I fixed some tests (relying on the actual value of `DefaultMailboxes.DEFAULT_MAILBOXES`), but there are some that still fails:

1) `org.apache.james.mpt.imapmailbox.suite.Condstore#condstoreShouldBeEnableWhenGivenAndTrue` - this one seems to be unrelated to default mailboxes

2) tests in `org.apache.james.jmap.rfc8621`, for example  `org.apache.james.jmap.rfc8621.contract.BackReferenceContract.resolvingAnUnexistingMethodCallIdShouldFail` - the difference in the length is obvious but the order of the list seems random at best?

```
net.javacrumbs.jsonunit.core.internal.Opentest4jExceptionFactory$JsonAssertError: JSON documents are different:
Array "methodResponses[0][1].list" has different length, expected: <6> but was: <7>.
Array "methodResponses[0][1].list" has different content. Extra values: [{"id":"7"}], expected: <[{"id":"1"},{"id":"5"},{"id":"2"},{"id":"3"},{"id":"4"},{"id":"6"}]> but was: <[{"id":"1"},{"id":"6"},{"id":"5"},{"id":"2"},{"id":"3"},{"id":"4"},{"id":"7"}]>
Different value found in node "methodResponses[0][1].list[1].id", expected: <"5"> but was: <"6">.
Different value found in node "methodResponses[0][1].list[2].id", expected: <"2"> but was: <"5">.
Different value found in node "methodResponses[0][1].list[3].id", expected: <"3"> but was: <"2">.
Different value found in node "methodResponses[0][1].list[4].id", expected: <"4"> but was: <"3">.
Different value found in node "methodResponses[0][1].list[5].id", expected: <"6"> but was: <"4">.
```

Btw. One question, why some tests are in `main` and not in `tests` directories?

woj-tek:
Looking at the history of the builds (https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1476/buildTimeTrend) the failure of the tests seems to be somewhat… random?

Arsnael:
> 1. `org.apache.james.mpt.imapmailbox.suite.Condstore#condstoreShouldBeEnableWhenGivenAndTrue` - this one seems to be unrelated to default mailboxes

Unfortunately I think it is... I ran that test on master branch -> success every time
I ran that test on your branch -> fails every time...

I tried to take a quick look at why it fails but I'm at a bit of a loss honestly. But it is related to the PR, I'm sorry.

> 2. tests in `org.apache.james.jmap.rfc8621`, for example  `org.apache.james.jmap.rfc8621.contract.BackReferenceContract.resolvingAnUnexistingMethodCallIdShouldFail` - the difference in the length is obvious but the order of the list seems random at best?

Between `assertThatJson` and `isEqualTo` you can add the method `.withOptions(new Options(IGNORING_ARRAY_ORDER))`. It will ignore the order of the list (we didn't clean up everything related to that thanks for spotting it)

woj-tek:
> Unfortunately I think it is... I ran that test on master branch -> success every time I ran that test on your branch -> fails every time...
> 
> I tried to take a quick look at why it fails but I'm at a bit of a loss honestly. But it is related to the PR, I'm sorry.

It's caused by this change https://github.com/apache/james-project/pull/1476/files#diff-1d7aee20797c7e30c6d737b31a644a146bb6e68bcd6ddbd599dd518dcf3da335R83 and then I made same changes in the problematic test: https://github.com/apache/james-project/pull/1476/files#diff-af9e56bd418cc71a19d530bb8b3d40b8d78f478dcbfaf6ec9355643fcfb466e5R65.

Most likely doing setting IMAP configuration in `JamesImapHostSystem` affects other IMAP processors hence difference in actual and expected list of capabilities. Will have to look more thoroughly into it.


> Between `assertThatJson` and `isEqualTo` you can add the method `.withOptions(new Options(IGNORING_ARRAY_ORDER))`. It will ignore the order of the list (we didn't clean up everything related to that thanks for spotting it)

Thanks! That did the trick :-)

woj-tek:
> > 1. `org.apache.james.mpt.imapmailbox.suite.Condstore#condstoreShouldBeEnableWhenGivenAndTrue` - this one seems to be unrelated to default mailboxes
> 
> Unfortunately I think it is... I ran that test on master branch -> success every time I ran that test on your branch -> fails every time...

I dug a bit more into it, and it seems it was caused by https://issues.apache.org/jira/browse/JAMES-2055 & https://github.com/linagora/james-project/pull/841/files and my change in testing when I used default global configuration but `Condstore` test was changing configuration as well (to check if disabling it works correctly). The issue was that if it got disabled once (as I did globally), it wasn't possible to re-enable it (which the `Condstore` was doing) hence the failing test. I'm not sure if reconfiguration is possible / expected, but I modified `CapabilityProcessor` so it's possible to disable/enable it on subsequent `configure()` calls and added unit test to verify it's working correctly.

woj-tek:
It seems all tests are fixed now :-)

What's weird is that previous build failed suddenly with a bunch of mailet tests failing (https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1476/8/)...

On that note - is there a plan to somehow tackle the issue with the randomly failing test? and also printing lots of exceptions in the output? I saw there is `<excludedGroups>unstable</excludedGroups>` configuration already but I'm not sure that adding more test to that category "solves" the problem.

And the second thing - why some testing code is in the `main` directory` instead of `test`?

chibenwa:
:green_apple:  so I am merging this.

Thanks for the contribution.

Apache CI somethimes fails launching docker containers which causes multiple test to break. It was not stable lately but it seems now to be back on line.

> On that note - is there a plan to somehow tackle the issue with the randomly failing test? and also printing lots of exceptions in the output? 

Yes. Contributors occasinally address some of them. 

> I saw there is <excludedGroups>unstable</excludedGroups> configuration already but I'm not sure that adding more test to that category "solves" the problem.

Non critical unstable tests are excluded yes. Though this group shall be small ideally. :+1: for not adding more tests in there.

> And the second thing - why some testing code is in the main directoryinstead oftest`?

You have a specific example in mind? 

We should not hesitate to move this discussion on the mailing list IMO.

chibenwa (approved):

chibenwa:
Second approuval. @Arsnael  maybe?

Arsnael (approved):
I'm good with it, thanks for the contribution :)

PR #1484, opened 2023-03-14
[FIX] Remove broken issueCommentTrigger

chibenwa:


chibenwa:
To re-enable builds. We can plan longer term action to fix the comment trigger. But at least we will have a working build...

Arsnael:
```
java.lang.NoSuchMethodError: No such DSL method 'cleanWs' found among steps
```
What kind of upgrade they did honestly on Jenkins :o this is gonna be painful...

chibenwa:
https://issues.apache.org/jira/projects/INFRA/issues/INFRA-24340

quantranhong1999 (approved):

quantranhong1999 (approved):

Arsnael:
Regarding cleanWs I remember it's something I did after the 3.7 release... (because maybe the builds were not cleaned well... but we were never sure about it)

 If you look at the backport, the build is running fine. Should we revert it for now? That could help us getting back on track quicker

Arsnael:
I did put back deleteDir() instead... let's see the result

Arsnael:
```
Creating placeholder flownodes because failed loading originals.
ERROR: Cannot resume build because FlowNode 84 for FlowHead 1 could not be loaded. This is expected to happen when using the PERFORMANCE_OPTIMIZED durability setting and Jenkins is not shut down cleanly. Consider investigating to understand if Jenkins was not shut down cleanly or switching to the MAX_SURVIVABILITY durability setting which should prevent this issue in most cases.
```
Ouch... :(

I think the build seems okish like that, should we merge this?

chibenwa:
I do not see a :green_apple: build on this...

PR #1495, opened 2023-03-20
[FIX] Shared mailbox revocation rights should be visible in Mailbox/C…

Arsnael:
…hanges

### DIFF
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxChange.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxChange.java
@@ -207,7 +207,7 @@ public class MailboxChange implements JmapChange {
                     .state(stateFactory.generate())
                     .date(now)
                     .isCountChange(false)
-                    .destroyed(ImmutableList.of(mailboxACLUpdated.getMailboxId()))
+                    .updated(ImmutableList.of(mailboxACLUpdated.getMailboxId()))
                     .delegated()
                     .build());
 
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxChange.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxChange.java
@@ -198,7 +198,21 @@ public class MailboxChange implements JmapChange {
                     .delegated()
                     .build());
 
-            return Stream.concat(Stream.of(ownerChange), shareeChanges)
+            Stream<MailboxChange> deletionChanges = mailboxACLUpdated.getAclDiff()
+                .removedEntries()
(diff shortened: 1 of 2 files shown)


server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxChange.java
```diff
@@ -226,7 +226,21 @@ public List<JmapChange> fromMailboxACLUpdated(MailboxACLUpdated mailboxACLUpdate
...
+            Stream<MailboxChange> deletionChanges = mailboxACLUpdated.getAclDiff()
+                .removedEntries()
+                .filter(entry -> entry.getKey().getNameType().equals(MailboxACL.NameType.user))
+                .filter(entry -> !entry.getKey().isNegative())
+                .map(entry -> MailboxChange.builder()
+                    .accountId(AccountId.fromString(entry.getKey().getName()))
+                    .state(stateFactory.generate())
+                    .date(now)
+                    .isCountChange(false)
+                    .destroyed(ImmutableList.of(mailboxACLUpdated.getMailboxId()))
```
chibenwa, line 238: I think let's call it updated because otherwise if I revoke then re-invite this will not synchronize well

Arsnael: Are you sure that with just updated though the client will understand that the user is not linked to the shared mailbox anymore?

Arsnael:
Rebase (build was green)

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxChange.java
chibenwa: Client will fetch the id to get the changes and get "not found".

If it is re-shared then it will again appear as "updated" and thus will eventually be synchronised.

With "destroyed" the client will drop the item. If I re-share it it might appear as still destroyed because we fold changes. So never synchronised.

If you do not believe me, write this test:

```
Given bob mailbox in state A
AND bob revokes ritghts on marketting@linagora.com team mailbox -> his mailboxes are in state B
AND alice re shares marketting@linagora.com -> his mailboxes are in state C
WHEN bob requests changes with state A
THEN marketting@linagora.com team mailbox appears as updated => synced
```

If it appears as destroyed it would not get synced ;-)

chibenwa (changes requested):
See above comment.

Otherwize good job!

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/change/MailboxChange.java
Arsnael: Indeed I managed to reproduce your case, now I get it.

Made the change to updated and added your case as a test as well

chibenwa (approved):

PR #1503, opened 2023-03-24
JAMES-2080 Add a dedicated Subject search criterion

chibenwa:
This enable hitting the subject field whose indexing can be separately tuned (language?)

This also avoids relying on nested entities for subject search.

https://github.com/apache/james-project/pull/988


chibenwa:
I think so

Arsnael (approved):

vttranlina (approved):

PR #1519, opened 2023-04-14
JAMES-3777 Mitigate performance problems with JMAP filters

chibenwa:
Uses a read projection (optional)

This prevents loading the full aggregate history (O(n2)) and doing LWTs upon receiving emails.

A route allows resetting the projection.

Possible enhancement:
 - Allow individual filter addition, removal (history in O(N) )
 - (Snapshots for the event sourcing system in James ?)


Arsnael (approved):

Arsnael:
Just realising... last commit is a fixup correct?

PR #1530, opened 2023-04-18
JAMES-3777 Incremental events for Filters

chibenwa:
Goals: capture most changes of the filtering aggregate to reduce the complexity of this aggregate.

This work (in progress) proposes an incremental update that covers:
 - Rules addition at the beginning of the rule set or at the end
 - Rules deletion
 - Rules modification
 
 Actions falling outside of this scope cannot be captured in an incremental fashion and would lead to a full reset (todays behaviour). This is the case when:
  - Rules are being reordered.
  - Rules are added in the middle of existing rules.
  
I suspect no UI allows this. Hence this is more that likely enough.

## Work remaining

 - [x] Plug the Incremental events into the aggregate
 - [x] Serialization for this incremental event
 - [x] Conditionally disable incremental change via ENV and explain that it should be disable during rolling update of the upgrade.
 - [x] Tests with 1.000 rule changes

### DIFF
--- a/server/apps/cassandra-app/sample-configuration/jvm.properties
+++ b/server/apps/cassandra-app/sample-configuration/jvm.properties
@@ -53,3 +53,8 @@ james.jmx.credential.generation=true
 # Disable Remote Code Execution feature from JMX
 # CF https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/19fb8f93c59dfd791f62d41f332db9e306bc1422/src/java.management/share/classes/com/sun/jmx/remote/security/MBeanServerAccessController.java#L646
 jmx.remote.x.mlet.allow.getMBeansFromURL=false
+
+# Disabling JMAP filters event source increments is necessary during rolling adoption of this change.
+# Defaults to true, meaning James will use JMAP filters event source increments, thus transparently and significantly
+# improving JMAP filter storage efficiency.
+# james.jmap.filters.eventsource.increments.enabled=true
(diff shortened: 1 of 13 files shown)


chibenwa:
## Benchmarks

### The bench

Create 1.000 sequencially, one by one:

```
    @Test
    void test(EventStore eventStore) {
        FilteringManagement testee = instantiateFilteringManagement(eventStore);

        ImmutableList.Builder<Rule> rules = ImmutableList.builder();

        final Stopwatch gstopwatch = Stopwatch.createStarted();
        for (int i = 0; i < 1000; i++) {
            final Stopwatch stopwatch = Stopwatch.createStarted();
            rules.add( Rule.builder()
                .id(Rule.Id.of("id-subject" + i))
                .name("name")
                .action(Rule.Action.of(Rule.Action.AppendInMailboxes.withMailboxIds("mbx1")))
                .condition(Rule.Condition.of(
                    Rule.Condition.Field.SUBJECT,
                    Rule.Condition.Comparator.NOT_CONTAINS,
                    "A value to match 2"))
                .build());

            Mono.from(testee.defineRulesForUser(USERNAME, rules.build(), Optional.empty())).block();

            System.out.println("Step " + i + ": " + stopwatch.elapsed(TimeUnit.MILLISECONDS));
        }
        System.out.println("Total: " + gstopwatch.elapsed(TimeUnit.MILLISECONDS));
    }
```

### Results: before

Failed at step 964 after 36 minutes (!!!) -> Cassandra timeout (2s)

Step 963 took 14.7s (for storing a single rule!)

Cassandra storage is very well compressed (biiiiig redundancy in the data) yet still the aggregate is big: 7MB on-disk, 210 MB uncompressed.

As such, long histories is clearly non-viable.

### Results: after

Storing the 1.000 rules sequentially took roughly 1 minute with last step taking only `95 ms`.

Cassandra storage is well compressed (twice less) yet still the aggregate is smaller: 26KB on-disk, 400KB uncompressed.

Event running 5000 updates the 5000+ update, the latest step is done in ~500ms.

Long histories for filter are thus more viable.

Further improvements in long histories can be unlocked via complementary approaches like **event sourcing snapshoting**.

chibenwa:
## Snapshots

The snapshots allows to easily improve insertions.

Before above benchmark for 5000 rules: `31 min 42` with last rue creation taking `954 ms`.

After `6 min 18` with last rule creation taking `124 ms`

TODO:
 - [ ] Split in 2 PRs
 - [ ] Tests for snapshots

Arsnael (approved):

PR #1532, opened 2023-04-19
JAMES-3777 Snapshots for eventSourcing JMAP filters

chibenwa:
## Snapshots

The snapshots allows to easily improve insertions.

Before above benchmark for 5000 rules: `31 min 42` with last rue creation taking `954 ms`.

After `6 min 18` with last rule creation taking `124 ms`

### DIFF
--- a/event-sourcing/event-store-cassandra/src/test/scala/org/apache/james/eventsourcing/eventstore/cassandra/CassandraEventStoreTest.scala
+++ b/event-sourcing/event-store-cassandra/src/test/scala/org/apache/james/eventsourcing/eventstore/cassandra/CassandraEventStoreTest.scala
@@ -18,8 +18,41 @@
  ****************************************************************/
 package org.apache.james.eventsourcing.eventstore.cassandra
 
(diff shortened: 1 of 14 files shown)


event-sourcing/event-store-cassandra/src/test/scala/org/apache/james/eventsourcing/eventstore/cassandra/CassandraEventStoreTest.scala
```diff
@@ -18,8 +18,41 @@
...
+    SMono(testee.append(event1)).block()
+    SMono(testee.append(event2)).block()
+    SMono(testee.append(event3)).block()
+
+    assertThat(SMono(testee.getEventsOfAggregate(EventStoreContract.AGGREGATE_1)).block())
+      .isEqualTo(History.of(event2, event3))
+  }
+
+  @Test
+  def test1(testee: EventStore) : Unit = {
```
Arsnael, line 46: idem

Arsnael (approved):

PR #1533, opened 2023-04-20
JAMES-3900 Snapshots for polled updates

chibenwa:
They fully capture the state of the aggregate.

This prevents loading the full history for looong
running tasks and prevents timeouts upon polling updates.

Before the aggregate was containing a lot of events, that were periodically loaded:

```
Widest Partitions:
   [Task/2d534232-fade-47aa-8c6f-1ec2d0f238f9] 1855
   [Task/d8499ec7-a62a-4541-970a-7339fccf23e8] 779
```

Eventually resulting in timeouts as the full partition was loaded...

```
reactor.core.Exceptions$ErrorCallbackNotImplemented: com.datastax.oss.driver.api.core.DriverTimeoutException: Query timed out after PT5S
Caused by: com.datastax.oss.driver.api.core.DriverTimeoutException: Query timed out after PT5S
	at com.datastax.oss.driver.internal.core.cql.CqlRequestHandler.lambda$scheduleTimeout$1(CqlRequestHandler.java:207)
	at io.netty.util.HashedWheelTimer$HashedWheelTimeout.run(HashedWheelTimer.java:715)
	at io.netty.util.concurrent.ImmediateExecutor.execute(ImmediateExecutor.java:34)
	at io.netty.util.HashedWheelTimer$HashedWheelTimeout.expire(HashedWheelTimer.java:703)
	at io.netty.util.HashedWheelTimer$HashedWheelBucket.expireTimeouts(HashedWheelTimer.java:790)
	at io.netty.util.HashedWheelTimer$Worker.run(HashedWheelTimer.java:503)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Unknown Source)
```

With this work, this is mitigated by loading only the latest event...

### DIFF
--- a/server/task/task-memory/src/main/scala/org/apache/james/task/eventsourcing/DecisionProjection.scala
+++ b/server/task/task-memory/src/main/scala/org/apache/james/task/eventsourcing/DecisionProjection.scala
@@ -32,7 +32,7 @@ case class DecisionProjection(status: Status, latestUpdateAdditionalInformationU
         case event: Cancelled => DecisionProjection(Status.CANCELLED, event.additionalInformation.map(_.timestamp))
         case event: Completed => DecisionProjection(Status.COMPLETED, event.additionalInformation.map(_.timestamp))
         case event: Failed => DecisionProjection(Status.FAILED, event.additionalInformation.map(_.timestamp))
-        case event: AdditionalInformationUpdated => DecisionProjection(status, Some(event.additionalInformation.timestamp))
+        case event: AdditionalInformationUpdated => DecisionProjection(Status.IN_PROGRESS, Some(event.additionalInformation.timestamp))
       }
   }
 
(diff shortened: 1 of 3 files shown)


server/task/task-memory/src/main/scala/org/apache/james/task/eventsourcing/DecisionProjection.scala
```diff
@@ -32,7 +32,7 @@ case class DecisionProjection(status: Status, latestUpdateAdditionalInformationU
         case event: Cancelled => DecisionProjection(Status.CANCELLED, event.additionalInformation.map(_.timestamp))
         case event: Completed => DecisionProjection(Status.COMPLETED, event.additionalInformation.map(_.timestamp))
         case event: Failed => DecisionProjection(Status.FAILED, event.additionalInformation.map(_.timestamp))
-        case event: AdditionalInformationUpdated => DecisionProjection(status, Some(event.additionalInformation.timestamp))
+        case event: AdditionalInformationUpdated => DecisionProjection(Status.IN_PROGRESS, Some(event.additionalInformation.timestamp))
```
Arsnael, line 35: I think I get the change, but what's the point of keeping status as an argument to the method update then?

chibenwa: > I think I get the change, but what's the point of keeping status as an argument to the method update then?

Status is not an argument of the update method but part of the state of DecisionProjection and as such we need it.

This change is that the event fully determines the Decision projection disregardless of previous state (prerequisite for snapshot) but do not change the fact that status is needed to take a decision.

Arsnael: Case class indeed, my bad

Arsnael (approved):

PR #1540, opened 2023-04-25
JAMES-3905 LDAP should allow per user base DN

chibenwa:


server/data/data-ldap/src/test/resources/ldif-files/populate.ldif
```diff
@@ -14,3 +18,12 @@ sn: james-user
 mail: james-user@james.org
 userPassword: secret
 description: James user
+
+dn: uid=bob, ou=whatever, dc=james,dc=org
```
Arsnael, line 22: My knowledge of ldap is limited but shouldn't it be more logical to have:

```suggestion
dn: uid=bob, ou=whatever, dc=extra,dc=org
```
instead?

server/data/data-ldap/src/test/resources/ldif-files/populate.ldif
```diff
@@ -6,6 +6,10 @@ dn: ou=empty, dc=james,dc=org
 ou: empty
 objectClass: organizationalUnit
 
+dn: ou=whatever, dc=james,dc=org
```
Arsnael, line 9: idem here? dc=james => dc=extra, if domain name is extra.org?

server/data/data-ldap/src/test/resources/ldif-files/populate.ldif
chibenwa: No, that's a limitation from dinkel container, I struggled a lot with it... OU needs to be based on the LDAP domain...

server/data/data-ldap/src/main/java/org/apache/james/user/ldap/LdapRepositoryConfiguration.java
```diff
@@ -250,12 +275,17 @@ public static LdapRepositoryConfiguration from(HierarchicalConfiguration<Immutab
...
+
     private final boolean trustAllCerts;
 
+    private final ImmutableMap<Domain, String> perDomainBaseDN;
+
     private LdapRepositoryConfiguration(String ldapHost, String principal, String credentials, String userBase, String userIdAttribute,
                                         String userObjectClass, int connectionTimeout, int readTimeout,
                                         boolean supportsVirtualHosting, int poolSize, ReadOnlyLDAPGroupRestriction restriction, String filter,
-                                        Optional<String> administratorId, boolean trustAllCerts) throws ConfigurationException {
+
```
quantranhong1999, line 286:
```suggestion
```

redundant line

quantranhong1999 (commented):
nothing to add

Arsnael (approved):

quantranhong1999 (approved):

PR #1550, opened 2023-05-04
POC Protocol restart - easy with current james code base

chibenwa:
To unwip this we would need:

 - [ ] Log based assertions that the IMAP servers are well restarted
 - [ ] Dedicated maven modules for webadmin routes for protocol restart (for webadmin, guice injections can go in existing guice modules for protocols I bet)

Complementary work:

 - [ ] Other protocols support: IMAP SMTP JMAP POP3 ManageSieve
 - [ ] Explore restarting the mailet container / spooler?

### DIFF
--- a/server/apps/memory-app/src/main/java/org/apache/james/ImapRoutes.java
+++ b/server/apps/memory-app/src/main/java/org/apache/james/ImapRoutes.java
@@ -1,0 +1,40 @@
+package org.apache.james;
+
+import javax.inject.Inject;
+
+import org.apache.james.imapserver.netty.IMAPServerFactory;
+import org.apache.james.server.core.configuration.ConfigurationProvider;
+import org.apache.james.webadmin.Routes;
+
+import com.google.common.base.Preconditions;
+
+import spark.Service;
+
+public class ImapRoutes implements Routes {
+    private final IMAPServerFactory imapServerFactory;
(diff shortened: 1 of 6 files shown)


testing/base/src/main/resources/logback-test.xml
```diff
@@ -17,10 +17,6 @@
                 <encoder>
                         <pattern>%d{HH:mm:ss.SSS} [%-5level] %logger{15} - %msg%n%rEx</pattern>
                 </encoder>
-                <immediateFlush>false</immediateFlush>
-                <filter class="ch.qos.logback.classic.filter.ThresholdFilter">
-                        <level>ERROR</level>
-                </filter>
```
chibenwa, line 23: Should not be kept

server/apps/memory-app/src/main/java/org/apache/james/ImapRoutes.java
```diff
@@ -0,0 +1,40 @@
+package org.apache.james;
```
chibenwa, line 1: License

chibenwa:
CF https://issues.apache.org/jira/browse/JAMES-3906

Consensus seems to lean toward a lighter solution that do not involve service interuption. ACME support and change propagation accress the cluster seems to be a goal too.

As such, I am closing this POC.

PR #1570, opened 2023-05-22
JAMES-3909 Delete all users data of a domain

quantranhong1999:


Arsnael (commented):
Shouldn't we add maybe a limit then to how many failed users we accept to report, to limit the issue? 
After a certain number as well I believe the task progress potentially becomes unreadable for the admin too

quantranhong1999:
> Shouldn't we add maybe a limit then to how many failed users we accept to report, to limit the issue?
After a certain number as well I believe the task progress potentially becomes unreadable for the admin too

Sounds good idea. Maybe I would just push a hard code limit say 100 failed users. That would be enough for an admin to handle at once...

Arsnael (approved):

chibenwa (approved):

PR #1579, opened 2023-06-02
[FIX] JPA blocking calls should be moved to bounded elastic

chibenwa:
## Before

openjpa (blocking) calls on the IMAP event loop

Screenshot from 2023-06-02 14-29-32

## After

No openjpa (blocking) calls on the IMAP event loop

Screenshot from 2023-06-02 14-28-42

### DIFF
--- a/mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/JPAMessageMapper.java
+++ b/mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/JPAMessageMapper.java
@@ -103,6 +103,12 @@ public class JPAMessageMapper extends JPATransactionalMapper implements MessageM
     }
 
     @Override
+    public Flux<MailboxMessage> findInMailboxReactive(Mailbox mailbox, MessageRange messageRange, FetchType ftype, int limitAsInt) {
+        return Flux.defer(Throwing.supplier(() -> Flux.fromIterable(findAsList(mailbox.getMailboxId(), messageRange, limitAsInt))).sneakyThrow())
+            .subscribeOn(Schedulers.boundedElastic());
+    }
+
+    @Override
     public Iterator<MailboxMessage> findInMailbox(Mailbox mailbox, MessageRange set, FetchType fType, int max)
             throws MailboxException {
 
--- a/mailbox/jpa/pom.xml
+++ b/mailbox/jpa/pom.xml
@@ -134,6 +134,17 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <reuseForks>false</reuseForks>
+                    <forkCount>1</forkCount>
+                    <argLine>-Djava.library.path=
+                        -javaagent:"${settings.localRepository}"/org/jacoco/org.jacoco.agent/${jacoco-maven-plugin.version}/org.jacoco.agent-${jacoco-maven-plugin.version}-runtime.jar=destfile=${basedir}/target/jacoco.exec
+                        -Xms512m -Xmx1024m -Dopenjpa.Multithreaded=true</argLine>
+                </configuration>
+            </plugin>
+            <plugin>
                 <groupId>org.apache.openjpa</groupId>
                 <artifactId>openjpa-maven-plugin</artifactId>
                 <version>${apache.openjpa.version}</version>
--- a/mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/JPAMailboxMapper.java
+++ b/mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/JPAMailboxMapper.java
@@ -19,6 +19,8 @@
 
 package org.apache.james.mailbox.jpa.mail;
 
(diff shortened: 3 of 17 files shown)


Arsnael (approved):

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1579/1/testReport/

Looks related

chibenwa:
So I tracked down and detected the leak:

Spawn a dockerized JPA james locally thanks to docker and imapsync my company emails to it...

### Before

Screenshot from 2023-06-02 10-04-33

GC is unable to reclaim memory, showing a leak that eventually builds up to an OOM.

## After

Screenshot from 2023-06-02 17-09-47


We are waaaay better at reclaiming the damn memory.

woj-tek:
I ran a test with the changes (though, on my local machine and not remote deployment) and it seems to be behaving better:
Captura de pantalla 2023-06-02 a las 18 29 22

While simultaneously running the profiler I noticed that, it seems, James/OpenJPA doesn't use any connection pool and simply re-opens connection to the database for each query (?!):
Captura de pantalla 2023-06-02 a las 18 38 07

Documentation (https://james.apache.org/server/config-system.html) states, that James delegates connectivity to `commons-dbcp`, but it doesn't seem to be the case. I even tried to force pool size but still - connections being opened and closed

mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/JPAMessageMapper.java
```diff
@@ -93,6 +94,12 @@ public Flux<MessageUid> listAllMessageUids(Mailbox mailbox) {
             .map(MailboxMessage::getUid);
     }
 
+    @Override
+    public Flux<MailboxMessage> findInMailboxReactive(Mailbox mailbox, MessageRange messageRange, FetchType ftype, int limitAsInt) {
+        return Flux.defer(Throwing.supplier(() -> Flux.fromIterable(findAsList(mailbox.getMailboxId(), messageRange, limitAsInt))).sneakyThrow())
```
woj-tek, line 99: Shouldn't internal `Flux` (`Flux.fromIterable(findAsList())`) have same `Schedulers.boundedElastic()`?

woj-tek:
Hmm... it seems something is still leaking:
Captura de pantalla 2023-06-02 a las 20 02 52

mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/JPAMessageMapper.java
chibenwa: No it is already running on this scheduler.

chibenwa:
I must confess I did only run a local derby database thus might have mess the connection churn...

chibenwa:
> While simultaneously running the profiler I noticed that, it seems, James/OpenJPA doesn't use any connection pool and simply re-opens connection to the database for each query (?!)

Confirmed by flame graphs running mariaDB...

Screenshot from 2023-06-05 10-52-30

chibenwa:
Adding DBCP2 on the classpath seems to be enough to solve the connection churn issue...

Screenshot from 2023-06-05 13-23-21

I no longer spot obvious connects in the flae graph of james + mariaDB...

Full flame graph:

[msql2.zip](https://github.com/apache/james-project/files/11649480/msql2.zip)

woj-tek:
With dbcp2 in classpath I was getting lot of "multithred" exceptions. I had `openjpa.Multithreaded=true` in `james-database.properties` but upon debugging I noticed it didn't take any effect (getting properties from injected EntityManager). Including it directly in `JPAEntityManagerModule` fixed that problem (most likely passing it via `-D` commandline parameter would work as well). Memory usage is way better and kept (relatively) low, though copying/appending messages is still quite slow (~2msg/s, with low CPU usage so still seems like IO-bound issue):
Captura de pantalla 2023-06-05 a las 12 07 53

chibenwa:
>  I had openjpa.Multithreaded=true in james-database.properties

Should be in jvm.properties

woj-tek:
> > I had openjpa.Multithreaded=true in james-database.properties
> 
> Should be in jvm.properties

That's true. When I searched for it, I didn't pay enough attention and saw "...properties" and assumed it was database properties as it seemed related...

chibenwa:
I did run some performance test on top of this branch (local mariaDB setup), 50 user simulation on top of 5 accounts with ~50 mailboxes and ~1000 emails...

Results were better than I expected:

Screenshot from 2023-06-06 12-57-51

Though `SELECT` and `STATUS` are shockingly slow (p50 -> p99) and the biggest contributor to latencies. I suspect those loads all messages... 

Also this PR as it stands breaks soe tests (did not had time to dig...)

Screenshot from 2023-06-06 12-57-29

chibenwa:
> Is it possible to to get exact sql queries/parameters? To possibly try to optimise it somehow?

We did get that with glowroot in the past for Cassandra...

Arsnael:
is it related? https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1579/9/testReport/

quantranhong1999 (approved):

Arsnael (approved):

Arsnael:
@chibenwa ?

PR #1580, opened 2023-06-06
Allow passing openjpa and underlying datasource (DBCP2) properties

woj-tek:
This is a little bit of POC referencing my last mail in thread "OpenJPA pool with DBCP".

It fixes:
* configuring `openjpa.Multithreaded` (with defaulting to `true` unless configured otherwise)
* configuring additional openjpa properties: originally it used ` openjpa.ConnectionFactoryProperties` but as configuration (https://openjpa.apache.org/builds/3.2.2/apache-openjpa/docs/index.html#ref_guide_dbsetup_config) indicates it's only for internal DataSource (and `MaxTotal` would not even be considered here) - used `openjpa.ConnectionProperties` instead

Added passing all `openjpa.*` and `datasource.*` properties from `james-database.properties` file to openjpa or to DBCP2 (via `openjpa.ConnectionProperties`) respectively. I left existing `connectionProperties` (`TestOnBorrow`, `ValidationTimeout`, `ValidationSQL` and `MaxTotal`) though they could probably be removed altogether at this point.


server/container/guice/jpa-common/src/main/java/org/apache/james/modules/data/JPAEntityManagerModule.java
```diff
@@ -68,15 +76,34 @@ public EntityManagerFactory provideEntityManagerFactory(JPAConfiguration jpaConf
...
+
+
         return JPAConfiguration.builder()
                 .driverName(dataSource.getString("database.driverClassName"))
                 .driverURL(dataSource.getString("database.url"))
                 .testOnBorrow(dataSource.getBoolean("datasource.testOnBorrow", false))
                 .validationQueryTimeoutSec(dataSource.getInteger("datasource.validationQueryTimeoutSec", null))
                 .validationQuery(dataSource.getString("datasource.validationQuery", null))
                 .maxConnections(dataSource.getInteger("datasource.maxTotal",null))
+                .multithreaded(dataSource.getBoolean(JPAConfiguration.JPA_MULTITHREADED,true))
```
chibenwa, line 91:
```suggestion
                .multithreaded(dataSource.getBoolean(JPAConfiguration.JPA_MULTITHREADED, true))
```

chibenwa (commented):
Should be documented: https://github.com/apache/james-project/blob/e00f08d06de956438b821ab9f087e7787e7cba5d/src/site/xdoc/server/config-system.xml#L55 ?

woj-tek:
> Should be documented:

I didn't add documentation earlier as I wasn't sure if the change is good/desired. I just pushed updated documentation and corrected formatting.

chibenwa (approved):

Arsnael (approved):

PR #1585, opened 2023-06-08
JAMES-3911 JPA: Alternative IMAP schedulers

chibenwa:
I do not like this but of course that's in the back of my mind as I do not succeed to get a working version of https://github.com/apache/james-project/pull/1579 ...


chibenwa:
FInaly nailed https://github.com/apache/james-project/pull/1579 !!!

PR #1600, opened 2023-06-20
Add simple null-check in JPAMailboxMessage #JAMES-3915

woj-tek:


Arsnael (approved):

mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/model/openjpa/JPAMailboxMessage.java
```diff
@@ -49,6 +49,7 @@
 public class JPAMailboxMessage extends AbstractJPAMailboxMessage {
 
     private static final byte[] EMPTY_BODY = new byte[] {};
+    private static final byte[] EMPTY_HEADER = new byte[] {};
```
chibenwa, line 52: Maybe we just need one of these `EMPTY_ARRAY` constant, returned in both cases?

chibenwa (approved):

mailbox/jpa/src/main/java/org/apache/james/mailbox/jpa/mail/model/openjpa/JPAMailboxMessage.java
woj-tek: Pushed change.

chibenwa (approved):

PR #1618, opened 2023-06-29
JAMES-3921: Use lenient parser, which should help parsing invalid datetime fields

woj-tek:
JAMES-3921: Use lenient parser, which should help parsing invalid datetime fields; more informative log entry

### DIFF
--- a/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LuceneMessageSearchIndex.java
+++ b/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LuceneMessageSearchIndex.java
@@ -74,6 +74,7 @@ import org.apache.james.mailbox.store.mail.model.MailboxMessage;
 import org.apache.james.mailbox.store.search.ListeningMessageSearchIndex;
 import org.apache.james.mailbox.store.search.SearchUtil;
 import org.apache.james.mime4j.MimeException;
+import org.apache.james.mime4j.codec.DecodeMonitor;
 import org.apache.james.mime4j.dom.Header;
 import org.apache.james.mime4j.dom.address.Address;
 import org.apache.james.mime4j.dom.address.AddressList;
@@ -81,6 +82,7 @@ import org.apache.james.mime4j.dom.address.Group;
 import org.apache.james.mime4j.dom.address.MailboxList;
 import org.apache.james.mime4j.dom.datetime.DateTime;
 import org.apache.james.mime4j.dom.field.DateTimeField;
+import org.apache.james.mime4j.field.LenientFieldParser;
 import org.apache.james.mime4j.field.address.AddressFormatter;
 import org.apache.james.mime4j.field.address.LenientAddressParser;
 import org.apache.james.mime4j.field.datetime.parser.DateTimeParser;
@@ -607,7 +609,7 @@ public class LuceneMessageSearchIndex extends ListeningMessageSearchIndex {
         doc.add(new NumericField(SIZE_FIELD,Store.YES, true).setLongValue(membership.getFullContentOctets()));
 
         // content handler which will index the headers and the body of the message
-        SimpleContentHandler handler = new SimpleContentHandler() {
+        SimpleContentHandler handler = new SimpleContentHandler(LenientFieldParser.getParser(), DecodeMonitor.SILENT) {
             
 
             @Override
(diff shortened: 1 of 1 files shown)


chibenwa (approved):
Do we need to open a 3.8.x backport for this?

woj-tek:
Not sure, but keep in mind that this PR is rather a WIP - I included a test in later commit to indicate that using lenient alone doesn't fix the issue - but fix in apache-mime4j does (more in Jira, would have been easier to keep it in one place :) )

mailbox/lucene/src/test/java/org/apache/james/mailbox/lucene/search/LuceneMailboxMessageSearchIndexTest.java
```diff
@@ -227,6 +229,27 @@ void bodySearchShouldNotMatchPhraseOnlyInSubject() throws Exception {
         assertThat(result).isEmpty();
     }
 
+    @Test
+    void dateHeaderParsingShouldNotImpactProcessing() throws Exception {
```
chibenwa, line 233: You mean, this test is failing?

woj-tek: Yes:

```
$ mvn test -Dtest=org.apache.james.mailbox.lucene.search.LuceneMailboxMessageSearchIndexTest#dateHeaderParsingShouldNotImpactProcessing
…
17:35:13,045 [ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.477 s <<< FAILURE! - in org.apache.james.mailbox.lucene.search.LuceneMailboxMessageSearchIndexTest
17:35:13,046 [ERROR] dateHeaderParsingShouldNotImpactProcessing  Time elapsed: 0.415 s  <<< ERROR!
org.apache.james.mime4j.field.datetime.parser.TokenMgrError: Lexical error at line 1, column 3.  Encountered: "." (46), after : ""
	at org.apache.james.mailbox.lucene.search.LuceneMailboxMessageSearchIndexTest.dateHeaderParsingShouldNotImpactProcessing(LuceneMailboxMessageSearchIndexTest.java:246)
```

And passes just fine with https://github.com/apache/james-mime4j/pull/88 (though, It's just ignoring the parsing problem).

Btw. what's the general consensus/flow when it comes to fixes in dependent libraries?

Arsnael: > Btw. what's the general consensus/flow when it comes to fixes in dependent libraries?

I would say similar to the main project?

Arsnael:
So need to wait for the fix in mime4j that you propose to be merged then, new mime4j release and upgrade in James, correct?

mailbox/lucene/src/test/java/org/apache/james/mailbox/lucene/search/LuceneMailboxMessageSearchIndexTest.java
chibenwa: TokenMgrError shows that lenient parsers are not actually being used here (if I am not wrong).

We need to dig deeper.

Corporate meetings on my side are other, I will have time to take a look today.

chibenwa:
Hello,

Taking a deeper look at this (turning on the IDE...) I realized the date is parsed twice and that our code did not need to manage GMT convertion because Lucene takes care of it for us.

Please see https://github.com/apache/james-project/pull/1621

mailbox/lucene/src/test/java/org/apache/james/mailbox/lucene/search/LuceneMailboxMessageSearchIndexTest.java
woj-tek: > I would say similar to the main project?

I meant - would it be correct to bump mime4j dependency version (in james) to `0.8.10-SNAPSHOT` to indicate that during release it should be also included? (though I'm not sure if SNAPSHOTS are published anywhere, so there's that)

woj-tek:
> Taking a deeper look at this (turning on the IDE...) I realized the date is parsed twice and that our code did not need to manage GMT convertion because Lucene takes care of it for us.

Oh, so the whole code wasn't needed altogether :-)

Btw. why use zoned GMT instead of UTC? (they are virtually the same but still…)

PR #1561, opened 2023-05-15
JAMES 3822 - FutureRelease for JMAP

thanhbv200585:

PR #1620, opened 2023-06-30
JAMES-3822: New properties for future release jmap in jmap.properties

thanhbv200585:
In order to config future release extension for jmap. we need a new properties `delay.sends.enabled` in `jmap.properties`

- `delay.sends.enabled = true` => support for future release 
- `delay.sends.enabled = false` => no support for future release 

IMO, we need add doc for this properties in https://james.apache.org/server/config-jmap.html


quantranhong1999:
This PR should be more than document `delay.sends.enabled` properties, because it is already in the code.

Also, this PR commits history is a mess. Please always create a new branch for every PR, and rebase on the latest master.

chibenwa (changes requested):
I just do not understand the diff that actually reshow some merged content of https://github.com/apache/james-project/pull/1561

I won't review until the history is clean.

quantranhong1999 (approved):

Arsnael (approved):

chibenwa (approved):

PR #1633, opened 2023-07-05
[FIX] JAMES-3919 RabbitMQMailQueue: clean up cassandra projection whe…

Arsnael:
…n we can't publish to Rabbit

Was a piece missing


Arsnael:
Was merged with https://github.com/apache/james-project/pull/1629

PR #1635, opened 2023-07-06
[JAMES-3822] Upgrade EmailSubmissionSetMethodFutureReleaseTest for distributed test

thanhbv200585:


server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailSubmissionSetMethodFutureReleaseContract.scala
```diff
@@ -297,8 +299,10 @@ trait EmailSubmissionSetMethodFutureReleaseContract {
       .body(request)
       .post
 
-    // Wait one second
-    updatableTickingClock.setInstant(DATE.plusSeconds(1))
+    // Wait ten seconds
+    updatableTickingClock.setInstant(DATE.plusSeconds(10))
+
+    Thread.sleep(5000)
```
chibenwa, line 305: Can we find other solutions than such a looooong wait time? Why do we need to wait for?

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailSubmissionSetMethodFutureReleaseContract.scala
```diff
@@ -712,6 +744,7 @@ trait EmailSubmissionSetMethodFutureReleaseContract {
       .body(request)
       .post
 
+    Thread.sleep(5000)
```
chibenwa, line 747: Can we find other solutions than such a looooong wait time? Why do we need to wait for?

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailSubmissionSetMethodFutureReleaseContract.scala
thanhbv200585: Should I decrease the waiting time? I try to sleep 1s and tests are still correct

chibenwa (changes requested):
No I disagree to alter the future release contract.

The contract relies on the fact that the MQ is fully `UpdatableTickingClock` aware, which the distributed mail queue is not.

Instead I would write a single test proving delays not to work in distributed mode:

```
1. Record time t0 with a StopWatch
2. Send a mail, delayed 5 seconds
3. Await the mail to arrive for 10 seconds (100 ms resolution)
4. After the mail is received record time t1
5. Asstert t1 - t0 > 5 seconds (will fail)
```

quantranhong1999:
>Instead I would write a single test proving delays not to work in distributed mode:

IMO your suggestion just proves something (delay send on Distributed) **could not** work. It does not change the fact that the current contract test suite is not strong enough to prove something (delay send on whatever platform) **could** work.

Thanh tried to run the test suite with James Memory with delay send **disabled**, and delay send tests are still passed...

What if in the future we have a JMAP server on top of the Pulsar mail queue, how do we make sure the delay send **could** work?

Exchanging 4 seconds of build time to get more reliable tests is a good trade IMO.

chibenwa:
> Exchanging 4 seconds of build time to get more reliable tests is a good trade IMO.

Ok

quantranhong1999:
@thanhbv200585  you are missing the James ticket number prefix. You can use the old JMAP future release number.

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailSubmissionSetMethodFutureReleaseContract.scala
```diff
@@ -312,24 +313,27 @@ trait EmailSubmissionSetMethodFutureReleaseContract {
...
+
+    awaitAtMostTenSeconds.untilAsserted { () =>
+      val response = `given`(
+        baseRequestSpecBuilder(server)
+          .setAuth(authScheme(UserCredential(ANDRE, ANDRE_PASSWORD)))
+          .addHeader(ACCEPT.toString, ACCEPT_RFC8621_VERSION_HEADER)
+          .setBody(requestAndre)
+          .build, new ResponseSpecBuilder().build)
+        .post
+        .`then`
```
Arsnael, line 325: indent

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailSubmissionSetMethodFutureReleaseContract.scala
```diff
@@ -395,6 +397,33 @@ trait EmailSubmissionSetMethodFutureReleaseContract {
...
+    Assertions.assertThatThrownBy(() => {
+      calmlyAwait.atMost(1, TimeUnit.SECONDS).untilAsserted { () =>
+        val response = `given`(
+          baseRequestSpecBuilder(server)
+            .setAuth(authScheme(UserCredential(ANDRE, ANDRE_PASSWORD)))
+            .addHeader(ACCEPT.toString, ACCEPT_RFC8621_VERSION_HEADER)
+            .setBody(requestAndre)
+            .build, new ResponseSpecBuilder().build)
+          .post
+          .`then`
```
Arsnael, line 410: ident

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailSubmissionSetMethodFutureReleaseContract.scala
```diff
@@ -811,6 +838,32 @@ trait EmailSubmissionSetMethodFutureReleaseContract {
...
```
Arsnael, line 851: indent

Arsnael (commented):
@thanhbv200585 please when you think you fixed a comment, think of marking it as resolved, before it becomes mess :)

chibenwa (approved):

quantranhong1999 (approved):

Arsnael (approved):

PR #1639, opened 2023-07-08
Move rabbitMQ channel close operation to boundedElastic thread

Arooba-git:
This prevents the call from blocking the reactive pipeline.

JDC Flight Control performance (CPU and latency) results before fix:
JAMES-BEFORE
Screen Shot 2023-07-08 at 11 10 30 AM

**CPU and latency after fix:**
JAMES-AFTER
Screen Shot 2023-07-08 at 11 07 35 AM


chibenwa (approved):

chibenwa:
Thanks a lot for this contribution!

PR #1645, opened 2023-07-12
TLS Host name verification was added in 3.7.3

Xcreen:
TLS Host name verification was added in 3.7.3 and not in 3.8.0.

Pull-Request: https://github.com/apache/james-project/pull/1247


Xcreen:
I saw now that the update-instructions say: `Changes to apply between 3.7.x and 3.8.0 are reported here.`
But i feel like it should be mentioned in 3.7.3 because this change can lead to undelivered mails and the change would be overseen (like i did) when upgrading to 3.7.3 / 3.7.4.

chibenwa (approved):

quantranhong1999 (approved):

Arsnael (approved):

chibenwa (approved):

PR #1650, opened 2023-07-18
Move blocking mail items polling operation to boundedElastic

Arooba-git:
Hi 👋 
Apparently the mail-items polling operation in `memory-app` is blocking, as detected by BlockHound.
Screen Shot 2023-07-19 at 5 43 01 AM

We also compared the performance before and after the fix.
Before:
james-2-latency-before
james2-cpu-before

After:

james-2-cpu-after
james2-latency-after


chibenwa (approved):
Nice catch!

Arsnael (approved):
Thanks for the catch @Arooba-git and the fix :)

quantranhong1999 (approved):

PR #1653, opened 2023-07-23
Move blocking readPair call to boundedElastic thread

Arooba-git:
Hi! 🙂 

We noticed you did a great job in ensuring the reactive modules indeed stay reactive end to end. The *mailbox* module, however, was discovered to still have a blocking call in `CassandraMessageMapper` class as detected by BlockHound:
james3-blocking

This PR fixes this code. We also re-ran the tests and verified the performance (in terms of heap usage) before and after the fix:

**Before**
james3-heap-before

**After**

james3-heap-after

### DIFF
--- a/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageMapper.java
+++ b/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageMapper.java
@@ -170,7 +170,16 @@ public class CassandraMessageMapper implements MessageMapper {
                 }
                 return Mono.just(counters);
             })
-            .doOnNext(counters -> readRepair(mailbox, counters));
+            .doOnNext(counters -> {
+                if (this.cassandraConfiguration.getMailboxReadRepair() > 0) {
+                    Mono.fromRunnable(() -> {
+                        readRepair(mailbox, counters);
+                    })
+                   .subscribeOn(Schedulers.boundedElastic())
+                   .subscribe();
+                }
+            });
+
     }
 
     public Mono<MailboxCounters> readMailboxCounters(CassandraId mailboxId) {


quantranhong1999 (approved):
Thanks for your contribution  <3

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageMapper.java
```diff
@@ -170,7 +170,14 @@ public Mono<MailboxCounters> getMailboxCountersReactive(Mailbox mailbox) {
                 }
                 return Mono.just(counters);
             })
-            .doOnNext(counters -> readRepair(mailbox, counters));
+            .doOnNext(counters -> {
+                Mono.fromRunnable(() -> {
+                    readRepair(mailbox, counters);
```
chibenwa, line 175: Can we manage to subscribe here **only if read repairs are enabled** ie probability > 0 ?

Arooba-git: Sure, you mean something like:
```
.doOnNext(counters -> {
       if (probability > 0) {
              Mono.fromRunnable(() -> {
                   readRepair(mailbox, counters);
              })
             .subscribeOn(Schedulers.boundedElastic())
             .subscribe();
       }
});
```

chibenwa: Yes

Arooba-git: @chibenwa I'm sorry could you guide how to access probability property here? 🤔

chibenwa: from the configuration

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageMapper.java
```diff
@@ -170,7 +170,16 @@ public Mono<MailboxCounters> getMailboxCountersReactive(Mailbox mailbox) {
                 }
                 return Mono.just(counters);
             })
-            .doOnNext(counters -> readRepair(mailbox, counters));
+            .doOnNext(counters -> {
+                if (this.cassandraConfiguration.getMailboxReadRepair() > 0) {
```
chibenwa, line 174: I'm oretty sure it should be

```suggestion
                if (this.cassandraConfiguration.getMailboxCountersReadRepair() > 0) {
```

Arooba-git: Oh, you mean we should update the name in the original configuration file too? 🤔 
Screen Shot 2023-07-25 at 12 13 15 AM

Arsnael: I believe what chibenwa is saying is:
- `getMailboxReadRepair()` is used in `CassandraMailboxMapper`
- for `CassandraMessageMapper `, we use `getMailboxCountersReadRepairChanceMax()` and/or `getMailboxCountersReadRepairChanceOneHundred()`

=> for example: https://github.com/apache/james-project/blob/master/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageMapper.java#L197

Arsnael: Maybe we could extract that line actually in a function to avoid duplication, that we could call from there and from your code change :)

chibenwa:
Also read repair could be a good place for pseudo random usage aka non blocking...

There is no need for cryptographioc grade randomness for such an application.

Maybe otherall this could be a better fix. WDYT?

chibenwa:
Please provide a pseudo random implementation for read repairs

PR #1666, opened 2023-08-02
DRAFT JAMES-4824 Refactor quota for Cassandra: create POJOs and DAOs

hungphan227:

PR #1670, opened 2023-08-08
JAMES-4824 Refactor quota for Cassandra: create POJOs

hungphan227:


core/src/main/java/org/apache/james/core/StringWrapper.java
```diff
@@ -0,0 +1,57 @@
...
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.core;
+
+import java.util.Objects;
+
+import com.google.common.base.MoreObjects;
+
+public class StringWrapper {
```
Arsnael, line 26: Not sure this was necessary. Allows to reduce boiler plate, reduces readability though a bit IMO.

But I'm not against it.

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
```diff
@@ -0,0 +1,74 @@
...
+import java.util.Objects;
+
+public class QuotaCurrentValue {
+
+    public static QuotaCurrentValue of(QuotaComponent quotaComponent, String identifier, QuotaType quotaType, long currentValue) {
+        return new QuotaCurrentValue(quotaComponent, identifier, quotaType, currentValue);
+    }
+
+    private final QuotaComponent quotaComponent;
+    private final String identifier;
```
Arsnael, line 31: A quota current value would be logically the value of a user's usage... I would expect here Username instead of String

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
```diff
@@ -0,0 +1,74 @@
...
+    private final QuotaType quotaType;
+    private final long currentValue;
```
Arsnael, line 33: Optional<Long> ? Maybe there is no quota yet for a user that just got created  and didnt receive any message yet for example?

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class QuotaLimit {
+
+    public static QuotaLimit of(QuotaComponent component, QuotaScope scope, String identifier, QuotaType quotaType, long limit) {
+        return new QuotaLimit(component, scope, identifier, quotaType, limit);
+    }
+
+    private final QuotaComponent quotaComponent;
+    private final QuotaScope quotaScope;
+    private final String identifier;
```
Arsnael, line 32: Thinking what we could do here...

Should be at least an Optional (global quota limit will not have an identifier for example)

After the identifier will be either a Username (account limit) or a Domain (domain limit)

But I guess if we want to be flexible to extensions even on the scope, String is likely fine I guess. But I would still make it Optional, for the global case

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
```diff
@@ -0,0 +1,81 @@
...
+    private final QuotaType quotaType;
+    private final long limit;
```
Arsnael, line 34: Optional<Long>?

We can have undefined limits

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
hungphan227: I think initially it should be zero

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
hungphan227: identifier is in primary key so it could not be null. In case of global, we could set a fixed value to the identifier ("APP" for example)

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
hungphan227: In fact, this value in in counter table. I think it should never be null

Arsnael: Not necessary? What if it's not? As said above user just got created for example..?

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
Arsnael: Correct! I think just "global" then

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
```diff
@@ -23,6 +23,14 @@
 
 public class QuotaComponent extends StringWrapper {
 
+    public static final QuotaComponent MAILBOX = of("mailbox");
+    public static final QuotaComponent SIEVE = of("sieve");
+    public static final QuotaComponent JMAP_UPLOADS = of("jmapUploads");
+
+    public static QuotaComponent of(String value) {
```
Arsnael, line 30: What's the point here?

Could just do `public static final QuotaComponent MAILBOX = new QuotaComponent("mailbox");` right?

core/src/main/java/org/apache/james/core/quota/QuotaScope.java
```diff
@@ -23,6 +23,14 @@
 
 public class QuotaScope extends StringWrapper {
 
+    public static final QuotaScope GLOBAL = of("global");
+    public static final QuotaScope DOMAIN = of("domain");
+    public static final QuotaScope USER = of("user");
+
+    public static QuotaScope of(String value) {
```
Arsnael, line 30: Idem

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
```diff
@@ -20,25 +20,26 @@
 package org.apache.james.core.quota;
 
 import java.util.Objects;
+import java.util.Optional;
 
 public class QuotaLimit {
 
-    public static QuotaLimit of(QuotaComponent component, QuotaScope scope, String identifier, QuotaType quotaType, long limit) {
+    public static QuotaLimit of(QuotaComponent component, QuotaScope scope, String identifier, QuotaType quotaType, Long limit) {
```
Arsnael, line 27: Missed that the first time but is that static method useful here as well?

core/src/main/java/org/apache/james/core/quota/QuotaType.java
```diff
@@ -23,6 +23,13 @@
 
 public class QuotaType extends StringWrapper {
 
+    public static final QuotaType COUNT = of("count");
+    public static final QuotaType SIZE = of("size");
+
+    public static QuotaType of(String value) {
```
Arsnael, line 29: idem

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
hungphan227: Just a shorter way to create new QuotaComponent. It does not matter which way I use, right?

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
hungphan227: I saw this method in other places of james. I think we prefer this way to create objects :))

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
hungphan227: Copied from cassandra official web page:
"The counter type is used to define counter columns. A counter column is a column whose value is a 64-bit signed integer and on which 2 operations are supported: incrementing and decrementing (see the [UPDATE](https://cassandra.apache.org/doc/latest/cassandra/cql/dml.html#update-statement) statement for syntax). Note that the value of a counter cannot be set: a counter does not exist until first incremented/decremented, and that first increment/decrement is made as if the prior value was 0"

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
Arsnael: I find adding a whole static method just to pass from `new QuotaComponent` to `of` is adding more verbosity for... not much any kind of gain?

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
Arsnael: Like where?

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
Arsnael: `a counter does not exist until first incremented/decremented`

Which was the point... what if the counter does not exist yet for that user? Turning in circles here :)

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
hungphan227: org/apache/james/jmap/api/filtering/Rule.java - class Condition - line 186

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
hungphan227: If the counter does not exist, the whole row in cassandra table also does not exist. Therefore, when converting from a row to a QuotaCurrentValue object, the field currentValue of QuotaCurrentValue object always has a value

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
```diff
@@ -0,0 +1,76 @@
...
+
+package org.apache.james.core.quota;
+
+import java.util.Objects;
+
+import org.apache.james.core.Username;
+
+public class QuotaCurrentValue {
+
+    public static QuotaCurrentValue of(QuotaComponent quotaComponent, Username identifier, QuotaType quotaType, long currentValue) {
```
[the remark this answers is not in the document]
hungphan227: I think it does not fit. the domain part in QuotaRoot is optional while the identifier in QuotaCurrentValue always has this part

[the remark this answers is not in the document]
Arsnael: Nope you are wrong here actually :) 

Look in UsersRepository code and configuration file... We have a parameter, `supportVirtualHosting`. If it's true we have domain names, if false we don't.

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
Arsnael: Missed your constructors were private. My bad

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
Arsnael: Make the constructor private then?

core/src/main/java/org/apache/james/core/quota/QuotaScope.java
Arsnael: Make the constructor private then?

core/src/main/java/org/apache/james/core/quota/QuotaType.java
Arsnael: Make the constructor private then?

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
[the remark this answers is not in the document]
hungphan227: I see that we still could use Username class that have the same logic as QuotaRoot in terms of username. Username class even provide more functions for username than QuotaRoot

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
```diff
@@ -0,0 +1,82 @@
...
+
+public class QuotaLimit {
+
+    public static QuotaLimit of(QuotaComponent component, QuotaScope scope, String identifier, QuotaType quotaType, Long limit) {
+        return new QuotaLimit(component, scope, identifier, quotaType, limit);
+    }
+
+    private final QuotaComponent quotaComponent;
+    private final QuotaScope quotaScope;
+    private final String identifier;
```
vttranlina, line 33: you should consistency type class of `identifier` here, 
```suggestion
    private final Username identifier;
```
same with QuotaCurrentValue.java

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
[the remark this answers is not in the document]
vttranlina: Not sure, but I just quickly check `com.linagora.tmail.team.TMailQuotaRootResolver`, Quotaroot is not only Username, It depends on which impl of `UserQuotaRootResolver` 
Example quotaRoot of TeamMailbox

```
def quotaRoot: QuotaRoot = QuotaRoot.quotaRoot(s"$TEAM_MAILBOX_NAMESPACE&${mailboxName.value.value}@${domain.asString()}", Some(domain).toJava)
```

core/src/main/java/org/apache/james/core/quota/QuotaLimit.java
hungphan227: The identifier of QuotaLimit is not necessarily username. It could be other things (domain name for example)

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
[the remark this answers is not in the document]
hungphan227: Yeah actually I am a little bit confused now. I found an example of quota root that looks strange in DefaultUserQuotaRootResolverTest, it is "private&benwa". Does anyone know what this mean?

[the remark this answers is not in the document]
quantranhong1999: IMO `#private` is namespace

core/src/main/java/org/apache/james/core/StringWrapper.java
quantranhong1999, line 26: From what I can see: `QuotaComponent` still can work without extending `StringWrapper`.

Q: What is the point of introducing this  `StringWrapper`?

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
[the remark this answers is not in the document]
quantranhong1999: I think the same as Rene.
`QuotaRoot` is a kind of ID for quota already.

> Username class even provide more functions for username than QuotaRoot

QuotaRoot exists for a reason, otherwise, the current quota code would use Username directly instead of QuotaRoot, no?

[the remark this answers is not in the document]
hungphan227: what is namespace? :))

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
```diff
@@ -0,0 +1,45 @@
...
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.core.quota;
+
+import org.apache.james.core.StringWrapper;
+
+public class QuotaComponent extends StringWrapper {
```
quantranhong1999, line 24: Isn't
```
public class QuotaComponent {
    private final String value;
    // constructor, getter, equals, hashcode, toString
}
```
Enough?

rf: https://github.com/apache/james-project/pull/1666#discussion_r1282614846

core/src/main/java/org/apache/james/core/StringWrapper.java
hungphan227: to reuse some code in other classes: QuotaType, QuotaScope

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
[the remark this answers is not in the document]
quantranhong1999: > what is namespace? :))

https://www.rfc-editor.org/rfc/rfc2342.html

QUOTAROOT IMAP commands spec: https://www.rfc-editor.org/rfc/rfc2087

QUOTAROOT IMAP example: https://github.com/apache/james-project/blob/master/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Quota.test

It looks like QuotaRoot is actually the id of quota, and is specially bound to the IMAP concept. As Rene said, I think that we should still respect the QuotaRoot here.

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
hungphan227: No it is not enough. We should restrict the value of QuotaComponent within three values mailbox, sieve, jmapUploads

quantranhong1999: The way how we blueprint the `QuotaComponent` class is not related to `restrict the value of QuotaComponent`, no?

How is that related to `StringWrapper` usage BTW?

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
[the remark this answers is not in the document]
hungphan227: If I use QuotaRoot, I will have to use something like "private&benwa" in which I do not really understand why we need the keywork private

core/src/main/java/org/apache/james/core/StringWrapper.java
quantranhong1999: If you mean "reuse some code" is reuse the String attribute then I am not convinced.

Why just not declare that string attribute directly in `QuotaType`, `QuotaScope`, and `QuotaComponent`?

IMO the types of classes `QuotaType`, `QuotaScope`, and `QuotaComponent` have nothing in common (imagine in the real world) that needs inheritance. String is just an implementation detail of those class, extending all of them from `StringWrapper` just add more complexity and potentially misunderstand binding IMO.

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
chibenwa: +1 with Quan. The `StringWrapper` is a workaround for a language limitation.

Complex hierarchy should be avoided and 'flat' POJOs shall be prefered.

If you really want to limit the boiler plate here then you could consider adding in some scala. Or waiting a Java 17+ switch.

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
```diff
@@ -0,0 +1,45 @@
...
+    public static QuotaComponent of(String value) {
+        return new QuotaComponent(value);
+    }
+
+    private QuotaComponent(String value) {
+        super(value);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
chibenwa, line 39: does this pass equals verifier? No hashCode, not final...

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
chibenwa, line 28: Builder please?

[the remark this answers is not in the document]
Arsnael: Forget the Quota Root actually :)

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
hungphan227: deleted

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
hungphan227: changed to quota key and currentvalue -> too few fields -> builder not needed

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
```diff
@@ -0,0 +1,124 @@
...
+
+import java.util.Objects;
+
+import org.apache.james.core.Username;
+
+import com.google.common.base.MoreObjects;
+
+public class QuotaCurrentValue {
+
+    public static class QuotaKey {
```
Arsnael, line 30: Those 2 QuotaKey classes that look similar but a bit different is looking wrong and confusing to me. 

Or we:
- find a way to unify that into one POJO (and separated, not a subclass of an other POJO) => then maybe using QuotaScope as identifier and finding a good structure would make sense then
- just sticking to builders here (need to add them), and have the QuotaKey for the DAO with current values only (next PR), if sticking to Username as identifier.

I start to think though now that refactoring QuotaScope to fit the identifier and scope needs might be the better way here

quantranhong1999 (commented):
read it. nothing to add.

core/src/main/java/org/apache/james/core/quota/QuotaCurrentValue.java
hungphan227: After reading Benoit's comment, I feel it is better to group these field in a subclass as a key. Not sure why you are confused.

Arsnael: He suggested to do one class QuotaKey, not two?

hungphan227: I am not sure what he mean exactly but it gives me the idea :))

Arsnael: I think you go the idea wrong. The fact that @quantranhong1999 put a thumb up on the comment above means he likely agrees to the comment as well

Please rework on this :)

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.commons.lang3.StringUtils;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Preconditions;
+
+public class QuotaComponent {
+
+    public static final QuotaComponent MAILBOX = of("mailbox");
+    public static final QuotaComponent SIEVE = of("sieve");
+    public static final QuotaComponent JMAP_UPLOADS = of("jmapUploads");
```
chibenwa, line 33: We don't need to know all the values here, and likely should not.

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
```diff
@@ -0,0 +1,70 @@
...
+        return value;
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(value);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
chibenwa, line 56: Why not final?

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
```diff
@@ -0,0 +1,70 @@
...
+    private QuotaComponent(String value) {
+        this.value = value;
+    }
+
+    public String getValue() {
+        return value;
+    }
+
+    @Override
+    public int hashCode() {
```
chibenwa, line 51: Why not final?

chibenwa (commented):
I do not get why these POJOs are in `/core` and not in `backend-cassandra`

hungphan227:
> I do not get why these POJOs are in `/core` and not in `backend-cassandra`

I see that old quota POJOs are in `/core` so I put new POJOs in there as well

core/src/main/java/org/apache/james/core/quota/QuotaComponent.java
hungphan227: These constants would be used in other functions. For example, JMAP_UPLOADS would be used in jmap upload function and recomputation function.

Arsnael (approved):
Looks good to me :)

quantranhong1999 (approved):

vttranlina (approved):

PR #1672, opened 2023-08-09
[Build] Better stability with Cassandra docker containers and integration tests

Arsnael:
A lots of builds seems failing recently on Docker Cassandra container failing to start up correctly when trying to run integration tests. It's not impossible that Cassandra 4 requires more time and resources to setup than version 3. So let's try with this, run a couple of builds and see if it gets better.

### DIFF
--- a/server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/S3MinioTest.java
+++ b/server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/S3MinioTest.java
@@ -119,5 +119,4 @@ public class S3MinioTest implements BlobStoreDAOContract {
             .then()
             .block();
     }
-
 }
--- a/backends-common/cassandra/src/test/java/org/apache/james/backends/cassandra/CassandraWaitStrategy.java
+++ b/backends-common/cassandra/src/test/java/org/apache/james/backends/cassandra/CassandraWaitStrategy.java
@@ -34,7 +34,7 @@ import com.google.common.primitives.Ints;
 
 public class CassandraWaitStrategy implements WaitStrategy {
 
-    private static final Duration DEFAULT_TIMEOUT = Duration.ofMinutes(2);
+    private static final Duration DEFAULT_TIMEOUT = Duration.ofMinutes(5);
     private final GenericContainer<?> cassandraContainer;
     private final Duration timeout;
 
--- a/backends-common/cassandra/src/test/java/org/apache/james/backends/cassandra/DockerCassandra.java
+++ b/backends-common/cassandra/src/test/java/org/apache/james/backends/cassandra/DockerCassandra.java
@@ -123,7 +123,7 @@ public class DockerCassandra {
     }
 
     private static final int CASSANDRA_PORT = 9042;
-    private static final int CASSANDRA_MEMORY = 750;
+    private static final int CASSANDRA_MEMORY = 1024;
 
     private static final String CASSANDRA_CONFIG_DIR = "$CASSANDRA_CONFIG";
     private static final String JVM_OPTIONS = CASSANDRA_CONFIG_DIR + "/jvm.options";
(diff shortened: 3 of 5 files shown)


vttranlina (approved):

chibenwa (approved):

hungphan227 (approved):

quantranhong1999 (approved):

Arsnael:
Been twice I get a timeout in the fork on webadmin-cli module, which I dont see it being related to any recent changes... Am I just unlucky here? :(

vttranlina:
on my local

image


on jenkin ci
image
only Skipped, no failed, no error

Arsnael:
Look at the logs, you will see this around the end:

```
14:06:54,732 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.1.2:test (default-test) on project webadmin-cli: There was a timeout in the fork -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:3.1.2:test (default-test) on project webadmin-cli: There was a timeout in the fork
```

quantranhong1999:
`on project webadmin-cli: There was a timeout in the fork` Well seems 3rd time in a row...

I doubted that increasing Cassandra's startup timeout leads to timeout for `webadmin-cli`. But actually, that module has no tests related to Cassandra and should be very lightweight.

Maybe spitting the S3Minio test fix commit to another PR first?

server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/S3MinioTest.java
```diff
@@ -120,4 +121,9 @@ void saveShouldWorkForAllPayloadsWithHash() {
             .block();
     }
 
+    @Disabled("Minio changed and returns a generic S3Exception instead of a NoSuchBucketException in its latest release")
```
chibenwa, line 124: Please better describe the impact on existing features: does this impact some features used by us? By our customers? 

Let's document MINIO limitations on the S3 config page on the Antora website maybe?

Arsnael: I dont think customer mentions should figure on an apache open source project comments?

But I will enhance that a bit. Difference is if you try to delete a bucket via our API with MinIO, instead of it being silenced if the bucket does not exist (making it transparent to the user), James will return an error.

Arsnael: => https://github.com/apache/james-project/pull/1675

Separated PR, as suggested by @quantranhong1999 , as this issue is a blocker on master branch but we seem to have something else suspicious on the build with other changes. 

But it addresses your concern.

Arsnael:
Had multiple issues of timeout in some modules with those commits (when we test, if it takes longer than 20 mins for a module, we consider it failing, in maven-surefire-plugin conf), like `webadmin-cli` and/or `james-server-cli-integration-test` that could take longer than 20mins, or even when less more than 10, while in latest master green build it was still between 1 and 2 minutes

This build is green, but still the results on those modules aren't good:

```
12:15:37,164 [INFO] Apache James :: Server :: Cli :: Integration tests . SUCCESS [14:31 min]
12:15:37,165 [INFO] Apache James :: Server :: Web Admin CLI ............ SUCCESS [19:53 min]
```

Limit of failing with webadmin-cli here, at a few seconds away...

However, https://github.com/apache/james-project/pull/1676 => back to between 1 and 2 minutes

```
10:58:14,058 [INFO] Apache James :: Server :: Cli :: Integration tests . SUCCESS [01:03 min]
10:58:14,058 [INFO] Apache James :: Server :: Web Admin CLI ............ SUCCESS [01:38 min]
```

The timeout increase on Cassandra waiting strategy seems to be the culprit... I know we fire tests on different modules in parallel, not completely sure how it works under the hood with the maven-surefire-plugin but maybe it created issues of blocking threads much longer when spawning Cassandra containers, in the middle of testing other modules? Or something around that... \

Anyway will merge  https://github.com/apache/james-project/pull/1676 asap as the master branch has been red for too long. If you have other comments to do feel free, will be checked upon later...

PR #1675, opened 2023-08-14
[FIX] Latest MinIO image is returning a generic S3Exception and not NoSuchBucketException when trying to get an unexisting bucket

Arsnael:


server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/S3MinioTest.java
```diff
@@ -120,4 +121,10 @@ void saveShouldWorkForAllPayloadsWithHash() {
             .block();
     }
 
+    @Disabled("Minio changed and returns a generic S3Exception instead of a NoSuchBucketException in its latest release." +
+        "Consequence is trying to delete a blob on a non existing bucket on MinIO will throw an error, instead of being silenced.")
+    @Override
+    public void deleteShouldNotThrowWhenBucketDoesNotExist() {
+
+    }
```
quantranhong1999, line 129: Adapt the test to this would fix the failure for MinIO:

```java
    @Test
    default void deleteShouldNotThrowWhenBucketDoesNotExist() {
        BlobStoreDAO store = testee();

        assertThatCode(() -> Mono.from(store.delete(BucketName.of("not-existing-bucket-name"), TEST_BLOB_ID)).block())
            .doesNotThrowAnyException();
    }
```

Rationale: Recently MinIO applies more strictly the rules for bucket names following AWS S3 standard (no underscore in bucket name)
rf: https://github.com/minio/minio/pull/17742, https://github.com/conda/conda/pull/12986

Arsnael: Oh well caught, the issue is actually the underscores and not something else... Alright!

quantranhong1999 (approved):

Arsnael:
cf https://github.com/apache/james-project/pull/1672#issuecomment-1678353087

PR #1676, opened 2023-08-14
[DUPLICATED] Better stability with Cassandra docker containers and integration tests

Arsnael:
Same as https://github.com/apache/james-project/pull/1672 but without the timeout increase of Cassandra waiting strategy, as I ain't sure about what exactly is creating those timeouts in webadmin-cli (which I'm surprised as none of those commits should have a change on that)


quantranhong1999:
green build :D

Arsnael:
cd https://github.com/apache/james-project/pull/1672#issuecomment-1678353087

PR #1678, opened 2023-08-14
[DUPLICATED] Better stability with Cassandra docker containers and integration tests

Arsnael:
Same as https://github.com/apache/james-project/pull/1672 but by trying to isolate webadmin-cli test by putting reuseForks to false


Arsnael:
cf https://github.com/apache/james-project/pull/1672#issuecomment-1678353087

PR #1681, opened 2023-08-16
[BUILD] Changing OpenSearch waiting strategy

Arsnael:
Let's try with this, idea taken from there: https://github.com/opensearch-project/opensearch-testcontainers/blob/main/src/main/java/org/opensearch/testcontainers/OpensearchContainer.java#L104


chibenwa (approved):
Good catch <3

PR #1685, opened 2023-08-17
[BUILD] Fix unstable SpamTrapHandlerTest.testSpamTrap test

quantranhong1999:
resolve https://github.com/linagora/james-project/issues/4844

Rely on UpdatableTickingClock to avoid CPU resource race condition on the CI that leads to test unstable.

### DIFF
--- a/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/fastfail/SpamTrapHandler.java
+++ b/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/fastfail/SpamTrapHandler.java
@@ -20,11 +20,14 @@
 
 package org.apache.james.protocols.smtp.core.fastfail;
 
+import java.time.Clock;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+
+import javax.inject.Inject;
 
 import org.apache.james.core.MailAddress;
 import org.apache.james.core.MaybeSender;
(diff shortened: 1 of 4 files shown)


protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/fastfail/SpamTrapHandler.java
```diff
@@ -42,13 +45,20 @@ public class SpamTrapHandler implements RcptHook {
...
+    private final Map<String,Long> blockedIps;
+    private final Clock clock;
     
     private Collection<String> spamTrapRecips = new ArrayList<>();
     
     /** Default blocktime 12 hours */
     protected long blockTime = 4320000;
 
+    @Inject
+    public SpamTrapHandler(Clock clock) {
```
chibenwa, line 57: This would beak Spring injection.

Please put two constructor: @Inject on a constructor with not arguments relying on system clock, and an other one to specify the clock in tests.

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/fastfail/SpamTrapHandler.java
```diff
@@ -42,13 +45,20 @@ public class SpamTrapHandler implements RcptHook {
     private static final Logger LOGGER = LoggerFactory.getLogger(SpamTrapHandler.class);
 
     /** Map which hold blockedIps and blockTime in memory */
-    private final Map<String,Long> blockedIps = new HashMap<>();
+    private final Map<String,Long> blockedIps;
```
chibenwa, line 48: ConcurrentHashMap + remove synchronze while we are at it please

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/fastfail/SpamTrapHandler.java
```diff
@@ -54,8 +56,14 @@ public class SpamTrapHandler implements RcptHook {
     protected long blockTime = 4320000;
 
     @Inject
+    public SpamTrapHandler() {
+        this.blockedIps = new ConcurrentHashMap<>();
+        this.clock = Clock.systemUTC();
```
chibenwa, line 61: Call the other constructor with the clock...

Arsnael (approved):

PR #1686, opened 2023-08-17
DRAFT JAMES-4839 jmap quota for uploads implement a web admin api to update limit of uploads

hungphan227:


server/container/guice/protocols/webadmin-mailbox/src/main/java/org/apache/james/modules/server/MailboxRoutesModule.java
```diff
@@ -53,6 +54,7 @@ protected void configure() {
         routesMultibinder.addBinding().to(EventDeadLettersRoutes.class);
         routesMultibinder.addBinding().to(GlobalQuotaRoutes.class);
         routesMultibinder.addBinding().to(UserQuotaRoutes.class);
+        routesMultibinder.addBinding().to(QuotaRoutes.class);
```
chibenwa, line 57: SHould be located in JMAP guice module.

server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/routes/QuotaRoutes.java
```diff
@@ -0,0 +1,130 @@
...
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.guava.GuavaModule;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+
+import spark.Service;
+
+public class QuotaRoutes implements Routes {
```
chibenwa, line 48: Should be named "JMAP QuotaRoutes" and should be located in webadmin-jmap

chibenwa (changes requested):
Stopped reading.

This exposes directly the cassandra backend for quota (IE a technical capability) while we MUDST expose applicative feature here. We need to have a JMAP interface for managing quota for uploads. We need to provide a memory implementation and a contract test for this API. We ned to implement the Cassandra version of that API with CassandraQuotaLimitDao, and the webadmin API needs to rely on the JMAP API, nt on storage level details.

My 2 cents: 
 - Start defning a JMAP API for quota for uploads, provide a memory implementation of it, and a contract test.
  - We will come back to the JMAP API later.

chibenwa:
Ticket number in PR titl and from commits is taken from the ASF JIRA, NOT from github issues.

MUST BE [JAMES-3925](https://issues.apache.org/jira/browse/JAMES-3925)

PR #1699, opened 2023-08-24
JAMES-3925 Define an API to store current values of JMAP uploads

hungphan227:


server/data/data-jmap-cassandra/src/test/java/org/apache/james/jmap/cassandra/upload/CassandraUploadUsageRepositoryTest.java
```diff
@@ -0,0 +1,62 @@
...
+        resetCounterToZero();
+    }
+
+    private void resetCounterToZero() {
+        Mono.from(cassandraUploadUsageRepository.increaseSpace(USER_NAME(), QuotaSizeUsage.size(0))).block();
+        QuotaSizeUsage quotaSizeUsage = Mono.from(cassandraUploadUsageRepository.getSpaceUsage(USER_NAME())).block();
+        Mono.from(cassandraUploadUsageRepository.decreaseSpace(USER_NAME(), quotaSizeUsage)).block();
+        QuotaSizeUsage actual = Mono.from(cassandraUploadUsageRepository.getSpaceUsage(USER_NAME())).block();
+        assertThat(actual.asLong()).isEqualTo(0l);
+    }
```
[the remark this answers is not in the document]
hungphan227: why?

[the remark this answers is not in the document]
chibenwa: Where is this called?

How is test isolation enforced? (by the cassandra tooling....)

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/upload/UploadUsageRepositoryContract.scala
```diff
@@ -0,0 +1,48 @@
...
+    Assertions.assertThat(SMono.fromPublisher(uploadUsageRepository.getSpaceUsage(USER_NAME)).block().asLong()).isEqualTo(0);
+    SMono.fromPublisher(uploadUsageRepository.increaseSpace(USER_NAME, QuotaSizeUsage.size(100))).block();
+    val expected = SMono.fromPublisher(uploadUsageRepository.getSpaceUsage(USER_NAME)).block();
+    Assertions.assertThat(expected.asLong()).isEqualTo(100);
+  }
+
+  @Test
+  def decreaseSpaceShouldDecreaseSuccessfully(): Unit = {
+    SMono.fromPublisher(uploadUsageRepository.increaseSpace(USER_NAME, QuotaSizeUsage.size(200))).block();
+    Assertions.assertThat(SMono.fromPublisher(uploadUsageRepository.getSpaceUsage(USER_NAME)).block().asLong()).isEqualTo(200);
```
Arsnael, line 28: Assert unecessary. You have a test to prove that increase works as expected already

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/upload/UploadUsageRepositoryContract.scala
```diff
@@ -0,0 +1,48 @@
...
+    SMono.fromPublisher(uploadUsageRepository.decreaseSpace(USER_NAME, QuotaSizeUsage.size(100))).block();
+    val expected = SMono.fromPublisher(uploadUsageRepository.getSpaceUsage(USER_NAME)).block();
+    Assertions.assertThat(expected.asLong()).isEqualTo(100);
+  }
+
+  @Test
+  def resetSpaceShouldResetSuccessfully(): Unit = {
+    SMono.fromPublisher(uploadUsageRepository.increaseSpace(USER_NAME, QuotaSizeUsage.size(200))).block();
+    Assertions.assertThat(SMono.fromPublisher(uploadUsageRepository.getSpaceUsage(USER_NAME)).block().asLong()).isEqualTo(200);
```
Arsnael, line 37: Idem

server/data/data-jmap-cassandra/src/test/java/org/apache/james/jmap/cassandra/upload/CassandraUploadUsageRepositoryTest.java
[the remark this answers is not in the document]
Arsnael: You definitely shouldn't have an assert in the init phase outside on the contract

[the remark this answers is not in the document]
Arsnael: You should put the init BeforeEach in the contract

server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/upload/CassandraUploadUsageRepository.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private CassandraQuotaCurrentValueDao cassandraQuotaCurrentValueDao;
+
+    @Inject
+    public CassandraUploadUsageRepository(CassandraQuotaCurrentValueDao cassandraQuotaCurrentValueDao) {
+        this.cassandraQuotaCurrentValueDao = cassandraQuotaCurrentValueDao;
+    }
+
+    @Override
+    public Publisher<Void> increaseSpace(Username username, QuotaSizeUsage usage) {
```
vttranlina, line 46: Publisher -> Mono
more easier when somewhere invokes it

server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/upload/CassandraUploadUsageRepository.java
```diff
@@ -0,0 +1,68 @@
...
+    public Publisher<Void> decreaseSpace(Username username, QuotaSizeUsage usage) {
+        return cassandraQuotaCurrentValueDao.decrease(CassandraQuotaCurrentValueDao.QuotaKey.of(QuotaComponent.JMAP_UPLOADS, username.asString(), QuotaType.SIZE),
+            usage.asLong());
+    }
+
+    @Override
+    public Publisher<QuotaSizeUsage> getSpaceUsage(Username username) {
+        return cassandraQuotaCurrentValueDao.getQuotaCurrentValue(CassandraQuotaCurrentValueDao.QuotaKey.of(QuotaComponent.JMAP_UPLOADS, username.asString(), QuotaType.SIZE))
+            .map(quotaCurrentValue -> QuotaSizeUsage.size(quotaCurrentValue.getCurrentValue())).defaultIfEmpty(QuotaSizeUsage.size(0));
+    }
```
vttranlina, line 61: `QuotaSizeUsage.size(0)` it should be a static variable

server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/upload/CassandraUploadUsageRepository.java
hungphan227: I saw in your code you use Publisher as well. Could you explain why?
```
    public Publisher<Void> delete(UploadId id, Username user) {
        return uploadDAO.delete(user, id);
    }
```

vttranlina: yes, it should be better if `public Mono<Void> delete`, my lazy when IDE auto-suggests complete code, 

My comment is because I see the `resetSpace` method use `Mono` to wrap the publisher in the same "internal" class

```
return Mono.from(getSpaceUsage(username)).flatMap(quotaSizeUsage -> Mono.from(decreaseSpace(username, quotaSizeUsage)))
            .then(Mono.from(increaseSpace(username, usage)));
```

it can be 
```
return getSpaceUsage(username).flatMap(quotaSizeUsage -> decreaseSpace(username, quotaSizeUsage))
            .then(increaseSpace(username, usage));
```

hungphan227: Actually this is the thing I has been confused so far. If we know for sure the return type (Mono or Flux), we should use Mono or Flux instead of Publisher, right?

Arsnael: Publisher -> API definition
Implementation -> Mono or Flux

vttranlina: yes, 
- Mono, Flux for java implement, 
- SMono, SFlux for scala implement
- Publisher for interface/ trait

Arsnael: Yes not always respected (but should try best to do so)
I think the rationale back in the day was well if somebody use something else than reactor, he could still do his own implem with his own reactive framework other than Reactor... 

Think of James as a big toolbox trying to be flexible enough to let people customize it a bit the way they want it :)

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/upload/UploadUsageRepository.java
```diff
@@ -21,14 +21,15 @@
 
 import org.apache.james.core.Username;
 import org.apache.james.core.quota.QuotaSizeUsage;
-import org.reactivestreams.Publisher;
+
+import reactor.core.publisher.Mono;
 
 public interface UploadUsageRepository {
-    Publisher<Void> increaseSpace(Username username, QuotaSizeUsage usage);
+    Mono<Void> increaseSpace(Username username, QuotaSizeUsage usage);
```
vttranlina: No, it is an interface class, should keep `Publisher`

vttranlina: resolved

vttranlina (approved):

Arsnael (approved):

server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/upload/CassandraUploadUsageRepository.java
```diff
@@ -0,0 +1,67 @@
...
+    public Mono<Void> decreaseSpace(Username username, QuotaSizeUsage usage) {
+        return cassandraQuotaCurrentValueDao.decrease(CassandraQuotaCurrentValueDao.QuotaKey.of(QuotaComponent.JMAP_UPLOADS, username.asString(), QuotaType.SIZE),
+            usage.asLong());
+    }
+
+    @Override
+    public Mono<QuotaSizeUsage> getSpaceUsage(Username username) {
+        return cassandraQuotaCurrentValueDao.getQuotaCurrentValue(CassandraQuotaCurrentValueDao.QuotaKey.of(QuotaComponent.JMAP_UPLOADS, username.asString(), QuotaType.SIZE))
+            .map(quotaCurrentValue -> QuotaSizeUsage.size(quotaCurrentValue.getCurrentValue())).defaultIfEmpty(DEFAULT_QUOTA_SIZE_USAGE);
+    }
```
quantranhong1999, line 60: From the ticket: 
```
    Publisher<QuotaSizeUsage> spaceUsag(User user); // Need to log WARNING upon negative values and sanitize them to ZERO
```

Need to handle this?

server/data/data-jmap/src/main/java/org/apache/james/jmap/memory/upload/InMemoryUploadUsageRepository.java
```diff
@@ -0,0 +1,74 @@
...
+                cache.put(username, new AtomicReference<>(QuotaSizeUsage.size(amount)));
+            } else {
+                quotaSizeUsageAtomicReference.updateAndGet(quotaSizeUsage -> quotaSizeUsage.add(amount));
+            }
+        });
+    }
+
+    @Override
+    public Mono<QuotaSizeUsage> getSpaceUsage(Username username) {
+        return Mono.just(cache.getOrDefault(username, new AtomicReference<>(DEFAULT_QUOTA_SIZE_USAGE))).map(quotaSizeUsageAtomicReference -> quotaSizeUsageAtomicReference.get());
```
quantranhong1999, line 66:
```suggestion
        return Mono.just(cache.getOrDefault(username, new AtomicReference<>(DEFAULT_QUOTA_SIZE_USAGE)))
                .map(quotaSizeUsageAtomicReference -> quotaSizeUsageAtomicReference.get());
```

"One dot per line" would be more readable IMO.

quantranhong1999 (approved):

server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/upload/CassandraUploadUsageRepository.java
chibenwa: IMO after thinking about it we should log a warning and recompute this one quota staight away...

hungphan227: ok so we still return 0, log a warning and run recomputation asynchronously immediately. Is it ok?

chibenwa: Yes that would be ok

PR #1702, opened 2023-08-25
JAMES-3926 Adopt Cassandra quota helpers for Mailbox (current) quota

Arsnael:


mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/quota/CassandraCurrentQuotaManagerV2.java
```diff
@@ -0,0 +1,111 @@
...
+            .map(QuotaCurrentValue::getCurrentValue)
+            .map(QuotaSizeUsage::size)
+            .defaultIfEmpty(QuotaSizeUsage.size(0L));
+    }
+
+    @Override
+    public Mono<CurrentQuotas> getCurrentQuotas(QuotaRoot quotaRoot) {
+        return Mono.zip(
+                getCurrentMessageCount(quotaRoot),
+                getCurrentStorage(quotaRoot))
```
chibenwa, line 68: Only the QotaType differ. If I'm not wrong we should be able to manage this in one primary key read...

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/quota/CassandraCurrentQuotaManagerV2.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public Mono<Void> setCurrentQuotas(QuotaOperation quotaOperation) {
+        return getCurrentQuotas(quotaOperation.quotaRoot())
+            .filter(Predicate.not(Predicate.isEqual(CurrentQuotas.from(quotaOperation))))
+            .flatMap(storedQuotas -> decrease(quotaOperation.quotaRoot(), storedQuotas.count(), storedQuotas.size())
+                .then(increase(quotaOperation)));
+
+    }
```
chibenwa, line 95: extra line

mailbox/cassandra/src/test/java/org/apache/james/mailbox/cassandra/quota/migration/CassandraCurrentQuotaManagerMigrationTest.java
```diff
@@ -0,0 +1,127 @@
...
+        CassandraQuotaModule.MODULE,
+        CassandraUsersRepositoryModule.MODULE,
+        CassandraDomainListModule.MODULE,
+        CassandraMutualizedQuotaModule.MODULE,
+        CassandraUidModule.MODULE,
+        CassandraBlobModule.MODULE,
+        CassandraMailboxModule.MODULE,
+        CassandraModSeqModule.MODULE,
+        CassandraThreadModule.MODULE,
+        CassandraSchemaVersionModule.MODULE));
```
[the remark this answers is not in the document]
chibenwa: maybe a mock might actually be simpler here? (for once?)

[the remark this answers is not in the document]
Arsnael: Thought about it, was trying to avoid it but... yeah :)

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/quota/CassandraCurrentQuotaManagerV2.java
Arsnael: Hmmm I agree with this. However, the mutualized quota DAO being in cassandra backends, and the POJO having both count and size (CurrentQuotas) being very mailbox specific, we don't want to import the mailbox-api in the backends I think?

I either need to create a new POJO in james-core (as it seems our generic quota structs are stored there) or move CurrentQuotas in james-core with a little refactoring I think to do that cleanly. But is that acceptable? Of course I would need to add an extra method in the dao as well

WDYT?

chibenwa: > move CurrentQuotas

We can try.

Arsnael: Tried something, not really completely convinced but I could not think of any better for now

Arsnael:
Ready for review

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/quota/CassandraCurrentQuotaManagerV2.java
```diff
@@ -0,0 +1,110 @@
...
+    public CassandraCurrentQuotaManagerV2(CassandraQuotaCurrentValueDao currentValueDao) {
+        this.currentValueDao = currentValueDao;
+    }
+
+    @Override
+    public Mono<QuotaCountUsage> getCurrentMessageCount(QuotaRoot quotaRoot) {
+        return currentValueDao.getQuotaCurrentValue(asQuotaKeyCount(quotaRoot))
+            .map(QuotaCurrentValue::getCurrentValue)
+            .map(QuotaCountUsage::count)
+            .defaultIfEmpty(QuotaCountUsage.count(0L));
```
quantranhong1999, line 55: <3 Shorten syntax for `.switchIfEmpty`

quantranhong1999 (approved):

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/quota/CassandraCurrentQuotaManagerV2.java
```diff
@@ -0,0 +1,129 @@
...
+    private Mono<Void> decrease(QuotaRoot quotaRoot, QuotaCountUsage count, QuotaSizeUsage size) {
+        return currentValueDao.decrease(asQuotaKeyCount(quotaRoot), count.asLong())
+            .then(currentValueDao.decrease(asQuotaKeySize(quotaRoot), size.asLong()));
+    }
+
+    @Override
+    public Mono<Void> setCurrentQuotas(QuotaOperation quotaOperation) {
+        return getCurrentQuotas(quotaOperation.quotaRoot())
+            .filter(Predicate.not(Predicate.isEqual(CurrentQuotas.from(quotaOperation))))
+            .flatMap(storedQuotas -> decrease(quotaOperation.quotaRoot(), storedQuotas.count(), storedQuotas.size())
```
vttranlina, line 111: `decrease` then `increase` => we need 4 queries
I think we can evaluate the "changed" before query to the database 
Ex: 
```suggestion
             .flatMap(storedQuotas -> {
                long count = quotaOperation.count().asLong() - storedQuotas.count().asLong();
                long size = quotaOperation.size().asLong() - storedQuotas.size().asLong();

                return currentValueDao.increase(asQuotaKeyCount(quotaOperation.quotaRoot()), count)
                    .then(currentValueDao.increase(asQuotaKeySize(quotaOperation.quotaRoot()), size));
            });
```
we need 2 queries with this way

Arsnael: cf fixed in https://github.com/apache/james-project/pull/1702/commits/85bef866fb552c5935b0273782f232dae61e2806

Didnt want to make a fixup, the CI seems too capricious lately, would be happy to not have to squash and wait again after a green build :)

Arsnael:
Don't pay attention to the branch build, I pushed a branch on apache by mistake and that triggered a build even if I didnt open a PR from it for some reason...

vttranlina (approved):

hungphan227 (approved):

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1702/13/testReport/

smells like not a good isolation between tests...

PR #1708, opened 2023-09-01
JAMES-3604 Enable to use quorum queues onto DTM work queue

chibenwa:


chibenwa:
CF https://github.com/linagora/james-project/pull/4858

Arsnael (approved):

PR #1729, opened 2023-09-15
JAMES-3940 Set up TTL on mailbox_change and email_change tables

quantranhong1999:


server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/change/CassandraChangesConfiguration.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+import org.apache.commons.configuration2.Configuration;
+import org.apache.james.util.DurationParser;
+
+import com.google.common.base.Preconditions;
+
+public class CassandraChangesConfiguration {
+    public static final Duration DEFAULT_TTL = Duration.ofDays(60);
```
chibenwa, line 33: Then may add a mention in upgrade instruction that newer versions of James will use a TTL by default and that in order to preserve the existing behaviour one needs to configure X, Y Z...

Arsnael (approved):

Arsnael:
It's not trivial in the doc btw that if you put just a number it's in seconds and that you can add the unit of duration as well alongside it for example ?

Arsnael (approved):

chibenwa (approved):

PR #1739, opened 2023-09-26
JAMES-3822 MailQueue::enqueueReactive is blocking and should be executed on the boundedElastic scheduler

quantranhong1999:
`enQueue` executed a `.block` on the Cassandra thread.

RabbitMQ was not effected thanks to its own reactive implementation. While other implementations e.g. Pulsar were blocked.


Arsnael (approved):

Arsnael:
Good catch :)

chibenwa (approved):

vttranlina (approved):

PR #1743, opened 2023-10-02
JAMES-3897 - Crowdsec integration for IMAP Authentication

thanhbv200585:


server/apps/distributed-app/docker-compose-crowdsec.yml
```diff
@@ -0,0 +1,113 @@
...
+        condition: service_healthy
+      opensearch:
+        condition: service_started
+      rabbitmq:
+        condition: service_started
+      s3:
+        condition: service_started
+      pulsar:
+        condition: service_started
+    image: apache/james:distributed-latest
```
quantranhong1999, line 17: James memory + CrowdSec should be more lightweight for a POC. Anyway not important.

thanhbv200585:
James will feed this log to Crowdsec. Since IP 172.12.1.4 tried to attack brute force login so Crowdsec implies to ban this IP.
image
image

server/apps/distributed-app/docker-compose-crowdsec.yml
```diff
@@ -0,0 +1,113 @@
...
+      - "110:110"
+      - "143:143"
+      - "465:465"
+      - "587:587"
+      - "993:993"
+      - "8000:8000"
+    volumes:
+      - ./sample-configuration/pulsar.properties:/root/conf/pulsar.properties
+      - ./sample-configuration/queue-pulsar.properties:/root/conf/queue.properties
+      - ./logs:/logs/
```
quantranhong1999, line 36: Create a docker volume that is shared between James container and CrowdSec, rather than volume to the host file system?

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
```diff
@@ -73,6 +73,7 @@ public void configure(ImapConfiguration imapConfiguration) {
 
     protected void doAuth(AuthenticationAttempt authenticationAttempt, ImapSession session, ImapRequest request, Responder responder, HumanReadableText failed) {
         Preconditions.checkArgument(!authenticationAttempt.isDelegation());
+        LOGGER.warn("Failed IMAP login from: 172.12.1.4");
```
quantranhong1999, line 76: Likely need to propagate the remoteIP from Netty to `IMAPSession` then.

quantranhong1999 (commented):
Looks cool to me! 
Any proposition work following this POC?

server/apps/distributed-app/docker-compose-crowdsec.yml
thanhbv200585: Make sense! I will fix it

server/apps/distributed-app/crowdsec/scenarios/james-bf-auth.yaml
```diff
@@ -0,0 +1,13 @@
+type: leaky
+name: crowdsecurity/james
+description: "Detect login james bruteforce"
+filter: "evt.Meta.log_type == 'james_failed-auth'"
+leakspeed: "10s"
+capacity: 5
```
quantranhong1999, line 6: Is this the configuration for the maximum failure authentication?

e.g. after 5 failure authentication in 10s -> we trigger this scenario.

vttranlina:
Good first shot
Can you explain a bit in the description?

server/apps/distributed-app/crowdsec/acquis.yaml
```diff
@@ -0,0 +1,5 @@
+source: file
```
chibenwa, line 1: Move this to `/third-party/crowdsec` please

server/apps/distributed-app/crowdsec/parsers/james-auth.yaml
```diff
@@ -0,0 +1,23 @@
+onsuccess: next_stage
+debug: true
+filter: "evt.Parsed.program == 'james'"
+name: linagora/james-connect-security-log
+description: "Parser for James Connect Security Log"
+
+pattern_syntax:
+  JAMES_AUTH_FAIL: '%{TIME:time} \[%{LOGLEVEL:level} \] %{JAVACLASS:class} - Failed IMAP login from: %{IPV4:IP}'
```
chibenwa, line 8: Please use JSON log formats, and not flat log formats

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
chibenwa: Ip should already be in the MDC. We should switch the logs into JSON to get access to it.

server/apps/distributed-app/docker-compose-crowdsec.yml
```diff
@@ -0,0 +1,113 @@
...
+    image: apachepulsar/pulsar:2.10.1
+    container_name: pulsar
+    command: >
+      bin/pulsar standalone
+    ports:
+      - "8081:8080"
+      - "6650:6650"
+    networks:
+      - james
+  crowdsec:
```
chibenwa, line 96:
```suggestion
      - james
      
  crowdsec:
```

server/apps/distributed-app/crowdsec/scenarios/james-bf-auth.yaml
thanhbv200585: No, there is a queue containing these failure, and its capacity is 5 times. 
- leakspeed means that every 10s, a failure will be released from this queue, like water leaks out the bucket

quantranhong1999: Is it not that within 10 seconds, the maximum failure is 5 (maximum 5 items in the queue)? And for the 6th failure, it trigger the scenario?

thanhbv200585: Obviously, within 10s, if there are 6 failures, it will trigger the scenario. But it is only one case. More generally, whenever the queue gets more than 5 failures, it will trigger the scenario

chibenwa:
Next step: have james apply the ban ^^

We could write a HELO hook looking for this?

server/apps/distributed-app/crowdsec/scenarios/james-bf-auth.yaml
Arsnael: Leaking bucket principle, as I remember, so yes

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
```diff
@@ -45,7 +45,7 @@
 
 import reactor.core.publisher.Mono;
 
-public abstract class AbstractAuthProcessor<R extends ImapRequest> extends AbstractMailboxProcessor<R> {
+    public abstract class AbstractAuthProcessor<R extends ImapRequest> extends AbstractMailboxProcessor<R> {
```
Arsnael, line 48: Careful of those indents :)

Arsnael (commented):
Nice job, it proves that it's not too hard to get crowdsec scrapping logs from James to get IPs and ban abusing usage

Just need to polish this now (with real IPs from json logs as suggested above) so we can merge it :)

third-party/crowdsec/docker-compose.yml
```diff
@@ -0,0 +1,113 @@
...
+        condition: service_healthy
+      opensearch:
+        condition: service_started
+      rabbitmq:
+        condition: service_started
+      s3:
+        condition: service_started
+      pulsar:
+        condition: service_started
+    image: apache/james:distributed-latest
```
chibenwa, line 17: Use `apache/james:memory-latest`

third-party/crowdsec/sample-configuration/parsers/james-auth.yaml
```diff
@@ -0,0 +1,27 @@
+onsuccess: next_stage
+debug: true
+filter: "evt.Parsed.program == 'james'"
+name: linagora/james-connect-security-log
+description: "Parser for James Connect Security Log"
+
+pattern_syntax:
+  JAMES_AUTH_FAIL: '\{"timestamp":"%{DATA:time}","level":"%{LOGLEVEL:level}","thread":"%{DATA:thread}","mdc":{"protocol":"%{DATA:protocol}","ip":"%{IPV4:ip}","host":"%{IPV4:host}","action":"LOGIN","login-user":"%{EMAILADDRESS:loginUser}","sessionId":"%{DATA:sessionId}","user":"%{DATA:user}"\},"logger":"%{JAVACLASS:logger}","message":"User %{DATA:loginUser} failed authentication","context":"%{DATA:context}"\}'
```
chibenwa, line 8: Not sure it is a good idea to rewrite a JSON parser with a REGEXP... 
How robust is this solution to field reordering for instance?

thanhbv200585: Because Crowdsec doesn't support to read JSON format so I have to read it with a Regex. 
IMO, it is seldom when the field reorder but maybe there will be some new added fields in the future so I will reconsider this pattern syntax

chibenwa: Did you open a bug report for your issue with JSON format?

> IMO, it is seldom when the field reorder but maybe there will be some new added fields in the future so I will reconsider this pattern syntax

We already know it will break hard in the future. We should reconsider it now.

What does it take to write a parser in CrowdSec?

Arsnael: From the mob / brainstorming, actually maybe two leads:

- from tung: we could add via logback an other entryof logs just for crowdsec with just the fields we are interested in in the format we want. Then we isolate the ones with AUTH_FAIL in crowdsec pattern stuff
- from rene:  we could maybe have a pattern syntax like `'%{DATA:data}failed authentication%{DATA:data}'` to just get logs with failed auth and then try to see if can use https://docs.crowdsec.net/docs/expr/helpers/#json-helpers to extract in a json way the fields we want from the logs

server/apps/memory-app/sample-configuration/logback.xml
```diff
@@ -20,39 +20,62 @@
 
 <configuration>
 
+
```
chibenwa, line 23: Don't change this file, overload it.

chibenwa: Via mounting a volume

chibenwa (commented):
Please only use logs from the docker console in crowdsec. No need to log to a file!

server/apps/memory-app/sample-configuration/logback.xml
```diff
@@ -52,29 +43,16 @@
...
-                                <!-- Importance for handling multiple lines log -->
-                                <appendLineSeparator>true</appendLineSeparator>
-
-                                <jsonFormatter class="ch.qos.logback.contrib.jackson.JacksonJsonFormatter">
-                                        <prettyPrint>false</prettyPrint>
-                                </jsonFormatter>
-                        </layout>
+                <encoder>
+                        <pattern>%d{HH:mm:ss.SSS} [%-5level] %logger{15} - %msg%n%rEx</pattern>
+                        <immediateFlush>false</immediateFlush>
```
Arsnael, line 48: Why not removing directly the appender `LOG_FILE` as you seem to only get logs from the console now anyways/ :)

server/apps/distributed-app/sample-configuration/logback.xml
```diff
@@ -33,41 +24,19 @@
...
-                                <!-- Importance for handling multiple lines log -->
-                                <appendLineSeparator>true</appendLineSeparator>
-
-                                <jsonFormatter class="ch.qos.logback.contrib.jackson.JacksonJsonFormatter">
-                                        <prettyPrint>false</prettyPrint>
-                                </jsonFormatter>
-                        </layout>
+                <encoder>
+                        <pattern>%d{HH:mm:ss.SSS} [%-5level] %logger{15} - %msg%n%rEx</pattern>
+                        <immediateFlush>false</immediateFlush>
```
Arsnael, line 29: idem

Arsnael (commented):
Does it work then with the JSON utils?

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
```diff
@@ -100,6 +100,7 @@ protected void doAuth(AuthenticationAttempt authenticationAttempt, ImapSession s
                     session.stopDetectingCommandInjection();
                 } catch (BadCredentialsException e) {
                     authFailure = true;
+                    LOGGER.info("User {} failed authentication", authenticationAttempt.getAuthenticationId().asString());
```
chibenwa, line 103: Now can we base our selves on the audit trail to not add another log?

chibenwa:
Let's do SMTP AUTHentification failure too.

Let's also analyse logs within ValidRCPTHandler failures (RCPT do not exist) in order to be able to fight Dictionary Enumeration Attacks too.

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
chibenwa: Why is remoteIp empty?

server/apps/distributed-app/sample-configuration/logback.xml
```diff
@@ -11,32 +11,11 @@
...
                 <appender-ref ref="CONSOLE" />
-                <appender-ref ref="LOG_FILE" />
         </root>
 
         <logger name="org.apache.james" level="INFO" />
         <logger name="com.datastax.driver.core.QueryLogger.SLOW" level="DEBUG" />
 
-
-</configuration>
+</configuration>
```
vttranlina, line 21: please don't edit the current logback file
Create one new, then mount it to docker-compose

Arsnael: He is just removing the appender that was only  created for crowdsec but not needed anymore? surely we dont want to keep that right? :)

Arsnael:
> Let's apply crowdsec decisions now.
>
> Let's do SMTP AUTHentification failure too.
> 
> Let's also analyse logs within ValidRCPTHandler failures (RCPT do not exist) in order to be able to fight Dictionary Enumeration Attacks too.

In other PRs perhaps? :)

server/apps/distributed-app/sample-configuration/logback.xml
thanhbv200585: @Arsnael  This is logback file in distribute-app, so we should not modify it, right? Similar with memory-app?

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
thanhbv200585: Because it's unavailable now. We need to change code to get remoteIP

server/apps/distributed-app/sample-configuration/logback.xml
vttranlina: Maybe we don't want to keep that, but  I think it will be better if has another logback.xml in `crowdsec` module.

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
quantranhong1999: @thanhbv200585 

I suggest to expose remoteIP via `ImapSession` interface:
```java
    /**
     * Return the {@link InetSocketAddress} of the remote peer
     */
    InetSocketAddress getRemoteAddress();
```

Implement it in `NettyImapSession`:
```java
    @Override
    public InetSocketAddress getRemoteAddress() {
        return (InetSocketAddress) channel.remoteAddress();
    }
```

Then add remoteIP in `AbstractAuthProcessor`' AuditTrail (s).

chibenwa: Please take also care of Proxy protocol!

Arsnael: Not solved IMO and this log should not be merged, as it brings nothing than just making work some poc/demo... 

I thought we were supposed to try to use audit trail logs (already in place) for this?

thanhbv200585: I think it makes sense, of course we need to add more some info such as remoteIP

server/apps/distributed-app/sample-configuration/logback.xml
Arsnael: Ok sorry, I thought this was introduced by this PR but it was already there, @vttranlina comment is valid actually :)

Arsnael (approved):

Arsnael:
```
08:39:36,371 [ERROR] /home/jenkins/jenkins-agent/workspace/james_ApacheJames_PR-1743/protocols/imap/src/main/java/org/apache/james/imap/encode/FakeImapSession.java:[41,8] org.apache.james.imap.encode.FakeImapSession is not abstract and does not override abstract method getRemoteAddress() in org.apache.james.imap.api.process.ImapSession
```

Arsnael (approved):

vttranlina (approved):

PR #1749, opened 2023-10-06
JAMES-3604 Fix inverted condition for RabbitMQWorkQueue quorum queue …

chibenwa:
…set up

Likely due to a badly solved rebase conflict

### DIFF
--- a/server/task/task-distributed/src/main/java/org/apache/james/task/eventsourcing/distributed/RabbitMQWorkQueue.java
+++ b/server/task/task-distributed/src/main/java/org/apache/james/task/eventsourcing/distributed/RabbitMQWorkQueue.java
@@ -75,6 +75,7 @@ public class RabbitMQWorkQueue implements WorkQueue {
 
     public static final int NUM_RETRIES = 8;
     public static final Duration FIRST_BACKOFF = Duration.ofMillis(100);
+    public static final boolean AUTO_DELETE_QUEUE = true;
 
     private final TaskManagerWorker worker;
     private final JsonTaskSerializer taskSerializer;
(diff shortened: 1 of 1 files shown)


server/task/task-distributed/src/main/java/org/apache/james/task/eventsourcing/distributed/RabbitMQWorkQueue.java
```diff
@@ -123,7 +124,7 @@ void declareQueue() {
         Mono<AMQP.Queue.DeclareOk> declareQueue = sender
             .declare(QueueSpecification.queue(QUEUE_NAME)
                 .durable(true)
-                .arguments(rabbitMQConfiguration.workQueueArgumentsBuilder(ALLOW_QUORUM)
+                .arguments(rabbitMQConfiguration.workQueueArgumentsBuilder(!AUTO_DELETE_QUEUE)
```
Arsnael, line 127: Hmmmm... I really don't see it here... `workQueueArgumentsBuilder` takes a `allowQuorum` boolean argument, so I don't see what the auto delete here is doing.

Also for the queue to be quorum, it needs that argument to be true and to have quorum enabled in the conf. But here you put it to false... so even if quorum is enabled in rabbitmq.properties, it will never be quorum... 

Why are you doing trying this? Cause I don't get it sorry

chibenwa: image

on 3.7.x

Arsnael: Fair point :)

Arsnael (approved):

quantranhong1999 (approved):

PR #1754, opened 2023-10-14
Introduce a ResolvedBucketName type to avoid temporal coupling (and e…

mbaechler:
…nhance readability)

### DIFF
--- a/server/blob/blob-api/src/main/java/org/apache/james/blob/api/ResolvedBucketName.java
+++ b/server/blob/blob-api/src/main/java/org/apache/james/blob/api/ResolvedBucketName.java
@@ -1,0 +1,66 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.blob.api;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Preconditions;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Objects;
+
+public final class ResolvedBucketName {
+    public static ResolvedBucketName of(String value) {
+        return new ResolvedBucketName(value);
+    }
+
+    private final String value;
+
+    private ResolvedBucketName(String value) {
+        Preconditions.checkNotNull(value);
+        Preconditions.checkArgument(StringUtils.isNotBlank(value), "`value` cannot be blank");
+
+        this.value = value;
+    }
+
+    public String asString() {
+        return value;
+    }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (o instanceof ResolvedBucketName) {
+            ResolvedBucketName that = (ResolvedBucketName) o;
+            return Objects.equals(this.value, that.value);
+        }
+        return false;
+    }
+
+    @Override
+    public final int hashCode() {
+        return Objects.hash(value);
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
+            .add("value", value)
+            .toString();
+    }
+}
--- a/server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/BucketNameResolver.java
+++ b/server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/BucketNameResolver.java
@@ -24,7 +24,6 @@ import java.util.Optional;
 import org.apache.james.blob.api.BucketName;
 
 import com.google.common.base.Preconditions;
-import org.apache.james.blob.api.ResolvedBucketName;
 
 public class BucketNameResolver {
     static class Builder {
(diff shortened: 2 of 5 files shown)


server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/BucketNameResolver.java
```diff
@@ -84,33 +85,39 @@ private BucketNameResolver(Optional<BucketName> namespace, Optional<String> pref
...
             }
             return Optional.<BucketName>empty();
-        }).orElse(Optional.of(bucketName));
+        }).orElse(Optional.of(BucketName.of(bucketName.asString())));
     }
 
     private boolean isNameSpace(BucketName bucketName) {
         return namespace
-            .map(existingNamespace -> existingNamespace.equals(bucketName))
+            .map(existingNamespace -> existingNamespace.asString().equals(bucketName.asString()))
```
jeantil, line 114: **💭 thought:** the namespace term is also called defaultBucketName in other places ( it is called namespace in the configuration file, the configuration builder accepts a namespace setter but when building the actual configuration object it injects the namespace into a field named `defaultBucketName` a similarly named field is also exposed by the S3 dao but in the bucketname resolver it is again  called namespace. It is unclear to me what this concept is supposed to represent and how it is different from bucketPrefix (whose name is fairly stable throughout the code)
I don't expect an action on this MR but seeing the name here reminded me of this thought.

jeantil: **❔ question:** ‏why change the equality from `BucketName#equals` to `asString().equals` I agree that the result is identical but don't think this change makes the code clearer

I understand why it is necessary in the  overload of `isNameSpace` that compares a `ResolvedBucketName` and a `BucketName`.

server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/S3BlobStoreDAO.java
```diff
@@ -411,7 +409,7 @@ public Mono<Void> deleteAllBuckets() {
         return Mono.fromFuture(client::listBuckets)
             .publishOn(Schedulers.parallel())
             .flatMapIterable(ListBucketsResponse::buckets)
-                .flatMap(bucket -> deleteResolvedBucket(BucketName.of(bucket.name())), DEFAULT_CONCURRENCY)
+                .flatMap(bucket -> deleteResolvedBucket(ResolvedBucketName.of(bucket.name())), DEFAULT_CONCURRENCY)
```
jeantil, line 412: **❔ question:** ‏ I realize this was the case before this MR but is it normal that we don't go through 
`ResolvedBucketName resolvedBucketName = bucketNameResolver.resolve(bucketName);` here whil we do for all other public calls ? I don't necessarily expect this to be fixed in this MR, just trying to understand if there is something that needs fixing at all :D

server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/S3BlobStoreDAO.java
```diff
@@ -420,7 +418,7 @@ public Publisher<BucketName> listBuckets() {
         return Mono.fromFuture(client::listBuckets)
             .flatMapIterable(ListBucketsResponse::buckets)
             .map(Bucket::name)
-            .handle((bucket, sink) -> bucketNameResolver.unresolve(BucketName.of(bucket))
+            .handle((bucket, sink) -> bucketNameResolver.unresolve(ResolvedBucketName.of(bucket))
```
jeantil, line 421: **❔ question:** ‏same as above, I realize this was the case before this MR but is it normal that we don't go through 
`ResolvedBucketName resolvedBucketName = bucketNameResolver.resolve(bucketName);` here whil we do for all other public calls ? I don't necessarily expect this to be fixed in this MR, just trying to understand if there is something that needs fixing at all :D

jeantil (approved):

server/blob/blob-api/src/main/java/org/apache/james/blob/api/ResolvedBucketName.java
```diff
@@ -0,0 +1,66 @@
...
+
+package org.apache.james.blob.api;
+
+import com.google.common.base.MoreObjects;
+import com.google.common.base.Preconditions;
+import org.apache.commons.lang3.StringUtils;
+
+import java.util.Objects;
+
+public final class ResolvedBucketName {
```
chibenwa, line 28: This POJO is defined in the API but only used in blob-s3.

I think introducing it is a good idea, but we should avoid poluting the API and move it into blob-s3.

Arsnael:
```
09:19:04,821 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.3.0:check (check-style) on project blob-api: You have 2 Checkstyle violations. -> [Help 1]
```

server/blob/blob-api/src/main/java/org/apache/james/blob/api/ResolvedBucketName.java
aduprat: Hi @chibenwa ,
Do you have a plan to migrate to Java 14 (or 16), this will prevent from lot of code like this one.

Arsnael: If we upgrade better a LTS version, so 17 (or even 21 would be better, but maybe too big of a jump?)

I would support such a migration, but I'm not sure the all community is ready to upgrade their java version :)

Could perhaps start a thread on the ML though?

chibenwa: >Hi @chibenwa ,
> Do you have a plan to migrate to Java 14 (or 16), this will prevent from lot of code like this one.

`You` is out of place, as a community we should embrasse a `we`.

I would personnalyy support a switch to Java 17 or even 21.

Do you think you can pull off a mail on this topic on the mailing lists?

Arsnael:
@mbaechler status on this? any plan on fixing soon checkstyle issues and answering to comments? :)

mbaechler:
> @mbaechler status on this? any plan on fixing soon checkstyle issues and answering comments? :)

I apologize for dropping PRs without a proper follow-up plan.
I will try this week.

server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/S3BlobStoreDAO.java
mbaechler: We are listing buckets with S3 API then we call a delete for each bucket.
In S3, there are only buckets with "resolved names".

server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/S3BlobStoreDAO.java
mbaechler: listBuckets returns some BucketName but list from S3 returns resolved bucket names.

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1754/4/testReport/

I think the errors in BucketNameResolverTest are related

PR #1756, opened 2023-10-17
JAMES-3828 ICALAttributeDTO: support missing uid, method and dtstamp fields

quantranhong1999:
resolve https://github.com/linagora/james-project/issues/4880


quantranhong1999:
> Breaking change needs to be documented.

I am not sure where to document. Those fields are documented as Optional in the docs already: https://github.com/apache/james-project/blob/master/mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICALToJsonAttribute.java#L71

Arsnael (approved):

chibenwa (approved):

PR #1758, opened 2023-10-19
JAMES-4888 FILTER: forward loop issue

hungphan227:


server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -0,0 +1,242 @@
...
+        jamesServer = TemporaryJamesServer.builder()
+            .withOverrides(binder -> Multibinder.newSetBinder(binder, GuiceProbe.class).addBinding().to(FilteringManagementProbeImpl.class))
+            .withMailetContainer(MailetContainer.builder()
+                .putProcessor(ProcessorConfiguration.root()
+                    .addMailet(MailetConfiguration.builder()
+                        .matcher(All.class)
+                        .mailet(JMAPFiltering.class))
+                    .addMailet(MailetConfiguration.builder()
+                        .matcher(All.class)
+                        .mailet(RecipientRewriteTable.class))
```
chibenwa, line 90: Put RRT first

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -226,8 +229,21 @@ static ForwardDecision sendACopy(MailetContext context, MailAddress originalReci
             return mail -> {
                 MailImpl copy = MailImpl.duplicate(mail);
                 try {
+                    Optional<Attribute> attribute = copy.getAttribute(AttributeName.FORWARDED_MAIL_ADDRESSES_ATTRIBUTE_NAME);
```
chibenwa, line 232: Can we check the attribute before doing the copy ?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -226,8 +229,21 @@ static ForwardDecision sendACopy(MailetContext context, MailAddress originalReci
...
+                    if (attribute.isPresent()) {
+                        Set<MailAddress> forwardedMailAddresses = getForwardedMailAddresses(attribute);
+                        List<MailAddress> newForwards = getNewForwards(forwards, forwardedMailAddresses);
+                        if (newForwards.isEmpty()) {
+                            return;
+                        } else {
+                            copy.setAttribute(createAttribute(forwardedMailAddresses, newForwards));
+                            copy.setRecipients(newForwards);
```
chibenwa, line 240: Too much indent level. 

Please extract.

server/protocols/jmap-draft/src/main/java/org/apache/james/jmap/mailet/filter/ActionApplier.java
```diff
@@ -182,4 +201,32 @@ private void sendACopy(MailetContext context, MailAddress originalRecipient, Lis
...
+    }
+
+    private Set<MailAddress> getForwardedMailAddresses(Optional<Attribute> attribute) {
+        Collection<AttributeValue> attributeValues = (Collection<AttributeValue>) attribute.get().getValue().getValue();
+        Set<MailAddress> forwardedMailAddresses = attributeValues
+            .stream()
+            .map(Throwing.function(attributeValue -> new MailAddress((String) attributeValue.getValue())))
+            .collect(ImmutableSet.toImmutableSet());
+        return forwardedMailAddresses;
+    }
```
chibenwa, line 231: Any solution to not duplicate that code?

hungphan227: The only possible place I can think of to put common code is the mallet api module. However, you said in the comment above that you do not want to pollute the API. Therefore, I do not have any solution :((

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        Awaitility.waitAtMost(1L, TimeUnit.HOURS).until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 2L);
+
+        SoftAssertions.assertSoftly(Throwing.consumer(softly -> {
+            List<Mail> mails = mailRepositoryProbe.listMails(CUSTOM_REPOSITORY)
+                .filter(Throwing.predicate(mail -> mail.getRecipients().contains(ALICE.asMailAddress())))
+                .collect(ImmutableList.toImmutableList());
+
+            softly.assertThat(mails.get(0).getRecipients()).containsOnly(ALICE.asMailAddress());
+            softly.assertThat(mails.get(1).getRecipients()).containsOnly(ALICE.asMailAddress());
```
Arsnael, line 169: When reading the ticket it seems Cedric should not send the mail back to alice (cause loop). One mail is because local copy, the second one I'm surprised though?

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -0,0 +1,247 @@
...
+                Mapping.forward(CEDRIC.asString()));
+        jamesServer.getProbe(DataProbeImpl.class)
+            .addMapping(MappingSource.fromUser(CEDRIC),
+                Mapping.forward(ALICE.asString()));
+
+        messageSender.connect(LOCALHOST_IP, jamesServer.getProbe(SmtpGuiceProbe.class).getSmtpPort())
+            .authenticate(SENDER.asString(), PASSWORD)
+            .sendMessage(SENDER.asString(), ALICE.asString());
+
+//        Awaitility.await().pollDelay(1, TimeUnit.SECONDS).until(() -> jamesServer.getProbe(SpoolerProbe.class).processingFinished());
```
Arsnael, line 192: remove this

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -0,0 +1,247 @@
...
+        Awaitility.await().until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 1L);
+
+        SoftAssertions.assertSoftly(Throwing.consumer(softly -> {
+            List<Mail> mails = mailRepositoryProbe.listMails(CUSTOM_REPOSITORY)
+                .collect(ImmutableList.toImmutableList());
+
+            softly.assertThat(mails).hasSize(1);
```
Arsnael, line 199: 2 things:
- `Awaitility.await().until` above already checks that?
- I would think cedric gets one and sender gets one too?

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        Awaitility.await().until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 2L);
+
+        SoftAssertions.assertSoftly(Throwing.consumer(softly -> {
+            List<Mail> mails = mailRepositoryProbe.listMails(CUSTOM_REPOSITORY)
+                .filter(Throwing.predicate(mail -> mail.getRecipients().contains(ALICE.asMailAddress())))
+                .collect(ImmutableList.toImmutableList());
+
+            softly.assertThat(mails.get(0).getRecipients()).containsOnly(ALICE.asMailAddress());
+            softly.assertThat(mails.get(1).getRecipients()).containsOnly(ALICE.asMailAddress());
```
Arsnael, line 243: same comment as the first one, I would expect one mail for alice and one for cedric?

mailet/api/src/main/java/org/apache/mailet/LoopPrevention.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Set;
+
+import org.apache.james.core.MailAddress;
+
+import com.github.fge.lambdas.Throwing;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+
+public class LoopPrevention {
```
chibenwa, line 32: Please unit test this class

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -157,7 +158,7 @@ void filterForwardShouldNotCreateLoopError() throws Exception {
             .authenticate(SENDER.asString(), PASSWORD)
             .sendMessage(SENDER.asString(), ALICE.asString());
 
-        Awaitility.await().until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 2L);
+        Awaitility.waitAtMost(1L, TimeUnit.HOURS).until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 2L);
```
chibenwa, line 161: Why 1 hour?

mailet/api/src/main/java/org/apache/mailet/LoopPrevention.java
hungphan227: Is it necessary to write unit test for a method like this?

```
public static Set<MailAddress> nonRecordedRecipients(Set<MailAddress> recipients, Set<MailAddress> recordedRecipients) {
        return Sets.difference(recipients, recordedRecipients);
    }
```

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
hungphan227: Initially I could not figure out how to prevent Alice from getting a duplicate mail. Just fixed it. Check the newest commit.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
hungphan227: They do not keep a copy by default

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
hungphan227: Cedric does not keep a copy

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
hungphan227: answered above

mailet/api/src/main/java/org/apache/mailet/LoopPrevention.java
chibenwa: Please write unit test for the relevant methods.

> Is it necessary to write unit test for a method like this?

Is it necessary to write a method like this in the first place?

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -158,15 +159,15 @@ void filterForwardShouldNotCreateLoopError() throws Exception {
...
+        Awaitility.await().until(() -> jamesServer.getProbe(SpoolerProbe.class).processingFinished());
+        Awaitility.await().until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 1L);
 
         SoftAssertions.assertSoftly(Throwing.consumer(softly -> {
             List<Mail> mails = mailRepositoryProbe.listMails(CUSTOM_REPOSITORY)
                 .filter(Throwing.predicate(mail -> mail.getRecipients().contains(ALICE.asMailAddress())))
                 .collect(ImmutableList.toImmutableList());
 
             softly.assertThat(mails.get(0).getRecipients()).containsOnly(ALICE.asMailAddress());
-            softly.assertThat(mails.get(1).getRecipients()).containsOnly(ALICE.asMailAddress());
```
Arsnael: Hmmm as discussed, not ok IMO...

Meaning in the case here that Alice does not keep a local copy, the mail would just not be stored anywhere and vanish. This is unacceptable for a mail server in any way! (I'm sure @chibenwa agrees on that^^)

I think we need to think a bit deeper here. Of course we want the mail to be stored somewhere. What do we want though?
- mail stop at the end of the chian before looping (here stored on cedric mailbox)
- mail is stored back at the original recipient when loop is detected (alice, which was the behavior before I asked the change)
- if stored at the original recipient and that recipient had localCopy enabled, how do we deal with that? I don't think we should store it twice right? (same in the case of cedric if he had localCopy, we don't want to store it twice either)

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
Arsnael: cf https://github.com/apache/james-project/pull/1758#discussion_r1368334183

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
Arsnael: cf https://github.com/apache/james-project/pull/1758#discussion_r1368334183

mailet/api/src/main/java/org/apache/mailet/LoopPrevention.java
hungphan227: added

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/FilterForwardIntegrationTest.java
```diff
@@ -0,0 +1,204 @@
...
+        Rule.Action.Forward forward = Rule.Action.Forward.of(forwardedMailAddresses, true);
+        filteringManagementProbe.defineRulesForUser(BOB,
+            Optional.empty(),
+            Rule.builder()
+                .id(Rule.Id.of("1"))
+                .name("rule 1")
+                .conditionGroup(Rule.ConditionGroup.of(Rule.ConditionCombiner.AND, Rule.Condition.of(FROM, CONTAINS, ALICE.asString())))
+                .action(Rule.Action.builder().setAppendInMailboxes(Rule.Action.AppendInMailboxes.withMailboxIds(ImmutableList.of()))
+                    .setWithKeywords(ImmutableList.of())
+                    .setForward(Optional.of(forward))
```
chibenwa, line 124: Let's add a method in the builder so that we could do:

```
                    .setForward(forward)
```

?

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/FilterForwardIntegrationTest.java
```diff
@@ -0,0 +1,204 @@
...
+        ImmutableList<MailAddress> forwardedMailAddresses = ImmutableList.of(CEDRIC.asMailAddress());
+        Rule.Action.Forward forward = Rule.Action.Forward.of(forwardedMailAddresses, true);
+        filteringManagementProbe.defineRulesForUser(BOB,
+            Optional.empty(),
+            Rule.builder()
+                .id(Rule.Id.of("1"))
+                .name("rule 1")
+                .conditionGroup(Rule.ConditionGroup.of(Rule.ConditionCombiner.AND, Rule.Condition.of(FROM, CONTAINS, ALICE.asString())))
+                .action(Rule.Action.builder().setAppendInMailboxes(Rule.Action.AppendInMailboxes.withMailboxIds(ImmutableList.of()))
+                    .setWithKeywords(ImmutableList.of())
```
chibenwa, line 123: This line do not add any value and thus should not be necessary.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/FilterForwardIntegrationTest.java
```diff
@@ -0,0 +1,204 @@
...
+    void tearDown() {
+        jamesServer.shutdown();
+    }
+
+    @Test
+    void forwardShouldWork() throws Exception {
+        ImmutableList<MailAddress> forwardedMailAddresses = ImmutableList.of(CEDRIC.asMailAddress());
+        Rule.Action.Forward forward = Rule.Action.Forward.of(forwardedMailAddresses, true);
+        filteringManagementProbe.defineRulesForUser(BOB,
+            Optional.empty(),
```
chibenwa, line 117: Name this. Is it `NO_VERSION` ?

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -0,0 +1,248 @@
...
+        filteringManagementProbe.defineRulesForUser(ALICE,
+            Optional.empty(),
+            Rule.builder()
+                .id(Rule.Id.of("1"))
+                .name("rule 1")
+                .conditionGroup(Rule.ConditionGroup.of(Rule.ConditionCombiner.AND, Rule.Condition.of(FROM, NOT_CONTAINS, "AAA")))
+                .action(Rule.Action.builder().setAppendInMailboxes(Rule.Action.AppendInMailboxes.withMailboxIds(ImmutableList.of()))
+                    .setWithKeywords(ImmutableList.of())
+                    .setForward(Optional.of(Rule.Action.Forward.of(ImmutableList.of(BOB.asMailAddress()), true)))
+                    .build())
```
chibenwa, line 133:
```suggestion
                .action(Rule.Action.builder()
                    .setForward(Optional.of(Rule.Action.Forward.of(ImmutableList.of(BOB.asMailAddress()), true)))
                    .build())
```

?

chibenwa:
Tests should be way easier to read.

I did let a few comments on how to achieve this.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
chibenwa: I don't think the mail is lost but it is possible that in case of loop the mail do not end up in the correct place...

hungphan227: > I don't think the mail is lost but it is possible that in case of loop the mail do not end up in the correct place...
> 
> Let's write more readable test cases then see if the behaviour is the one that we want. Today it is not that clear to me to be honnest...

If we setup in the way that A forward mail to B, B forward to C and C forward to A but no one set keepACopy to be true, the mail would be lost.

quantranhong1999: > If we setup in the way that A forward mail to B, B forward to C and C forward to A but no one set keepACopy to be true, the mail would be lost.

How about C is the end of the loop and C is the one who keeps the final mail?

Even if C does not want to `keepACopy`, standing as a user, I say when forwarding me this mail, please do not keep the mail for me. In the case that C does not forward the mail to A, it means there is no forwarding done, therefore no need to obey `keepACopy` for C which is reasonable IMO.

Maybe `Bounce` a mail to announce C that his mail can not be forward anymore as the end of the loop as well?

mailet/api/src/main/java/org/apache/mailet/LoopPrevention.java
```diff
@@ -0,0 +1,70 @@
...
+                                        Set<MailAddress> newRecipients,
+                                        MailAddress originalRecipient) {
+        mail.setAttribute(createAttribute(ImmutableSet.<MailAddress>builder()
+            .addAll(recordedRecipients)
+            .addAll(newRecipients)
+            .add(originalRecipient)
+            .build()));
+    }
+
+    public static Set<MailAddress> retrieveRecordedRecipients(Mail mail) {
```
aduprat, line 50: Could be better to return a real object instead of a set (which will be a wrapper around the current `set`)
```java
public static RecordedRecipients retrieveRecordedRecipients(Mail mail) {
...
```

then you can remove previous method:
```
public static Set<MailAddress> nonRecordedRecipients(Set<MailAddress> recipients, RecordedRecipients recordedRecipients)
```

and add it in the newly created class.

That way it will prevent you from confusion on calling `nonRecordedRecipients` between the two parameters,
plus it will give us the state of the current processing rather than calling static methods in mailet processing.

hungphan227: Thank for your suggestion. However, I do not really see that applying your idea would make my code better. To distinguish between two parameters, we can check parameter names (recipients, recordedRecipients).

chibenwa: Well if I understand what @aduprat  suggests it is just applying some strong typing here, and moving some of the computations into this new type.

chibenwa: I did give a shot at it.

Arsnael:
Needs a rebase though :)

mailet/api/src/main/java/org/apache/mailet/LoopPrevention.java
aduprat: I got my idea @chibenwa .
Thank you for your modifications. 👏

chibenwa: Looks better indeed ;-)

chibenwa (approved):
Like this would be okish to me.

aduprat (approved):

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @AfterEach
+    void tearDown() {
+        jamesServer.shutdown();
+    }
+
+    @Test
+    void filterForwardShouldNotCreateLoopError() throws Exception {
+        filteringManagementProbe.defineRulesForUser(ALICE, asRule(Forward.to(BOB.asMailAddress()).keepACopy()));
```
Arsnael, line 134: I maintain a previous comment on this... Try honestly writing the same test  with alice doing a forward without a copy, and I would not be surprised that we just get no email being recorded in James in the end.

I think we agreed previously to store the mail at the end of the chain, and I don't see that here.

Arsnael (changes requested):
Been a bit too fast actually on approval... I still think we are loosing a mail here on a case.

Also did you potentially force push over a PR without fetching latest update first @chibenwa ? Cause I thought this was fixed on latest by @hungphan227 actually... But maybe I'm wrong and he didn't push it yet (in which case I'm sorry).

@hungphan227 is all your initial code in there, can you confirm?

hungphan227:
I have pushed a new commit to resolve some issues

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
hungphan227: fixed. In case A have to forward mail to B but B has already received the mail before, A will keep a local copy no matter what keepACopy is set

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/Rule.java
```diff
@@ -390,14 +390,20 @@ public Builder setForward(Optional<Forward> forward) {
             }
 
             public Builder setForward(Forward forward) {
-                this.forward = Optional.of(forward);
+                this.forward = Optional.ofNullable(forward);
```
chibenwa, line 393: Heresy. `null` value shall be avoided at all cost.

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/Rule.java
```diff
@@ -390,14 +390,20 @@ public Builder setForward(Optional<Forward> forward) {
...
-                Preconditions.checkNotNull(forward, "forward should not be null");
+                if (Objects.isNull(appendInMailboxes)) {
+                    appendInMailboxes = Rule.Action.AppendInMailboxes.withMailboxIds(ImmutableList.of());
+                }
+                if (Objects.isNull(withKeywords)) {
+                    withKeywords = ImmutableList.of();
+                }
+                if (Objects.isNull(forward)) {
+                    forward = Optional.empty();
+                }
```
chibenwa, line 406: Then use Optionals

mailet/api/src/main/java/org/apache/mailet/LoopPrevention.java
```diff
@@ -84,6 +84,14 @@ public RecordedRecipients merge(MailAddress... other) {
             return merge(ImmutableSet.copyOf(other));
         }
 
+        public RecordedRecipients mergeIfEmpty(MailAddress... other) {
```
chibenwa, line 87: Why merge only if empty? Why not merge all the time?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -223,37 +223,27 @@ static ForwardDecision removeRecipient(MailAddress recipient) {
                 .collect(ImmutableList.toImmutableList()));
         }
 
-        static ForwardDecision sendACopy(MailetContext context, MailAddress originalRecipient, List<MailAddress> forwards) {
+        static ForwardDecision sendACopy(MailetContext context,
+                                         MailAddress originalRecipient,
+                                         List<MailAddress> forwards,
+                                         LoopPrevention.RecordedRecipients recordedRecipients,
+                                         Set<MailAddress> newRecipients) {
```
chibenwa, line 230: Why are forwards and original recipients different?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -272,36 +262,50 @@ private static void recordInAuditTrail(Mail mail, MailImpl copy, MailAddress ori
...
+        ImmutableSet<Mapping> forwards = getForwards(recipient);
         if (forwards.isEmpty()) {
             return Stream.of();
         }
         Mapping localCopyMapping = Mapping.forward(recipient.asString());
         boolean localCopy = forwards.contains(localCopyMapping);
-        List<MailAddress> forwardedRecipients = forwards.stream()
+        List<MailAddress> forwardedRecipients = getForwardedMailAddressesWithLocalCopyExcluded(forwards, localCopyMapping);
+
+        Set<MailAddress> newRecipients = recordedRecipients.nonRecordedRecipients(ImmutableSet.copyOf(forwardedRecipients));
```
chibenwa, line 284: Please find a way to write 

```suggestion
        Set<MailAddress> newRecipients = recordedRecipients.nonRecordedRecipients(forwardedRecipients);
```

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -272,36 +262,50 @@ private static void recordInAuditTrail(Mail mail, MailImpl copy, MailAddress ori
...
+        boolean shouldMailBeForwarded = !newRecipients.isEmpty();
+
+        if (localCopy && shouldMailBeForwarded) {
+            return Stream.of(ForwardDecision.sendACopy(mailetContext, recipient, forwardedRecipients, recordedRecipients, newRecipients));
+        }
+        if (!shouldMailBeForwarded) {
+            return Stream.of();
+        }
+        return Stream.of(ForwardDecision.removeRecipient(recipient),
+            ForwardDecision.sendACopy(mailetContext, recipient, forwardedRecipients, recordedRecipients, newRecipients));
```
chibenwa, line 294: Too complicated to me. 

How about something more functionnal?

```
    Optional<Decision> localCopyDecision(ImmutableSet<Mapping> forwards, MailAddress recipient) {
        Mapping localCopyMapping = Mapping.forward(recipient.asString());
        if (forwards.contains(localCopyMapping) {
            return Optional.empty();
        }
        return Optional.of(ForwardDecision.removeRecipient(recipient));
    }

    Optional<Decision> forwardDecision(ImmutableSet<Mapping> forwards, MailAddress recipient) {
        Mapping localCopyMapping = Mapping.forward(recipient.asString());
        List<MailAddress> forwardedRecipients = getForwardedMailAddressesWithLocalCopyExcluded(forwards, localCopyMapping);
        if (forwardedRecipients.isEmpty()) {
            return Optional.empty();
        }
        return Optional.of(sendACopy(...));
    }

    
    Then
    
    return Stream.of(localCopyDecision(...),
        forwardDecision(...))
        .flatMap(Optional::stream);
```

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
chibenwa, line 230: Way too much arguments. With 4 different disguised list of mail address. 
That's not reasonnable.

chibenwa (commented):
IMO we should still significantly invest in code quality here.

It will take the time that it takes.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
hungphan227: for example, if the original recipient is A, forwards will be the list of mail addresses that A has to forward email to

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -144,16 +145,14 @@ void filterForwardShouldNotCreateLoopError() throws Exception {
...
                 .collect(ImmutableList.toImmutableList());
 
-            softly.assertThat(mails.get(0).getRecipients()).containsOnly(ALICE.asMailAddress());
+            softly.assertThat(mails.get(0).getRecipients()).containsOnly(CEDRIC.asMailAddress());
         }));
     }
 
     @Test
     void regularForwardShouldNotCreateLoopError() throws Exception {
-        dataProbe.addMapping(MappingSource.fromUser(ALICE), Mapping.forward(SENDER.asString()));
```
Arsnael: You could leave this one, and check you get a mail in cedric and sender. Covers an extra case IMO :)

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -131,7 +132,7 @@ void tearDown() {
 
     @Test
     void filterForwardShouldNotCreateLoopError() throws Exception {
-        filteringManagementProbe.defineRulesForUser(ALICE, asRule(Forward.to(BOB.asMailAddress()).keepACopy()));
+        filteringManagementProbe.defineRulesForUser(ALICE, asRule(Forward.to(BOB.asMailAddress()).withoutACopy()));
```
Arsnael, line 135: Add maybe an other test, where you keep a copy, and you have a mail stored for alice and cedric :)

mailet/api/src/main/java/org/apache/mailet/LoopPrevention.java
hungphan227: for example, we setup to make A forward mail to B, B to C, C to A.

someone send email to A, then james receive an email with A as recipient, we have to add A to the recorded recipient list (before adding, the list is empty)
when james receive an email with B as recipient, adding B to the recorded recipient list is not necessary (before adding, the list has already included A and B)

chibenwa: Isn't it managed naturrally by the `Set` data structure?

hungphan227: if I just merge it without checking empty, there would be two issues:
- when others read code, it could make them confused because the originalRecipient is always merged. By using mergeIfEmpty, they could understand how this actually work.
- create unnecessary operations. It does not impact performance much but it is good if we can avoid.

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/Rule.java
hungphan227: fixed

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
hungphan227: fixed

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
hungphan227: removed one argument

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
hungphan227: added a similar test case

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
hungphan227: added

server/container/guice/common/src/main/java/org/apache/james/utils/MailRepositoryProbeImpl.java
```diff
@@ -64,6 +65,11 @@ public Stream<Mail> listMails(MailRepositoryUrl url) throws Exception {
             .map(Throwing.function(key -> getMail(url, key)));
     }
 
+    public List<Mail> listMails(MailRepositoryUrl url, MailAddress recipient) throws Exception {
```
chibenwa, line 68:
```suggestion
    public List<Mail> listMailsForRecipient(MailRepositoryUrl url, MailAddress recipient) throws Exception {
```

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -222,65 +223,107 @@ static ForwardDecision removeRecipient(MailAddress recipient) {
...
+    private Optional<ForwardDecision> localCopyDecision(boolean doForwardsContainLocalCopy,
+                                                        boolean shouldMailBeForwarded,
+                                                        MailAddress recipient) {
+        if (!shouldMailBeForwarded) {
+            return Optional.empty();
+        } else if (doForwardsContainLocalCopy) {
+            return Optional.empty();
+        } else {
+            return Optional.of(ForwardDecision.removeRecipient(recipient));
+        }
```
chibenwa, line 300: Too complex

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -222,65 +223,107 @@ static ForwardDecision removeRecipient(MailAddress recipient) {
...
+        Set<MailAddress> newRecipients = recordedRecipients.nonRecordedRecipients(forwardedRecipients);
+        boolean shouldMailBeForwarded = !newRecipients.isEmpty();
+
+        return Stream.of(localCopyDecision(doForwardsContainLocalCopy, shouldMailBeForwarded, recipient),
+                forwardDecision(shouldMailBeForwarded, mailetContext, recipient, recordedRecipients, newRecipients))
+            .flatMap(Optional::stream);
+    }
+
+    private Optional<ForwardDecision> localCopyDecision(boolean doForwardsContainLocalCopy,
+                                                        boolean shouldMailBeForwarded,
```
chibenwa, line 292: Taking 2 boolean parameters is a good sign of something wrong.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -222,65 +223,107 @@ static ForwardDecision removeRecipient(MailAddress recipient) {
...
+        if (!shouldMailBeForwarded) {
+            return Optional.empty();
+        } else if (doForwardsContainLocalCopy) {
+            return Optional.empty();
+        } else {
+            return Optional.of(ForwardDecision.removeRecipient(recipient));
+        }
+    }
+
+    private Optional<ForwardDecision> forwardDecision(boolean shouldMailBeForwarded,
```
chibenwa, line 303: this boolean parameter is evil

chibenwa:
`JMAPFilteringTest` is failing on this branch. 

I let you debug it...

chibenwa:
- Fixed `JMAPFilteringTest` issue
 - Simplified the overall handling by tracking only recipients that are converted to forward. This is easier to do if the localCopy is done with the forwarded mail.
 - Misc code style improvments based on earlier remarks.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -225,14 +225,14 @@ static ForwardDecision removeRecipient(MailAddress recipient) {
...
-                                         LoopPrevention.RecordedRecipients recordedRecipients,
                                          Set<MailAddress> newRecipients) {
             return mail -> {
                 MailImpl copy = MailImpl.duplicate(mail);
+                LoopPrevention.RecordedRecipients recordedRecipients = LoopPrevention.RecordedRecipients.fromMail(mail);
                 try {
                     copy.setSender(originalRecipient);
                     copy.setRecipients(newRecipients);
-                    recordedRecipients.mergeIfEmpty(originalRecipient).merge(newRecipients).recordOn(copy);
+                    recordedRecipients.merge(originalRecipient).recordOn(copy);
```
hungphan227, line 235: Setup: A forward to B, A to C, B to C, C to B
if we only save originalRecipient to the list, we could not prevent B and C from resending email to each other. However, if you want to make it simple, saving only originalRecipient is still acceptable.

server/protocols/jmap-draft/src/main/java/org/apache/james/jmap/mailet/filter/ActionApplier.java
```diff
@@ -127,21 +128,33 @@ private void applyReject(Rule.Action action) {
...
-                return true;
-            }
-        })).orElse(false);
+    private void applyForward(Rule.Action action) {
+        action.getForward()
+            .filter(any -> shouldProcessingContinue())
+            .ifPresent(Throwing.consumer(forward -> {
+                LoopPrevention.RecordedRecipients recordedRecipients = LoopPrevention.RecordedRecipients.fromMail(mail);
+
+                if (recordedRecipients.getRecipients().contains(mailAddress)) {
```
hungphan227, line 137: I believe with current way, this block never run

server/protocols/jmap-draft/src/main/java/org/apache/james/jmap/mailet/filter/ActionApplier.java
```diff
@@ -127,21 +128,33 @@ private void applyReject(Rule.Action action) {
...
+                }
+            }));
+    }
+
+
+    private Set<MailAddress> getNewRecipients(Rule.Action.Forward forward, LoopPrevention.RecordedRecipients recordedRecipients) {
+        ImmutableSet.Builder<MailAddress> newRecipientsBuilder = ImmutableSet.<MailAddress>builder()
+            .addAll(recordedRecipients.nonRecordedRecipients(forward.getAddresses()));
+        if (forward.isKeepACopy()) {
+            newRecipientsBuilder.add(mailAddress);
```
hungphan227, line 155: Current user forward email to himself??

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardIntegrationTest.java
```diff
@@ -323,19 +323,13 @@ void forwardShouldSplitLocalCopy() throws Exception {
...
                 .findAny().get();
-            Mail localCopy = mailRepositoryProbe.listMails(CUSTOM_REPOSITORY)
-                .filter(Throwing.predicate(mail -> mail.getRecipients().contains(BOB.asMailAddress())))
-                .findAny().get();
-            softly.assertThat(forwardedMail.getRecipients()).containsOnly(DELPHINE.asMailAddress());
+            softly.assertThat(forwardedMail.getRecipients()).containsOnly(DELPHINE.asMailAddress(), BOB.asMailAddress());
             softly.assertThat(forwardedMail.getMaybeSender().asOptional()).contains(BOB.asMailAddress());
-            softly.assertThat(localCopy.getRecipients()).containsOnly(BOB.asMailAddress());
-            softly.assertThat(localCopy.getMaybeSender().asOptional()).contains(new MailAddress(FROM));
         }));
```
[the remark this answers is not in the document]
chibenwa: Because now local copy is handled in the forwarded mail. 
This ease the overall logic a lot, and the end result is the same (message in both INBOX)

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -263,60 +263,40 @@ private static void recordInAuditTrail(Mail mail, MailImpl copy, MailAddress ori
...
-                                                      LoopPrevention.RecordedRecipients recordedRecipients,
-                                                      Set<MailAddress> newRecipients) {
-        if (shouldMailBeForwarded) {
-            return Optional.of(ForwardDecision.sendACopy(context, originalRecipient, recordedRecipients, newRecipients));
-        } else {
-            return Optional.empty();
-        }
-    }
+    private Stream<ForwardDecision> forwardDecision(Mail mail, ImmutableSet<Mapping> forwards, MailAddress originalRecipient) {
+        LoopPrevention.RecordedRecipients recordedRecipients = LoopPrevention.RecordedRecipients.fromMail(mail);
```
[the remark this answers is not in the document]
chibenwa: IMO this avoids passing parameters everywhere so it is worth it.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
chibenwa: With the new logic you just need to carry over the original recipient. Easier...

server/protocols/jmap-draft/src/main/java/org/apache/james/jmap/mailet/filter/ActionApplier.java
chibenwa: As explained, loclCopy is now handled by keeping the original recipient on the forwarded mail: easier...

chibenwa:
Also please fix ticket number: **JAMES-4448 is NOT a valid JIRA...**

chibenwa:
IMO could be closed in favor of https://github.com/apache/james-project/pull/1762

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardIntegrationTest.java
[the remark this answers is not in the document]
hungphan227: two concerns:
- when Bob see the mail, the sender is him. does it make sense?
- why doesn't mailRepositoryProbe have two identical mails? (because the mail is forwarded to Bob and Delphine)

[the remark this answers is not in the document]
chibenwa: > when Bob see the mail, the sender is him. does it make sense?

You confuse envelope and headers

PR #1764, opened 2023-10-27
JAMES 3897 - Crowdsec integration for Dictionary Attack

thanhbv200585:


third-party/crowdsec/sample-configuration/parsers/james-auth.yaml
```diff
@@ -73,3 +74,19 @@ nodes:
           expression: evt.Parsed.mdc_remoteIP
         - meta: user
           expression: evt.Parsed.mdc_username
+  - grok:
+      name: "DICTIONARY_ATTACK"
```
chibenwa, line 78: Maybe rename the file james.yml? It no longer is just about auth...

quantranhong1999 (approved):

third-party/crowdsec/sample-configuration/parsers/james-auth.yaml
thanhbv200585: @chibenwa  A file yaml is a parser, so how about creating a new parser for Dictionary attack?

chibenwa: Ok

Arsnael (approved):

PR #1765, opened 2023-10-29
JAMES-3874 LMTP should handle overQuota exceptions

Maxxx873:


server/protocols/protocols-lmtp/src/main/java/org/apache/james/lmtpserver/hook/MailboxDeliverToRecipientHandler.java
```diff
@@ -105,6 +105,13 @@ public long size() {
                 .smtpReturnCode(SMTPRetCode.MAIL_OK)
                 .smtpDescription(DSNStatus.getStatus(DSNStatus.SUCCESS, DSNStatus.CONTENT_OTHER) + " Message received <" + recipient.asString() + ">")
                 .build();
+        } catch (OverQuotaException e) {
+            LOGGER.error("OverQuota error handling ", e);
```
chibenwa, line 109: Imo this is eventually expected upon regular operation, INFO  is likely ok.

"{} is overquota" is likely a more infirmative message. Pass target username as a parameter too to maximize usefullness.

server/protocols/protocols-lmtp/src/test/java/org/apache/james/lmtpserver/LmtpServerTest.java
```diff
@@ -73,15 +80,16 @@
...
 import org.apache.mailet.Mail;
-import org.junit.jupiter.api.AfterEach;
-import org.junit.jupiter.api.BeforeEach;
-import org.junit.jupiter.api.Nested;
-import org.junit.jupiter.api.Test;
 
 import com.github.fge.lambdas.Throwing;
 import com.google.inject.name.Names;
 
+import org.junit.jupiter.api.AfterEach;
```
chibenwa, line 87: Careful to import order changes: you likely break checkstyles.

chibenwa (commented):
Nice little contib!
Thanks!

Arsnael (commented):
Read it, nothing to add. After the little fixes asked by @chibenwa , it should be approved without issues I think.
Thanks for the contribution :)

Arsnael (approved):

chibenwa (approved):

PR #1771, opened 2023-11-01
JAMES-3491 Deserialized keywords should return empty list and not nul…

Arsnael:
…l when missing

### DIFF
--- a/server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/filtering/RuleDTO.java
+++ b/server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/filtering/RuleDTO.java
@@ -28,6 +28,8 @@ import org.apache.james.jmap.api.filtering.Rule;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSetter;
+import com.fasterxml.jackson.annotation.Nulls;
 import com.github.fge.lambdas.Throwing;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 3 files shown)


server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/filtering/RuleDTO.java
```diff
@@ -237,7 +239,7 @@ public ActionDTO(@JsonProperty("appendIn") AppendInMailboxesDTO appendIn,
                          @JsonProperty("seen") boolean seen,
                          @JsonProperty("important") boolean important,
                          @JsonProperty("reject") boolean reject,
-                         @JsonProperty("keywords") List<String> keyworkds,
+                         @JsonProperty("keywords") @JsonSetter(nulls = Nulls.AS_EMPTY)  List<String> keyworkds,
```
quantranhong1999, line 242:
```suggestion
                         @JsonProperty("keywords") @JsonSetter(nulls = Nulls.AS_EMPTY)  List<String> keywords,
```

Well, there is a concern that persisted data using `keyworkds` would be lost. Should be acceptable IMO because this keyword introduction is recent, with no FE implementation yet so no real usage yet.

chibenwa (approved):

server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/filtering/RuleDTO.java
Arsnael: Honestly I spotted it and... I'm not so sure... We have potentially real usage on staging and prod for example. This action has been introduced in a previous release already, we were saving them, just not deserializing them correctly

Arsnael: Likely not prod but staging at least... Maybe we could think of fixing it bt being retrocompatible with keyworkds on at least read to not break anything at least. 

But that should be treated in an other ticket/PR then

quantranhong1999 (approved):

PR #1781, opened 2023-11-07
[Build] Few stability improvements...

Arsnael:
I keep seeing from time to time some builds failing because a module spent 25 minutes to test instead of less than 20. I think we can relax it to 30 minutes.

Also sometimes I saw some builds failing because in MailImplTest we test that we gate the last updated date of created email close to 1s to the now Date. A bit flacky, with all the heavy tests we run in parallel. So increased it to 2s


quantranhong1999 (approved):

chibenwa (approved):

vttranlina (approved):

Arsnael:
Last commit reason: https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1781/1/testReport/junit/org.apache.james.blob.cassandra.cache/CachedBlobStoreTest$MetricsTest/readBytesShouldRecordDistinctTimingsWhenRepeatAndBackendRead/

PR #1784, opened 2023-11-08
JAMES-3959 Starting distributed James without OpenSearch

chibenwa:


Arsnael (commented):
Looks good, but I think we need documentation for it as well, and a little note in upgrade instructions? :)

chibenwa:
> and a little note in upgrade instructions

Not needed: it is not a breaking change...

> we need documentation for it as well

That's the case for the all `search.properties` file as well then...

Arsnael:
> That's the case for the all `search.properties` file as well then...

Well people should know from somewhere that this is possible to do :)

quantranhong1999:
Maybe add a doc at `server/apps/distributed-app/docs/modules/ROOT/pages/configure/search.adoc` note this?

Arsnael (approved):

PR #1813, opened 2023-11-23
JAMES-2586 Implement PostgresQuotaCurrentValueDAO

quantranhong1999:
resolve https://github.com/linagora/james-project/issues/4938

### DIFF
--- a/backends-common/postgres/src/main/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAO.java
+++ b/backends-common/postgres/src/main/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAO.java
@@ -22,12 +22,10 @@ package org.apache.james.backends.postgres.quota;
 import static org.apache.james.backends.postgres.quota.PostgresQuotaModule.PostgresQuotaCurrentValueTable.COMPONENT;
 import static org.apache.james.backends.postgres.quota.PostgresQuotaModule.PostgresQuotaCurrentValueTable.CURRENT_VALUE;
 import static org.apache.james.backends.postgres.quota.PostgresQuotaModule.PostgresQuotaCurrentValueTable.IDENTIFIER;
+import static org.apache.james.backends.postgres.quota.PostgresQuotaModule.PostgresQuotaCurrentValueTable.PRIMARY_KEY_CONSTRAINT_NAME;
 import static org.apache.james.backends.postgres.quota.PostgresQuotaModule.PostgresQuotaCurrentValueTable.TABLE_NAME;
 import static org.apache.james.backends.postgres.quota.PostgresQuotaModule.PostgresQuotaCurrentValueTable.TYPE;
-import static org.jooq.impl.DSL.greatest;
-import static org.jooq.impl.DSL.zero;
 
-import java.util.Objects;
 import java.util.function.Function;
 
 import org.apache.james.backends.postgres.utils.PostgresExecutor;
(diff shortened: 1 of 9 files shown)


chibenwa (approved):

backends-common/postgres/src/test/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAOTest.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Test
+    void decreaseQuotaCurrentValueDownToNegativeShouldSetValueToZero() {
+        postgresQuotaCurrentValueDAO.increase(QUOTA_KEY, 100L).block();
+        postgresQuotaCurrentValueDAO.decrease(QUOTA_KEY, 1000L).block();
+
+        assertThat(postgresQuotaCurrentValueDAO.getQuotaCurrentValue(QUOTA_KEY).block().getCurrentValue())
+            .isZero();
+    }
```
quantranhong1999, line 100: Cassandra implementation allows negative value.

Here Postgres implementation I enforce the value to never be negative. 
It is good to enforce that, no? TBH I am not sure if we accept negative quota counter and how we handle them. I feel pretty strange having a negative counter (maybe by a decreasing mistake), it is like a cheat to the quota itself (having a quota bigger than the quota).

chibenwa: Please accept negative value.

Why? Because increase / decrease are likely not to be indempotant, and likely not to be retriable.

When this happens, emit a log and sanitize to 0.

We have tooling for recomputing quota.

I am unsure I want to have extremely strong consistency (transactions) only for this and likely can live with this.

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAO.java
```diff
@@ -0,0 +1,164 @@
...
+        this.postgresExecutor = postgresExecutor;
+    }
+
+    public Mono<Void> increase(QuotaKey quotaKey, long amount) {
+        return postgresExecutor.executeVoid(dslContext -> Mono.from(dslContext.insertInto(TABLE_NAME)
+            .set(IDENTIFIER, quotaKey.getIdentifier())
+            .set(COMPONENT, quotaKey.getQuotaComponent().getValue())
+            .set(TYPE, quotaKey.getQuotaType().getValue())
+            .set(CURRENT_VALUE, amount)
+            .onConflict(IDENTIFIER, COMPONENT, TYPE)
```
hungphan227, line 110: You can use the same way in my PR to avoid enumerating again all components of the primary key

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAO.java
```diff
@@ -0,0 +1,164 @@
...
+import org.apache.james.core.quota.QuotaType;
+import org.jooq.Record;
+
+import com.google.common.base.MoreObjects;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public class PostgresQuotaCurrentValueDAO {
+    public static class QuotaKey {
```
hungphan227, line 45: IMO this class should be created somewhere else and used by both PostgresQuotaCurrentValueDAO and CassandraQuotaCurrentValueDao to avoid code duplication

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAO.java
```diff
@@ -0,0 +1,120 @@
...
+        this.postgresExecutor = postgresExecutor;
+    }
+
+    public Mono<Void> increase(QuotaCurrentValue.Key quotaKey, long amount) {
+        return postgresExecutor.executeVoid(dslContext -> Mono.from(dslContext.insertInto(TABLE_NAME)
+            .set(IDENTIFIER, quotaKey.getIdentifier())
+            .set(COMPONENT, quotaKey.getQuotaComponent().getValue())
+            .set(TYPE, quotaKey.getQuotaType().getValue())
+            .set(CURRENT_VALUE, amount)
+            .onConflictOnConstraint(PRIMARY_KEY_CONSTRAINT_NAME)
```
Arsnael, line 57: Oh... does that allow you to try an insert and fallback on an update if already exists? Could not find that (honestly Jooq doc is terrible) and I'm happy to see this, will be useful with annotations and the hstore :)

backends-common/postgres/src/test/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAOTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+import org.apache.james.backends.postgres.PostgresExtension;
+import org.apache.james.core.quota.QuotaComponent;
+import org.apache.james.core.quota.QuotaCurrentValue;
+import org.apache.james.core.quota.QuotaType;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+class PostgresQuotaCurrentValueDAOTest {
```
Arsnael, line 34: Shouldn't we create a contract shared by cassandra and postgres?

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAO.java
quantranhong1999: Yes. Thanks to ChatGPT helped me and @hungphan227 on this yesterday :))

ChatGPT can help to write the SQL queries and translate them to jooq code :D

image

Arsnael: Damn does everybody but me has an account now? :rofl: ok thanks :)

backends-common/postgres/src/test/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAOTest.java
quantranhong1999: Likely we should. But now doing that would require refactoring Cassandra code (might need to introduce an interface `CurrentValueDAO` and reorganizing code). We disable the tests for Cassandra now so I would not want to risk that now TBH.

Maybe create a Polish ticket with points that we can deal with later?

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAO.java
hungphan227: I do not have it too. I am considering getting one. It is helpful :))

chibenwa: Please also ask him to be sure this is an atomic PG query -> No transaction involved?

backends-common/postgres/src/test/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAOTest.java
chibenwa: Let's wait for a 3rd implem for such a refactoring?

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/quota/PostgresQuotaCurrentValueDAO.java
quantranhong1999: image

It is expected though: every SQL query is an auto-commit transaction.

PR #1815, opened 2023-11-24
JAMES-2586 implement postgres quota limit dao

hungphan227:


core/src/main/java/org/apache/james/core/quota/QuotaLimitKey.java
```diff
@@ -0,0 +1,84 @@
...
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.core.quota;
+
+import java.util.Objects;
+
+import com.google.common.base.MoreObjects;
+
+public class QuotaLimitKey {
```
Arsnael, line 26: Ok we gonna have to do the choice of doing this one way and not two: https://github.com/apache/james-project/pull/1813

@quantranhong1999 adds the key as a subclass of QuotaCurrentValue, here it's a separate class. Personally I think this could be a subclass of QuotaLimit?

mailbox/postgres/src/test/java/org/apache/james/mailbox/postgres/quota/PostgresQuotaLimitDaoTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.apache.james.core.quota.QuotaComponent;
+import org.apache.james.core.quota.QuotaLimit;
+import org.apache.james.core.quota.QuotaLimitKey;
+import org.apache.james.core.quota.QuotaScope;
+import org.apache.james.core.quota.QuotaType;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+public class PostgresQuotaLimitDaoTest {
```
Arsnael, line 34: Feels like we could have a contract here as well, that would be shared by postgres and cassandra

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/quota/PostgresQuotaLimitDAO.java
```diff
@@ -0,0 +1,98 @@
...
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.mailbox.postgres.quota;
```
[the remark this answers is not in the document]
hungphan227: fixed

mailbox/postgres/src/test/java/org/apache/james/mailbox/postgres/quota/PostgresQuotaLimitDaoTest.java
Arsnael: Ok for later, as discussed

PR #1824, opened 2023-11-29
JAMES 3897: CrowdsecImapConnectionCheck

thanhbv200585:
image
I have problem in duplicated binding. Hope team to find out and catch this bug :)))


quantranhong1999:
> I have problem in duplicated binding. Hope team to find out and catch this bug :)))

Reason: Guice does not accept 2 beans for the same default `Set<ConnectionCheck>`, we need to aggregate James provided connection checks with additional connection checks (e.g. `CrowdSecConnectionCheck`).

Likely you would need to do something similar with `additional.healthchecks`: https://github.com/apache/james-project/blob/master/examples/custom-healthcheck/README.md#L34. e.g.  `additional.connectionchecks=org.apache.james.CustomConnectionCheck`.

And aggregate the extension connection checks to the `Set<ConnectionCheck>` bean like what we did with additional healthcheck: https://github.com/apache/james-project/blob/master/server/container/guice/common/src/main/java/org/apache/james/modules/PeriodicalHealthChecksModule.java#L64

Maybe other people can have a better idea as well...

protocols/imap/src/main/java/org/apache/james/imap/api/ConnectionCheck.java
```diff
@@ -0,0 +1,28 @@
...
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.imap.api;
+
+import java.net.InetSocketAddress;
+
+import org.reactivestreams.Publisher;
+
+public interface ConnectionCheck {
```
chibenwa, line 26: @FunctionalInterface ?

protocols/imap/src/main/java/org/apache/james/imap/api/AllowConnectionCheck.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.apache.james.imap.api;
+
+import java.net.InetSocketAddress;
+
+import org.reactivestreams.Publisher;
+
+import reactor.core.publisher.Mono;
+
+public class AllowConnectionCheck implements ConnectionCheck {
```
chibenwa, line 28: Could be a constant defined in `ConnectionCheck`. We do not need a separate class...

server/container/guice/protocols/imap/src/main/java/org/apache/james/modules/protocols/IMAPServerModule.java
```diff
@@ -103,6 +106,10 @@ protected void configure() {
         Multibinder.newSetBinder(binder(), GuiceProbe.class).addBinding().to(ImapGuiceProbe.class);
 
         Multibinder.newSetBinder(binder(), CertificateReloadable.Factory.class).addBinding().to(IMAPServerFactory.class);
+
+        Multibinder.newSetBinder(binder(), ConnectionCheck.class)
+            .addBinding()
+            .to(AllowConnectionCheck.class);
```
chibenwa, line 112: Just define the multibinder.

```suggestion
        Multibinder.newSetBinder(binder(), ConnectionCheck.class);
```

The binding to AllowConnectionCheck is useless as it is a Noop. Remove it.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServer.java
```diff
@@ -132,6 +134,7 @@ public Optional<OidcSASLConfiguration> getOidcSASLConfiguration() {
     private final ImapDecoder decoder;
     private final ImapMetrics imapMetrics;
     private final GaugeRegistry gaugeRegistry;
+    private final Set<ConnectionCheck> connectionCheckSet;
```
chibenwa, line 137:
```suggestion
    private final Set<ConnectionCheck> connectionChecks;
```

Here and everywhere.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapChannelUpstreamHandler.java
```diff
@@ -115,6 +118,11 @@ public ImapChannelUpstreamHandlerBuilder authenticationConfiguration(IMAPServer.
             return this;
         }
 
+        public ImapChannelUpstreamHandlerBuilder connectionCheckSet(Set<ConnectionCheck> connectionCheckSet) {
```
chibenwa, line 121:
```suggestion
        public ImapChannelUpstreamHandlerBuilder connectionChecks(Set<ConnectionCheck> connectionChecks) {

```

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapChannelUpstreamHandler.java
```diff
@@ -173,6 +184,9 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
         ImapSession imapsession = new NettyImapSession(ctx.channel(), secure, compress, authenticationConfiguration.isSSLRequired(),
             authenticationConfiguration.isPlainAuthEnabled(), sessionId,
             authenticationConfiguration.getOidcSASLConfiguration());
+
+        connectionCheckSet.forEach(connectionCheck -> Mono.from(connectionCheck.validate(imapsession.getRemoteAddress())).block());
```
chibenwa, line 188: Calling .block is FORBIDDEN

server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/DenyConnectionCheck.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.apache.james.imap.api.ConnectionCheck;
+import org.reactivestreams.Publisher;
+
+import reactor.core.publisher.Mono;
+
+public class DenyConnectionCheck implements ConnectionCheck {
+    @Override
+    public Publisher<Void> validate(InetSocketAddress remoteAddress) {
+        return Mono.error(() -> new RuntimeException());
```
chibenwa, line 32: Define this as a constant (lambda) in the `ConnectionCheck` interface please.

server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IpConnectionCheck.java
```diff
@@ -0,0 +1,48 @@
...
+public class IpConnectionCheck implements ConnectionCheck {
+    private Set<String> bannedIps = Set.of();
+
+    @Override
+    public Publisher<Void> validate(InetSocketAddress remoteAddress) {
+        String ip = remoteAddress.toString().split("/")[1].split(":")[0];
+
+        // check against bannedIps
+        if (bannedIps.stream().anyMatch(bannedIp -> bannedIp.equals(ip))) {
+            return Mono.error(() -> new RuntimeException("Invalid"));
```
chibenwa, line 39:
```suggestion
            return Mono.error(() -> new RuntimeException("Banned"));
```

server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IMAPServerTest.java
```diff
@@ -178,6 +182,63 @@ private IMAPServer createImapServer(HierarchicalConfiguration<ImmutableNode> con
         return imapServer;
     }
 
+    private IMAPServer createImapServer(HierarchicalConfiguration<ImmutableNode> config,
```
chibenwa, line 185: Please let's not duplicate all that setup code. Just for a test.

third-party/crowdsec/docker-compose.yml
```diff
@@ -9,7 +9,9 @@ services:
     command:
       - --generate-keystore
     volumes:
-      - ./sample-configuration/logback.xml:/root/conf/logback.xml
+#      - ./sample-configuration/logback.xml:/root/conf/logback.xml
```
chibenwa, line 12: Why comment this?

third-party/crowdsec/pom.xml
```diff
@@ -42,11 +42,42 @@
             <version>${project.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-guice-imap</artifactId>
```
chibenwa, line 47: scope: provided

third-party/crowdsec/pom.xml
```diff
@@ -42,11 +42,42 @@
...
         </dependency>
+        <dependency>
+            <groupId>${james.protocols.groupId}</groupId>
+            <artifactId>protocols-imap</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>${james.protocols.groupId}</groupId>
+            <artifactId>protocols-smtp</artifactId>
+            <version>${project.version}</version>
```
chibenwa, line 67: scope: provided

third-party/crowdsec/pom.xml
```diff
@@ -67,6 +98,10 @@
             <artifactId>guice</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>commons-net</groupId>
+            <artifactId>commons-net</artifactId>
```
chibenwa, line 103: scope: provided

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
```diff
@@ -0,0 +1,71 @@
...
+    private final CrowdsecClientConfiguration crowdsecClientConfiguration;
+
+    @Inject
+    public CrowdsecImapConnectionCheck(CrowdsecClientConfiguration crowdsecClientConfiguration) {
+        this.crowdsecClientConfiguration = crowdsecClientConfiguration;
+    }
+
+    @Override
+    public Publisher<Void> validate(InetSocketAddress remoteAddress) {
+        String ip = remoteAddress.toString().split("/")[1].split(":")[0];
```
chibenwa, line 46: This is way too hacky for me.

1/ split underneeth uses regexps and as such must be ban from production code. Use `Splitter.on('/')` instead.
2/ Extract the parsing of the IP into a dedicated method

```
String extractIp(InetSocketAddress address) {...}
```

And write dedicated tests for it.

3/ All this parsing ceremony is useless. Use third party APIs to do the job and read java doc. `address.getAddress().getHostAddress()` https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/net/InetSocketAddress.html#getAddress() https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F%2F/java/net/InetAddress.html#getHostAddress()

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
```diff
@@ -0,0 +1,71 @@
...
+        CrowdsecHttpClient client = new CrowdsecHttpClient(crowdsecClientConfiguration);
+        return client.getCrowdsecDecisions()
+            .filter(decision -> decision.stream().anyMatch(d -> isBanned(d, ip)))
```
chibenwa, line 49: What is `d` ?

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
```diff
@@ -0,0 +1,71 @@
...
+            .map(decision -> {
+                throw new RuntimeException("Ip " + ip + " is not allowed to connect to IMAP server");
```
chibenwa, line 51: Why? Because it was banned on crowdsec? Maybe we can even add a decision identifier in the log?

third-party/crowdsec/src/test/java/org/apache/james/CrowdsecImapConnectionCheckTest.java
```diff
@@ -0,0 +1,72 @@
...
+    @Test
+    void test() throws IOException, InterruptedException {
+        banIP("--ip", "127.0.0.3");
+        int port = crowdsecExtension.getCrowdsecContainer().getMappedPort(CROWDSEC_PORT);
+        CrowdsecClientConfiguration crowdsecClientConfiguration = new CrowdsecClientConfiguration(new URL("http://localhost:" + port + "/v1"), DEFAULT_API_KEY);
+
+        CrowdsecImapConnectionCheck connectionCheck = new CrowdsecImapConnectionCheck(crowdsecClientConfiguration);
+        assertThatThrownBy(() -> Mono.from(connectionCheck.validate(new InetSocketAddress("127.0.0.3", 8800))).block())
+            .hasMessage("Ip 127.0.0.3 is not allowed to connect to IMAP server");
+        System.out.println(fakeImapSession.getUserName());
```
chibenwa, line 64: Debug spotted

third-party/crowdsec/src/test/java/org/apache/james/CrowdsecImapConnectionCheckTest.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.james.imap.encode.FakeImapSession;
+import org.apache.james.model.CrowdsecClientConfiguration;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import reactor.core.publisher.Mono;
+
+public class CrowdsecImapConnectionCheckTest {
```
chibenwa, line 39: One more test with banning ipA but allow ipB
Test ranges: one test in the range, one test at the limit of the range, one test outside the range.

third-party/crowdsec/docker-compose.yml
thanhbv200585: Because it is easier for me to check log. I will uncomment it later

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
thanhbv200585: image
We know why this ip is banned. Is that enough info?

chibenwa: No it is not. Give as much info in James logs. Operators will thank you!

server/container/guice/protocols/imap/pom.xml
```diff
@@ -57,5 +57,11 @@
             <groupId>com.google.inject</groupId>
             <artifactId>guice</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.james</groupId>
+            <artifactId>apache-james-crowdsec</artifactId>
+            <version>3.9.0-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
```
quantranhong1999, line 65: please remove this

quantranhong1999:
> Likely you would need to do something similar with additional.healthchecks: https://github.com/apache/james-project/blob/master/examples/custom-healthcheck/README.md#L34. e.g. additional.connectionchecks=org.apache.james.CustomConnectionCheck

Maybe no need to declare a new configuration file just for connection check, we can declare it in `imapserver.xml`:

```xml
<imapservers>
    <imapserver enabled="true">
        <jmxName>imapserver</jmxName>
        <bind>0.0.0.0:143</bind>
...
     <additionalConnectionChecks>org.apache.james.connection.check.A,org.apache.james.connection.check.B</additionalConnectionChecks>
    </imapserver>
</imapservers>
```

Then in `IMAPServerModule`, use `GuiceGenericLoader` to load the connection check classes into instances. e.g. like what 
James did to extend the `ImapPackage`:
- https://github.com/apache/james-project/blob/master/server/container/guice/protocols/imap/src/main/java/org/apache/james/modules/protocols/IMAPServerModule.java#L147
- https://github.com/apache/james-project/blob/master/examples/custom-imap/README.md#L1

I can imagine splitting this into smaller tasks:
- Introduce ConnectionCheck interface and noop guice binding for it
- Implement `additionalConnectionChecks` extension for IMAP (configuration + guice loading + binding + example and documentation)
- Implement `CrowdSecConnectionCheck` and binding to James using the `additionalConnectionChecks` extension.

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
chibenwa: Still not applied...

third-party/crowdsec/src/test/java/org/apache/james/CrowdsecImapConnectionCheckTest.java
chibenwa: Once again, this test class is very poor.
We can do better.

third-party/crowdsec/src/main/java/org/apache/james/module/CrowdsecModule.java
```diff
@@ -36,6 +37,7 @@ public class CrowdsecModule extends AbstractModule {
     @Override
     protected void configure() {
         Multibinder.newSetBinder(binder(), CrowdsecImapConnectionCheck.class);
+        bind(ConnectionCheck.class).to(CrowdsecImapConnectionCheck.class);
```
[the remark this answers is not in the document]
thanhbv200585: > Multibinder.newSetBinder(binder(), CrowdsecImapConnectionCheck.class)
   .addBinding().to(CrowdsecImapConnectionCheck);

Can you explain more, what is the difference?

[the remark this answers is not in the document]
quantranhong1999: Should be 
```
Multibinder.newSetBinder(binder(), ConnectionCheck.class)
   .addBinding().to(CrowdsecImapConnectionCheck);
``` 
, no?

BTW we guided Guice to load `CrowdsecImapConnectionCheck` via `additionalConnectionChecks` extension mechanism already, we do not need to declare Guice binding for `CrowdsecImapConnectionCheck` in `CrowdSecModule` anymore IMO. (you can test though).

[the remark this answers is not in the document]
chibenwa: > Should be Multibinder.newSetBinder(binder(), ConnectionCheck.class).addBinding().to(CrowdsecImapConnectionCheck);

Yes

> BTW we guided Guice to load CrowdsecImapConnectionCheck via additionalConnectionChecks extension mechanism already

Ok thus completly uneeded

> additionalConnectionChecks extension mechanism 

Is `additionalConnectionChecks` documented?

Is setting up Crowdsec extensions with `additionalConnectionChecks` documented?

server/apps/distributed-app/docs/modules/ROOT/pages/extending/imap.adoc
```diff
@@ -22,4 +22,25 @@ Custom configuration can be obtained through `ImapConfiguration` class via the `
...
+```
+
+- `validate` method is used to check the connecting IP is secured.
+
+ConnectionCheck can be configured in `imapserver.xml` file:
+```
+<additionalConnectionCheck>org.apache.james.CrowdsecImapConnectionCheck</additionalConnectionCheck>
+```
+
+An example for configuration is available xref:james-project::../../../../../james-project/third-party/crowdsec/sample-configuration/imapserver.xml [here].
```
chibenwa, line 46: Put a github link instead.

server/container/guice/protocols/imap/src/main/java/org/apache/james/modules/protocols/IMAPServerModule.java
```diff
@@ -82,6 +87,7 @@
 import com.google.inject.multibindings.ProvidesIntoSet;
 
 public class IMAPServerModule extends AbstractModule {
+    public static final Logger LOGGER = LoggerFactory.getLogger(IMAPServerModule.class);
```
chibenwa, line 90: Not needed?

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
```diff
@@ -43,28 +45,32 @@ public class IMAPServerFactory extends AbstractServerFactory {
     protected final ThrowingFunction<HierarchicalConfiguration<ImmutableNode>, ImapSuite> imapSuiteProvider;
     protected final ImapMetrics imapMetrics;
     protected final GaugeRegistry gaugeRegistry;
+    protected final ThrowingFunction<HierarchicalConfiguration<ImmutableNode>, Set<ConnectionCheck>> connectionCheckProvider;
 
     @Inject
     @Deprecated
     public IMAPServerFactory(FileSystem fileSystem, ImapDecoder decoder, ImapEncoder encoder, ImapProcessor processor,
-                             MetricFactory metricFactory, GaugeRegistry gaugeRegistry) {
+                             MetricFactory metricFactory, GaugeRegistry gaugeRegistry, ThrowingFunction<HierarchicalConfiguration<ImmutableNode>, Set<ConnectionCheck>> connectionCheckProvider) {
```
chibenwa, line 53:
```suggestion
                             MetricFactory metricFactory, GaugeRegistry gaugeRegistry, Function<HierarchicalConfiguration<ImmutableNode>, Set<ConnectionCheck>> connectionCheckProvider
```

That's lighter. Tips: Sneaky throw.

Or even better: name this explicitly with an interface:

```
@FunctionalInterface
interface ConnectionCheckFactory {
    Set<ConnectionCheck> create(ImapConfiguration imapConfiguration)
}
```

BTW that's mabe a refactoring but we would need to convert the XML into an ImapConfiguration in this class as soon as we can.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapChannelUpstreamHandler.java
```diff
@@ -173,6 +184,9 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
         ImapSession imapsession = new NettyImapSession(ctx.channel(), secure, compress, authenticationConfiguration.isSSLRequired(),
             authenticationConfiguration.isPlainAuthEnabled(), sessionId,
             authenticationConfiguration.getOidcSASLConfiguration());
+        connectionChecks.forEach(connectionCheck -> Mono.from(connectionCheck.validate(imapsession.getRemoteAddress())).block());
+        connectionChecks.forEach(connectionCheck -> LOGGER.info("ConnectionCheck: {}", connectionCheck));
```
chibenwa, line 188: THis log is useless ?

```suggestion

```

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapChannelUpstreamHandler.java
```diff
@@ -173,6 +184,9 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
         ImapSession imapsession = new NettyImapSession(ctx.channel(), secure, compress, authenticationConfiguration.isSSLRequired(),
             authenticationConfiguration.isPlainAuthEnabled(), sessionId,
             authenticationConfiguration.getOidcSASLConfiguration());
+        connectionChecks.forEach(connectionCheck -> Mono.from(connectionCheck.validate(imapsession.getRemoteAddress())).block());
+        connectionChecks.forEach(connectionCheck -> LOGGER.info("ConnectionCheck: {}", connectionCheck));
+        LOGGER.info("ConnectionCheck {}", connectionChecks);
```
chibenwa, line 189: Idem, log this when starting the IMAP server. Not on every connection please.

```suggestion

```

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public Publisher<Void> validate(InetSocketAddress remoteAddress) {
+        String ip = remoteAddress.getAddress().getHostAddress();
+        CrowdsecHttpClient client = new CrowdsecHttpClient(crowdsecClientConfiguration);
+        return client.getCrowdsecDecisions()
+            .filter(decisions -> decisions.stream().anyMatch(decision -> isBanned(decision, ip)))
+            .map(decision -> {
+                throw new RuntimeException("Ip " + ip + " is not allowed to connect to IMAP server by Crowdsec");
+            });
```
chibenwa, line 52: throwing in a map step is an ignomy that denotes a partial understanding of reactor APIs.

That's defended. Fobidden.

Actually any side effect in a map step in strictly forbidden. 

Use `handle` instead.

chibenwa, line 52: We do not have better than a runtime exception? Can we create an exception please ?

server/container/guice/protocols/imap/src/main/java/org/apache/james/modules/protocols/IMAPServerModule.java
```diff
@@ -160,6 +168,22 @@ private ImapPackage retrievePackages(GuiceGenericLoader loader, HierarchicalConf
...
+            .flatMap(Arrays::stream)
+            .map(ClassName::new)
+            .map(Throwing.function(loader::instantiate))
+            .map(ConnectionCheck.class::cast)
+            .collect(ImmutableSet.toImmutableSet());
+    }
+
+    private ThrowingFunction<HierarchicalConfiguration<ImmutableNode>, Set<ConnectionCheck>> retrieveConnectionChecks(GuiceGenericLoader loader) {
+        return configuration -> retrieveConnectionChecks(loader, configuration);
+    }
```
chibenwa, line 185: -> Merge in one method and inject `ConnectionCheckFactory` instead...

chibenwa (commented):
Still some job to be done...

server/apps/distributed-app/docs/modules/ROOT/pages/extending/imap.adoc
```diff
@@ -22,4 +22,25 @@ Custom configuration can be obtained through `ImapConfiguration` class via the `
...
 
-See this page for xref:configure/imap.adoc#_extending_imap[more details on configuring IMAP extensions].
\ No newline at end of file
+See this page for xref:configure/imap.adoc#_extending_imap[more details on configuring IMAP extensions].
+
+== IMAP additional Connection Checks
+
+James allows defining your own additional connection checks to guarantee that the connecting IP is secured.
+
+An Connection check is a functional interface:
```
Arsnael, line 31:
```suggestion
A custom connection check should implement the following functional interface:
```

server/apps/distributed-app/docs/modules/ROOT/pages/extending/imap.adoc
```diff
@@ -22,4 +22,25 @@ Custom configuration can be obtained through `ImapConfiguration` class via the `
...
+```
+@FunctionalInterface
+public interface ConnectionCheck {
+    Publisher<Void> validate(InetSocketAddress remoteAddress);
+}
+```
+
+- `validate` method is used to check the connecting IP is secured.
+
+ConnectionCheck can be configured in `imapserver.xml` file:
```
Arsnael, line 41:
```suggestion
Then the custom defined ConnectionCheck can be added in `imapserver.xml` file:
```

server/apps/distributed-app/docs/modules/ROOT/pages/extending/imap.adoc
```diff
@@ -22,4 +22,25 @@ Custom configuration can be obtained through `ImapConfiguration` class via the `
...
+```
+<additionalConnectionCheck>org.apache.james.CrowdsecImapConnectionCheck</additionalConnectionCheck>
+```
+
+An example for configuration is available link:https://github.com/apache/james-project/blob/master/third-party/crowdsec/sample-configuration/acquis.yaml[here].
```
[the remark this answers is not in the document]
thanhbv200585: > Put a github link instead.

This link will exist in this PR

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
```diff
@@ -45,31 +45,30 @@ public class IMAPServerFactory extends AbstractServerFactory {
...
-        this.connectionCheckProvider = connectionCheckProvider;
+        this.connectionChecks = connectionChecks;
         this.imapSuiteProvider = any -> new ImapSuite(decoder, encoder, processor);
         this.imapMetrics = new ImapMetrics(metricFactory);
         this.gaugeRegistry = gaugeRegistry;
     }
 
     public IMAPServerFactory(FileSystem fileSystem, ThrowingFunction<HierarchicalConfiguration<ImmutableNode>, ImapSuite> imapSuiteProvider,
-                             MetricFactory metricFactory, GaugeRegistry gaugeRegistry, ThrowingFunction<HierarchicalConfiguration<ImmutableNode>, Set<ConnectionCheck>> connectionCheckProvider) {
+                             MetricFactory metricFactory, GaugeRegistry gaugeRegistry,Set<ConnectionCheck> connectionChecks) {
```
chibenwa, line 62:
```suggestion
                             MetricFactory metricFactory, GaugeRegistry gaugeRegistry, Set<ConnectionCheck> connectionChecks) {

```

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
```diff
@@ -45,31 +45,30 @@ public class IMAPServerFactory extends AbstractServerFactory {
...
     protected final ImapMetrics imapMetrics;
     protected final GaugeRegistry gaugeRegistry;
-    protected final ThrowingFunction<HierarchicalConfiguration<ImmutableNode>, Set<ConnectionCheck>> connectionCheckProvider;
+    protected final Set<ConnectionCheck> connectionChecks;
 
     @Inject
     @Deprecated
     public IMAPServerFactory(FileSystem fileSystem, ImapDecoder decoder, ImapEncoder encoder, ImapProcessor processor,
-                             MetricFactory metricFactory, GaugeRegistry gaugeRegistry, ThrowingFunction<HierarchicalConfiguration<ImmutableNode>, Set<ConnectionCheck>> connectionCheckProvider) {
+                             MetricFactory metricFactory, GaugeRegistry gaugeRegistry, Set<ConnectionCheck> connectionChecks) {
```
chibenwa, line 53: No It was better before: one set of connections set per ports. More flexibility.

chibenwa (commented):
The suggestions I did were not applied.

protocols/imap/src/main/java/org/apache/james/imap/api/ConnectionCheckFactory.java
```diff
@@ -0,0 +1,37 @@
...
+    Set<ConnectionCheck> create(GuiceGenericLoader loader);
+
+    public static class Impl implements ConnectionCheckFactory {
+
+        @Override
+        public Set<ConnectionCheck> create(GuiceGenericLoader loader) {
+            return null;
+        }
+    }
+}
```
chibenwa, line 37: That's not what I asked for

```
@FunctionalInterface
interface ConnectionCheckFactory {
    Set<ConnectionCheck> create(ImapConfiguration imapConfiguration)
}
```

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
thanhbv200585: You mean that parse imapserver.xml to get set of ConnectionCheck, then add them into  ImapConfig?

quantranhong1999: Let's put `additionalConnectionChecks` into `ImapConfiguration`. And init `Set<ConnectionCheck>` from `ImapConfiguration`, as Benoit suggestion.

chibenwa:
Please add tests regarding PROXY protocol tests (we should ban the source IP and not th proxy ip)

protocols/imap/src/main/java/org/apache/james/imap/api/ImapConfiguration.java
```diff
@@ -173,8 +181,9 @@ public ImapConfiguration build() {
     private final boolean isCondstoreEnable;
     private final boolean provisionDefaultMailboxes;
     private final Properties customProperties;
+    private final ImmutableSet<String> connectionChecks;
```
quantranhong1999, line 184:
```suggestion
    private final ImmutableSet<String> additionalConnectionChecks;
```

to distinguish with built-in connection check.

Shift+F6

protocols/imap/pom.xml
```diff
@@ -60,6 +60,11 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-guice-utils</artifactId>
+            <scope>provided</scope>
+        </dependency>
```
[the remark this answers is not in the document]
quantranhong1999: not resolved yet.

quantranhong1999 (commented):
And PROXY protocol tests left?

LGTM overall though.

Arsnael (commented):
Read it, nothing to add

server/container/guice/protocols/imap/src/main/java/org/apache/james/modules/protocols/IMAPServerModule.java
```diff
@@ -160,6 +166,23 @@ private ImapPackage retrievePackages(GuiceGenericLoader loader, HierarchicalConf
         return ImapPackage.and(packages);
     }
 
+    private static ImmutableSet<ConnectionCheck> getConnectionChecks(HierarchicalConfiguration<ImmutableNode> configuration) {
```
quantranhong1999, line 169: Where is this being used?

protocols/imap/src/main/java/org/apache/james/imap/api/ConnectionCheckFactory.java
```diff
@@ -0,0 +1,56 @@
...
+        public Set<ConnectionCheck> create(ImapConfiguration imapConfiguration) {
+            return Optional.ofNullable(imapConfiguration.getConnectionChecks())
+                .orElse(ImmutableSet.of())
+                .stream()
+                .map(ClassName::new)
+                .map(Throwing.function(loader::instantiate))
+                .map(ConnectionCheck.class::cast)
+                .collect(ImmutableSet.toImmutableSet());
+        }
+    }
```
quantranhong1999, line 55: Remove this impl class to another module that handles guice loading/binding stuff e.g. at the same module with `server/container/guice/protocols/imap/src/main/java/org/apache/james/modules/protocols/IMAPServerModule.java`? So we can remove the `guice-utils` out of `protocols/imap`.

The idea is `protocols/imap` does not need to care about guice loading-binding duty.

protocols/imap/pom.xml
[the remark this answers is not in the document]
quantranhong1999: Wrong placing for `GuiceGenericLoader` led to this dependency declaration. Please find a way to remove it out of `protocols/imap/pom.xml`.

protocols/imap/src/main/java/org/apache/james/imap/api/ConnectionCheckFactory.java
thanhbv200585: So can we move `Impl` class into another module? E.g: `server/container/util/src/main/java/org/apache/james/util`

quantranhong1999: wherever appropriate (I can sit down with u tmr), but for sure not in `protocols/imap` as ATM.

protocols/imap/src/main/java/org/apache/james/imap/api/ImapConfiguration.java
```diff
@@ -167,7 +167,7 @@ public ImapConfiguration build() {
                     isCondstoreEnable.orElse(DEFAULT_CONDSTORE_DISABLE),
                     provisionDefaultMailboxes.orElse(DEFAULT_PROVISION_DEFAULT_MAILBOXES),
                     customProperties.orElseGet(Properties::new),
-                    connectionChecks);
+                additionalConnectionChecks);
```
Arsnael, line 170: indent

Arsnael (commented):
Proxy protocol tests still missing?

And please when something is resolved, mark it as resolved, before it becomes a mess

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/HAProxyMessageHandler.java
```diff
@@ -60,6 +69,7 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                 // Refresh MDC info to account for proxying
                 MDCBuilder boundMDC = IMAPMDCContext.boundMDC(ctx);
 
+                connectionChecks.forEach(connectionCheck -> Mono.from(connectionCheck.validate((InetSocketAddress) ctx.channel().remoteAddress())).block());
```
quantranhong1999, line 72: IMO we should get the client IP from `HAProxyMessage`, not parsing again.

Maybe `haproxyMsg.sourceAddress()` is the client IP instead? 

TBH I am not sure. Please test more: at least try docker-compose with haproxy and try to see which one is the client IP. The current IMAPServerTest can not be fully trusted IMO.

After being sure about the client IP, this work should be good IMO.

thanhbv200585: Yep, you're right, `haproxyMsg.sourceAddress()` is the client IP 
I will adapt the code later. Thank for your comments!!
image

image

chibenwa (commented):
With proxy protocol are connection checks executed to time?

 -> One time upon initial connection
 -> One time upon receiving the PROXY protocol message
 
 ?

thanhbv200585:
>With proxy protocol are connection checks executed to time?
> -> One time upon initial connection
> -> One time upon receiving the PROXY protocol message

connection checks will be executed upon receiving the PROXY protocol message

server/container/guice/protocols/imap/src/main/java/org/apache/james/modules/protocols/IMAPServerModule.java
```diff
@@ -103,6 +105,8 @@ protected void configure() {
         Multibinder.newSetBinder(binder(), GuiceProbe.class).addBinding().to(ImapGuiceProbe.class);
 
         Multibinder.newSetBinder(binder(), CertificateReloadable.Factory.class).addBinding().to(IMAPServerFactory.class);
+        Multibinder.newSetBinder(binder(), ConnectionCheck.class);
```
chibenwa, line 108:
```suggestion
```

Why is this line needed?

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/HAProxyMessageHandler.java
chibenwa: Flux.fromIterable(...).concatMap(...)>then().block();

Please use reactor control structures.

chibenwa: > I will adapt the code later.

Now is about the time.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapChannelUpstreamHandler.java
```diff
@@ -173,6 +184,7 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
         ImapSession imapsession = new NettyImapSession(ctx.channel(), secure, compress, authenticationConfiguration.isSSLRequired(),
             authenticationConfiguration.isPlainAuthEnabled(), sessionId,
             authenticationConfiguration.getOidcSASLConfiguration());
+        connectionChecks.forEach(connectionCheck -> Mono.from(connectionCheck.validate(imapsession.getRemoteAddress())).block());
```
chibenwa, line 187: This is call is useless at best, extermely dangerous at worst if HA proxy is enabled...
channelActive is called of course also when establishing a connection when proxy protocol is on, then called again after receiving the proxy protocol payload.
Please do not call this here if the proxy protocol is on.

chibenwa (commented):
Proxy protocol support is needed for SMTP too.

Also I would liek to se integration tests: 

Scenari (IMAP & SMTP):

```
Start James memory + crowdsec
 (connect, bad login x3)
 Ensure connection is rejected
 ```

```
Start James memory + crowdsec + proxy protocol
 (connect, proxy message, bad login x3)
 Ensure (connect + proxy message) is rejected
 + add another test that other people connecting through the proxy should be rejected.
 ```
 
 Unless I see this I would assume we blacklist the proxy IP which of course is not acceptable.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/HAProxyMessageHandler.java
```diff
@@ -68,8 +68,8 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                 LOGGER.info("Connection from {} runs through {} proxy", haproxyMsg.sourceAddress(), haproxyMsg.destinationAddress());
                 // Refresh MDC info to account for proxying
                 MDCBuilder boundMDC = IMAPMDCContext.boundMDC(ctx);
+                Flux.fromIterable(connectionChecks).concatMap(connectionCheck -> connectionCheck.validate(InetSocketAddress.createUnresolved(haproxyMsg.sourceAddress(), haproxyMsg.sourcePort()))).then().block();
```
quantranhong1999, line 71: Why not just `InetSocketAddress abcxyz = new InetSocketAddress(haproxyMsg.sourceAddress(), haproxyMsg.sourcePort())` and reuse it above also to recalculate the variable again?

examples/proxy-smtp/logback.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
```
quantranhong1999, line 2: Why?

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
```diff
@@ -41,7 +41,7 @@ public CrowdsecImapConnectionCheck(CrowdsecClientConfiguration crowdsecClientCon
 
     @Override
     public Publisher<Void> validate(InetSocketAddress remoteAddress) {
-        String ip = remoteAddress.getAddress().getHostAddress();
+        String ip = remoteAddress.getHostName();
```
quantranhong1999, line 44: Q: Are they the same?

thanhbv200585: There is a bit difference.

```
InetSocketAddress.createUnresolved(haproxyMsg.sourceAddress(), haproxyMsg.sourcePort())
```

I created `InetSocketAddress` by this way, so addr will be null. If I call `getAddress`, it will throw NullPointerException. That's why I changed it
```
private InetAddress getAddress() {
    return addr;
}
```

quantranhong1999: Wrong choice led to regression is my feeling about this...

thanhbv200585: Yep, but one positive thing is that it looks more succinct :)

quantranhong1999: I disagree with keeping this change. `succinct` has no point if it is a regression.

see:
- https://stackoverflow.com/questions/55521964/different-behavior-of-inetsocketaddressgethostname-in-windows-and-linux-for-12
- https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#getHostName--

hostname is different from IP address. Be careful with what you are doing.

quantranhong1999:
@thanhbv200585 I helped you setting up the CrowdSec integration test and clean some code [here](https://github.com/quantranhong1999/james-project/commits/thanh-crowdsec-imap/). Please have a look and finish the remaining proxy test case.

Context of the commit:
- Better handle Crowdsec timeout: wrong CrowdSec endpoint configuration could lead to hanging CrowdSec connection check
- Better log for refused connection from an IP
- Better reuse `CrowdsecHttpClient` object in `CrowdsecImapConnectionCheck`. Do not create a new client for every check.
- Stay away from evil regression `remoteAddress.getHostName`
- Only perform connection check if needed. Most of the cases it would be empty connection check therefore no need Flux manipulation.
- Set up Integration test for CrowdSec: Plug James log file into CrowdSec, then let CrowdSec and James do their work.

quantranhong1999:
> Proxy protocol support is needed for SMTP too.

For the SMTP proxy case, I propose another ticket: https://github.com/linagora/james-project/issues/5003

third-party/crowdsec/src/test/java/org/apache/james/CrowdsecIntegrationTest.java
```diff
@@ -0,0 +1,91 @@
...
+                    .login(BOB, BAD_PASSWORD))
+                    .isInstanceOf(IOException.class)
+                    .hasMessage("Login failed"));
+
+        // THEN connection from the IP would be blocked. CrowdSec takes time to processing the ban decision therefore the await below.
+        CALMLY_AWAIT.atMost(Durations.TEN_SECONDS)
+            .untilAsserted(() -> assertThatThrownBy(() -> testIMAPClient.connect("127.0.0.1", server.getProbe(ImapGuiceProbe.class).getImapPort()))
+                .isInstanceOf(EOFException.class)
+                .hasMessage("Connection closed without indication."));
+    }
```
quantranhong1999, line 90: Where is the proxy test?

thanhbv200585: Sorry, I misunderstood it would be in another ticket

Arsnael: No part of this ticket and PR, fix it here please

server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IpConnectionCheck.java
```diff
@@ -0,0 +1,47 @@
...
+import reactor.core.publisher.Mono;
+
+public class IpConnectionCheck implements ConnectionCheck {
+    private Set<String> bannedIps = Set.of();
+
+    @Override
+    public Publisher<Void> validate(InetSocketAddress remoteAddress) {
+        String ip = remoteAddress.getAddress().getHostAddress();
+        // check against bannedIps
+        if (bannedIps.stream().anyMatch(bannedIp -> bannedIp.equals(ip))) {
```
vttranlina, line 37: very easy with 
`if (bannedIps.contains(ip)) {` 

and if you don't add a new value to `bannedIps` somewhere, then it always return false

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
```diff
@@ -36,37 +39,45 @@
...
         this.gaugeRegistry = gaugeRegistry;
+        this.connectionCheckFactory = connectionCheckFactory;
     }
 
     protected IMAPServer createServer(HierarchicalConfiguration<ImmutableNode> config) {
         ImapSuite imapSuite = imapSuiteProvider.apply(config);
-        return new IMAPServer(imapSuite.getDecoder(), imapSuite.getEncoder(), imapSuite.getProcessor(), imapMetrics, gaugeRegistry);
+        ImmutableSet<String> connectionChecks = Arrays.stream(config.getStringArray("additionalConnectionChecks")).collect(ImmutableSet.toImmutableSet());
+
+        return new IMAPServer(imapSuite.getDecoder(), imapSuite.getEncoder(), imapSuite.getProcessor(), imapMetrics, gaugeRegistry, connectionCheckFactory.create(ImapConfiguration.builder()
```
vttranlina, line 76: IMO
```
ImapConfiguration.builder()
            .connectionChecks(connectionChecks)
            .build()
```
it looks quite mock,  
- It should be full (real) "ImapConfiguration", not just a empty ImapConfiguration with only `connectionChecks` value
- or the input of `ConnectionCheckFactory::create` should be `HierarchicalConfiguration<ImmutableNode> config` or another

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
```diff
@@ -0,0 +1,80 @@
...
+    @Inject
+    public CrowdsecImapConnectionCheck(CrowdsecClientConfiguration crowdsecClientConfiguration) {
+        this.client = new CrowdsecHttpClient(crowdsecClientConfiguration);
+    }
+
+    @Override
+    public Publisher<Void> validate(InetSocketAddress remoteAddress) {
+        String ip = remoteAddress.getAddress().getHostAddress();
+
+        return client.getCrowdsecDecisions()
```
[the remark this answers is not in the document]
vttranlina: Cache example: https://github.com/apache/james-project/blob/eff08aed533af2cb79035a05d81e5d698de48979/server/container/guice/configuration/src/main/java/org/apache/james/utils/PropertiesProvider.java#L83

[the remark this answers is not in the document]
chibenwa: IMO caching is overkill.

Connection churn rate is very low for IMAP and SMTP so I expect no noticeable effect on performance. 
Caching brings possible inconsistencies and do not seem like strictly needed.

However I would agree agressively lowering the timeout to say 500ms. If crowdsec do not answer in the given time then it would be ok to me to bypass it. Even better: make the timeout configurable.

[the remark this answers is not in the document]
vttranlina: @thanhbv200585 
if we don't cache it, 
can you check the http api of Crowdsec support is reactive or not?
if yes, the `client.getCrowdsecDecisions()` return `Flux<Result>` is better than `Mono<List<Result>>`

server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IpConnectionCheck.java
```diff
@@ -0,0 +1,47 @@
...
+            return Mono.error(() -> new RuntimeException("Banned"));
+        } else {
+            return Mono.empty();
+        }
+    }
+
+    public void setBannedIps(Set<String> bannedIps) {
```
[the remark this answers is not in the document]
thanhbv200585: Because I need to add bannedIps for testing
```
private void addBannedIps(String clientIp) {
            imapServer.getConnectionChecks().stream()
                .filter(check -> check instanceof IpConnectionCheck)
                .map(check -> (IpConnectionCheck) check)
                .forEach(ipCheck -> ipCheck.setBannedIps(Set.of(clientIp)));
        }
```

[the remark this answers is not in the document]
vttranlina: If it is a method for testing

1.  you can add annotation it by @VisibleForTesting
2. the `bannedIps` is always empty on prod -> the class `IpConnectionCheck` does not make sense

third-party/crowdsec/src/test/java/org/apache/james/CrowdsecIntegrationTest.java
```diff
@@ -87,5 +99,55 @@ void ipShouldBeBannedByCrowdSecWhenFailingToImapLoginThreeTimes(GuiceJamesServer
...
+                }});
+        clientConnection.finishConnect();
+        // THEN connection from the IP would be blocked. CrowdSec takes time to processing the ban decision therefore the await below.
+        CALMLY_AWAIT.atMost(Durations.TEN_SECONDS)
+            .untilAsserted(() -> assertThatThrownBy(() ->  clientConnection.connect(new InetSocketAddress(haProxyExtension.getHaproxyContainer().getHost(), haProxyExtension.getHaproxyContainer().getMappedPort(IMAPS_PORT))))
+                .isInstanceOf(EOFException.class)
+                .hasMessage("Connection closed without indication."));
+
+    }
+
```
Arsnael, line 132: Can we have a test as well for proxy where like above you have an IP that attempts to login through the proxy 3 times with wrong creds. Then an other IP tries to login via the same proxy with right creds and it should succeed. Just to make sure the ip banned is the one from the client and not the proxy? :) Should be similar to test above with an extra step, easy enough :)

server/protocols/protocols-imap4/src/test/java/org/apache/james/imapserver/netty/IpConnectionCheck.java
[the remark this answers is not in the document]
vttranlina: ah, my mistake, it is in the "test" package. 
forget it

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
thanhbv200585: I prefer the option 2

third-party/crowdsec/src/test/java/org/apache/james/CrowdsecIntegrationTest.java
thanhbv200585: Yep, I understand the test script but there is a programming problem.

third-party/crowdsec/src/test/resources/haproxy.cfg
```diff
@@ -0,0 +1,24 @@
+global
```
quantranhong1999, line 1: Q: Why do we need this file?

third-party/crowdsec/src/test/java/org/apache/james/CrowdsecIntegrationTest.java
```diff
@@ -0,0 +1,153 @@
...
+        haProxyExtension.getHaproxyContainer().execInContainer(createCommandHAProxyConfig(server));
+        clientConnection = SocketChannel.open();
+
+        clientConnection.connect(new InetSocketAddress(haProxyExtension.getHaproxyContainer().getHost(), haProxyExtension.getHaproxyContainer().getMappedPort(IMAPS_PORT)));
+        IntStream.range(0, 3)
+            .forEach(any -> {
+                try {
+                    clientConnection.write(ByteBuffer.wrap(String.format("a0 LOGIN %s %s\r\n",
+                        "TCP4", CLIENT_IP, haProxyExtension.getHaproxyContainer().getHost(), 65535, haProxyExtension.getHaproxyContainer().getMappedPort(IMAPS_PORT),
+                        BOB, BAD_PASSWORD).getBytes(StandardCharsets.UTF_8)));
```
quantranhong1999, line 118: Why are there too many string format objects while there are just 2 `%s`?

I believe we just need `BOB, BAD_PASSWORD`.

third-party/crowdsec/src/main/java/org/apache/james/CrowdsecImapConnectionCheck.java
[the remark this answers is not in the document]
thanhbv200585: Maybe in another ticket?

quantranhong1999:
Related:
```java
Existing failures - 2
testServer – org.apache.james.app.spring.JamesAppSpringMainTest

Error creating bean with name 'imapserver' defined in class path resource [META-INF/spring/imapserver-context.xml]: Unsatisfied dependency expressed through constructor parameter 6; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.apache.james.imap.api.ConnectionCheckFactory' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
Stacktrace
org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'imapserver' defined in class path resource [META-INF/spring/imapserver-context.xml]: Unsatisfied dependency expressed through constructor parameter 6; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.apache.james.imap.api.ConnectionCheckFactory' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

cc @thanhbv200585 please pay attention to the CI build and fix failed tests.

quantranhong1999:
Nothing to add. Waiting for a green build...

Arsnael (approved):

PR #1825, opened 2023-11-29
[Build] Do not build on ephemeral nodes

Arsnael:
Infra team put some extra nodes called ephemeral are just overflow nodes, kicking in when no buildsxx nodes are available for new builds.

We can't run on those as there is less resources on those than buildsxx, which is not enough for James heavy builds.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,7 @@
  *
  */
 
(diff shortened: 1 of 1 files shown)


Arsnael:
cf https://issues.apache.org/jira/browse/INFRA-25053

chibenwa (approved):

PR #1839, opened 2023-12-05
JAMES-2586 Implement PostgresPerUserMaxQuotaManager

hungphan227:


mailbox/api/src/main/java/org/apache/james/mailbox/quota/Limits.java
```diff
@@ -17,7 +17,7 @@
  * under the License.                                           *
  ****************************************************************/
 
-package org.apache.james.mailbox.cassandra.quota;
+package org.apache.james.mailbox.quota;
```
chibenwa, line 20: Did you remove the cassandra classes?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/quota/JPAPerUserMaxQuotaDAO.java
```diff
@@ -1,238 +0,0 @@
-/****************************************************************
```
chibenwa, line 1: Please kill MaxDomainMessageCount MaxDomainStorage MaxGlobalMessageCount MaxGlobalStorage MaxUserMessageCount and MaxUserStorage as well

chibenwa (approved):

Arsnael (approved):

PR #1845, opened 2023-12-08
JAMES-3539 PushSubscription/get should return expired subscriptions

quantranhong1999:
Otherwise, client would not be able to extend or destroy expired subscriptions.

### DIFF
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/pushsubscription/PushSubscriptionHelpers.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/api/pushsubscription/PushSubscriptionHelpers.java
@@ -48,8 +48,8 @@ public class PushSubscriptionHelpers {
             .orElse(maxExpiresTime));
     }
 
-    public static boolean isNotOutdatedSubscription(PushSubscription subscription, Clock clock) {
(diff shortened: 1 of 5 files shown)


server/data/data-jmap/src/main/java/org/apache/james/jmap/api/pushsubscription/PushSubscriptionHelpers.java
```diff
@@ -48,8 +48,8 @@ public static PushSubscriptionExpiredTime evaluateExpiresTime(Optional<ZonedDate
             .orElse(maxExpiresTime));
     }
 
-    public static boolean isNotOutdatedSubscription(PushSubscription subscription, Clock clock) {
-        return subscription.expires().isAfter(ZonedDateTime.now(clock));
+    public static boolean isNotOutdatedSubscription(PushSubscription subscription) {
+        return subscription.expires().isAfter(ZonedDateTime.now());
```
chibenwa, line 52: Why? Sounds like a regression...

Arsnael (commented):
Read it, I have same remark as @chibenwa , why not keeping the clock?

Arsnael (approved):

chibenwa (approved):

vttranlina (approved):

PR #1849, opened 2023-12-11
JAMES-2586 Implement PostgresSieveScriptDAO + PostgresSieveRepository

quantranhong1999:
TODO optimize + guice binding + remove JPA code.


server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
```diff
@@ -0,0 +1,58 @@
...
+
+        PostgresTable TABLE = PostgresTable.name(TABLE_NAME.getName())
+            .createTableStep(((dsl, tableName) -> dsl.createTableIfNotExists(tableName)
+                .column(USERNAME)
+                .column(SCRIPT_NAME)
+                .column(SCRIPT_SIZE)
+                .column(SCRIPT_CONTENT)
+                .column(IS_ACTIVE)
+                .column(ACTIVATION_DATE_TIME)
+                .primaryKey(USERNAME, SCRIPT_NAME)))
```
chibenwa, line 51: We have constraints, let's try to encode them on the DB:

 - For a given user not two scripts with the same name
 - For a given user, only one script should be active
 
 Potentially encoding this into the data layer would ease the code a lot!

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jooq.impl.SQLDataType;
+
+public interface PostgresSieveModule {
+    interface PostgresSieveScriptTable {
+        Table<Record> TABLE_NAME = DSL.table("sieve_scripts");
+
+        Field<String> USERNAME = DSL.field("username", SQLDataType.VARCHAR(255).notNull());
+        Field<String> SCRIPT_NAME = DSL.field("script_name", SQLDataType.VARCHAR.notNull());
+        Field<Long> SCRIPT_SIZE = DSL.field("script_size", SQLDataType.BIGINT);
+        Field<String> SCRIPT_CONTENT = DSL.field("script_content", SQLDataType.VARCHAR.notNull());
```
chibenwa, line 39: Text please

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
```diff
@@ -0,0 +1,58 @@
...
+        Field<Boolean> IS_ACTIVE = DSL.field("is_active", SQLDataType.BOOLEAN.notNull());
+        Field<OffsetDateTime> ACTIVATION_DATE_TIME = DSL.field("activation_date_time", SQLDataType.OFFSETDATETIME);
```
chibenwa, line 41: Maybe it is not an issue but be aware time zone is not stored but only the offset is...

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveRepository.java
```diff
@@ -100,166 +80,154 @@ private boolean overQuotaAfterModification(long usedSpace, long size, QuotaSizeL
...
-                rollbackTransactionIfActive(entityManager.getTransaction());
-                throw e;
-            }
-        }).sneakyThrow(), throwStorageExceptionConsumer("Unable to put script for user " + username.asString()));
+    public void putScript(Username username, ScriptName name, ScriptContent content) throws QuotaExceededException {
+        reThrowQuotaExceededException(() -> spaceThatWillBeUsedByNewScript(username, name, content.length())
+            .flatMap(spaceToUse -> throwOnOverQuota(username, spaceToUse)
+                .thenEmpty(Flux.merge(
+                        updateSpaceUsed(username, spaceToUse),
+                        postgresSieveScriptDAO.upsertScript(PostgresSieveScript.builder()
```
chibenwa, line 88: update space after inserting the script, using feedback of the insertion (`RETURNING` keyword)?

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveRepository.java
```diff
@@ -100,166 +80,154 @@ private boolean overQuotaAfterModification(long usedSpace, long size, QuotaSizeL
...
-            }
+        postgresSieveScriptDAO.getScript(username, oldName)
+            .blockOptional()
+            .orElseThrow(ScriptNotFoundException::new);
 
-            JPASieveScript sieveScriptToRename = sieveScript.get();
-            sieveScriptToRename.renameTo(newName);
-        }).sneakyThrow(), throwStorageExceptionConsumer("Unable to rename script " + oldName.getValue() + " for user " + username.asString()));
+        postgresSieveScriptDAO.renameScript(username, oldName, newName)
+            .block();
```
chibenwa, line 223: One single update with returning instead ?

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveRepository.java
```diff
@@ -100,166 +80,154 @@ private boolean overQuotaAfterModification(long usedSpace, long size, QuotaSizeL
...
     @Override
-    public void renameScript(Username username, ScriptName oldName, ScriptName newName) {
-        transactionRunner.runAndHandleException(Throwing.<EntityManager>consumer(entityManager -> {
-            Optional<JPASieveScript> sieveScript = findSieveScript(username, oldName, entityManager);
-            if (!sieveScript.isPresent()) {
-                rollbackTransactionIfActive(entityManager.getTransaction());
-                throw new ScriptNotFoundException("Unable to find script " + oldName.getValue() + " for user " + username.asString());
-            }
+    public void renameScript(Username username, ScriptName oldName, ScriptName newName) throws DuplicateException, ScriptNotFoundException {
+        checkIfNewScriptNameDuplicated(username, newName);
```
chibenwa, line 216: We should rather use a constraint on the DB and not check this ourselves.

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
quantranhong1999: This uses `timestamp(p)` data types of Postgres which stores the time zone.
image

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
```diff
@@ -0,0 +1,58 @@
...
+import org.jooq.Table;
+import org.jooq.impl.DSL;
+import org.jooq.impl.SQLDataType;
+
+public interface PostgresSieveModule {
+    interface PostgresSieveScriptTable {
+        Table<Record> TABLE_NAME = DSL.table("sieve_scripts");
+
+        Field<String> USERNAME = DSL.field("username", SQLDataType.VARCHAR(255).notNull());
+        Field<String> SCRIPT_NAME = DSL.field("script_name", SQLDataType.VARCHAR.notNull());
```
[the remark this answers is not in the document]
quantranhong1999: There is no `TEXT` data type in jooq.

This declares the `VARCHAR` data type in Postgres, which basically likes `TEXT`.

`If you do not specify the n integer for the VARCHAR data type, it behaves like the TEXT datatype. The performance of the VARCHAR (without the size n) and TEXT are the same.`

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
chibenwa: An offset is not a timezone, Quan.

chibenwa: Example of timezone CEST is offset UTC+1 in winter and offset UTC+2 in summer

quantranhong1999: Ok, I got it. Thanks for sharing ^^

Arsnael:
Why so many commits?

Arsnael (changes requested):
Just blocking this the time to fix the branch :)

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
```diff
@@ -0,0 +1,58 @@
...
+        Field<Long> SCRIPT_SIZE = DSL.field("script_size", SQLDataType.BIGINT);
```
vttranlina, line 38: `.notNull()` ?

I see the method `recordToPostgresSieveScript`, it will throw NullPointerException if ScripSize is null

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveScriptDAO.java
```diff
@@ -0,0 +1,136 @@
...
+        return postgresExecutor.dslContext()
+            .flatMap(dslContext -> Mono.from(dslContext.update(TABLE_NAME)
+                .set(IS_ACTIVE, true)
+                .set(ACTIVATION_DATE_TIME, OffsetDateTime.now())
+                .where(USERNAME.eq(username.asString()),
+                    SCRIPT_NAME.eq(scriptName.getValue()))))
+            .cast(Long.class);
+    }
+
+    public Mono<Void> deactivateScript(Username username, ScriptName scriptName) {
```
vttranlina, line 100: `Method 'deactivateScript(org.apache.james.core.Username, org.apache.james.sieverepository.api.ScriptName)' is never used `

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveRepository.java
quantranhong1999: On that, I will tend to rely on the number of inserted/updated rows which is always returned by Postgres to know if the upsert succeeded. RETURNING requires extra data transfer while we do not really need them in this case.

Arsnael:
@quantranhong1999 can you rebase on latest postgresql branch?

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
Arsnael: > For a given user not two scripts with the same name

Normally by declaring SCRIPT_NAME part of the primary key, it should have an implicit unique constraint on it as well

chibenwa:
I did propose a little refactoring making the not so reactive code easier to read (minus 20 lines of code - not bad)

quantranhong1999:
> I did propose a little refactoring making the not so reactive code easier to read (minus 20 lines of code - not bad)

Well, I have done some refactoring too. Let me see how to rebase on your work...

chibenwa (approved):

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveRepository.java
```diff
@@ -58,42 +58,24 @@ public PostgresSieveRepository(PostgresSieveQuotaDAO postgresSieveQuotaDAO,
...
-                            .username(username.asString())
-                            .scriptName(name.getValue())
-                            .scriptContent(content.getValue())
-                            .scriptSize(content.length())
-                            .isActive(false)
-                            .build()))
-                    .then()))
-            .block());
+        long sizeDifference = spaceThatWillBeUsedByNewScript(username, name, content.length()).block();
+        throwOnOverQuota(username, sizeDifference);
```
Arsnael, line 68: why not calling haveSpace that has the exact same two lines then? And make the function return the space of the script?

chibenwa: Because we would still need to check if a script exist and what its size was

quantranhong1999: `haveSpace` is an Override method. We should not change the method signature.

quantranhong1999:
Two latest commits provides fixup for comments: https://github.com/apache/james-project/pull/1849/commits/4e9a9d7155ed92b2776bb0ad53b0be2aa067caf6 and guice binding: https://github.com/apache/james-project/pull/1849/commits/4e3f41b3985dfb6ceac36f9e0c41931e294f6325

Please review them also.

Arsnael (approved):

chibenwa:
(I really hate those "rethrow" methods...)

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveScriptDAO.java
```diff
@@ -39,20 +39,43 @@
...
+            })
+            .block();
+    }
+
     private final PostgresExecutor postgresExecutor;
 
     @Inject
     public PostgresSieveScriptDAO(@Named(DEFAULT_INJECT) PostgresExecutor postgresExecutor) {
         this.postgresExecutor = postgresExecutor;
+        enforceMaximumOneActiveScriptPerUserConstraint(postgresExecutor);
```
chibenwa, line 74: Please find a way to do this at table creation

quantranhong1999: The rationale behind that is jooq does not have DSL support for EXCLUDE constraint (only for primary key, foreign key and unique constraints), and no raw SQL to create constraint at the same time table creation.

quantranhong1999: Maybe I would think about adding raw sql query in the `PostgresModule`:
```
    List<String> additionalConstraints();
```

and upon `PostgresTableManager` initiation, create the constraints using the raw query too.

chibenwa: I mean: alter statement is OK but find a way to attach it to the table declaration and execute it just after table creation.

chibenwa: > and upon PostgresTableManager initiation, create the constraints using the raw query too.

Yes that's what I want!

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
```diff
@@ -0,0 +1,62 @@
...
+            .createTableStep(((dsl, tableName) -> dsl.createTableIfNotExists(tableName)
+                .column(USERNAME)
+                .column(SCRIPT_NAME)
+                .column(SCRIPT_SIZE)
+                .column(SCRIPT_CONTENT)
+                .column(IS_ACTIVE)
+                .column(ACTIVATION_DATE_TIME)
+                .primaryKey(USERNAME, SCRIPT_NAME)))
+            .disableRowLevelSecurity()
+            .addAdditionalConstraints(MAXIMUM_ONE_ACTIVE_SCRIPT_PER_USER_CONSTRAINT)
```
vttranlina, line 55: How about unique index?

```
        PostgresIndex INDEX = PostgresIndex.name("sieve_scripts_username_active_index")
            .createIndexStep((dsl, indexName) -> dsl.createIndex(indexName)
                .on(TABLE_NAME, USERNAME, IS_ACTIVE)
                .where(IS_ACTIVE.isTrue()));
```

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/PostgresTable.java
```diff
@@ -39,31 +41,60 @@ public interface CreateTableFunction {
...
+        default FinalStage supportsRowLevelSecurity() {
             return supportsRowLevelSecurity(true);
         }
     }
 
+    public static class FinalStage {
+        private final String tableName;
+        private final boolean supportsRowLevelSecurity;
+        private final Function<DSLContext, DDLQuery> createTableStepFunction;
+        private final ImmutableList.Builder<String> additionalConstraints;
```
vttranlina, line 59: https://github.com/apache/james-project/pull/1849#discussion_r1424833536 If the PostgresIndex is acceptable, we don't need `additionalConstraints` here. 
If it is still necessary, We should rename it to -> `additionalAlterQueries`, it will be used in same cases that we want to use native sql

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveModule.java
quantranhong1999: I did not know that the unique index can enforce constraint.

```
        PostgresIndex MAXIMUM_ONE_ACTIVE_SCRIPT_PER_USER_INDEX = PostgresIndex.name("maximum_one_active_script_per_user")
            .createIndexStep(((dsl, indexName) -> dsl.createUniqueIndexIfNotExists(indexName)
                .on(TABLE_NAME, USERNAME)
                .where(IS_ACTIVE)));
```

It worked. Thanks!

Bear in mind that the unique index only makes constraint on the index itself and not the table. So it is not a really strong constraint as the tradditional constraint. e.g. if one day we want the indexes to be updated in asynchronous mode, then the unique index can not enforce constraint correctly.

But I guess that won't be the case soon...

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/PostgresTable.java
quantranhong1999: Following the unique index adoption, I would split the `additionalAlterQueries` support into another PR. I think it still brings value as jOOQ DSL does not support all the cases.

chibenwa: As long as we keep a fluent API for index creation I am fine

server/data/data-postgres/src/main/java/org/apache/james/sieve/postgres/PostgresSieveRepository.java
```diff
@@ -49,217 +43,167 @@
...
 
-    private QuotaSizeLimit limitToUser(Username username) {
-        return postgresSieveQuotaDAO.getQuota(username)
-            .filter(Optional::isPresent)
-            .switchIfEmpty(postgresSieveQuotaDAO.getGlobalQuota())
-            .block()
-            .orElse(QuotaSizeLimit.unlimited());
+    private Mono<Long> spaceThatWillBeUsedByNewScript(Username username, ScriptName name, long scriptSize) {
+        return postgresSieveScriptDAO.getScript(username, name)
+            .map(PostgresSieveScript::getScriptSize)
```
hungphan227, line 96: should we just get script size from db? the whole record is unnecessary, especially script content could be big

quantranhong1999: Yes. We could optimize that one indeed.

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/PostgresTable.java
quantranhong1999: > I would split the additionalAlterQueries support into another PR

https://github.com/apache/james-project/pull/1852

hungphan227 (approved):

Arsnael (approved):

vttranlina (approved):

quantranhong1999:
Squashed + rebased.

PR #1851, opened 2023-12-13
JAMES-2586 PostgresDelegationStore

hungphan227:


server/container/guice/postgres-common/src/main/java/org/apache/james/modules/data/PostgresDelegationStoreModule.java
```diff
@@ -0,0 +1,62 @@
...
+import com.google.inject.Provides;
+import com.google.inject.Scopes;
+import com.google.inject.Singleton;
+import com.google.inject.multibindings.Multibinder;
+
+public class PostgresDelegationStoreModule extends AbstractModule {
+    @Override
+    public void configure() {
+        bind(DelegationStore.class).to(PostgresDelegationStore.class);
+        bind(PostgresDelegationStore.UserExistencePredicate.class).to(PostgresDelegationStore.UserExistencePredicateImplementation.class);
```
chibenwa, line 44: The user existance predicate depends on the user repository implementation, thus should be in `PostgresUsersRepositoryModule`. (We might need the PGDelegationStore with LDAP...)

server/data/data-postgres/src/main/java/org/apache/james/user/postgres/PostgresUsersDAO.java
```diff
@@ -141,4 +150,93 @@ public void addUser(Username username, String password) {
...
+    private Mono<Void> addUserToList(Field<String[]> field, Username baseUser, Username targetUser) {
+        String fullAuthorizedUsersColumnName = TABLE.getName() + "." + field.getName();
+        return postgresExecutor.executeVoid(dslContext ->
+            Mono.from(dslContext.insertInto(TABLE_NAME)
+                .set(USERNAME, baseUser.asString())
+                .set(field, DSL.array(targetUser.asString()))
+                .onConflict(USERNAME)
+                .doUpdate()
+                .set(DSL.field(field.getName()),
+                    (Object) DSL.field("array_append(coalesce(" + fullAuthorizedUsersColumnName + ", array[]::varchar[]), ?)",
```
chibenwa, line 176: string concatenation in a SQL query on a field potentially controlled by a user spotted.

Extra care handling this is definitively needed.

server/data/data-postgres/src/main/java/org/apache/james/user/postgres/PostgresUsersDAO.java
```diff
@@ -141,4 +150,93 @@ public void addUser(Username username, String password) {
...
+
+    private Mono<Void> removeUserFromList(Field<String[]> field, Username baseUser, Username targetUser) {
+        return postgresExecutor.executeVoid(dslContext ->
+            Mono.from(createQueryRemoveUserFromList(dslContext, field, baseUser, targetUser)));
+    }
+
+    private UpdateConditionStep<Record> createQueryRemoveUserFromList(DSLContext dslContext, Field<String[]> field, Username baseUser, Username targetUser) {
+        return dslContext.update(TABLE_NAME)
+            .set(DSL.field(field.getName()),
+                (Object) DSL.field("array_remove(" + field.getName() + ", ?)",
```
chibenwa, line 208: Idem.

server/data/data-postgres/src/test/java/org/apache/james/user/postgres/PostgresDelegationStoreTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.james.backends.postgres.PostgresExtension;
+import org.apache.james.core.Username;
+import org.apache.james.user.api.DelegationStore;
+import org.apache.james.user.api.DelegationStoreContract;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+import reactor.core.publisher.Mono;
+
+public class PostgresDelegationStoreTest implements DelegationStoreContract {
```
chibenwa, line 31: Please add

```
    @Test
    void virtualUsersShouldNotBeListed() {
        testee = new CassandraDelegationStore(cassandraUsersDAO, any -> Mono.just(false));
        addUser(BOB);

        Mono.from(testee().addAuthorizedUser(ALICE).forUser(BOB)).block();

        assertThat(cassandraUsersDAO.listReactive().collectList().block())
            .containsOnly(BOB);
    }
```

server/data/data-postgres/src/main/java/org/apache/james/user/postgres/PostgresUsersDAO.java
hungphan227: String fullAuthorizedUsersColumnName = TABLE.getName() + "." + field.getName();

TABLE is constant. field could be AUTHORIZED_USERS or DELEGATED_USERS which are all constant. Therefore, it is not vulnerable.

server/data/data-postgres/src/main/java/org/apache/james/user/postgres/PostgresUsersDAO.java
hungphan227: Similar to above

server/data/data-postgres/src/test/java/org/apache/james/user/postgres/PostgresDelegationStoreTest.java
hungphan227: added

server/data/data-postgres/src/main/java/org/apache/james/user/postgres/PostgresUsersDAO.java
chibenwa: I did miss read.

Yes of course that's ok.

chibenwa (approved):

Arsnael (approved):

PR #1852, opened 2023-12-14
JAMES-2586 Add an `addAdditionalAlterQueries` option when declaring Postgres table

quantranhong1999:
Would be useful in case jOOQ DSL does not provide support for some queries e.g. create an EXCLUDE constraint.


chibenwa:
Is this feature used?

If yes, let's merge it.

If no, let's wait...

chibenwa:
(so that we do not merge by mistake :-p

chibenwa:
Needed.

vttranlina (approved):

Arsnael (approved):

PR #1856, opened 2023-12-15
JAMES-3944 Add more tests with filter forwards with self forward use …

Arsnael:
…cases

### DIFF
--- a/server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardIntegrationTest.java
+++ b/server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardIntegrationTest.java
@@ -27,6 +27,7 @@ import static org.apache.james.mailets.configuration.Constants.PASSWORD;
 import java.io.File;
 
 import org.apache.james.core.MailAddress;
+import org.apache.james.core.MaybeSender;
 import org.apache.james.core.Username;
 import org.apache.james.mailets.configuration.CommonProcessors;
 import org.apache.james.mailets.configuration.MailetConfiguration;
@@ -323,13 +324,19 @@ public class ForwardIntegrationTest {
             .authenticate(FROM, PASSWORD)
             .sendMessage(FROM, BOB.asString());
 
-        Awaitility.await().until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 1L);
+        Awaitility.await().until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 2L);
 
         SoftAssertions.assertSoftly(Throwing.consumer(softly -> {
             Mail forwardedMail = mailRepositoryProbe.listMails(CUSTOM_REPOSITORY)
+                .filter(mail -> mail.getMaybeSender().asString().equals(BOB.asString()))
                 .findAny().get();
-            softly.assertThat(forwardedMail.getRecipients()).containsOnly(DELPHINE.asMailAddress(), BOB.asMailAddress());
+            Mail originalMail = mailRepositoryProbe.listMails(CUSTOM_REPOSITORY)
+                .filter(mail -> mail.getMaybeSender().asString().equals(FROM))
+                .findAny().get();
+            softly.assertThat(forwardedMail.getRecipients()).containsOnly(DELPHINE.asMailAddress());
             softly.assertThat(forwardedMail.getMaybeSender().asOptional()).contains(BOB.asMailAddress());
+            softly.assertThat(originalMail.getRecipients()).containsOnly(BOB.asMailAddress());
+            softly.assertThat(originalMail.getMaybeSender().asOptional().map(MailAddress::asString)).contains(FROM);
         }));
     }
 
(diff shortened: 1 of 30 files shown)


quantranhong1999:
I think we should try to add `.mailet(LocalDelivery.class)` into the `setUp` and see what happens.

Currently, the setup seems generally only to make sure there is no loop within a single mailet pipeline (just end at move mail to the custom repository).

Maybe the forward loop could happen through an infinite mailet loop? For example:
- mailet pipeline 1st: get the mail from the mail queue, RRT to myself, and `LocalDelivery` to me (enqueue the mail)
- mailet pipeline 2nd: get the mail from the mail queue, RRT to myself, and `LocalDelivery` to me (enqueue the mail)
... keeping the loop

Arsnael:
@quantranhong1999 could be interesting to check too, thanks for the suggestion.

I just made some breaks though when mixing filterung rule and mapping for forwards without a loop and local copy... does not work as expected.

I think also the test `regularForwardShouldNotCreateLoopErrorAndSendEmailToAppropriateReceivers` is wrong.

It's only forward mappings but we have:
```
ALICE forwards to BOB
BOB forwards to CEDRIC
CEDRIC forwards to ALICE
CEDRIC forwards to SENDER
```
We just check that mail arrives to sender... I would expect I think as well cedric gets a mail as one should stop there with the loop detection? But it's not the case

quantranhong1999:
Sorry please forget what I proposed. `RecipientRewriteTable` is the one sending the forward mail already, not `LocalDelivery`.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardIntegrationTest.java
```diff
@@ -333,6 +333,26 @@ void forwardShouldHandleLocalCopy() throws Exception {
...
+            .authenticate(FROM, PASSWORD)
+            .sendMessage(FROM, BOB.asString());
+
+        Awaitility.await().until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 1L);
+
+        SoftAssertions.assertSoftly(Throwing.consumer(softly -> {
+            Mail forwardedMail = mailRepositoryProbe.listMails(CUSTOM_REPOSITORY)
+                .findAny().get();
+            softly.assertThat(forwardedMail.getRecipients()).containsOnly(BOB.asMailAddress());
+            softly.assertThat(forwardedMail.getMaybeSender().asOptional()).contains(BOB.asMailAddress());
```
chibenwa, line 352: If we only have the local copy maybe we should not rewrite the sender?

Arsnael: I agree, it should be fixed.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
```diff
@@ -186,19 +200,89 @@ void regularForwardShouldNotCreateLoopError() throws Exception {
...
+        filteringManagementProbe.defineRulesForUser(ALICE, asRule(Forward.to(BOB.asMailAddress()).withoutACopy()));
+        dataProbe.addMapping(MappingSource.fromUser(BOB), Mapping.forward(CEDRIC.asString()));
+        filteringManagementProbe.defineRulesForUser(CEDRIC, asRule(Forward.to(ALICE.asMailAddress()).withoutACopy()));
+
+        messageSender.connect(LOCALHOST_IP, jamesServer.getProbe(SmtpGuiceProbe.class).getSmtpPort())
+            .authenticate(SENDER.asString(), PASSWORD)
+            .sendMessage(SENDER.asString(), ALICE.asString());
+
+        Awaitility.await().until(() -> jamesServer.getProbe(SpoolerProbe.class).processingFinished());
+        Awaitility.await().until(() -> mailRepositoryProbe.getRepositoryMailCount(CUSTOM_REPOSITORY) == 0L);
```
Arsnael, line 280: I think this test leads to rrt error?

Arsnael (commented):
If I understand the new approach is: 

If a loop is detected => we store the mail in rrt-error, we don't try to store it in the last recipient of the loop anymore, correct? Letting then the manual intervention of an admin if that happens to a user.

I'm ok with it. Just a small remark but overall ok

Arsnael:
```
Your configurations specifies the following undefined processors: [rrt-error]
```

https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1856/6/ => Cassandra and memory james test server

chibenwa:
> If a loop is detected => we store the mail in rrt-error, we don't try to store it in the last recipient of the loop anymore, correct? Letting then the manual intervention of an admin if that happens to a user.

Correct

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/ForwardLoopIntegrationTest.java
chibenwa: Yes. That's tested in another test. Here we ensure that nothing gets delivered.

Arsnael (commented):
I can't approve my own PR but... I do approve @chibenwa's hard work on it. Thank you :)

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1856/9/#showFailuresLink

webadmin memory tests are definitely related

Arsnael:
It's green!!! Let me squash the fixups...

chibenwa (approved):

Arsnael:
Rebased to fix conflicts

PR #1870, opened 2023-12-20
JAMES-2586 Module chooser: S3, file blobStore

quantranhong1999:
resolve [[PGSQL] Module chooser: S3, file blobStore](https://github.com/linagora/james-project/issues/4984)

I tested the normal docker-compose and distributed one, they work normally too.

### DIFF
--- a/server/apps/postgres-app/pom.xml
+++ b/server/apps/postgres-app/pom.xml
@@ -80,6 +80,18 @@
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
+            <artifactId>blob-s3</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>blob-s3-guice</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
             <artifactId>james-server-cassandra-app</artifactId>
             <type>test-jar</type>
             <scope>test</scope>
@@ -98,6 +110,18 @@
         <dependency>
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-data-postgres</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-distributed-app</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>${james.groupId}</groupId>
+                    <artifactId>james-server-guice-distributed</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
@@ -142,6 +166,10 @@
         <dependency>
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-guice-managedsieve</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-guice-memory</artifactId>
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
--- a/server/apps/postgres-app/src/main/java/org/apache/james/PostgresJamesServerMain.java
+++ b/server/apps/postgres-app/src/main/java/org/apache/james/PostgresJamesServerMain.java
@@ -19,14 +19,20 @@
 
 package org.apache.james;
 
+import java.util.List;
+
+import org.apache.james.blob.api.BlobReferenceSource;
 import org.apache.james.data.UsersRepositoryModuleChooser;
 import org.apache.james.modules.MailboxModule;
 import org.apache.james.modules.MailetProcessingModule;
 import org.apache.james.modules.RunArgumentsModule;
+import org.apache.james.modules.blobstore.BlobStoreCacheModulesChooser;
+import org.apache.james.modules.blobstore.BlobStoreModulesChooser;
 import org.apache.james.modules.data.PostgresDataModule;
 import org.apache.james.modules.data.PostgresDelegationStoreModule;
 import org.apache.james.modules.data.PostgresUsersRepositoryModule;
 import org.apache.james.modules.data.SievePostgresRepositoryModules;
+import org.apache.james.modules.eventstore.MemoryEventStoreModule;
 import org.apache.james.modules.mailbox.DefaultEventModule;
 import org.apache.james.modules.mailbox.MemoryDeadLetterModule;
 import org.apache.james.modules.mailbox.PostgresMailboxModule;
@@ -52,8 +58,11 @@ import org.apache.james.modules.server.SieveRoutesModule;
 import org.apache.james.modules.server.TaskManagerModule;
 import org.apache.james.modules.server.WebAdminReIndexingTaskSerializationModule;
 import org.apache.james.modules.server.WebAdminServerModule;
+import org.apache.james.server.blob.deduplication.StorageStrategy;
 
+import com.google.common.collect.ImmutableList;
 import com.google.inject.Module;
+import com.google.inject.multibindings.Multibinder;
 import com.google.inject.util.Modules;
 
 public class PostgresJamesServerMain implements JamesServerMain {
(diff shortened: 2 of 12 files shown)


server/apps/postgres-app/pom.xml
```diff
@@ -60,6 +60,24 @@
...
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>blob-s3-guice</artifactId>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-cassandra-app</artifactId>
```
chibenwa, line 77: Suggestion: move S3extension into guice-s3 maven module? So that we do not need those testing dependencies onto `james-server-cassandra-app` and `james-server-distributed-app` ?

server/container/guice/postgres-common/src/main/java/org/apache/james/modules/blobstore/BlobStoreConfiguration.java
```diff
@@ -0,0 +1,229 @@
...
+import org.apache.james.modules.mailbox.ConfigurationComponent;
+import org.apache.james.server.blob.deduplication.StorageStrategy;
+import org.apache.james.server.core.filesystem.FileSystemImpl;
+import org.apache.james.utils.PropertiesProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.common.base.MoreObjects;
+
+public class BlobStoreConfiguration {
```
chibenwa, line 41: Please do NOT duplicate this file. 

IMO we must reuse the one defined in `server/container/guice/distributed/src/main/java/org/apache/james/modules/blobstore/BlobStoreConfiguration.java`.

We then can validate later, EG in a `StartUpCheck` that the blob store cache is not activated for the postgres application.

server/apps/postgres-app/pom.xml
Arsnael: I had same dependencies issues with some test classes depending on the cassandra app test scope

It might be a bit of a mess if we start relocating everything same time... A refactoring later maybe?

server/container/guice/postgres-common/src/main/java/org/apache/james/modules/blobstore/BlobStoreConfiguration.java
quantranhong1999: I found it hard to reuse. Because the configuration could be diverged. e.g in `server/container/guice/distributed/src/main/java/org/apache/james/modules/blobstore/BlobStoreConfiguration.java`.
```
       } catch (FileNotFoundException e) {
            LOGGER.warn("Could not find " + ConfigurationComponent.NAME + " configuration file, using cassandra blobstore as the default");
            return BlobStoreConfiguration.builder()
                .cassandra()
```

Put a configurable default blobStore choice (e.g. file/postgres) to the method?

The same thing for cache configuration. It is not the case for postgres-app now (unless we plan to implement blob cache with Postgres too?). Make the cache configuration optional? 

My feeling is that it is do-able to reuse, but it could be misleading too.

server/apps/postgres-app/pom.xml
chibenwa: Ok I buy the later refactoring as long as it is a ticket of this sprint ;-)

Arsnael (commented):
Read it, nothing to add

server/container/guice/postgres-common/src/main/java/org/apache/james/modules/blobstore/BlobStoreConfiguration.java
Arsnael: Maybe a bit tricky to refactor. I agree it would be cleaner to have something like the SearchModuleChooser that we can reuse on any implem. But it requires some refactoring here.

Ticket?

server/apps/postgres-app/pom.xml
Arsnael: https://github.com/linagora/james-project/issues/4999

server/container/guice/postgres-common/src/main/java/org/apache/james/modules/blobstore/BlobStoreConfiguration.java
quantranhong1999: >? Maybe a bit tricky to refactor. I agree it would be cleaner to have something like the SearchModuleChooser that we can reuse on any implem. But it requires some refactoring here.

Let me try to handle this in this PR.

quantranhong1999: fully reuse reuse `BlobStoreConfiguration` + `BlobStoreModulesChooser` now.

server/apps/postgres-app/src/main/java/org/apache/james/PostgresJamesServerMain.java
```diff
@@ -116,7 +125,20 @@ static GuiceJamesServer createServer(PostgresJamesConfiguration configuration) {
...
                 .chooseModules(configuration.getUsersRepositoryImplementation()))
-            .combineWith(BlobStoreModulesChooser.chooseModules(configuration.blobStoreConfiguration()))
+            .combineWith(chooseBlobStoreModules(configuration))
             .combineWith(POSTGRES_MODULE_AGGREGATE);
     }
+
+    private static List<Module> chooseBlobStoreModules(PostgresJamesConfiguration configuration) {
+        ImmutableList.Builder<Module> builder = ImmutableList.<Module>builder()
+            .addAll(BlobStoreModulesChooser.chooseModules(configuration.blobStoreConfiguration()))
+            .addAll(BlobStoreCacheModulesChooser.chooseModules(configuration.blobStoreConfiguration()));
```
Arsnael, line 135: We don't have the blob_cache table yet with postgresql... Is that pertinent here?
Should we add a task for the cache with Postgresql?

server/apps/postgres-app/src/main/java/org/apache/james/PostgresJamesServerMain.java
```diff
@@ -92,7 +100,8 @@ public class PostgresJamesServerMain implements JamesServerMain {
         new SievePostgresRepositoryModules(),
         new DefaultEventModule(),
         new TaskManagerModule(),
-        new MemoryDeadLetterModule());
+        new MemoryDeadLetterModule(),
+        new MemoryEventStoreModule());
```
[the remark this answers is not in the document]
quantranhong1999: Yes. Likely EventStore was needed for deduplication module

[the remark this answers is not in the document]
quantranhong1999: Yes. Likely EventStore was needed for deduplication module

server/apps/postgres-app/src/main/java/org/apache/james/PostgresJamesServerMain.java
quantranhong1999: > We don't have the blob_cache table yet with postgresql... Is that pertinent here?

On that, right now we need CacheDisabledModule which tags the `@Named(MetricableBlobStore.BLOB_STORE_IMPLEMENTATION)` so the blob store binding can work and IT pass. It is how `BlobStoreModulesChooser` and `BlobStoreCacheModulesChooser` designed, I can not against it.

But we enforce disable blob cache configuration for postgres-app for now, so the binding will be fine. Don't worry :-)

>Should we add a task for the cache with Postgresql?

cc @chibenwa do you want it or not?

Arsnael: I think it would make sense to have a task for the cache yeah, I think like with cassandra it could be beneficial?

Arsnael:
Ah and it needs a rebase :)

Arsnael (approved):

server/apps/postgres-app/src/main/java/org/apache/james/PostgresJamesServerMain.java
chibenwa: >> Should we add a task for the cache with Postgresql?

> cc @chibenwa do you want it or not?

No it do not make sense as headers are a bytea on the message table

Blob store cache thing is a clumsy design around the simple fact that we wanted metadata for distributed app to be stored in Cassandra.

Arsnael: Alright... Let's remove this then and just enforce directly the no blob store cache bindings :)

vttranlina:
conflict, neet to rebase

Arsnael (approved):

vttranlina (approved):

PR #1874, opened 2023-12-21
JAMES-2586 Implement Postgresql search overrides

Arsnael:
Todo:

- [x] AllSearchOverride
- [x] DeletedSearchOverride
- [x] DeletedWithRangeSearchOverride
- [x] NotDeletedWithRangeSearchOverride
- [x] UidSearchOverride
- [x] UnseenSearchOverride

Technically, if I'm not wrong, with the searchModuleChooser integrated, the conf is already there, and there is no need to care about bindings?

### DIFF
--- a/mailbox/postgres/src/test/java/org/apache/james/mailbox/postgres/search/SearchOverrideFixture.java
+++ b/mailbox/postgres/src/test/java/org/apache/james/mailbox/postgres/search/SearchOverrideFixture.java
@@ -1,0 +1,71 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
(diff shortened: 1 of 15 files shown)


chibenwa (approved):

mailbox/postgres/src/test/java/org/apache/james/mailbox/postgres/search/SearchOverrideFixture.java
```diff
@@ -0,0 +1,52 @@
+package org.apache.james.mailbox.postgres.search;
```
Arsnael, line 1: License missing :)

Arsnael:
Rebased, added a missing license

quantranhong1999 (commented):
we should use Postgres indexes where appropriate IMO.

Arsnael:
> we should use Postgres indexes where appropriate IMO.

For the SearchOverrides the indexes are already in place, I'm between branches but check the modules where the PostgresMailboxMessageDAO is declared, there is 5 indexes declared with it, covering the cases of the search overrides IMO

quantranhong1999 (approved):

PR #1891, opened 2024-01-02
JAMES-2586 Implement Deleted Message Vault backed by Postgres

quantranhong1999:
resolve https://github.com/linagora/james-project/issues/5000

TODO: plug `DeletedMessageVaultDeletionCallback`.


mailbox/plugin/deleted-messages-vault-postgres/src/main/java/org/apache/james/vault/metadata/PostgresDeletedMessageMetadataModule.java
```diff
@@ -0,0 +1,56 @@
...
+        Field<JSONB> METADATA = DSL.field("metadata", SQLDataType.JSONB.notNull());
+
+        PostgresTable TABLE = PostgresTable.name(TABLE_NAME.getName())
+            .createTableStep(((dsl, tableName) -> dsl.createTableIfNotExists(tableName)
+                .column(BUCKET_NAME)
+                .column(OWNER)
+                .column(MESSAGE_ID)
+                .column(BLOB_ID)
+                .column(METADATA)
+                .primaryKey(BUCKET_NAME, OWNER, MESSAGE_ID)))
```
chibenwa, line 48: Index on `BUCKET_NAME, OWNER` ?

We might need to easily iterate entries for a given user...

mailbox/plugin/deleted-messages-vault-postgres/src/main/java/org/apache/james/vault/metadata/PostgresDeletedMessageMetadataVault.java
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    @Override
+    public Publisher<Void> store(DeletedMessageWithStorageInformation deletedMessage) {
+        return postgresExecutor.executeVoid(context -> Mono.from(context.insertInto(TABLE_NAME)
+            .set(OWNER, deletedMessage.getDeletedMessage().getOwner().asString())
+            .set(MESSAGE_ID, deletedMessage.getDeletedMessage().getMessageId().serialize())
+            .set(BUCKET_NAME, deletedMessage.getStorageInformation().getBucketName().asString())
+            .set(BLOB_ID, deletedMessage.getStorageInformation().getBlobId().asString())
+            .set(METADATA, jsonb(metadataSerializer.serialize(deletedMessage)))));
```
chibenwa, line 67: We are having a POJO that we convert into a JSON tree that we convert into a String that we parse again (jsonb) as a JSON tree...

I wonder if it would be doable to bypass the intermediary string representation if for instance using things like `jooq-jackson-extensions`...

Maybe not that significant here but it is worth a look...

chibenwa (commented):
I am missing a test...

Like:

```
GIVEN bob deletes a message 
WHEN bob restores deleted messages using webadmin endpoints
THEN bob has one message into the restored mailbox
```

mailbox/plugin/deleted-messages-vault-postgres/src/main/java/org/apache/james/vault/metadata/PostgresDeletedMessageMetadataModule.java
quantranhong1999: > Index on BUCKET_NAME, OWNER ?

As I researched, given we indexed `BUCKET_NAME + OWNER + MESSAGE_ID` composite index, then the performance when query `BUCKET_NAME + OWNER` is slower than the exact trio, but still better than no index at all.

The question is Is it worth to explicitly index for every possible composite? If the insert time and storage size is not an issue, yeh why we do not index them?

quantranhong1999:
> I am missing a test...

I thought about it as integration test `DeletedMessageVaultIntegrationTest` as part of the https://github.com/linagora/james-project/issues/5005.

But I can do the the vault IT here if you want ^^

chibenwa:
> Maybe we can modify SinglePostgresConnectionFactory to be backed by the pool connection (and rename it)

That would be enough to me. No need of an extra integration test then.

Arsnael (commented):
Read it, nothing to add :)

vttranlina:
read it

mailbox/plugin/deleted-messages-vault-postgres/src/main/java/org/apache/james/vault/metadata/PostgresDeletedMessageMetadataModule.java
hungphan227: IMO it is ok for BUCKET_NAME + OWNER but OWNER + MESSAGE_ID should be another index

mailbox/plugin/deleted-messages-vault-postgres/src/main/java/org/apache/james/vault/metadata/PostgresDeletedMessageMetadataVault.java
quantranhong1999: I tried the `jooq-jackson-extensions` but it seems to convert Java object to String just like what we are doing, nothing more special.

image

> that we convert into a String that we parse again (jsonb) as a JSON tree...

BTW `JSONB` class seems to be just a string wrapper, and SQL query to insert jsonb just using a normal text value also. So likely no `parse again (jsonb) as a JSON tree` part.

chibenwa:
Rebase needed...

vttranlina:
```
The file /home/tungtv/workplace/JAMES/1james-project/server/protocols/webadmin-integration-test/postgres-webadmin-integration-test/pom.xml is not sorted -> [Help 1]
```

mailbox/plugin/deleted-messages-vault-postgres/src/main/java/org/apache/james/vault/metadata/PostgresDeletedMessageMetadataModule.java
```diff
@@ -0,0 +1,65 @@
...
+            .disableRowLevelSecurity()
+            .build();
+
+        PostgresIndex OWNER_MESSAGE_ID_INDEX = PostgresIndex.name("owner_messageId_index")
+            .createIndexStep((dsl, indexName) -> dsl.createIndexIfNotExists(indexName)
```
vttranlina, line 57: `createIndexIfNotExists` or `createUniqueIndexIfNotExists`  ?

server/apps/postgres-app/src/main/java/org/apache/james/PostgresJamesServerMain.java
```diff
@@ -158,4 +164,14 @@ public static List<Module> chooseEventBusModules(PostgresJamesConfiguration conf
                 throw new RuntimeException("Unsupported event-bus implementation " + configuration.eventBusImpl().name());
         }
     }
+
+    private static Module chooseDeletedMessageVaultModules(VaultConfiguration vaultConfiguration) {
+        if (vaultConfiguration.isEnabled()) {
+            return Modules.combine(new PostgresDeletedMessageVaultModule(), new DeletedMessageVaultRoutesModule());
+        }
+
+        return binder -> {
```
vttranlina, line 173: I like `Modules.EMPTY_MODULE;` than create  an anonymous lambda
But I'm ok with 2

chibenwa (approved):

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/MessageRepresentation.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    private final MessageId messageId;
+    private final Date internalDate;
+    private final Long size;
+    private final Content headerContent;
+    private final BlobId bodyBlobId;
+
+    private MessageRepresentation(MessageId messageId, Date internalDate, Long size,
+                                 Content headerContent, BlobId bodyBlobId) {
```
vttranlina, line 86: indent

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/MessageRepresentation.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        public MessageRepresentation.Builder bodyBlobId(BlobId bodyBlobId) {
+            this.bodyBlobId = bodyBlobId;
+            return this;
+        }
+
+        public MessageRepresentation build() {
+            Preconditions.checkNotNull(messageId, "messageId is required");
+            Preconditions.checkNotNull(internalDate, "internalDate is required");
```
vttranlina, line 70: Q:
I see the `internal_date` in the database accepts NULL value
is it ok when checkNotNull here?

quantranhong1999: I did not know that. Q: In which chance `internal_date` could be null?

vttranlina: > I did not know that. Q: In which chance `internal_date` could be null?

yes, It can be null

Arsnael (approved):

mailbox/plugin/deleted-messages-vault-postgres/src/main/java/org/apache/james/vault/metadata/PostgresDeletedMessageMetadataModule.java
quantranhong1999: `createIndexIfNotExists` IMO. `OWNER, MESSAGE_ID` is not unique but `BUCKET_NAME, OWNER, MESSAGE_ID`.

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/MessageRepresentation.java
quantranhong1999: FYI, `SimpleMailboxMessage:build()` (which Postgres is using to build `SimpleMailboxMessage`) requires `internalDate`.

IMO `internalDate` or even `saveDate` should not be null in the `message` table.

For `saveDate` in Cassandra, it was possiblly null because of retro-compatibility (no saveDate column before), while it is not the case for Postgres.

mailbox/plugin/deleted-messages-vault-postgres/src/main/java/org/apache/james/vault/metadata/PostgresDeletedMessageMetadataModule.java
vttranlina: But how about the method `retrieveStorageInformation`
I see you use Mono, not FLux

quantranhong1999: You are right. I just recheck Cassandra and `OWNER, MESSAGE_ID` should be enough for unique.

vttranlina (approved):

Arsnael (approved):

quantranhong1999:
@Arsnael Can we merge this?

PR #1906, opened 2024-01-09
JAMES-3953 Performance enhancements

chibenwa:
Avoid partial reads and array concatenations.

CF: https://github.com/linagora/james-project/pull/5032

### DIFF
--- a/server/blob/blob-file/src/main/java/org/apache/james/blob/file/FileBlobStoreDAO.java
+++ b/server/blob/blob-file/src/main/java/org/apache/james/blob/file/FileBlobStoreDAO.java
@@ -106,10 +106,12 @@ public class FileBlobStoreDAO implements BlobStoreDAO {
     @Override
     public Mono<Void> save(BucketName bucketName, BlobId blobId, byte[] data) {
         Preconditions.checkNotNull(data);
-        File bucketRoot = getBucketRoot(bucketName);
-        File blob = new File(bucketRoot, blobId.asString());
 
-        return Mono.fromRunnable(() -> save(data, blob))
+        return Mono.fromRunnable(() -> {
+                File bucketRoot = getBucketRoot(bucketName);
+                File blob = new File(bucketRoot, blobId.asString());
+                save(data, blob);
+            })
             .subscribeOn(Schedulers.boundedElastic())
             .then();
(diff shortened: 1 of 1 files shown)


Arsnael (approved):

server/blob/blob-file/src/main/java/org/apache/james/blob/file/FileBlobStoreDAO.java
```diff
@@ -97,13 +95,23 @@ public Mono<InputStream> readReactive(BucketName bucketName, BlobId blobId) {
...
+        }).onErrorResume(NoSuchFileException.class, e -> Mono.error(new ObjectNotFoundException(String.format("Cannot locate %s within %s", blobId.asString(), bucketName.asString()), e)))
+            .subscribeOn(Schedulers.boundedElastic());
     }
 
     @Override
     public Mono<Void> save(BucketName bucketName, BlobId blobId, byte[] data) {
-        return save(bucketName, blobId, new ByteArrayInputStream(data));
+        Preconditions.checkNotNull(data);
+        File bucketRoot = getBucketRoot(bucketName);
+        File blob = new File(bucketRoot, blobId.asString());
```
chibenwa, line 110: Needs to be wrapped in a mono....

-> Beginner mistake!

chibenwa:
@vttranlina I think the perf tests you did for choosing the default BlobStoreDAO are invalidated by this change.

Don't you mind bench again this VS PGBlobStore?

vttranlina:
> @vttranlina I think the perf tests you did for choosing the default BlobStoreDAO are invalidated by this change.
> 
> Don't you mind bench again this VS PGBlobStore?

postgresVsFile
The gap between min and max of Postgres is very high  
postgres: https://vttranlina.github.io/gatling_20240117/postgres/index.html
file: https://vttranlina.github.io/gatling_20240117/file/index.html

My screnario
```
export DURATION=7
export USER_COUNT=3000
export MAX_DURATION=16
```

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1906/8/

Related or not? I've seen lots of instability on those tests around those changes for the past few days

PR #1908, opened 2024-01-09
JAMES-3897 CrowdSec integration test for SMTP

quantranhong1999:
resolve https://github.com/linagora/james-project/issues/5003

SMTP proxy support with CrowdSec should already be the case.

### DIFF
--- a/third-party/crowdsec/pom.xml
+++ b/third-party/crowdsec/pom.xml
@@ -44,6 +44,11 @@
         </dependency>
         <dependency>
             <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-memory-app</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
             <artifactId>testing-base</artifactId>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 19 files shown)


chibenwa (approved):

Arsnael (approved):
Looks g ood to me.

Just remove the `third-party/crowdsec/src/test/resources/log/james.log` empty file and we are good :)

thanhbv200585:
There is also integration test for IMAP as well, should we put it CrowdsecIntegrationTest with SMTP or separate them?

quantranhong1999:
> There is also integration test for IMAP as well, should we put it CrowdsecIntegrationTest with SMTP or separate them?

The ticket name says it all.

thanhbv200585:
Read it. Look good to me

thanhbv200585 (approved):

quantranhong1999:
> Just remove the third-party/crowdsec/src/test/resources/log/james.log empty file and we are good :)

That empty file was intended. During tests, it is used to store James log and share with CrowdSec. After tests, it would be reset to empty again.

Sry I did not find a solution without an empty file yet. Maybe one day I feel bored :D. Or if you have any idea for me ^^.

PR #1913, opened 2024-01-10
JAMES-3968 Fix mail loss due to RabbitMQ ack failure (3.7.x)

ottoka:
One-liner to correctly propagate the exception up the chain. 3.7.x backport.


chibenwa (approved):

server/queue/queue-rabbitmq/src/main/java/org/apache/james/queue/rabbitmq/Enqueuer.java
```diff
@@ -119,7 +119,7 @@ private Mono<Void> publishReferenceToRabbit(MailReference mailReference) throws
             .build();
 
         OutboundMessage data = new OutboundMessage(
-            name.toRabbitExchangeName().asString(),
+            name.   toRabbitExchangeName().asString(),
```
Arsnael, line 122: Mistake?

ottoka: Oops. Fixed in push.

vttranlina (approved):

Arsnael (approved):

PR #1915, opened 2024-01-10
JAMES-3968 Fix mail loss due to RabbitMQ ack failure

chibenwa:


server/queue/queue-rabbitmq/src/main/java/org/apache/james/queue/rabbitmq/Enqueuer.java
```diff
@@ -122,7 +122,7 @@ private Mono<Void> publishReferenceToRabbit(MailReference mailReference) throws
             .build();
 
         OutboundMessage data = new OutboundMessage(
-            name.toRabbitExchangeName().asString(),
+            name.   toRabbitExchangeName().asString(),
```
Arsnael, line 125: Mistake?

ottoka (approved):
Not sure why I forgot this branch. LGTM.

PR #1928, opened 2024-01-17
JAMES-2586 Wire StoreMessageIdManager on top of the PostgresMessageIdMapper

Arsnael:
Reasons of the first commit (before I get asked^^):

- PostgresCombinationManagerTestSystem returns a createMessageManager, that has a protected access in PostgresmessageManager. I could have either move the test in mail subpackage to match it, or move the Managers out of the mail package.

In the end, I moved them out. Two reasons:

- tests related to managers are out of mail subpackage, and the tests should be on same package as the code implem
- took the decision by matching with Cassandra, where managers are out of mail subpackage.


mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresMailboxSessionMapperFactory.java
```diff
@@ -122,4 +124,11 @@ public AttachmentMapper createAttachmentMapper(MailboxSession session) {
     public AttachmentMapper getAttachmentMapper(MailboxSession session) {
         throw new NotImplementedException("not implemented");
     }
+
+    public DeleteMessageListener deleteMessageListener() {
```
hungphan227, line 128: is this method just for supporting testing?

chibenwa (approved):

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresMailboxSessionMapperFactory.java
quantranhong1999: Mark `@VisibleForTesting` then?

quantranhong1999 (approved):

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresMailboxSessionMapperFactory.java
Arsnael: Little refactoring to have this protected (solves the issue then and no need of the tag right?) and using this method as well in the PostgresMailboxManagerProvider

Good now? :)

quantranhong1999 (approved):

PR #1931, opened 2024-01-17
Fix for NoSuchElement in SPFHandler

lwolters:
The original code assumes that there's always a value present, which in my case, it isn't. As a rule of thumb you should always check an Optional<> for presence of the value, instead of directing calling .get() since you'll never know if the value is truly present.

That said, I don't know the implications of not having a SPF Attribute Value present later up in the chain. Maybe some other logic should be implemented when no SPF Attribute Value is present.


Arsnael (approved):
Good catch! Thank you for the fix @lwolters :)

quantranhong1999 (approved):

chibenwa (approved):

PR #1932, opened 2024-01-17
[FIX] Enable dynamic translation of LDAP entries

chibenwa:
# Why

I want a user to authenticate with a login old by one attribute eg `uid` and then use another attribute as a end username eg `mail`.

That way we can set up complex login architecture, eg:

```
uid: device1_bob
mail: bob@linagora.com
password: abc

uid: device2_bob
mail: bob@linagora.com
password: def
```


server/data/data-ldap/src/main/java/org/apache/james/user/ldap/LdapRepositoryConfiguration.java
```diff
@@ -282,10 +287,16 @@ public static LdapRepositoryConfiguration from(HierarchicalConfiguration<Immutab
     /**
      * The value of this field is taken from the configuration attribute
      * &quot;userIdAttribute&quot;. This is the LDAP attribute type which holds
-     * the userId value. Note that this is not the same as the email address
-     * attribute.
+     * the userId value. This value is used to identify the LDAP entry used for authentication,
+     * and by default is used to retrie the associated username.
```
Arsnael, line 291:
```suggestion
     * and by default is used to retrieve the associated username.
```

Arsnael (approved):

PR #1933, opened 2024-01-18
[FIX] SMTP stack should recompute relaying rights upon PROXY message

chibenwa:
Otherwise, the relaying information is taken from the proxy and not from the source.

### DIFF
--- a/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/SMTPSessionImpl.java
+++ b/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/SMTPSessionImpl.java
@@ -23,6 +23,7 @@ import java.util.Optional;
 
 import org.apache.james.protocols.api.ProtocolSessionImpl;
 import org.apache.james.protocols.api.ProtocolTransport;
+import org.apache.james.protocols.api.ProxyInformation;
 import org.apache.james.protocols.api.Response;
 
 /**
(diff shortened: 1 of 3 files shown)


Arsnael (approved):

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/SMTPSessionImpl.java
```diff
@@ -41,8 +42,12 @@ public class SMTPSessionImpl extends ProtocolSessionImpl implements SMTPSession
...
     public SMTPSessionImpl(ProtocolTransport transport, SMTPConfiguration config) {
         super(transport, config);
-        relayingAllowed = Optional.ofNullable(getRemoteAddress())
-            .flatMap(addres -> Optional.ofNullable(addres.getAddress()))
+        relayingAllowed = computeRelayingAllowed(config);
+    }
+
+    private Boolean computeRelayingAllowed(SMTPConfiguration config) {
+        return Optional.ofNullable(getRemoteAddress())
+            .flatMap(address -> Optional.ofNullable(address.getAddress()))
```
quantranhong1999, line 50:
```suggestion
            .map(address -> address.getAddress())
```

as `Optional.map` already handles null case:

image

(I did not know that before)

quantranhong1999 (approved):

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/SMTPSessionImpl.java
chibenwa: I prefer being explicit to be honnest

chibenwa: But thanks for sharing.

PR #1941, opened 2024-01-24
JAMES-2586 Implement PostgresThreadIdGuessingAlgorithm

hungphan227:


mailbox/cassandra/src/test/java/org/apache/james/mailbox/cassandra/CassandraThreadIdGuessingAlgorithmTest.java
```diff
@@ -94,8 +94,10 @@ protected MessageId initOtherBasedMessageId() {
     }
 
     @Override
-    protected Flux<Void> saveThreadData(Username username, Set<MimeMessageId> mimeMessageIds, MessageId messageId, ThreadId threadId, Optional<Subject> baseSubject) {
-        return threadDAO.insertSome(username, hashMimeMessagesIds(mimeMessageIds), messageId, threadId, hashSubject(baseSubject));
+    protected void saveThreadData(Username username, Set<MimeMessageId> mimeMessageIds, MessageId messageId, ThreadId threadId, Optional<Subject> baseSubject) {
+        threadDAO.insertSome(username, hashMimeMessagesIds(mimeMessageIds), messageId, threadId, hashSubject(baseSubject))
+            .then()
+            .block();
     }
```
chibenwa, line 101: Why?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresMailboxFactory.java
```diff
@@ -0,0 +1,43 @@
...
+    @Inject
+    @Singleton
+    public PostgresMailboxFactory(PostgresExecutor.Factory executorFactory) {
+        this.executorFactory = executorFactory;
+    }
+
+    public PostgresThreadDAO createThreadDAO(Optional<Domain> domain) {
+        return new PostgresThreadDAO(executorFactory.create(domain));
+    }
+}
```
chibenwa, line 43: It is not a `MailboxFactory` if it builds ThreadDAO...

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresThreadIdGuessingAlgorithm.java
```diff
@@ -0,0 +1,103 @@
...
+            .andCriteria(SearchQuery.threadId(threadId))
+            .sorts(new SearchQuery.Sort(SearchQuery.Sort.SortClause.Arrival, SearchQuery.Sort.Order.NATURAL))
+            .build();
+
+        MultimailboxesSearchQuery expression = MultimailboxesSearchQuery
+            .from(searchQuery)
+            .build();
+
+        return Flux.from(mailboxManager.search(expression, session, Integer.MAX_VALUE))
+            .switchIfEmpty(Mono.error(() -> new ThreadNotFoundException(threadId)));
```
chibenwa, line 93: I would love to be able to do this against PG only.
Here I understand this would be done against OpenSearch.

I bet this would be as easy as reading the `thread` table by `username` and `threadId` ?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
```diff
@@ -0,0 +1,80 @@
...
+public class PostgresThreadDAO {
+    private final PostgresExecutor postgresExecutor;
+
+    public PostgresThreadDAO(PostgresExecutor postgresExecutor) {
+        this.postgresExecutor = postgresExecutor;
+    }
+
+    public Mono<Void> insertSome(Username username, Set<Integer> hashMimeMessageIds, MessageId messageId, ThreadId threadId, Optional<Integer> hashBaseSubject) {
+        return postgresExecutor.executeVoid(dslContext -> Mono.from(dslContext.batch(
+            hashMimeMessageIds.stream().map(hashMimeMessageId -> dslContext.insertInto(TABLE_NAME)
```
chibenwa, line 54: Use multi insert instead: https://stackoverflow.com/questions/20815028/how-do-i-insert-multiple-values-into-a-postgres-table-at-once

CF https://www.jooq.org/doc/latest/manual/sql-building/sql-statements/insert-statement/insert-values/

```
create.insertInto(AUTHOR,
        AUTHOR.ID, AUTHOR.FIRST_NAME, AUTHOR.LAST_NAME)
      .values(100, "Hermann", "Hesse")
      .values(101, "Alfred", "Döblin")
      .execute();
```

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
```diff
@@ -0,0 +1,80 @@
...
+                .set(THREAD_ID, ((PostgresMessageId) threadId.getBaseMessageId()).asUuid())
+                .set(HASH_BASE_SUBJECT, hashBaseSubject.orElse(null)))
+                .collect(ImmutableList.toImmutableList()))));
+    }
+
+    public Flux<Pair<Optional<Integer>, ThreadId>> findThreads(Username username, Set<Integer> hashMimeMessageIds) {
+        return postgresExecutor.executeRows(dslContext -> Flux.from(dslContext.select(THREAD_ID, HASH_BASE_SUBJECT)
+            .from(TABLE_NAME)
+            .where(USERNAME.eq(username.asString()))
+            .and(HASH_MIME_MESSAGE_ID.in(hashMimeMessageIds))))
```
chibenwa, line 67: Can I break this by supplying a message with an absurdly large number of references?

hungphan227: How many?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresThreadIdGuessingAlgorithm.java
hungphan227: In the current implementation, data will be sorted by arrival date. In postgres implementation, sorting data by MESSAGE_ID could be good option. And to do that, MESSAGE_ID must be generated by UuidCreator.getTimeOrderedEpoch(). WDYT?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresMailboxFactory.java
Arsnael: Yeah I don't understand that class either... Why you don't follow the current pattern? Should have a Factory subclass in PostgresThreadDAO, to make `PostgresThreadDAO.Factory`, like the other generated factories in `PostgresMailboxSessionMappoerFactory`?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadModule.java
```diff
@@ -0,0 +1,66 @@
...
+                .column(MESSAGE_ID)
+                .column(THREAD_ID)
+                .column(HASH_BASE_SUBJECT)
+                .constraint(DSL.primaryKey(USERNAME, HASH_MIME_MESSAGE_ID, MESSAGE_ID))))
+            .supportsRowLevelSecurity()
+            .build();
+
+        PostgresIndex MESSAGE_ID_INDEX = PostgresIndex.name("thread_message_id_index")
+            .createIndexStep((dsl, indexName) -> dsl.createIndexIfNotExists(indexName)
+                .on(TABLE_NAME, MESSAGE_ID));
```
Arsnael, line 59: Wouldn't it make sense to have THREAD_ID being indexed as well?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadModule.java
```diff
@@ -0,0 +1,66 @@
...
+
+        PostgresTable TABLE = PostgresTable.name(TABLE_NAME.getName())
+            .createTableStep(((dsl, tableName) -> dsl.createTableIfNotExists(tableName)
+                .column(USERNAME)
+                .column(HASH_MIME_MESSAGE_ID)
+                .column(MESSAGE_ID)
+                .column(THREAD_ID)
+                .column(HASH_BASE_SUBJECT)
+                .constraint(DSL.primaryKey(USERNAME, HASH_MIME_MESSAGE_ID, MESSAGE_ID))))
+            .supportsRowLevelSecurity()
```
quantranhong1999, line 54: No need RLS IMO, as information in this table is hashed and not critical data subject to tenant data leak issue.

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadModule.java
hungphan227: Will be added if https://github.com/apache/james-project/pull/1941#discussion_r1464356005 is implemented

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresMailboxFactory.java
hungphan227: I think it would be better to gather all DAOs creation in one feactory instead of one factory per DAO. WDYT?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadModule.java
quantranhong1999: Yeah. I am not fully aware of the RLS scope either...

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresMailboxFactory.java
Arsnael: I think the factories being subclasses of their DAO is easier to follow IMO

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
chibenwa: Maybe we can bathc them by group of 16?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
chibenwa: Other opinions?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadModule.java
chibenwa: Linked to a user -> RLS is doable ?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
hungphan227: It seems that Postgres is still able to handle hundreds elements in IN clause without any issues. What is maximum length of hashMimeMessageIds?

quantranhong1999: > What is maximum length of hashMimeMessageIds?

Can not say for sure. If the conversation keeps being looped then the length could be very big :D

On IN limitation, I remember @vttranlina apply it somewhere in the mailbox component?

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
quantranhong1999: IMO we should use a single INSERT Statement with Multiple Sets of Values.

ChatGPT:
```
1. Single INSERT Statement with Multiple Sets of Values:

sql
Copy code
INSERT INTO your_table (column1, column2, column3)
VALUES
    (value1_1, value1_2, value1_3),
    (value2_1, value2_2, value2_3),
    ...
    (valueN_1, valueN_2, valueN_3);
Pros:

Single statement execution, which can be more efficient and have lower overhead.
Easier to read and maintain for a small number of rows.

Generally performs better than a batch of individual INSERT statements.

Atomic execution, meaning either all rows are inserted, or none are.

Better for inserting large sets of data.

Cons:

Limited by the maximum allowed query size, which might lead to issues if the number of rows is very large.
Might become harder to read and maintain as the number of rows increases.

2. Batch INSERT Statements:

sql
Copy code
INSERT INTO your_table (column1, column2, column3) VALUES (value1_1, value1_2, value1_3);
INSERT INTO your_table (column1, column2, column3) VALUES (value2_1, value2_2, value2_3);
...
INSERT INTO your_table (column1, column2, column3) VALUES (valueN_1, valueN_2, valueN_3);
Pros:

Easier to generate dynamically if the data is coming from various sources or systems.
More control over the execution of each individual statement.
Cons:

Higher overhead due to multiple statement executions.
Might be slower for a large number of rows compared to a single INSERT statement.
Not as readable or maintainable for a large number of rows.
Each statement is executed independently, meaning some rows may be inserted while others fail.
```

TLDR:
- multi values INSERT seems to have better performance than a BATCH statement.
- multi values INSERT is atomic, while BATCH statement is not.

hungphan227: I asked ChatGPT by Vietnamese and it gave the opposite answer :((

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadModule.java
chibenwa: Should be added given that https://github.com/apache/james-project/pull/1941#discussion_r1464356005 will be implemented.

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
hungphan227: one more thing, batch should be used together with prepared statement, not alone

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/PostgresThreadIdGuessingAlgorithm.java
hungphan227: updated

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
hungphan227: partition added

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadModule.java
hungphan227: added

Arsnael (approved):

mailbox/postgres/src/test/java/org/apache/james/mailbox/postgres/PostgresThreadIdGuessingAlgorithmTest.java
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    @Override
+    protected void saveThreadData(Username username, Set<MimeMessageId> mimeMessageIds, MessageId messageId, ThreadId threadId, Optional<Subject> baseSubject) {
+        PostgresThreadDAO threadDAO = threadDAOFactory.create(username.getDomainPart());
+        threadDAO.insertSome(username, hashMimeMessagesIds(mimeMessageIds), PostgresMessageId.class.cast(messageId), threadId, hashSubject(baseSubject)).block();
+    }
+
+    @Test
+    void givenAMailInAThreadThenGetThreadShouldReturnAListWithOnlyOneMessageIdInThatThread() throws MailboxException {
```
chibenwa, line 101: Why those tests are not part of the contract?

vttranlina (approved):

mailbox/postgres/src/test/java/org/apache/james/mailbox/postgres/PostgresThreadIdGuessingAlgorithmTest.java
hungphan227: Instead of using threadDAO, getMessageIdsInThread method in CassandraThreadIdGuessingAlgorithm use mailboxManager to get data. The contract use messageMapper to create data for testing getMessageIdsInThread method. 
PostgresThreadIdGuessingAlgorithm use threadDAO to get data in getMessageIdsInThread method. Therefore, I also use threadDAO to create data for testing getMessageIdsInThread method. Thus, I have to override givenAMailInAThreadThenGetThreadShouldReturnAListWithOnlyOneMessageIdInThatThread of the contract

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/dao/PostgresThreadDAO.java
hungphan227: ChatGPT:

1.  **Single Insert Query with Multiple Rows:**
    
    *   **Advantages:**
        
        *   Simplicity: Writing a single insert query for multiple rows is straightforward and easy to implement.
        *   Fewer round-trips: You send one query to the database server, reducing the number of round-trips between your application and the database.
    *   **Considerations:**
        
        *   Scalability: For a large number of rows, a single insert query can become unwieldy and may lead to performance issues.
        *   Limited reuse: Each row's values need to be explicitly specified in the query, limiting the reuse of the query for different sets of data.
        
2.  **Batch Insert with Prepared Statements:**
    
    *   **Advantages:**
        
        *   Performance: Prepared statements can be more efficient, especially when inserting a large volume of data, as the query is precompiled and can be reused with different parameter values.
        *   Parameterized queries: Prepared statements allow you to use parameterized queries, enhancing security by protecting against SQL injection.
    *   **Considerations:**
        
        *   Complexity: Implementing batch inserts with prepared statements may require more code and complexity in your application.
        *   Overhead for small datasets: For a small number of rows, the overhead of using prepared statements might outweigh the performance benefits.

**Recommendations:**

*   For small to moderate datasets, a single insert query with multiple rows may be simpler and sufficient.
*   For large datasets or when performance is critical, consider using batch inserts with prepared statements. This can be particularly beneficial when inserting thousands or millions of rows, as it reduces the overhead of repeatedly sending similar queries.

Always profile and test the performance of your specific use case to determine the most suitable approach. Additionally, keep in mind that other factors, such as indexing, database configuration, and network latency, can also influence the overall performance.

hungphan227: In conclusion, Single Insert Query with Multiple Rows may not work well with a large number of rows. Single Insert Query with Multiple Rows ensure atomicity while Batch Insert with Prepared Statements does not.

Regarding atomicity, CassandraThreadIdGuessingAlgorithm does not ensure atomicity so I think it is not necessary to ensure atomicity in PostgresThreadIdGuessingAlgorithm

quantranhong1999 (approved):

PR #1943, opened 2024-01-24
JAMES-2586 Implement PostgresMailboxChangeRepository

quantranhong1999:


server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/change/PostgresMailboxChangeModule.java
```diff
@@ -0,0 +1,73 @@
...
+                .column(ACCOUNT_ID)
+                .column(STATE)
+                .column(DATE)
+                .column(IS_SHARED)
+                .column(IS_COUNT_CHANGE)
+                .column(CREATED)
+                .column(UPDATED)
+                .column(DESTROYED)
+                .constraint(DSL.primaryKey(ACCOUNT_ID, STATE))))
+            .supportsRowLevelSecurity()
```
quantranhong1999, line 61: Q: Should we disable RLS for email change and mailbox change?

( I am not sure but seeing DAOs keep being created during runtime makes me concerned about performance)
Likely the data in those two tables is not sensitive for domain isolation.

Arsnael: Might agree with this. The info also is bound to disappear eventually with the TTL

quantranhong1999: Likely can keep the RLS wherever we can: https://github.com/apache/james-project/pull/1941#discussion_r1465162089

Arsnael: @chibenwa thoughts about this one? I know we said RLSs deeply linked to users so technically it should apply there, but as those entries  are for jmap notifications and not persistent (TTL), does it really make sense?

chibenwa: To be honnest the accountId technically belong to a user linked to a domain. Which means that with this asumption in mind it should be doable to apply RLS (though we likely cannot get the domain from the AccountId).

So without an API change I think applying RLS would not be achieved.

As said above this is not that critical: just state changes... As such I'd agree not using RLS here if the setup is hard to achieve.

quantranhong1999: > though we likely cannot get the domain from the AccountId

Hmm likely we can: `Username.of(accountId.getIdentifier()).getDomainPart()`
cf https://github.com/apache/james-project/blob/postgresql/server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/change/PostgresEmailChangeRepository.java#L113

chibenwa: Then...


Let's go!

vttranlina (approved):

Arsnael (approved):

Arsnael:
Can you squash the fixup?

quantranhong1999:
> Conflict

Resolved.

quantranhong1999:
cc @Arsnael this is good to merge now.

PR #1947, opened 2024-01-25
[UPDATE] Update spark version: Jetty update

chibenwa:


pom.xml
```diff
@@ -2210,6 +2210,11 @@
                 <artifactId>guice-assistedinject</artifactId>
                 <version>${guice.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.intellisrc</groupId>
+                <artifactId>spark-core</artifactId>
+                <version>2.9.4-unofficial-5</version>
```
quantranhong1999, line 2216: <3 good catch
It is good to mitigate our Spark pain.

Arsnael (approved):

PR #1951, opened 2024-01-25
JAMES-2372 Fixing multiple accounts with Fetchmail

MaximeBrunetPolyconseil:
This PR aims to fix https://issues.apache.org/jira/browse/JAMES-2372.

There was a bug with the increment of one variable but also with getting the configuration for each account.

I did a test but I'm not certain about the correct use and configuration of users with different types for accounts.

### DIFF
--- a/server/protocols/fetchmail/pom.xml
+++ b/server/protocols/fetchmail/pom.xml
@@ -83,6 +83,11 @@
             <artifactId>commons-configuration2</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.apache.james</groupId>
+            <artifactId>james-server-data-memory</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
--- a/server/protocols/fetchmail/pom.xml
+++ b/server/protocols/fetchmail/pom.xml
@@ -83,6 +83,11 @@
             <artifactId>commons-configuration2</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>jcl-over-slf4j</artifactId>
         </dependency>
@@ -94,12 +99,5 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
-
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
-            <scope>test</scope>
-        </dependency>
     </dependencies>
-
 </project>
--- a/server/protocols/fetchmail/pom.xml
+++ b/server/protocols/fetchmail/pom.xml
@@ -94,6 +94,12 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
 </project>
--- a/server/protocols/fetchmail/src/test/java/org/apache/james/fetchmail/FetchMailTest.java
+++ b/server/protocols/fetchmail/src/test/java/org/apache/james/fetchmail/FetchMailTest.java
@@ -1,0 +1,68 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.fetchmail;
+
+import org.apache.commons.configuration2.BaseHierarchicalConfiguration;
+import org.apache.commons.configuration2.HierarchicalConfiguration;
+import org.apache.commons.configuration2.tree.ImmutableNode;
+import org.apache.james.core.Username;
+import org.apache.james.user.api.UsersRepository;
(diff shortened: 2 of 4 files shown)


server/protocols/fetchmail/src/test/java/org/apache/james/fetchmail/FetchMailTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.mockito.Mockito.when;
+
+class FetchMailTest {
+
+    @Test
+    void shouldAddAccountWithGoodDataFetchMail() throws Exception {
+        try (MockedConstruction<ParsedConfiguration> mockParsedConfiguration = Mockito.mockConstruction(ParsedConfiguration.class)) {
+            UsersRepository fieldLocalUsers = mock(UsersRepository.class);
+            when(fieldLocalUsers.countUsers()).thenReturn(4);
+            when(fieldLocalUsers.list()).thenReturn(List.of(Username.of("user23@domain.com")).iterator()).thenReturn(List.of(Username.of("user23@domain.com")).iterator());
```
Arsnael, line 45:
```suggestion
            when(fieldLocalUsers.list()).thenReturn(List.of(Username.of("user23@domain.com")).iterator());
```
think is enough :)

Arsnael (commented):
Except a minor remark to fix, I think looks good enough to me. Thank you for this!

server/protocols/fetchmail/pom.xml
```diff
@@ -94,6 +94,12 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.mockito</groupId>
```
chibenwa, line 99: POM file needs alphabetic ordering. THis thus need to ge before SLF4J.

server/protocols/fetchmail/src/test/java/org/apache/james/fetchmail/FetchMailTest.java
chibenwa: Could we indent this?

chibenwa (commented):
Hi. Thanks for the fix!

I am quite curious on why and how you encountered this bug: FetchMail is not a very used component of James, is not wired with James.

Did you suceeded to run it with Spring?

What do you use FetchMail for?

The pull request otherwize looks fine to me.

server/protocols/fetchmail/src/test/java/org/apache/james/fetchmail/FetchMailTest.java
MaximeBrunetPolyconseil: Thank you for the review. The thenReturn is added twice because it is called twice in the test. Otherwise the first iterator is returned on second call so it has already been iterated over and returns no value on hasNext.

I have indented the file using the formatting.xml mentioned https://james.apache.org/contribute.html. Am i missing something?

MaximeBrunetPolyconseil:
> Hi. Thanks for the fix!
> 
> I am quite curious on why and how you encountered this bug: FetchMail is not a very used component of James, is not wired with James.
> 
> Did you suceeded to run it with Spring?
> 
> What do you use FetchMail for?
> 
> The pull request otherwize looks fine to me.

I have not been using FetchMail myself, i wanted to contribute to the project and went for some tickets in Jira that were tagged as newbie. It seemed like a useful ticket to do i suppose.

server/protocols/fetchmail/src/test/java/org/apache/james/fetchmail/FetchMailTest.java
Arsnael: What @chibenwa means by indentation is something looking like this:

```
when(fieldLocalUsers.list())
    .thenReturn(List.of(Username.of("user23@domain.com")).iterator())
    .thenReturn(List.of(Username.of("user23@domain.com")).iterator());
```

> The thenReturn is added twice because it is called twice in the test.

Ah maybe yes, those iterators can be tricky sometimes :) Fair enough!

Arsnael (approved):

MaximeBrunetPolyconseil:
> > Hi. Thanks for the fix!
> > I am quite curious on why and how you encountered this bug: FetchMail is not a very used component of James, is not wired with James.
> > Did you suceeded to run it with Spring?
> > What do you use FetchMail for?
> > The pull request otherwize looks fine to me.
> 
> I have not been using FetchMail myself, i wanted to contribute to the project and went for some tickets in Jira that were tagged as newbie. It seemed like a useful ticket to do i suppose.
> 
> I'm going to try and run it with Spring to make sure it's working correctly.

Hello, I ran the Spring application with FetchMail enabled with two accounts added and they were both added with the right values and the fieldSequence incremented. So it is working.

image

Though I'm not sure why some unit tests are failing. They have nothing to do with this MR (other modules) and they also fail on master for me.

server/protocols/fetchmail/src/test/java/org/apache/james/fetchmail/FetchMailTest.java
```diff
@@ -0,0 +1,76 @@
...
+class FetchMailTest {
+
+    @Test
+    void shouldAddAccountWithGoodDataFetchMail() throws Exception {
+        try (MockedConstruction<ParsedConfiguration> mockParsedConfiguration = Mockito.mockConstruction(ParsedConfiguration.class)) {
+            UsersRepository fieldLocalUsers = mock(UsersRepository.class);
+            when(fieldLocalUsers.countUsers()).thenReturn(4);
+            when(fieldLocalUsers.list())
+                    .thenReturn(List.of(Username.of("user23@domain.com")).iterator())
+                    .thenReturn(List.of(Username.of("user23@domain.com")).iterator());
```
chibenwa, line 47: Maybe use `MemoryUserRepository` in order to avoid mocked data?

This way we would avoid using a mock and all related errors.

chibenwa (commented):
Which error do you encounter running the unit tests?

Arsnael:
> Though I'm not sure why some unit tests are failing. They have nothing to do with this MR (other modules) and they also fail on master for me.

master branch is broken at the moment and we work on fixing it. Don't worry too much for the moment, we will rerun your PR when things are fixed :) and sorry for the temporary inconvenience

server/protocols/fetchmail/src/test/java/org/apache/james/fetchmail/FetchMailTest.java
MaximeBrunetPolyconseil: Thank you for your review. I removed the mock for `UserRepository` and used `MemoryUserRepository`. I also improved the test by adding only one alllocal account as it is only possible to have one for a given fetchmail configuration.

They are no errors otherwise in the new tests. The errors are coming from other modules, the detail is in the CI run. But it will be fix eventually.

chibenwa (commented):
As we no longer use mockito we can cleanup the imports / the dependencies?

MaximeBrunetPolyconseil:
> As we no longer use mockito we can cleanup the imports / the dependencies?

Yes my bad. As I removed auto indentation, I also removed the cleanup of imports. I still use mockito to mock the construction of `ParsedConfiguration` since it would need way too much setup for no reason in the test.

Arsnael (approved):

Arsnael:
@chibenwa green! are you ok with latest changes now? Can we merge this?

chibenwa (approved):
Yes ;-)

Arsnael:
Merged! Thank you for your kind contribution @MaximeBrunetPolyconseil  :)

PR #1952, opened 2024-01-25
[boyscout] fixes missing mdc values in smtp audit logs

jeantil:
Both the HA proxy path and the non behaviour overrides do build the mdc. unfortunately unless behind an HA proxy, SMTP server will go the behaviour overrides path and lose its MDC in the process.

I discovered this while attempting to configure crowdsec on my SMTP server instance and noticed we lacked the remote IP in the logs.

I could not find a sane way to write tests for this patch.


Arsnael (approved):

chibenwa (approved):

chibenwa:
We could imagine starting a SMTP server with for instance an extra HELO hook that just perform a LOG

Then we could do assertions in an integration tests using https://projects.lidalia.org.uk/slf4j-test/ we likely can catch the MDC in the logging events.

That's how I would do if I am contractualy required to write such a test. But TBH I am very likely to just do like you did...

jeantil:
thank you :D

PR #1946, opened 2024-01-25
[UPDATE] Update OpenSearch driver (fixes minor CVEs and drop code)

chibenwa:

PR #1953, opened 2024-01-26
[UPDATE] Update OpenSearch driver (fixes minor CVEs and drop code)

Arsnael:
Credits to https://github.com/apache/james-project/pull/1946 , just to make the CI catch that build, which wasn't on the other one, no idea why...


chibenwa:
The CI crashed yesterday (503 code)

PR #1956, opened 2024-01-26
[ENHANCEMENT] Move Sent mailbox up the sort order

chibenwa:


Arsnael (approved):
Why not but how is that really an enhancement?

quantranhong1999:
> Why not but how is that really an enhancement?

https://github.com/linagora/tmail-flutter/issues/2328

Arsnael:
Not sure an arbitrary decision regarding TMail sorting mailbox options should influence Apache James like that though

Arsnael (approved):

quantranhong1999 (approved):

chibenwa:
Can someone take care of the follow up PR onto TMail?

quantranhong1999:
> Can someone take care of the follow up PR onto TMail?

Let me.

PR #1960, opened 2024-01-29
JAMES-2586 Implement PostgresCustomIdentityDAO

hungphan227:


Arsnael (approved):

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
```diff
@@ -0,0 +1,233 @@
...
+import scala.collection.JavaConverters;
+import scala.collection.immutable.List;
+import scala.collection.immutable.Seq;
+import scala.jdk.javaapi.OptionConverters;
+import scala.runtime.BoxedUnit;
+
+public class PostgresCustomIdentityDAO implements CustomIdentityDAO {
+    static class Email {
+        private String name;
+        private String email;
```
chibenwa, line 70: Why don't we get an immutable object? 
We can do a POJO copy if we need to change something in it.
Or a fancy builder?

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
```diff
@@ -0,0 +1,233 @@
...
+                .where(USERNAME.eq(user.asString()))))
+            .map(Throwing.function(this::readRecord));
+    }
+
+    @Override
+    public SMono<Identity> findByIdentityId(Username user, IdentityId identityId) {
+        return SMono.fromPublisher(findByIdentityIdReturnPublisher(user, identityId));
+    }
+
+    private Publisher<Identity> findByIdentityIdReturnPublisher(Username user, IdentityId identityId) {
```
vttranlina, line 129: return `SMono<Identity>` here, and you don't need to one more wrapping it in `SMono.fromPublisher`

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
```diff
@@ -0,0 +1,233 @@
...
+        return executorFactory.create(user.getDomainPart())
+            .executeRow(dslContext -> Mono.from(dslContext.selectFrom(TABLE_NAME)
+                .where(USERNAME.eq(user.asString()))
+                .and(ID.eq(identityId.id()))))
+            .map(Throwing.function(this::readRecord));
+    }
+
+    @Override
+    public Publisher<BoxedUnit> update(Username user, IdentityId identityId, IdentityUpdate identityUpdate) {
+        return Mono.from(findByIdentityIdReturnPublisher(user, identityId))
```
vttranlina, line 139: Do you think we should refactor here,
We don't need `find` and `update`, we can `update` directly in a single request?
I checked the `IdentityUpdate` class, it is quite simple, no dep business.
WDYT?

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
```diff
@@ -0,0 +1,233 @@
...
+        return OptionConverters.toScala(Optional.of(JavaConverters.asScalaIterator(Arrays.stream(jsonArray)
+                .map(Throwing.function(this::convertToObject))
+                .map(Throwing.function(email -> EmailAddress.from(Optional.ofNullable(email.getName()), new MailAddress(email.getEmail()))))
+                .iterator())
+            .toList()));
+    }
+
+    private String[] convertToStringArray(Option<List<EmailAddress>> maybeEmailAddresses) {
+        return OptionConverters.toJava(maybeEmailAddresses).map(emailAddresses ->
+                JavaConverters.asJavaCollection(emailAddresses).stream()
```
vttranlina, line 216:
```
@deprecated("Use `scala.jdk.CollectionConverters` instead", "2.13.0")
```

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
```diff
@@ -0,0 +1,233 @@
...
+                    .map(emailAddress -> convertToJson(new Email(emailAddress.nameAsString(),
+                        emailAddress.email().asString()))).toArray(String[]::new))
+            .orElse(new String[0]);
+    }
+
+    private String convertToJson(Email email) {
```
vttranlina, line 222: Why we don't use Postgresql `JSON` type directly?
Instead of `convertToJson` (asString) -> and save to columm `Array String`

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityModule.java
```diff
@@ -0,0 +1,69 @@
...
+                .column(SORT_ORDER)
+                .column(MAY_DELETE)
+                .constraint(DSL.primaryKey(USERNAME, ID))))
+            .supportsRowLevelSecurity()
+            .build();
+    }
+
+    PostgresModule MODULE = PostgresModule.builder()
+        .addTable(TABLE)
+        .build();
```
vttranlina, line 68: on more Index for only `username` column?

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
hungphan227: It require refactoring Identity which is quite complicated IMO. I think let it be for now

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
hungphan227: In cassandra, it is the array of udt. Therefore, I think string array in Postgres is equivalent. array of json is not supported by JOOQ. array json does not fit very well I think

vttranlina: Json of list?

hungphan227: uh

vttranlina: My meaning is we can serialize/ deserialize directly `Json of list`, we don't need array json.

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityModule.java
hungphan227: I think it is covered in index (USERNAME, ID)

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
hungphan227: What is the difference when using Json of list? I think the only difference is that Postgres check if it is valid json

quantranhong1999: > array json (the whole array in json format) does not fit very well I think

Just curious but why do you feel like that? 
IMO storing an array in a JSON would be better.  If we work and store data under JSON, we should use JSONB to enforce JSON validation. (we did the same in other PRs).
Storing JSON under Array[String] can not prevent inserting malformed strings and leads to deserialization issues.

hungphan227: Regarding validation, json library (jackson, ...) has already validated strings

quantranhong1999: Another story like mutable/ immutable objects for example.

We use a mutable object and we are confident that we would not change that object in the middle. But who knows?

That is why we follow best practices to avoid using mutable objects for example.

hungphan227: well ok If you guys insist on that, I will change

hungphan227: updated

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityModule.java
chibenwa: Link for such a claim?

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
chibenwa: Nope wrong.

Annotate the constructor with `@JsonCreator`, annotate parameters with `@JsonParameter` and use an immutable POJO please.

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityModule.java
hungphan227: "A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns"

https://www.postgresql.org/docs/current/indexes-multicolumn.html

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityDAO.java
hungphan227: updated. sorry for mistake :))

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityModule.java
vttranlina: I had a quick lab for testing:
Here is the result of query `explain select * from custom_identity where  username='user9811';`:

- Query where `username` with only `custom_identity_pkey`
Screenshot_20240130_185832

- Query where `username` with `custom_identity__index_username`  (Which I tried to create with only `username` column)
image

We can see the `cost` of `custom_identity__index_username` is better

_________

// Script for generating sample data if anyone interested:

```sql
INSERT INTO custom_identity (username, id, name, email, reply_to, bcc, text_signature, html_signature, sort_order, may_delete, domain)
SELECT
    'user' || generate_series(1, 10000) AS username,
    uuid_generate_v4(),
    'Name' || generate_series(1, 10000) AS name,
    'email' || generate_series(1, 10000) || '@example.com' AS email,
    '{}' AS reply_to,
    '{}' AS bcc,
    'Text Signature' AS text_signature,
    'HTML Signature' AS html_signature,
    null AS sort_order,
    true AS may_delete,
    'example.com' AS domain
FROM generate_series(1, 10000);
```

Arsnael:
Ok read it. Please rebase as well

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/identity/PostgresCustomIdentityModule.java
hungphan227: Probably in this case, we have a lot of ids per username so it is better to create index for usename only. Index added

vttranlina:
Conflict, please rebase it

vttranlina (approved):

Arsnael (approved):

PR #1957, opened 2024-01-26
Revert "[UPDATE] Update spark version: Jetty update"

chibenwa:
This reverts commit 69e45e4ddf164cff38764bcc4860b47b1274abb1.

```
org.apache.james.webadmin.routes.UserMailboxesRoutesTest$NormalBehaviour.putShouldReturnNotFoundWhenNoMailboxName

Request method:	PUT
Request URI:	http://localhost:40949/users/username/mailboxes
Proxy:			<none>
Request params:	<none>
Query params:	<none>
Form params:	<none>
Path params:	<none>
Headers:		Accept=application/json, application/javascript, text/javascript, text/json
				Content-Type=application/json; charset=UTF-8
Cookies:		<none>
Multiparts:		<none>
Body:			<none>

HTTP/1.1 405 Method Not Allowed
Date: Fri, 26 Jan 2024 11:28:59 GMT
Access-Control-Allow-Origin: *
Access-Control-Request-Method: DELETE, GET, POST, PUT
Access-Control-Allow-Headers: Content-Type, Authorization, Accept
Content-Type: application/json
Transfer-Encoding: chunked

<html><body><h2>405 Method Not Allowed</h2></body></html>
```

https://github.com/apache/james-project/pull/1947 likely merged too fast.

PR #1974, opened 2024-02-05
[UPDATE] Update spark version: Jetty update

Arsnael:
Getting back on this: https://github.com/apache/james-project/pull/1957


chibenwa (approved):

chibenwa:
Still many error

```
org.apache.james.webadmin.routes.AddressMappingRoutesTest.addAddressMappingShouldSupportOneTimeEncodedAddress
```

Arsnael:
Hmmm... I understand the need to move out a project that is not being maintained anymore to one of its forks, but a bit trickier here actually.

There seems to be issues with encoded slash, and we have a lot of routes that can allow that and related tests, like this one: https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-1974/3/testReport/junit/org.apache.james.webadmin.routes/AliasRoutesTest$FilteringOtherRewriteRuleTypes/putUserForAliasWithEncodedSlashShouldReturnNoContent/

It feels like it interprets the encoded slash as an actual slash... 

@chibenwa the code of that fork definitely raises issues on our project. This is trickier and would require caution. Maybe potentially investigating and someone could contribute to the fork if finding a solution? It would require time though.

WDYT?

chibenwa:
> WDYT?

This was opportunistic trial but if it do not work let's forget about this.

PR #1978, opened 2024-02-05
[Build][Fix] Fix tag for Minio testing image

Arsnael:
I saw lots of builds failing on Minio today. decided to investigate and yeah locally it was very unstable on latest minio image.

Not the first time it happens, so I went back on the latest stable tag that I could get when testing locally and pushed this PR. IDK if the image is buggy or if something slightly changed in minio behavior, but not the issue here.


chibenwa (approved):

jeantil:
thank you !

chibenwa:
> thank you !

HUGE +1

Though do that mean that James is broken in some subtle ways with latest releases of MinIO? 

:-/

Arsnael:
> Though do that mean that James is broken in some subtle ways with latest releases of MinIO?

I did not have the time to investigate much yesterday, but locally I had some save blob test failing on the big blobs, with something like a connection reset .

I don't know if there is a default max size value lower than our blob for example that we need to conf or if there is an issue on MinIO latest images. 

Will open a ticket though, worth checking.

PR #1986, opened 2024-02-07
[JAMES-3978] avoid forked javadoc goals and associated build scans

jeantil:
After enabling the build scans  quickly noticed that each full clean install generates extra build scans for mpt-core and testing
image
I tracked it to a behavior of the javadoc plugin which forks maven goals for dependencies
symptoms are 
```
[INFO] --- javadoc:3.1.1:javadoc (create-javadocs) @ apache-james-mpt-app ---
[INFO] The goal 'org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:javadoc' has not been previously called for the module: 'org.apache.james:testing-base:jar:3.9.0-SNAPSHOT'. Trying to invoke it...
[INFO] The goal 'org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:javadoc' has not been previously called for the module: 'org.apache.james:apache-james-mpt-core:jar:3.9.0-SNAPSHOT'. Trying to invoke it...
[INFO]
```
and 
```
[INFO] --- javadoc:3.1.1:jar (create-javadocs) @ apache-james-mpt-app ---
[INFO] The goal 'org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:javadoc' has not been previously called for the module: 'org.apache.james:testing-base:jar:3.9.0-SNAPSHOT'. Trying to invoke it...
[WARNING] Creating fake javadoc directory to prevent repeated invocations: /home/jean/perso/dev/secretjames/develocity/testing/base/target/apidocs
[INFO] The goal 'org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:javadoc' has not been previously called for the module: 'org.apache.james:apache-james-mpt-core:jar:3.9.0-SNAPSHOT'. Trying to invoke it...
[WARNING] Creating fake javadoc directory to prevent repeated invocations: /home/jean/perso/dev/secretjames/develocity/mpt/core/target/apidocs
[ERROR] Error fetching link: /home/jean/perso/dev/secretjames/develocity/testing/base/target/apidocs. Ignored it.
[ERROR] Error fetching link: /home/jean/perso/dev/secretjames/develocity/mpt/core/target/apidocs. Ignored it.
[INFO]
```
in the logs

Investigating led me to an apache bookkeeper issue that ended up disabling the `<detectOfflineLinks>false</detectOfflineLinks>` option but for performance reasons.

I haven't found another way to avoid the forks or to prevent the forks to emit build scans and it doesn´t  seem to be a problem for bookkeeper. So I propose to follow them and disable that feature.

I added a few cleanup commits that
- solve the toplevel maven warnings 
- remove the felix bundle plugin despite it not being the root cause for the forks as benoit seemed to feel it was a good thing :D

### DIFF
--- a/mpt/core/pom.xml
+++ b/mpt/core/pom.xml
@@ -87,7 +87,7 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.felix</groupId>
-                <artifactId>maven-bundle-plugin</artifactId>
+                <artifactId>maven-bundle-plugin </artifactId>
                 <version>${felix.plugin.version}</version>
                 <extensions>true</extensions>
                 <configuration>
--- a/mpt/core/pom.xml
+++ b/mpt/core/pom.xml
@@ -64,7 +64,6 @@
                 <artifactId>maven-jar-plugin</artifactId>
                 <configuration>
                     <archive>
-                        <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
                         <manifestEntries>
                             <Specification-Title>Apache James MPT</Specification-Title>
                             <Specification-Version>${project.version}</Specification-Version>
@@ -73,6 +72,9 @@
                             <Implementation-Version>${project.version}</Implementation-Version>
                             <Implementation-Vendor>The Apache Software Foundation</Implementation-Vendor>
                             <Implementation-Vendor-Id>org.apache</Implementation-Vendor-Id>
+                            <Implementation-Url>${project.url}</Implementation-Url>
+                            <Embed-Dependency>*;scope=runtime</Embed-Dependency>
+                            <Export-Package>org.apache.james.mpt;version="${project.version}"</Export-Package>
                             <url>${project.url}</url>
                         </manifestEntries>
                     </archive>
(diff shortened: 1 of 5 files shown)


mpt/core/pom.xml
```diff
@@ -85,9 +85,9 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
+            <!-- <plugin>
                 <groupId>org.apache.felix</groupId>
-                <artifactId>maven-bundle-plugin</artifactId>
+                <artifactId>maven-bundle-plugin </artifactId>
```
chibenwa, line 90: +1 to turn this plugin off

jeantil: hmm actually we can't at least not easily 
- the output of this plugin is used as in input in the jar plugin see line 67
- this plugin is not responsible for the forks :D I was going to revert that change

jeantil: I POCed dropping the plugin and letting jar-plugin create the manifest 
the diff in the manifest file is quite heavy and I have no idea what the consequences could be
image
some trivial properties can be replicated or ignored but no so sure about Import-Package

chibenwa (approved):

PR #1987, opened 2024-02-07
[JAMES-3978] activate flaky tests detection

jeantil:
Flaky test detection in develocity requires allowing test retry at least once. While this will probably make our test suite even longer than it already is, it could help prioritize the tests to be fixed to enabled test result remote caching

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3969,9 +3969,15 @@
                                 </execution>
                             </executions>
                         </plugin>
-
                         <plugin>
-                            <groupId>pl.project13.maven</groupId>
+                            <groupId>org.apache.maven.plugins</groupId>
+                            <artifactId>maven-surefire-plugin</artifactId>
+                            <configuration>
+                                <rerunFailingTestsCount>2</rerunFailingTestsCount>
+                            </configuration>
+                        </plugin>
+                        <plugin>
+                        <groupId>pl.project13.maven</groupId>
                             <artifactId>git-commit-id-plugin</artifactId>
                             <executions>
                                 <execution>


chibenwa (approved):

chibenwa:
> Flaky test detection in develocity requires allowing test retry at least once.

It will not catch "the docker did not start" kind of flackyness though, unless the retry is done in a new JVM...

jeantil:
> It will not catch "the docker did not start" kind of flackyness though, unless the retry is done in a new JVM...

I would need to look at the surefire options : maybe there is a way to ensure that a retry occurs in a fresh fork ?

chibenwa:
404 on https://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html

Let's merge?

pom.xml
```diff
@@ -664,6 +664,8 @@
         <mockito.version>5.2.0</mockito.version>
         <jacoco-maven-plugin.version>0.8.9</jacoco-maven-plugin.version>
         <javacrumbs.json-unit.version>2.38.0</javacrumbs.json-unit.version>
+
+        <surefire.rerunFailingTestsCount>2</surefire.rerunFailingTestsCount>
```
chibenwa, line 668: Having this as a general property, I would not be sure this is the way.

Shouldn't we rather pass it as a property of the surefire maven plugin?

jeantil:
wierd https://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html works fine for me ...

pom.xml
jeantil: actually your comment made me pause: do we want this behavior locally or only on CI ? in the latter case I would change the Jenkins file to pass the property to the command line

PR #1989, opened 2024-02-07
upgraded activemq to address CVE-2023-46604

LucasHouska:
Patch updated activemq version from 5.18.2 to 5.18.3 to address CVE-2023-46604


chibenwa (approved):
+1 though I do not think James is affected as activemq is embedded and we rely on a invm connection...

Correct?

chibenwa:
Thanks for the contribution!

PR #1994, opened 2024-02-11
[JAMES-3897] Crowdsec SMTP connect hook

jeantil:
EHLO is not required before sending AUTH, therefore blocking on EHLO does't work very well against bruteforce attempts.
This PR introduces a hard connection close when a banned ip attempts to connect again to james even before the IP can send any SMTP command.

### DIFF
--- a/third-party/crowdsec/src/main/java/org/apache/james/crowdsec/CrowdsecSMTPConnectHandler.java
+++ b/third-party/crowdsec/src/main/java/org/apache/james/crowdsec/CrowdsecSMTPConnectHandler.java
@@ -1,0 +1,88 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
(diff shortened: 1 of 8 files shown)


third-party/crowdsec/src/main/java/org/apache/james/crowdsec/CrowdsecSMTPConnectHandler.java
```diff
@@ -0,0 +1,88 @@
...
+import javax.inject.Inject;
+
+import org.apache.james.crowdsec.model.CrowdsecDecision;
+import org.apache.james.protocols.api.Response;
+import org.apache.james.protocols.api.handler.ConnectHandler;
+import org.apache.james.protocols.smtp.SMTPSession;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CrowdsecSMTPConnectHandler implements ConnectHandler<SMTPSession> {
```
chibenwa, line 34: Can this be documented?  At least in the crowdsec extension README...

chibenwa (commented):
Huge +1. Thanks for this!

third-party/crowdsec/src/main/java/org/apache/james/crowdsec/CrowdsecSMTPConnectHandler.java
jeantil: You are right, I added a small blurb in the crowdsec module README

chibenwa (approved):

jeantil:
There was a green build before I rebased to include the readme fix. master has not changed since then so I will consider this as flakyness in the test suite

PR #1998, opened 2024-02-13
[FIX] Lower log level of SSL handshake exceptions

chibenwa:
Client requested protocol TLSv1.1 is not enabled or supported in server context

This benign event should likely not lead to an error log.


chibenwa:
>  does the SMTP stack use the BasicChannelInboundHandler ? 

Yes it does.

quantranhong1999 (approved):

Arsnael (approved):

PR #1999, opened 2024-02-13
Feature/fixing tupo truststoretype

lwolters:
It looks if something is  wrong with the current code. I  think it should be like this, but please correct me if I'm wrong.
After looking in the code, I noticed that the `getTruststoreType()` is never actually used, which can't be correct.

But I might be wrong.


Arsnael:
Hi @lwolters 

Looks legit to me, thanks for the contribution! But I'm not too well versed with the sll part, maybe @chibenwa or @ottoka can have a look at it too?

ottoka:
Yes, you are right, since we are configuring the trust store we should use the trust store type for it. Probably nobody noticed before since in practice people rarely mix the store types, either use JKS (in compat mode) or PKCS12 for both. But nevertheless we should do this right. Nice catch!

ottoka (approved):

Arsnael:
Merging it, thanks @ottoka for the review and thanks @lwolters for your kind contribution :)

PR #2000, opened 2024-02-14
Fix and test indexation of multipart/alternative mails

chibenwa:
- simple multipart/alternative did lead to double indexed body and thus wasted significant disk space
 - But multipart/alternative with attachment had no text indexed at all, causing major miss in the search...
 
 We fix this by not indexing the htmlBody if the textBody is present, and by walking recursively the mime tree.


chibenwa:
Screenshot from 2024-02-14 00-50-33

Arsnael (approved):

PR #2002, opened 2024-02-14
JAMES-3991 Vacation handling should be case insensitive

ottoka:
Easiest way to achieve this goal is to always use lower case in the AccountId wrapper. That should fix the issue everywhere.
Also extended the DAO contracts with tests for case insensitive access.

### DIFF
--- a/server/data/data-api/src/main/java/org/apache/james/vacation/api/AccountId.java
+++ b/server/data/data-api/src/main/java/org/apache/james/vacation/api/AccountId.java
@@ -40,7 +40,7 @@ public class AccountId {
     private final String identifier;
 
     private AccountId(String identifier) {
-        this.identifier = identifier;
+        this.identifier = identifier.toLowerCase();
     }
 
     public String getIdentifier() {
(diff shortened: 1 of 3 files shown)


server/data/data-api/src/main/java/org/apache/james/vacation/api/AccountId.java
```diff
@@ -40,7 +40,7 @@ public static AccountId fromUsername(Username username) {
     private final String identifier;
 
     private AccountId(String identifier) {
-        this.identifier = identifier;
+        this.identifier = identifier.toLowerCase();
```
chibenwa, line 43: I would rather see this in the NotificationRegistry if possible.

chibenwa (approved):

ottoka:
Re your vanished(?) comment: I thought about doing this in the DAOs instead of AccountId, but then you have to do it in multiple places (error prone) and potentially repeated for each storage backend. Thus I decided to go with this centralized solution.

chibenwa:
Yes and after commenting I weighted the pros and cons and decided that putting up a gas factory for this was likely not worth it.

quantranhong1999 (approved):

Arsnael (approved):

PR #2005, opened 2024-02-15
JAMES-3995 Optimize Email/get

chibenwa:
On many very large message we might consume excessive memory.

Avoid buffering of the binary parts in order to save memory.

@Arsnael help wanted to help polish this.

Apparently I did broke some stuff in bodySTructure

### DIFF
--- a/server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/SizeUtils.java
+++ b/server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/SizeUtils.java
@@ -1,0 +1,152 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.jmap.mime4j;
+
+import java.io.IOException;
+import java.io.OutputStream;
+
+import org.apache.james.mime4j.dom.Body;
+import org.apache.james.mime4j.dom.Entity;
+import org.apache.james.mime4j.dom.Header;
+import org.apache.james.mime4j.dom.Message;
+import org.apache.james.mime4j.dom.Multipart;
+import org.apache.james.mime4j.dom.SingleBody;
+import org.apache.james.mime4j.dom.field.ContentTypeField;
+import org.apache.james.mime4j.dom.field.FieldName;
+import org.apache.james.mime4j.message.BodyPart;
+import org.apache.james.mime4j.message.DefaultMessageWriter;
+import org.apache.james.mime4j.message.MessageImpl;
+import org.apache.james.mime4j.message.MultipartImpl;
+import org.apache.james.mime4j.util.ByteSequence;
+import org.apache.james.mime4j.util.ContentUtil;
+
+import com.google.common.io.CountingOutputStream;
+
+public class SizeUtils {
+    public static long sizeOf(Entity entity) throws IOException {
+        if (entity instanceof BodyPart) {
+            BodyPart bodyPart = (BodyPart) entity;
+
+            return sizeOf(bodyPart.getBody());
+        }
+        if (entity instanceof MessageImpl) {
+            MessageImpl bodyPart = (MessageImpl) entity;
+
+            return sizeOf(bodyPart.getBody());
+        }
+        CountingOutputStream countingOutputStream = new CountingOutputStream(OutputStream.nullOutputStream());
+        DefaultMessageWriter defaultMessageWriter = new DefaultMessageWriter();
+        defaultMessageWriter.writeEntity(entity, countingOutputStream);
+        return countingOutputStream.getCount();
+    }
+
+    public static long sizeOf(Header header) throws IOException {
+        CountingOutputStream countingOutputStream = new CountingOutputStream(OutputStream.nullOutputStream());
+        DefaultMessageWriter defaultMessageWriter = new DefaultMessageWriter();
+        defaultMessageWriter.writeHeader(header, countingOutputStream);
+        return countingOutputStream.getCount();
+    }
+
+    public static long sizeOf(Body body) throws IOException {
+        if (body instanceof FakeBinaryBody) {
+            return ((FakeBinaryBody) body).getSize();
+        }
+        if (body instanceof SingleBody) {
+            return ((SingleBody) body).size();
+        }
+        if (body instanceof Multipart) {
+            return sizeOfMultipart((Multipart) body);
+        }
+        if (body instanceof Message) {
+            Message message = (Message) body;
+            return sizeOf(message.getHeader()) + sizeOf(message.getBody());
+        }
+        CountingOutputStream countingOutputStream = new CountingOutputStream(OutputStream.nullOutputStream());
+        DefaultMessageWriter defaultMessageWriter = new DefaultMessageWriter();
+        defaultMessageWriter.writeBody(body, countingOutputStream);
+        return countingOutputStream.getCount();
+    }
+
+    // Inspired from DefaultMessageWriter
+    public static long sizeOfMultipart(Multipart multipart)
+        throws IOException {
+        long result = 0;
+        ContentTypeField contentType = getContentType(multipart);
+
+        ByteSequence boundary = getBoundary(contentType);
+
+        ByteSequence preamble;
+        ByteSequence epilogue;
+        if (multipart instanceof MultipartImpl) {
+            preamble = ((MultipartImpl) multipart).getPreambleRaw();
+            epilogue = ((MultipartImpl) multipart).getEpilogueRaw();
+        } else {
+            preamble = multipart.getPreamble() != null ? ContentUtil.encode(multipart.getPreamble()) : null;
+            epilogue = multipart.getEpilogue() != null ? ContentUtil.encode(multipart.getEpilogue()) : null;
+        }
+        if (preamble != null) {
+            result += preamble.length() + 2;
+        }
+
+        for (Entity bodyPart : multipart.getBodyParts()) {
+            result += 2 + boundary.length() + 2; // -- boudary CRLF
+            result += sizeOf(bodyPart.getHeader());
+            result += sizeOf(bodyPart);
+            result += 2; // CRLF
+        }
+
+        result += 2 + boundary.length() + 2 + 2; // -- boudary -- CRLF
+        if (epilogue != null) {
+            result += epilogue.length();
+        }
+        return result;
+    }
+
+
+    // Taken from DefaultMessageWriter
+    private static ContentTypeField getContentType(Multipart multipart) {
+        Entity parent = multipart.getParent();
+        if (parent == null) {
+            throw new IllegalArgumentException("Missing parent entity in multipart");
+        }
+
+        Header header = parent.getHeader();
+        if (header == null) {
+            throw new IllegalArgumentException("Missing header in parent entity");
+        }
+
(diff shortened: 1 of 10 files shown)


Arsnael:
Alright let me have a look...

server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/JamesBodyFactory.java
```diff
@@ -0,0 +1,366 @@
...
+        CountingOutputStream out = new CountingOutputStream(OutputStream.nullOutputStream());
+        is.transferTo(out);
+        return new FakeBinaryBody(out.getCount());
+    }
+
+    public BinaryBody binaryBody(final byte[] buf) {
+        return new BinaryBody1(buf);
+    }
+
+    static class StringBody1 extends TextBody {
```
Arsnael, line 156: We should probably find better name for those static classes than `1`, `2`, ...

server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/SizeUtils.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+        System.out.println(body.getClass());
+        CountingOutputStream countingOutputStream = new CountingOutputStream(OutputStream.nullOutputStream());
+        DefaultMessageWriter defaultMessageWriter = new DefaultMessageWriter();
+        defaultMessageWriter.writeBody(body, countingOutputStream);
+        return countingOutputStream.getCount();
+    }
+
+    /*
+      private def size(entity: Entity): Try[Size] = {
```
Arsnael, line 70: Commented code block spotted

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailGetMethodContract.scala
```diff
@@ -4591,7 +4591,7 @@ trait EmailGetMethodContract {
       .header(ACCEPT.toString, ACCEPT_RFC8621_VERSION_HEADER)
       .body(request)
     .when
-      .post
+      .post.prettyPeek()
```
Arsnael, line 4594: debug spotted

Arsnael:
Fixed the build

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailGetMethodContract.scala
Arsnael: Note: yes scalafix throws on that... ^^'

Arsnael:
Alright get it email get contract tests failing too...

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/mail/Email.scala
```diff
@@ -610,7 +610,7 @@ private class EmailFullViewFactory @Inject()(zoneIdProvider: ZoneIdProvider, pre
         .getOrElse(Failure(new IllegalArgumentException("No message supplied")))
       mime4JMessage <- Email.parseAsMime4JMessage(firstMessage)
       blobId <- BlobId.of(messageId)
-      bodyStructure <- EmailBodyPart.ofMessage(request.bodyProperties, zoneIdProvider.get(), blobId, mime4JMessage, firstMessage.getSize)
+      bodyStructure <- EmailBodyPart.ofMessage(request.bodyProperties, zoneIdProvider.get(), blobId, mime4JMessage, firstMessage.getBody.size)
```
Arsnael, line 613: Was putting the full message size for the body one

Reduced failing tests from 5 to 1 on EmailGetMethodContract

There is still a problem with `bodyStructureForSimpleMultipart` test, size difference of 2 (actual 2285, expected 2287). Dont know if it's right and previous code wrong or if there is a little trick here.

Will check back after lunch

ottoka: Just a guess, maybe someone counts the CR+LF between headers and body towards the body size?

Arsnael: Maybe there is something like that... Honestly could not put my finger on it.
Thanks for having a guess at it though :)

I think I will pretend it's acceptable. @chibenwa don't hesitate to disagree with this decision if you think it's better

server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/SizeUtils.java
```diff
@@ -0,0 +1,99 @@
...
+import org.apache.james.mime4j.dom.Entity;
+import org.apache.james.mime4j.dom.SingleBody;
+import org.apache.james.mime4j.message.BodyPart;
+import org.apache.james.mime4j.message.DefaultMessageWriter;
+import org.apache.james.mime4j.message.MessageImpl;
+
+import com.google.common.io.CountingOutputStream;
+
+public class SizeUtils {
+    public static long sizeOf(Entity entity) throws IOException {
```
chibenwa, line 35: The `Email/get` test suite was imcomplete. I got on a production workload a call for `MultipartImp` with some FakeBinaryBody in it.

Regarding the JMAP spec, `partId` and `blobId` are null for multiparts so there is no downloads.

To quote the spec:

```
(as referenced by the blobId, i.e., the number of octets in the file the user would download).
```

So I fixed this by returning `0` for multiparts as they cannot be downloaded...

server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/SizeUtils.java
chibenwa: Can be removed.

Sorry I was in a hurry

server/protocols/jmap-rfc-8621-integration-tests/jmap-rfc-8621-integration-tests-common/src/main/scala/org/apache/james/jmap/rfc8621/contract/EmailGetMethodContract.scala
```diff
@@ -3865,7 +3865,7 @@ trait EmailGetMethodContract {
          |                                    "value": "$contentType"
          |                                }
          |                            ],
-         |                            "size": 2287,
+         |                            "size": 2285,
```
chibenwa, line 3868: As said above the only one reason this do not blow up is that the parts are text body and not binary body...

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/EmailGetMethod.scala
```diff
@@ -89,6 +89,7 @@ class EmailGetMethod @Inject() (readerFactory: EmailViewReaderFactory,
       case e: IllegalArgumentException => SMono.just(Invocation.error(ErrorCode.InvalidArguments, e.getMessage, invocation.invocation.methodCallId))
       case e: Throwable => SMono.error(e)
     }).map(invocationResult => InvocationWithContext(invocationResult, invocation.processingContext))
+      .doOnError(e => e.printStackTrace())
```
chibenwa, line 92: Oups

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/mail/Email.scala
chibenwa: Quite likely.

However that's the size of the message as stored in the mailbox.

I suggest downloading the EML and checking the size and validating which on it the right one...

It is not impossible the the MessageWriter does some sanitizing like adding a trailing CRLF if missing...

server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/JamesBodyFactory.java
```diff
@@ -0,0 +1,366 @@
...
+import org.apache.james.mime4j.message.BodyFactory;
+import org.apache.james.mime4j.util.ByteArrayOutputStreamRecycler;
+import org.apache.james.mime4j.util.ContentUtil;
+
+import com.google.common.io.CountingOutputStream;
+
+/**
+ * Factory for creating message bodies.
+ */
+public class JamesBodyFactory implements BodyFactory {
```
chibenwa, line 48: Better name needed: how about `AvoidBinaryBodyBufferingBodyFactory` ?

Arsnael:
Seems creating issues with Email/parse method as well: https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2005/7/testReport/

Will check...

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/mail/Email.scala
Arsnael: Might be that... seems to be 2285 to me when I download the eml and check the multipart/mixed part. Maybe some sanitizing at the work yes. So I'm fine with this honestly

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/EmailParseMethod.scala
```diff
@@ -103,7 +103,7 @@ class EmailParseMethod @Inject()(val blobResolvers: BlobResolvers,
   private def parse(request: EmailParseRequest, blobId: BlobId, blobContent: => InputStream, size: Size): SMono[EmailParseResults] = {
     val result: SMono[EmailParseView] = for {
       mime4JMessage <- SMono.fromTry(Email.parseStreamAsMime4JMessage(blobContent))
-      bodyStructure <- SMono.fromTry(EmailBodyPart.of(request.bodyProperties, zoneIdProvider.get(), blobId, mime4JMessage))
+      bodyStructure <- SMono.fromTry(EmailBodyPart.ofMessage(request.bodyProperties, zoneIdProvider.get(), blobId, mime4JMessage, size))
```
chibenwa, line 106: <3 cool catch thanks Rene!

Arsnael: Well the Email/parse test were failing without this :)

server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/SizeUtils.java
chibenwa: Is this handled?

Arsnael: When I read your comment, I thought you did, as it finished with a sentence `So I fixed this...`

Did I misunderstand something?

Arsnael:
> Could we tripple check exotic multiparts?

I ain't too sure about all of that honestly, but could try

chibenwa:
Email/get on these two mails:

[multipart_report.txt](https://github.com/apache/james-project/files/14350673/multipart_report.txt)

[mixed-alternative.txt](https://github.com/apache/james-project/files/14350758/mixed-alternative.txt)

Let's see if this pass!

And I insist that if we call size on a multipart, we return explicitly `0`...

server/data/data-jmap/src/main/java/org/apache/james/jmap/mime4j/SizeUtils.java
Arsnael: Well then you need to be more clear in your comments then. 

Will take a look, with your suggested tests for other multiparts that seem to test this

chibenwa:
I found the faulty input!

[nested_mixed_alternative.txt](https://github.com/apache/james-project/files/14365621/nested_mixed_alternative.txt)

That's a pretty big deal... Because we need to give a size for the nested message that could be downloaded. `0` is actually not the answer.

More work needed here. I suggest making the mime4J DefaultMessageWriter able to compute size...

chibenwa:
Edit: I did find my way around?

Arsnael:
```
02:47:02,846 [ERROR] Failed to execute goal io.github.evis:scalafix-maven-plugin_2.13:0.1.8_0.11.0:scalafix (scala-check-style) on project jmap-rfc-8621-integration-tests-common: Scalafix invoked with errors. Check logs for details. -> [Help 1]
```

chibenwa:
Some fine tunning needed apparently...

`org.apache.james.jmap.rfc8621.memory.MemoryEmailGetMethodTest.shouldSupportAttachedMessageWithComplexMultipart(GuiceJamesServer)`

Arsnael (approved):

PR #2007, opened 2024-02-16
JAMES-2586 Implement PostgresEmailQueryView

hungphan227:


server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/projections/PostgresEmailQueryViewDAO.java
```diff
@@ -0,0 +1,145 @@
...
+    public Mono<Void> save(PostgresMailboxId mailboxId, ZonedDateTime sentAt, ZonedDateTime receivedAt, PostgresMessageId messageId) {
+        return postgresExecutor.executeVoid(dslContext -> Mono.from(dslContext.insertInto(TABLE_NAME)
+            .set(MAILBOX_ID, mailboxId.asUuid())
+            .set(MESSAGE_ID, messageId.asUuid())
+            .set(SENT_AT, sentAt.toOffsetDateTime())
+            .set(RECEIVED_AT, receivedAt.toOffsetDateTime())
+            .onConflictOnConstraint(PK_CONSTRAINT_NAME)
+            .doUpdate()
+            .set(SENT_AT, sentAt.toOffsetDateTime())
+            .set(RECEIVED_AT, receivedAt.toOffsetDateTime())));
```
chibenwa, line 143: This is supposed to be indempotent so I bet in case of conflict we can just do nothing?

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/projections/PostgresEmailQueryViewModule.java
```diff
@@ -0,0 +1,66 @@
...
+            .createTableStep(((dsl, tableName) -> dsl.createTableIfNotExists(tableName)
+                .column(MAILBOX_ID)
+                .column(MESSAGE_ID)
+                .column(RECEIVED_AT)
+                .column(SENT_AT)
+                .constraint(DSL.constraint(PK_CONSTRAINT_NAME).primaryKey(MAILBOX_ID, MESSAGE_ID))))
+            .supportsRowLevelSecurity()
+            .build();
+
+        PostgresIndex INDEX = PostgresIndex.name("email_query_view_mailbox_id_index")
```
chibenwa, line 57: We want to have an index on sentAt? receivedAt?

chibenwa (commented):
Otherwize it looks great.

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/projections/PostgresEmailQueryViewModule.java
hungphan227: oh I forgot to add :))

server/data/data-jmap-postgres/src/main/java/org/apache/james/jmap/postgres/projections/PostgresEmailQueryViewDAO.java
hungphan227: actually in the contract, there is a test:

```
@Test
    default void saveShouldBeIdempotent() {
        testee().save(mailboxId1(), DATE_1, DATE_2, messageId1()).block();
        testee().save(mailboxId1(), DATE_1, DATE_2, messageId1()).block();

        assertThat(testee().listMailboxContentSortedBySentAt(mailboxId1(), Limit.limit(12)).collectList().block())
            .containsExactly(messageId1());
    }
```

Does this mean in case of conflict we just do nothing?

Arsnael: Idempotent means that if you insert the same record twice you still have the same result. The sent_at and received_at here are not supposed to change, thus why you can simply just do nothing, result will be the same, as for the test.

Arsnael (approved):

server/data/data-jmap/src/test/java/org/apache/james/jmap/api/projections/EmailQueryViewContract.java
```diff
@@ -200,7 +200,7 @@ default void datesCanBeDuplicated() {
         testee().save(mailboxId1(), DATE_1, DATE_2, messageId2()).block();
 
         assertThat(testee().listMailboxContentSortedBySentAt(mailboxId1(), Limit.limit(12)).collectList().block())
-            .containsExactly(messageId1(), messageId2());
+            .containsExactlyInAnyOrder(messageId1(), messageId2());
```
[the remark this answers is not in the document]
hungphan227: Date of msg1 and msg2 are equal so the order does not matter. The reason I change to containsExactlyInAnyOrder is that the order in Postgres implementation is different that Cassandra implementation

vttranlina (approved):

PR #2018, opened 2024-02-18
Do not index body

chibenwa:
This relax pressure on the AES blob store and APPEND

Also this is helpfull upon mass data ingestion.

TODO make this configurable

### DIFF
--- a/mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/events/OpenSearchListeningMessageSearchIndex.java
+++ b/mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/events/OpenSearchListeningMessageSearchIndex.java
@@ -58,6 +58,7 @@ import org.apache.james.mailbox.model.MessageId;
 import org.apache.james.mailbox.model.MessageMetaData;
 import org.apache.james.mailbox.model.SearchQuery;
 import org.apache.james.mailbox.model.UpdatedFlags;
+import org.apache.james.mailbox.opensearch.IndexBody;
 import org.apache.james.mailbox.opensearch.MailboxOpenSearchConstants;
 import org.apache.james.mailbox.opensearch.OpenSearchMailboxConfiguration;
 import org.apache.james.mailbox.opensearch.json.MessageToOpenSearchJson;
@@ -227,6 +228,7 @@ public class OpenSearchListeningMessageSearchIndex extends ListeningMessageSearc
     
     private final Metric reIndexNotFoundMetric;
     private final IndexingStrategy indexingStrategy;
+    private final IndexBody indexBody;
 
     @Inject
     public OpenSearchListeningMessageSearchIndex(MailboxSessionMapperFactory factory,
@@ -248,6 +250,7 @@ public class OpenSearchListeningMessageSearchIndex extends ListeningMessageSearc
         } else {
             this.indexingStrategy = new NaiveIndexingStrategy();
         }
+        this.indexBody = configuration.getIndexBody();
         this.reIndexNotFoundMetric = metricFactory.generate("opensearch_reindex_not_found");
 
         LOGGER.info("OpenSearchMessageSearchIndex activated with index strategy: {}", indexingStrategy.getClass().getSimpleName());
@@ -296,7 +299,14 @@ public class OpenSearchListeningMessageSearchIndex extends ListeningMessageSearc
     private Mono<Void> processAddedEvent(MailboxSession session, MailboxEvents.Added addedEvent, MailboxId mailboxId) {
         return factory.getMailboxMapper(session)
             .findMailboxById(mailboxId)
-            .flatMap(mailbox -> handleAdded(session, mailbox, addedEvent));
+            .flatMap(mailbox -> processAdded(session, mailbox, addedEvent));
+    }
+
+    private Mono<Void> processAdded(MailboxSession session, Mailbox mailbox, MailboxEvents.Added addedEvent) {
+        if (IndexBody.YES.equals(indexBody)) {
+            return handleAdded(session, mailbox, addedEvent, FetchType.FULL);
+        }
+        return handleAdded(session, mailbox, addedEvent, FetchType.HEADERS);
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


mailbox/store/src/main/java/org/apache/james/mailbox/store/search/ListeningMessageSearchIndex.java
```diff
@@ -132,7 +132,7 @@ protected Mono<Void> handleAdded(MailboxSession session, Mailbox mailbox, Added
 
     private Flux<MailboxMessage> retrieveMailboxMessages(MailboxSession session, Mailbox mailbox, MessageRange range) {
         return factory.getMessageMapper(session)
-            .findInMailboxReactive(mailbox, range, FetchType.FULL, UNLIMITED);
+            .findInMailboxReactive(mailbox, range, FetchType.HEADERS, UNLIMITED);
```
Arsnael, line 135: Wouldn't that make the search on the mail body unavailable though? I don't see why a client using James would find that acceptable?

chibenwa:
Unavailable on body  but you get all the other search properties.

This is SUPER handy as a "panic fix" or for people wanting to compromise on the cost of search.

(you can still search from, to cc, bcc, subject, dates, etc...)

Arsnael:
I get you can search on other header fields. I can admit maybe in some cases it could be wanted. But the non indexing of the body should not be enabled by default

chibenwa:
>  But the non indexing of the body should not be enabled by default

Of course.

I'm asking for an emergency safety trigger. 
Not a default value.
Like acl.enabled in casandra.properties.

mailbox/opensearch/src/test/java/org/apache/james/mailbox/opensearch/OpenSearchNoIndexBodyIntegrationTest.java
```diff
@@ -0,0 +1,197 @@
...
+    @Test
+    void searchingByBodyContentShouldNotReturnMessageWhenNoIndexBody() throws Exception {
+        addMessage(session, inboxPath).block();
+
+        awaitForOpenSearch(QueryBuilders.matchAll().build()._toQuery(), 1L);
+
+        SearchQuery searchQuery = SearchQuery.of(SearchQuery.bodyContains("Hello"));
+
+        assertThat(messageSearchIndex.search(session, mailbox, searchQuery).collectList().block())
+            .isEmpty();
```
Arsnael, line 159: Well fails:

```
java.lang.AssertionError: 
Expecting empty but was: [MessageUid{uid=1}]
```

Need to debug that, not sure where I messed up...

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/events/OpenSearchListeningMessageSearchIndex.java
```diff
@@ -296,7 +299,14 @@ private Mono<Void> handleMailboxEvent(Event event, MailboxSession session, Mailb
...
             .findMailboxById(mailboxId)
-            .flatMap(mailbox -> handleAdded(session, mailbox, addedEvent));
+            .flatMap(mailbox -> processAdded(session, mailbox, addedEvent));
+    }
+
+    private Mono<Void> processAdded(MailboxSession session, Mailbox mailbox, MailboxEvents.Added addedEvent) {
+        if (IndexBody.YES.equals(indexBody)) {
+            return handleAdded(session, mailbox, addedEvent, FetchType.FULL);
+        }
+        return handleAdded(session, mailbox, addedEvent, FetchType.HEADERS);
```
chibenwa, line 309: Only choose the FetchType amd avoid code duplication please

mailbox/opensearch/src/test/java/org/apache/james/mailbox/opensearch/OpenSearchNoIndexBodyIntegrationTest.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    @Test
+    void searchingAllShoulReturnMessageWhenNoIndexBody() throws Exception {
+        ComposedMessageId composedMessageId = addMessage(session, inboxPath).block();
+
+        awaitForOpenSearch(QueryBuilders.matchAll().build()._toQuery(), 1L);
+
+        SearchQuery searchQuery = SearchQuery.of(SearchQuery.bodyContains("Hello"));
```
chibenwa, line 168:
```suggestion
        SearchQuery searchQuery = SearchQuery.of(SearchQuery.all());
```

To match the test name?

src/site/xdoc/server/config-opensearch.xml
```diff
@@ -122,6 +122,8 @@
           <dd>When set to true, James will attempt to reindex from the indexed message when moved.
               If the message is not found, it will fall back to the old behavior (The message will be indexed from the blobStore source).
               Default to false.</dd>
+          <dt><strong>opensearch.indexBody</strong></dt>
+          <dd>Indicates if you wish to index body or not (default: true).</dd>
```
chibenwa, line 126:
```suggestion
          <dd>Indicates if you wish to index body or not (default: true). This can be used to decrease the performance cost associated with indexing.</dd>
```

server/apps/distributed-app/docs/modules/ROOT/pages/configure/opensearch.adoc
```diff
@@ -96,6 +96,9 @@ turning off headers indexing result in non-strict compliance with the IMAP / JMA
 | When set to true, James will attempt to reindex from the indexed message when moved.
 If the message is not found, it will fall back to the old behavior (The message will be indexed from the blobStore source)
 Default to false.
+
+| opensearch.indexBody
+| Indicates if you wish to index body or not (default: true).
```
chibenwa, line 101:
```suggestion
| Indicates if you wish to index body or not (default: true).  This can be used to decrease the performance cost associated with indexing.
```

mailbox/opensearch/src/test/java/org/apache/james/mailbox/opensearch/OpenSearchNoIndexBodyIntegrationTest.java
chibenwa: Memory backend is not compliant with FetchTyle and always return the full message. CF `InMemoryMessageMapper::findInMailbox`.

We should cary other the indexed body config deep into `MimePartParser::parse` and use `message.getHeaderContent()` instead of `message.getFullContent()` in order to make it work.

Arsnael (approved):

chibenwa:
https://issues.apache.org/jira/browse/JAMES-4010

PR #2022, opened 2024-02-18
JAMES-4003 [FIX] Avoid forwarding bounces as it might create infinite loops

chibenwa:
**Why** Destination service might consider Header from do not match sender envelop and reject with a bounce.

This creates an infinite loop.

=> This is **DIRTY FIXING**

Real fix: do like outlook...

- For local domains: like today.
- For remote domains: Wrap the forwarded message ina nested mime part, set subject to the value of the mail.

### DIFF
--- a/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
+++ b/server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
@@ -71,6 +71,7 @@ import com.google.common.collect.Sets;
 public class RecipientRewriteTableProcessor {
     private static final Logger LOGGER = LoggerFactory.getLogger(RecipientRewriteTableProcessor.class);
     private static final boolean REWRITE_SENDER_UPON_FORWARD = true;
+    private static final boolean FORWARD_AUTOMATED_EMAILS = true;
 
     private static class Decision {
         private final MailAddress originalAddress;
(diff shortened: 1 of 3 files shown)


Arsnael:
This test failed on the last two builds: https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2022/5/testReport/org.apache.james.transport.mailets/RemoteDeliveryErrorHandlingTest/remoteDeliveryShouldStoreTemporaryFailureAsPermanentWhenExceedsMaximumRetries_SMTPMessageSender__DockerMockSmtp_/

Issue? Flaky? Might need to double check

chibenwa:
I suspect the recent build change to not retrigger the test and report it as failed...

chibenwa:
Any thoughts on this work?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/RecipientRewriteTableProcessor.java
```diff
@@ -182,7 +186,8 @@ public RecipientRewriteTableProcessor(RecipientRewriteTable virtualTableStore, D
     }
 
     public RecipientRewriteTableProcessor(RecipientRewriteTable virtualTableStore, DomainList domainList, MailetContext mailetContext) {
-        this(virtualTableStore, domainList, mailetContext, new ProcessingState(Mail.ERROR), !REWRITE_SENDER_UPON_FORWARD);
+        this(virtualTableStore, domainList, mailetContext, new ProcessingState(Mail.ERROR), !REWRITE_SENDER_UPON_FORWARD,
+            false);
```
Arsnael, line 190:
```suggestion
            !FORWARD_AUTOMATED_EMAILS);
```

That or delete `FORWARD_AUTOMATED_EMAILS` as it does not seem used if not for this

chibenwa: Ok with the comment.

Please push your fixup ;-)

Arsnael (approved):

PR #2023, opened 2024-02-18
[FIX] Apply batch size to prefetch

chibenwa:
Default value 32 might be too much

### DIFF
--- a/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdMapper.java
+++ b/mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/mail/CassandraMessageIdMapper.java
@@ -121,10 +121,11 @@ public class CassandraMessageIdMapper implements MessageIdMapper {
 
     @Override
     public Flux<MailboxMessage> findReactive(Collection<MessageId> messageIds, FetchType fetchType) {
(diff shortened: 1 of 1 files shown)


quantranhong1999 (approved):

Arsnael:
I'm confused here... You applying the batchSize to prefetch saying 32 might be too much... that implies that configration has been changed for batchSizes, with the assumption the user running James put it lower than 32. However, default batch size value in the code is defined at 200 if conf is missing in batchsize.properties, and also default values in the sample conf files are defined like this:

```
fetch.metadata=200
fetch.headers=200
fetch.body=100
fetch.full=50
```

Wouldn't this change make things worse as is in the end?

chibenwa:
I agressively decreased this on CNB

```
fetch.metadata=1000

fetch.headers=30
fetch.body=1
fetch.full=1

# IMAP COPY command
copy=10
```

Arsnael:
> I agressively decreased this on CNB

James is an ASF project, not a Linagora/CNB customer project :)

I understand you likely did that on that platform, my question is what happens for other people using James with default batch values here in that case? Not sure it's beneficial in this state

Arsnael:
Misunderstood on my end, sorry. Alright then

Arsnael (approved):

PR #2034, opened 2024-02-21
JAMES-4006 Bouncer should pass DSN to bounce processor

ottoka:
See https://issues.apache.org/jira/browse/JAMES-4006
Just moved the "if (!mail.hasSender())" condition to the bounceWithMailetContext() method where it is actually needed.
Adjusted the respective test to expect the null sender mail being passed to the bounce processor.


chibenwa (approved):
+1 we do not risk a loop as DSNBounce do not send bounce if the mail has no sender

Arsnael (approved):

PR #2036, opened 2024-02-21
Fixed issue with incorrect JPA properties values

hantu85:
Due to changes introduced with "JAMES-2156 Add configuration option to enable attachment storage for JPA", values for keys `openjpa.Multithreaded` and `attachmentStorage.enabled` set in properties passed to `Persistence.createEntityManagerFactory()` method have incorrect values.

Due to usage of `toString()` method on `Optional<Boolean>` instead of calling it directly on `Boolean`, values are set as `Optional[true]` if they had value `true` causing it to evaluate inside persistence to `false`.

This change fixes this regression restoring ability to set  `openjpa.Multithreaded` and `attachmentStorage.enabled` to values other than `false`.


chibenwa (approved):

vttranlina (approved):

quantranhong1999 (approved):

Arsnael (approved):

Arsnael:
Thanks for the contribution @hantu85 !

PR #2038, opened 2024-02-22
[FIX] MailReceptionCheck do not send the mail before we actively list…

chibenwa:
…en to the results

**Why** : Sometime mail reception checks fails without reason

I suspect we do actually on busy envs receives the mail BEFORE we actively listen on the listener


chibenwa:
(caused some unecessary pod reboots and forced us to us less meaningfull livenessProbes in K8S)

Arsnael (approved):

quantranhong1999 (approved):

PR #2049, opened 2024-02-26
Revert git commit id maven plugin

Arsnael:
@jeantil I'm sorry to ask for this, but do you mind if we revert the upgrade of the git-commit-id-maven-plugin for now?

It seems tocreate issues with projects using James as a submodule (like TMail-backend) and makes the build impossible.

I opened an issue on the plugin backlog if you want to have a look for more context/details: https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/701

But for now this is a blocker for us unfortunately^^'

Mind that I still put back for `injectAllReactorProjects` to true after the revert, hoping it keeps similar gain in build performance :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3497,9 +3497,9 @@
                     </configuration>
                 </plugin>
                 <plugin>
-                    <groupId>io.github.git-commit-id</groupId>
-                    <artifactId>git-commit-id-maven-plugin</artifactId>
-                    <version>7.0.0</version>
+                    <groupId>pl.project13.maven</groupId>
+                    <artifactId>git-commit-id-plugin</artifactId>
+                    <version>4.9.10</version>
                 </plugin>
             </plugins>
         </pluginManagement>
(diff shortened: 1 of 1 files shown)


chibenwa:
Have we tried disactivating it in tmail build?

Arsnael:
> Have we tried disactivating it in tmail build?

I tried but wasn't successful. You are free to give it a shot though :)

chibenwa:
To be honnest I'd be more in favour of removing that plugin from global build and configuring it only in `server/apps` modules.

jeantil:
Hi,
I'm on limited connectivity this week.

I wasn't aware that the James project build used git submodules. There are
several limitations in the plugin with regard to submodules I'm not
surprised it fails.

You can revert or solve this differently if you want.

I' m interested in the reason for using git submodules if you have the time
:)

Jean



Le lun. 26 févr. 2024 à 08:28, Benoit TELLIER ***@***.***> a
écrit :

chibenwa:
> I wasn't aware that the James project build used git submodules.

Not in james but the use linagora uses James.

TLDR: it allows repackaging james, cross repository changes without the need of a fork.

chibenwa:
https://github.com/apache/james-project/pull/2063

Arsnael:
Thanks @jeantil for the suggestion, I think that might do it actually and it's rather clean. I'm in favor, thanks for making the PR too @chibenwa 

Closing this as a result.

quantranhong1999 (approved):

Arsnael:
@jeantil would you be ok if we maybe merge this first, open a ticket in parallel for thinking again on this, and then solve it on an other PR?

Also I did add back in this PR the param `injectAllReactorProjects` to true after the revert, as I think it was what gave the perf boost :)

If not it's ok, we will adapt in an other way, no big deal :)

chibenwa:
+1 to revert, keep injectAllReactorProjects and wait https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/701 for the version upgrade...

Arsnael:
Well about that... : https://github.com/git-commit-id/git-commit-id-maven-plugin/releases/tag/v8.0.0

Seems it's released with the fix, I will take the time to test around and see if it works. If it does we can just do the upgrade :)

Arsnael:
Closing this (again) in favor of https://github.com/apache/james-project/pull/2085

PR #2060, opened 2024-02-27
[FIX] Enforce a separate keyspace for blob store cache

chibenwa:
This was broken, likely upon the cassandra driver upgrade

### DIFF
--- a/server/container/guice/cassandra/src/main/java/org/apache/james/modules/mailbox/CassandraCacheSessionModule.java
+++ b/server/container/guice/cassandra/src/main/java/org/apache/james/modules/mailbox/CassandraCacheSessionModule.java
@@ -22,9 +22,8 @@ package org.apache.james.modules.mailbox;
 import java.util.Set;
 
 import org.apache.james.backends.cassandra.components.CassandraModule;
-import org.apache.james.backends.cassandra.init.KeyspaceFactory;
+import org.apache.james.backends.cassandra.init.ResilientClusterProvider;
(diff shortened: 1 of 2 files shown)


server/container/guice/cassandra/src/main/java/org/apache/james/modules/mailbox/CassandraCacheSessionModule.java
```diff
@@ -72,12 +71,8 @@ static class InitializedCacheCluster {
...
 
         @Inject
-        private InitializedCacheCluster(CqlSession cluster, ClusterConfiguration clusterConfiguration, KeyspacesConfiguration keyspacesConfiguration) {
-            this.cluster = cluster;
-
-            if (clusterConfiguration.shouldCreateKeyspace()) {
-                KeyspaceFactory.createKeyspace(keyspacesConfiguration.cacheKeyspaceConfiguration(), cluster).block();
-            }
+        private InitializedCacheCluster(ResilientClusterProvider sessionProvider, KeyspacesConfiguration keyspacesConfiguration) {
+            this.cluster = sessionProvider.get(keyspacesConfiguration.cacheKeyspaceConfiguration());
```
Arsnael, line 75: we don't create the keyspace anymore if it's necessary?

chibenwa: Session provider creates it implicitly?

Arsnael (approved):
Can we just remove the WIP part of the commit description?

PR #2063, opened 2024-02-27
[BUILD] Use git-commit-id-maven-plugin in ci-test

chibenwa:


Arsnael (approved):

quantranhong1999 (approved):

vttranlina (approved):

jeantil:
`ci-test` is only applied for test execution, neither build nor deploy
stage will execute the plugin.

I think we reuse the output of the previous stage to execute the next so it
might work and unblock you. I'll revisit this when I'm back from vacation

PR #2065, opened 2024-02-28
POC Make all queues on Rabbitmq quorum queue when quorum option is en…

Arsnael:
…abled

### DIFF
--- a/event-bus/distributed/src/main/java/org/apache/james/events/KeyReconnectionHandler.java
+++ b/event-bus/distributed/src/main/java/org/apache/james/events/KeyReconnectionHandler.java
@@ -19,7 +19,6 @@
 
 package org.apache.james.events;
 
-import static org.apache.james.backends.rabbitmq.Constants.ALLOW_QUORUM;
 import static org.apache.james.backends.rabbitmq.Constants.AUTO_DELETE;
 import static org.apache.james.backends.rabbitmq.Constants.DURABLE;
 import static org.apache.james.backends.rabbitmq.Constants.EXCLUSIVE;
@@ -56,7 +55,7 @@ public class KeyReconnectionHandler implements SimpleConnectionPool.Reconnection
     public Publisher<Void> handleReconnection(Connection connection) {
         return Mono.fromRunnable(() -> {
             try (Channel channel = connection.createChannel()) {
-                QueueArguments.Builder builder = configuration.workQueueArgumentsBuilder(!ALLOW_QUORUM);
+                QueueArguments.Builder builder = configuration.workQueueArgumentsBuilder();
                 configuration.getQueueTTL().ifPresent(builder::queueTTL);
                 channel.queueDeclare(namingStrategy.queueName(eventBusId).asString(), DURABLE, !EXCLUSIVE, AUTO_DELETE, builder.build());
             } catch (Exception e) {
--- a/event-bus/distributed/src/main/java/org/apache/james/events/KeyRegistrationHandler.java
+++ b/event-bus/distributed/src/main/java/org/apache/james/events/KeyRegistrationHandler.java
@@ -128,7 +128,7 @@ class KeyRegistrationHandler {
             QueueSpecification.queue(registrationQueue.asString())
                 .durable(configuration.isEventBusNotificationDurabilityEnabled())
                 .exclusive(!EXCLUSIVE)
-                .autoDelete(AUTO_DELETE)
+                .autoDelete(!AUTO_DELETE)
                 .arguments(builder.build()))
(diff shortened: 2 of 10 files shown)


event-bus/distributed/src/main/java/org/apache/james/events/KeyReconnectionHandler.java
```diff
@@ -56,7 +55,7 @@ public KeyReconnectionHandler(NamingStrategy namingStrategy, EventBusId eventBus
     public Publisher<Void> handleReconnection(Connection connection) {
         return Mono.fromRunnable(() -> {
             try (Channel channel = connection.createChannel()) {
-                QueueArguments.Builder builder = configuration.workQueueArgumentsBuilder(!ALLOW_QUORUM);
+                QueueArguments.Builder builder = configuration.workQueueArgumentsBuilder();
                 configuration.getQueueTTL().ifPresent(builder::queueTTL);
```
chibenwa, line 59: Are those options compatible?

quantranhong1999 (commented):
read it

vttranlina (commented):
read it

event-bus/distributed/src/main/java/org/apache/james/events/KeyReconnectionHandler.java
Arsnael: Seems not, good catch

event-bus/distributed/src/main/java/org/apache/james/events/KeyRegistrationHandler.java
```diff
@@ -128,7 +128,7 @@ private void declareQueue(Sender sender) {
             QueueSpecification.queue(registrationQueue.asString())
                 .durable(configuration.isEventBusNotificationDurabilityEnabled())
                 .exclusive(!EXCLUSIVE)
-                .autoDelete(AUTO_DELETE)
+                .autoDelete(!AUTO_DELETE)
```
quantranhong1999, line 131: the key queue seems volatile and can accept message lost to me -> Maybe we do not need to apply quorum queue on key queues?

After James restarts, a key queue would not be consumed anymore. So I think we should keep autoDelete if possible to avoid flushing RabbitMQ with unused key queues.

quantranhong1999: But thinking more about that, not applying quorum queue to the key queues could again break James if a RabbitMQ node is down...

quantranhong1999: Likely we can write a task to clean up inactive RabbitMQ queues (no consumers)

quantranhong1999: Hmm normally key queues should be cleaned up by James upon James shutdown: https://github.com/apache/james-project/blob/master/event-bus/distributed/src/main/java/org/apache/james/events/KeyRegistrationHandler.java#L140

event-bus/distributed/src/main/java/org/apache/james/events/KeyReconnectionHandler.java
quantranhong1999: Indeed TTL seems to work with quorum queue cf: https://www.rabbitmq.com/docs/quorum-queues#queue-and-per-message-ttl-since-rabbitmq-310

chibenwa:
> the key queue seems volatile and can accept message lost to me -> Maybe we do not need to apply quorum queue on key queues?
> After James restarts, a key queue would not be consumed anymore. So I think we should keep autoDelete if possible to avoid flushing RabbitMQ with unused key queues.

Do not Guess. Do not use the word maybe.

If needed experiment. How does this break things? Reject upon publish? Loose the notification? Are we able to recover connection loss afterward?

`RabbitMQClusterTest` might prove valuable...

> Hmm normally key queues should be cleaned up by James upon James shutdown

How about "not clean shutdown?"

> Likely we can write a task to clean up inactive RabbitMQ queues (no consumers)

In the best of the world I would prefer we do not come down to that...

Also keep in mind that RabbitMQ pubsub architecture (keys) could be replaced by redis. Do that work presently allows RabbitHA when using Redis on the pubsub infrastructure?

Arsnael:
Work moved here => https://github.com/apache/james-project/pull/2191

Closing this.

quantranhong1999:
> How about "not clean shutdown?"

Queue TTL still can apply on quorum queues -> dangling key queues would be cleared by RabbitMQ.

I just checked: we already apply the queue TTL on the ephemeral queues.

PR #2071, opened 2024-02-29
JAMES-4014 Cannot Email/set destroy with twice the same id

chibenwa:


mailbox/store/src/test/java/org/apache/james/mailbox/store/AbstractMessageIdManagerStorageTest.java
```diff
@@ -115,15 +112,6 @@ void deleteMessageShouldReturnNotFoundWhenMessageDoesNotExist() throws MailboxEx
             .isEqualTo(DeleteResult.notFound(messageId));
     }
 
-    @Disabled("Fails for cassandra app")
-    @Test
-    void deleteMessageShouldSupportDuplicatedEntries() throws MailboxException {
```
Arsnael: Why creating a test to prove there is an issue to then remove it along with the fix? Or did I misunderstand something here?

Arsnael (approved):

mailbox/store/src/test/java/org/apache/james/mailbox/store/AbstractMessageIdManagerStorageTest.java
chibenwa: Because the method signature changed we can no longer pass duplicates. Once modified the test no longer made sens as it exercised ImmutableSet. So I removed it.

I commited the test first to show I did TDD before fixing the issue so I do not get accused of not testing my stuff!

PR #2073, opened 2024-03-01
MIME4J-326 Adopt Mime4J 0.8.11-SNAPSHOT

chibenwa:


Arsnael (changes requested):
Thought should never use snapshots in production code? That could create issues if something unstable gets merged later in mime4j project no?

What's the rational behind this?

quantranhong1999:
> Thought should never use snapshots in production code? That could create issues if something unstable gets merged later in mime4j project no?

I have no idea about that. But in case we want to override the mime4j version just on tmail for example:
`mvn clean install -DskipTests -Dmaven.skip.doc=true -Dapache-mime4j.version=0.8.11-SNAPSHOT`

chibenwa:
> Thought should never use snapshots in production code? 

Well we are in for a release in MIME4J ^^

PR #2082, opened 2024-03-04
DRAFT JAMES-2586 Index for RLS column

hungphan227:


chibenwa (approved):

Arsnael (approved):

backends-common/postgres/src/main/java/org/apache/james/backends/postgres/PostgresTableManager.java
```diff
@@ -36,6 +36,36 @@
...
+                                            "    LOOP\n" +
+                                            "        SELECT pg_get_indexdef(indexrelid)\n" +
+                                            "            INTO index_definition\n" +
+                                            "            FROM pg_index\n" +
+                                            "            WHERE indexrelid = index_record.indexname::regclass;\n" +
+                                            "        column_list_string := substring(index_definition FROM '\\((.*)\\)');\n" +
+                                            "        column_list := string_to_array(column_list_string, ', ');\n" +
+                                            "        IF NOT column_list @> ARRAY[new_column_name] then\n" +
+                                            "            index_definition := regexp_replace(index_definition, '\\)$', ', ' || new_column_name || ')');\n" +
+                                            "            EXECUTE 'DROP INDEX IF EXISTS ' || index_record.indexname;\n" +
```
quantranhong1999, line 62: As discussed yesterday, I would prefer to control the indexes within our code at the index creation phase rather than control it at Postgres DB (without much app context): creating the indexes (at James startup) then droping them then creating new altered indexes.

Hard to unit test the PL/SQL function BTW. I would like to see at least some tests (the cases I concern about) e.g.:
- Should not DROP primary key and unique indexes
- Should not DROP correct RLS indexes and recreate them again upon every James startup:
  - When data is big in Postgres. I believe recreating the indexes has a not cheap fee.
  - Scenario: If 1 node James restarting or being spawned, that would likely affect index(ex) which disrupts query performance of other James nodes

hungphan227: - Should not DROP primary key and unique indexes -----> I noticed that when Postgres create index for primary key or unique constraint, it always put 'key' at the end of index name. I can use this to ignore indexes for primary key and unique constraint.
- Should not DROP correct RLS indexes and recreate them again upon every James startup -------> In fact, recreating index only run once. Index is recreated only when it does not contain RLS column.

quantranhong1999: > t always put 'key' at the end of index name. I can use this to ignore indexes for primary key and unique constraint.

Edge case: nothing can prevent an application index could end with the "key" postfix -> then that index would not be altered to have the `domain` column.

But I won't be more picky.

Arsnael:
After discussions in https://github.com/linagora/james-project/issues/5054, it's been proven that this does not really give much benefits.

Closing it

PR #2084, opened 2024-03-04
[FIX] Calendar parsing: be less verbose

chibenwa:
- Avoid logging the whole ICS in INFO mode
 - Reduce the log level to info
 
 This avoids super verbose logs for things that anyway we cannot process...

### DIFF
--- a/mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
+++ b/mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
@@ -142,13 +142,17 @@ public class ICalendarParser extends GenericMailet {
             ByteArrayInputStream inputStream = new ByteArrayInputStream(icsContent);
             return Stream.of(Pair.of(key, builder.build(inputStream)));
         } catch (IOException e) {
-            if (LOGGER.isErrorEnabled()) {
-                LOGGER.error("Error while reading input: " + new String(icsContent, StandardCharsets.UTF_8), e);
+            if (LOGGER.isDebugEnabled()) {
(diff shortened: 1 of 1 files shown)


mailet/icalendar/src/main/java/org/apache/james/transport/mailets/ICalendarParser.java
```diff
@@ -142,13 +142,17 @@ private Stream<Pair<String, Calendar>> createCalendar(String key, byte[] icsCont
             ByteArrayInputStream inputStream = new ByteArrayInputStream(icsContent);
             return Stream.of(Pair.of(key, builder.build(inputStream)));
         } catch (IOException e) {
-            if (LOGGER.isErrorEnabled()) {
-                LOGGER.error("Error while reading input: " + new String(icsContent, StandardCharsets.UTF_8), e);
+            if (LOGGER.isDebugEnabled()) {
+                LOGGER.debug("Error while reading calendar input: " + new String(icsContent, StandardCharsets.UTF_8), e);
+            } else {
+                LOGGER.info("Error while reading calendar input", e);
```
Arsnael, line 148: Does that represent really that many logs? 


I find weird TBH having a log message starting by `Error...` and not being on error level but info instead?

chibenwa: Well ERROR logs means "wo po po something CRITICAL happened and you as an admin need to get an urgent look"

Benign failed parsing clearly do not qualify as such.

Arsnael (approved):

quantranhong1999 (approved):

PR #2072, opened 2024-03-01
JAMES-2586 Fix PostgresAuthenticationTest

hungphan227:

PR #2086, opened 2024-03-05
Update JamesServerExtension

hungphan227:
This PR is used to test effect of changes in postgresql branch (https://github.com/apache/james-project/pull/2072) on master branch

### DIFF
--- a/server/container/guice/common/src/test/java/org/apache/james/JamesServerExtension.java
+++ b/server/container/guice/common/src/test/java/org/apache/james/JamesServerExtension.java
@@ -56,11 +56,22 @@ public class JamesServerExtension implements BeforeAllCallback, BeforeEachCallba
 
     public enum Lifecycle {
         // Restarts the server for each class, including nested classes
-        PER_CLASS(JamesServerExtension::start, (extension, context) -> { }, (extension, context) -> { }, JamesServerExtension::stop),
+        PER_CLASS(
+            (extension, context) -> {
+                extension.setup(context);
+                extension.start(context);
+            },
+            (extension, context) -> { },
+            (extension, context) -> { },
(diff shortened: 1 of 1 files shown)


chibenwa (commented):
Can you please better explain why this change is needed?

hungphan227:
> Can you please better explain why this change is needed?

The old code of JamesServerExtension:

```
PER_ENCLOSING_CLASS(
            (extension, context) -> {
                if (!isNested(context)) {
                    extension.start(context);
                }
            },
            (extension, context) -> { },
            (extension, context) -> { },
            (extension, context) -> {
                if (!isNested(context)) {
                    extension.stop(context);
                }
            }),
```

With the old code, when running test class that have nested classes, the beforeAll and afterAll method of all registrableExtensions run several times (corresponding to number of nested classes) during the test class execution while beforeEach and afterEach method of registrableExtensions only run once at the beginning and the end -----> wrong order


With the new code:
```
PER_ENCLOSING_CLASS(
            (extension, context) -> {
                if (!isNested(context)) {
                    extension.setup(context);
                    extension.start(context);
                }
            },
            (extension, context) -> { },
            (extension, context) -> { },
            (extension, context) -> {
                if (!isNested(context)) {
                    extension.stop(context);
                    extension.tearDown(context);
                }
            }),
```
when running test class that have nested classes, beforeAll, afterAll, beforeEach and afterEach methods of all registrableExtensions only run once with the right order beforeAll -> beforeEach -> afterEach -> afterAll

In addition, I believe the new code is easier to understand than the old code. The old code really confused me.

chibenwa (approved):

quantranhong1999 (approved):

PR #2095, opened 2024-03-07
Add tmpf to postgres docker container test

Arsnael:


Arsnael:
What I noticed, last logs of all hanging forever builds finish by something like this: 

```
08:19:30.425 [INFO ] o.a.j.b.p.DockerPostgresSingleton - 2024-03-06 08:19:30.425 UTC [62] LOG:  checkpoint starting: time
08:23:29.249 [INFO ] o.a.j.b.p.DockerPostgresSingleton - 2024-03-06 08:23:29.249 UTC [62] LOG:  checkpoint complete: wrote 2391 buffers (14.6%); 0 WAL file(s) added, 0 removed, 3 recycled; write=238.594 s, sync=0.002 s, total=238.824 s; sync files=0, longest=0.000 s, average=0.000 s; distance=49321 kB, estimate=49321 kB; lsn=0/493D4F8, redo lsn=0/493D4C0
```
=> https://www.postgresql.org/docs/current/wal-configuration.html

Honestly, in the build, lots of those checkpoint happen when sutting down the machine and it's fine. The problem is when it kicks before shutdown cause the `checkpoint_timeout` expired (5 minutes by default...) I checked the times between this one and the previous shutdown checkpoint... 5 minutes.

It seems somehow for some reason that this checkpoint time is making our build hang forever. It seems happening sometimes (but not all the time) in mpt imap tests. I think we play on a very narrow limit here.

So i'm taking the bet of increasing this timeout to 10 minutes instead, let's see.

Also while I'm at it, I added a tmpfs to it...

Arsnael:
Still occurred at 10 minutes interval... Not sure to understand what's the difference with the main branch, as it seems occuring in the mpt imap tests anyways and does not even reach the integration tests... I put 1 hour, we don't need those to kick in anyways on PG for testing purposes

Arsnael:
Checkpoint has nothing to do with it in fact... what's hanging the mpt tests is on the search, the `Optimize AttachmentLoader` commit is faulty, could reproduce locally on PostgresUidSearchOnIndexTest for example.

Will investigate

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2095/3/testReport/

I can reproduce two of those 3 locally (fails everytime):

- PostgresFastViewProjectionHealthCheckIntegrationTest
- PostgresMailboxSetMethodTest

Those fixes have not been well tested I'm afraid...

Arsnael:
Just reducing to the tmpf addition, which is still nice to have I think

chibenwa (approved):

PR #2099, opened 2024-03-10
[FIX] Ship mail over web webadmin routes into more apps

chibenwa:
Reported from the matric channel by البهي

@muhammadbahey_twitter:gitter.im

> I tried to james:jpa-3.8.0 and tried to call curl -XPOST /mail-transfer-service but it gives 404 NOT found


quantranhong1999 (approved):
Do we need to backport this for 3.8.x btw?

Arsnael (approved):

vttranlina (approved):

PR #2102, opened 2024-03-10
[ENHANCEMENT] Decrease log level for "NotSslRecordException"

chibenwa:
This pollutes the logs.

ERROR logs are for really bad thing needing immediate admin attention, ideally triggering an alarm.

Those are not to be triggered when a guy try telnet on a SMTPS endpoint.

Reducing the log level to info would help making james easier to monitor


quantranhong1999 (approved):

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapChannelUpstreamHandler.java
```diff
@@ -30,6 +30,7 @@
 import java.util.Optional;
 import java.util.Set;
 
+import io.netty.handler.ssl.NotSslRecordException;
```
quantranhong1999, line 33: Checkstyle: Wrong order for 'javax.net.ssl.SSLHandshakeException' import.

I could not make a quick suggestion, sorry.

Arsnael (approved):
Except for the checkstyle to fix, as commented above, lgtm

PR #2106, opened 2024-03-12
[REFACTORING] Run Intellij unused inspection

chibenwa:
Though need to be carefull not to remove code relied upon by jackson, JMX or other runtime calls. This makes the inspection tricky to run...


chibenwa:
I won't continue this...

PR #2131, opened 2024-03-18
[ENHANCEMENT] Workqueue for the deleted message vault

chibenwa:
Deleted message vault copy can be long, doing it
asynchronously would make the feature more reliable.


mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/DeleteMessageListener.java
```diff
@@ -86,9 +89,71 @@ public static class DeleteMessageListenerGroup extends Group {
 
     }
 
+    public static class DelectedMessageCopyCommand {
```
quantranhong1999, line 92:
```suggestion
    public static class DeletedMessageCopyCommand {
```

chibenwa:
Main issue with this design is that it requires "dedup" to work, as otherwise the blob payload may not be accessible.

Reflact this into the module chooser?

chibenwa:
Squashed, added documentation.

Arsnael (approved):

PR #2137, opened 2024-03-20
POC IMAP FETCH with large body should not rely on in-memory buffering

chibenwa:
- [x] Runned the Distributed AES tests successfully, S3 too

Real question on non-aes tests and I think `Flux<ButeBuffer> -> InputStream` needs to be improved in order not to be blocking

### DIFF
--- a/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/impl/SimpleMailboxMessage.java
+++ b/mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/impl/SimpleMailboxMessage.java
@@ -20,6 +20,7 @@
 package org.apache.james.mailbox.store.mail.model.impl;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
@@ -28,6 +29,7 @@ import java.util.Optional;
 import jakarta.mail.Flags;
 
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.NotImplementedException;
 import org.apache.james.mailbox.MessageUid;
 import org.apache.james.mailbox.ModSeq;
 import org.apache.james.mailbox.exception.MailboxException;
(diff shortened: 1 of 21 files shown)


mailbox/store/src/main/java/org/apache/james/mailbox/store/mail/model/impl/SimpleMailboxMessage.java
```diff
@@ -241,7 +256,7 @@ public SimpleMailboxMessage(MessageId messageId, ThreadId threadId, Date interna
                                 Properties properties, MailboxId mailboxId) {
         this(messageId, threadId, internalDate, size, bodyStartOctet,
                 content, flags,
-                properties, mailboxId, ImmutableList.of(), EMPTY_SAVE_DATE);
+                properties, mailboxId, ImmutableList.of(), Mono.error(() -> new NotImplementedException("SimpleMailboxMessage constructor")), EMPTY_SAVE_DATE);
```
chibenwa, line 259: OUPS

chibenwa:
image

As expected, the ChunckStream is blocking on the eventLoop...

chibenwa:
https://github.com/apache/james-project/pull/2138 is NOT a solution. 

ChunkStream is necessarily blocking on slow data supplier cf https://github.com/netty/netty/issues/2984

I will be closing this as such.

So far I see 2 possible solution:

 - Switch to a distinct EventExecutor for some parts of the Netty pipeline. Early tries shows at least fine tunning would be needed. Benefit: limited refactoring needs. Limits: realies on blocking abstractions, costs of switching to a distinct thread.

 - Carry other the `Flux<ByteBuffer>` and write it in a backpressure aware fashion. Fully compatible with reactive code. Requires refactoring of BlobStore, Mailbox and IMAP. More importantly the AES blob store would be non complying as it relies on `InputStream` absctraction...

chibenwa:
Blocking reads is a major issue I do not feel confident solving for the AES blobstore

(which I **need** to do)

See https://github.com/apache/james-project/pull/2149 as an alternative for reasonable memory usage on IMAP FETCH

PR #2138, opened 2024-03-20
POC Potentially non blocking Flux<ByteBuffer> => InputStream implementation

chibenwa:
The ChunkStream Netty abstraction makes it clear it needs a reliable implementation of `available()` method which is not the case today: always return 0 and needlessly block most of the time.

This attempt tries to enhance the overall working of this class.

Note: Unit test passes but the Distributed integration tests randomly fails. 5% of the time part of the message is missing causing failures. I suspect a data race between the subscription and consumption which happens on distinct threads.


chibenwa:
Closing 

 - It does not make https://github.com/apache/james-project/pull/2137 much better (ChunkedStream is still blocking)
 - As explained above there is some subtle data races making this failing integration tests

PR #2145, opened 2024-03-21
[Netty] Size estimator for ChunkedStream

quantranhong1999:
Rationale: `DefaultMessageSizeEstimator` does not support estimating size for `ChunkedStream` (the wrapper type JAMES IMAP is using to wrap ByteBuf over Netty channel), therefore Netty can not know if it should delay writing this message yet (e.g. IMAP server memory is full and the message size is big then Netty would delay the write).

Implementing a `MessageSizeEstimator` support `ChunkedStream` would help.

Before: the back-pressure is triggered at minimum (1-2 times over transferring 500 **big** messages)
cf: [without chunked stream size estimator.txt](https://github.com/apache/james-project/files/14692758/without.chunked.stream.size.estimator.txt)

After: the back-pressure is triggered more often (over 20 times over transferring 500 **big** messages)
cf: [with chunked stream size estimator.txt](https://github.com/apache/james-project/files/14692764/with.chunked.stream.size.estimator.txt)

cc @vttranlina @chibenwa

### DIFF
--- a/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ChunkedStreamWithSize.java
+++ b/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ChunkedStreamWithSize.java
@@ -1,0 +1,150 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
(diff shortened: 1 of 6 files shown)


chibenwa:
Cool.

With this we could potentially increase watermarks without risking things to blow up...

Arsnael (commented):
Read it

quantranhong1999:
> With this we could potentially increase watermarks without risking things to blow up...

Likely yes.
I tested with a lower `WriteBufferWaterMark` => more back-pressure triggered, while a higher `WriteBufferWaterMark` => less back-pressure triggered.

Selecting a big enough `WriteBufferWaterMark` likely would avoid being throttled too soon.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ChunkedStreamWithSize.java
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    @Override
+    public long progress() {
+        return offset;
+    }
+
+    public int getChunkSize() {
+        return chunkSize;
+    }
```
chibenwa, line 149: Well James do actually chucks some kinda transformations of "ByteArrayInputStream".

So the chuck size is NOT representative of overall consummed memory by the write task.

chibenwa:
See https://github.com/apache/james-project/pull/2149

Instead of inferring the size of the stream, which needs knowledge on the fact it is memory based is infered.

Preserving the knowledge that it is memory backed all along the chain, along with a tiny fix, enables accurate back pressure.

PR #2150, opened 2024-03-23
JAMES-2586 Unbind AttachmentBlobResolver

hungphan227:
Unbind unused AttachmentBlobResolver

### DIFF
--- a/server/container/guice/protocols/jmap/src/main/java/org/apache/james/jmap/rfc8621/RFC8621MethodsModule.java
+++ b/server/container/guice/protocols/jmap/src/main/java/org/apache/james/jmap/rfc8621/RFC8621MethodsModule.java
@@ -86,7 +86,6 @@ import org.apache.james.jmap.method.ZoneIdProvider;
 import org.apache.james.jmap.pushsubscription.DefaultWebPushClient;
 import org.apache.james.jmap.pushsubscription.PushClientConfiguration;
 import org.apache.james.jmap.pushsubscription.WebPushClient;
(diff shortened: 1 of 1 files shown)


server/container/guice/protocols/jmap/src/main/java/org/apache/james/jmap/rfc8621/RFC8621MethodsModule.java
```diff
@@ -198,7 +197,6 @@ protected void configure() {
         blobResolverMultibinder.addBinding().to(MessageBlobResolver.class);
         blobResolverMultibinder.addBinding().to(UploadResolver.class);
         blobResolverMultibinder.addBinding().to(MessagePartBlobResolver.class);
-        blobResolverMultibinder.addBinding().to(AttachmentBlobResolver.class);
```
quantranhong1999, line 201: Q: if it is unused, why dont we remove the class code also?

hungphan227: Asked:
anyone knows if AttachmentBlobResolver is still in use or not? It seems that tests still pass without it

Benoit:
Not used.
But keep it arround.
25% salary raise to the genius that would make attachment download use it in JMAP RFC-8621

50% salary raise if temporal coupling is avoided and the solution not too hacky

Arsnael (approved):

vttranlina (approved):

quantranhong1999 (approved):

PR #2152, opened 2024-03-25
[FIX] Solve weave/rest-smtp-sink: Docker image manifest v2 schema 1 deprecation issue

quantranhong1999:
cf: https://docs.docker.com/engine/deprecated/#pushing-and-pulling-with-image-manifest-v2-schema-1

Solution: https://hub.docker.com/r/quanth99/rest-smtp-sink

### DIFF
--- a/server/testing/src/main/java/org/apache/james/util/docker/Images.java
+++ b/server/testing/src/main/java/org/apache/james/util/docker/Images.java
@@ -20,7 +20,7 @@
 package org.apache.james.util.docker;
 
 public interface Images {
-    String FAKE_SMTP = "quanth99/rest-smtp-sink:1.0";
+    String FAKE_SMTP = "quanth99/rest-smtp-sink:1.0"; // Original Dockerfile: https://github.com/ambled/rest-smtp-sink/blob/master/Dockerfile
     String RABBITMQ = "rabbitmq:3.12.1-management";
     String ELASTICSEARCH_2 = "elasticsearch:2.4.6";
     String ELASTICSEARCH_6 = "docker.elastic.co/elasticsearch/elasticsearch:6.3.2";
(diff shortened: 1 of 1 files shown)


Arsnael (approved):
Not optimal I think but ok. Good catch and thanks for this fix :)

server/testing/src/main/java/org/apache/james/util/docker/Images.java
```diff
@@ -20,7 +20,7 @@
 package org.apache.james.util.docker;
 
 public interface Images {
-    String FAKE_SMTP = "weave/rest-smtp-sink:latest";
+    String FAKE_SMTP = "quanth99/rest-smtp-sink:1.0";
```
chibenwa, line 23: If you can comment the URL to the Dockerfile project on github then that's perfect!

quantranhong1999: done

server/testing/src/main/java/org/apache/james/util/docker/Images.java
```diff
@@ -20,7 +20,7 @@
 package org.apache.james.util.docker;
 
 public interface Images {
-    String FAKE_SMTP = "quanth99/rest-smtp-sink:1.0";
+    String FAKE_SMTP = "quanth99/rest-smtp-sink:1.0"; // Original Dockerfile: https://github.com/ambled/rest-smtp-sink/blob/master/Dockerfile
```
chibenwa, line 23: No I mean the docker file of `quanth99/rest-smtp-sink:1.0`

quantranhong1999: That one I documented in https://hub.docker.com/r/quanth99/rest-smtp-sink docs

chibenwa (approved):

PR #2158, opened 2024-03-26
[FIX] IMAP STATUS should better handle MailboxNotFoundException

chibenwa:
Some MUA requests mailboxes that no longer exist nor are subscribed.

 - Explicit IMAP message
 - Stacktrace is not necessary in the error message
 - INFO log


chibenwa:
(doubled checked against cyrus for the actual behaviour)

Arsnael (approved):

protocols/imap/src/main/java/org/apache/james/imap/processor/StatusProcessor.java
```diff
@@ -104,8 +104,8 @@ protected Mono<Void> processRequestReactive(StatusRequest request, ImapSession s
             .then(unsolicitedResponses(session, responder, false))
             .then(Mono.fromRunnable(() -> okComplete(request, responder)))
             .onErrorResume(MailboxException.class, e -> {
```
quantranhong1999, line 106: Should this be explicitly `MailboxNotFoundException`?

I see lots of sub classes of `MailboxException`. For the rest, I think we still can fallback to `STATUS_FAILED`.

chibenwa: Very right.

Sorry I have been too fast.

quantranhong1999 (approved):

Arsnael (approved):

PR #2179, opened 2024-04-04
JAMES-2586 PopulateEmailQueryViewTask should not hang for postgres-app

quantranhong1999:
Root cause: likely a jooq reactive bug. Likely the query that lists a big amount of users somehow blocks the jooq thread.

Mitigate fix: `collectList` to await the list users query to finish before executing other queries...

A similar fix: https://github.com/apache/james-project/commit/03e4b363409adf197ae3af667996b537a69e934f

### DIFF
--- a/server/data/data-postgres/src/main/java/org/apache/james/user/postgres/PostgresUsersDAO.java
+++ b/server/data/data-postgres/src/main/java/org/apache/james/user/postgres/PostgresUsersDAO.java
@@ -32,6 +32,7 @@ import static org.jooq.impl.DSL.count;
 
 import java.util.Iterator;
 import java.util.Optional;
+import java.util.function.Function;
 
 import javax.inject.Inject;
 import javax.inject.Named;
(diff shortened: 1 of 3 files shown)


chibenwa (commented):
Question:

Does adding `.publishOn(Schedulers.parrallel())` in `PG UsersRepository::listReactive` fixes the issue?

quantranhong1999:
> Does adding .publishOn(Schedulers.parrallel()) in PG UsersRepository::listReactive fixes the issue?

It does not.

quantranhong1999:
Rethink about it, I think putting `.collectList` in `PG UsersRepository::listReactive` may be a more common fix.

chibenwa:
> Rethink about it, I think putting .collectList in PG UsersRepository::listReactive may be a more common fix.

Agreed. THis avoids leaking this everywhere.

My 2 cents is try to reproduce the bug into PGUsersRepositoryTest too.

That way we would limit coupling and ease testing possible alternative solutions...

Arsnael (approved):

vttranlina (approved):

hungphan227 (approved):

quantranhong1999:
> My 2 cents is try to reproduce the bug into PGUsersRepositoryTest too.

I succeed in reproducing that indeed.

chibenwa:
If I put, say, 5 users, do I actually encounter the bug?

quantranhong1999:
> If I put, say, 5 users, do I actually encounter the bug?

It does not happen with a small amount of users. 
BTW likely this is not paging related bug as we are not using it.  
`One common way to implement pagination in SQL is to use the LIMIT and OFFSET clauses in a SELECT statement.`

chibenwa:
No I did not meant paging as in a UI

I meant paging as in a SQL driver. For instance Cassandra driver fetch a first page of 5000 items then when necessary to provide you some more results it fetches some more.

PR #2181, opened 2024-04-04
JAMES-3946 DropLists

Maxxx873:

### DIFF
--- a/server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
+++ b/server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
@@ -102,8 +102,12 @@ public class DropListEntry {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
         DropListEntry that = (DropListEntry) o;
         return ownerScope == that.ownerScope &&
             Objects.equals(owner, that.owner) &&
--- a/server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
+++ b/server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
@@ -18,9 +18,13 @@
  ****************************************************************/
 package org.apache.james.droplists.api;
 
+import static org.apache.james.core.Domain.MAXIMUM_DOMAIN_LENGTH;
+import static org.apache.james.droplists.api.OwnerScope.GLOBAL;
+
 import java.util.Objects;
 import java.util.Optional;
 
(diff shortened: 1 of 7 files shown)


server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
```diff
@@ -0,0 +1,122 @@
...
+        return deniedEntityType;
+    }
+
+    public Optional<String> getDeniedEntity() {
+        return deniedEntity;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
```
chibenwa, line 105: Please write a POJO test with EqualsVelifier. I bet this POJO is non-compliant...

server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
```diff
@@ -0,0 +1,122 @@
...
+
+        public DropListEntry build() {
+            return new DropListEntry(ownerScope, owner, deniedEntityType, deniedEntity);
+        }
+    }
+
+    private Optional<OwnerScope> ownerScope;
+    private Optional<String> owner;
+    private Optional<String> deniedEntityType;
+    private Optional<String> deniedEntity;
```
chibenwa, line 78: Aren't those all mandatory?

Can we really have a drop list entry without a target entity?

server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        public DropListEntry build() {
+            return new DropListEntry(ownerScope, owner, deniedEntityType, deniedEntity);
+        }
+    }
+
+    private Optional<OwnerScope> ownerScope;
+    private Optional<String> owner;
+    private Optional<String> deniedEntityType;
```
chibenwa, line 77: We shall have an enum for deniedEntityType; IMO

chibenwa (commented):
Thanks!

As this only is an API, and as we will need storage implementations, webadmin utilities and mailet tooling to make sense of this, I propose not to merge this work  too early.

Maybe we can set up a feature branch?

Also, I have a couple of cosmetic remarks.

Maxxx873:
> Thanks!
> 
> As this only is an API, and as we will need storage implementations, webadmin utilities and mailet tooling to make sense of this, I propose not to merge this work too early.
> 
> Maybe we can set up a feature branch?
> 
> Also, I have a couple of cosmetic remarks.

Thanks for the helpful remarks! Yes, I think when you create a new feature branch I will edit the PR.

chibenwa:
Here you are: https://github.com/apache/james-project/tree/droplist

Arsnael (approved):

chibenwa (approved):

Arsnael:
Thanks for the api contribution for this :) I merge this in the feature branch

PR #2185, opened 2024-04-05
JAMES-2586 Apply reactor timeout for jOOQ

quantranhong1999:
Allow James to recover upon jOOQ reactive hanging bug. Also log error so we can monitor and fix the potential issues.

### DIFF
--- a/backends-common/postgres/src/main/java/org/apache/james/backends/postgres/utils/PostgresExecutor.java
+++ b/backends-common/postgres/src/main/java/org/apache/james/backends/postgres/utils/PostgresExecutor.java
@@ -24,11 +24,13 @@ import static org.jooq.impl.DSL.field;
 
 import java.time.Duration;
 import java.util.Optional;
+import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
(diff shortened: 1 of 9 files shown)


backends-common/postgres/src/main/java/org/apache/james/backends/postgres/utils/PostgresExecutor.java
```diff
@@ -55,6 +58,9 @@ public class PostgresExecutor {
     public static final String NON_RLS_INJECT = "non_rls";
     public static final int MAX_RETRY_ATTEMPTS = 5;
     public static final Duration MIN_BACKOFF = Duration.ofMillis(1);
+    private static final Logger LOGGER = LoggerFactory.getLogger(PostgresExecutor.class);
+    private static final Duration JOOQ_REACTIVE_TIMEOUT = Duration.ofSeconds(10);
```
chibenwa, line 62: Please make this query timeout configurable

Arsnael (commented):
Read it nothing to add

Arsnael (approved):

Arsnael (approved):

PR #2189, opened 2024-04-07
[FIX] JWT should not attempt to unzip data by default

chibenwa:
jjwt support zip by default yet uncompression of untrusted data can result in excessive memory/cpu usage for special crafted malicious payload.

### DIFF
--- a/server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
+++ b/server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
@@ -25,16 +25,36 @@ import java.util.Optional;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 
 import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.CompressionCodecResolver;
 import io.jsonwebtoken.Jws;
 import io.jsonwebtoken.JwtException;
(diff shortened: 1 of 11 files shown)


server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
```diff
@@ -25,16 +25,36 @@
...
 import io.jsonwebtoken.MalformedJwtException;
+import io.jsonwebtoken.impl.compression.DefaultCompressionCodecResolver;
 
 public class JwtTokenVerifier {
+    private static final CompressionCodecResolver DEFAULT_COMPRESSION_CODEC_RESOLVER = new DefaultCompressionCodecResolver();
+    private static final CompressionCodecResolver SECURE_COMPRESSION_CODEC_RESOLVER = header -> {
+        if (Optional.ofNullable(header.getCompressionAlgorithm()).isPresent()) {
+            throw new RuntimeException("Rejecting a ZIP JWT. Usage of ZIPPED JWT can result in " +
+                "excessive memory usage with malicious JWT tokens. To activate support for ZIPPed" +
+                "JWT please run James with the -Djames.jwt.zip.allow=true system property.");
```
ottoka, line 46: I suggest you put this in jvm.properties with the other tweakable odds and ends. Gives users a chance to find this later without checking the source code.

quantranhong1999 (approved):

quantranhong1999:
Nothing to add

ottoka (approved):

Arsnael (approved):

PR #2197, opened 2024-04-11
JAMES-3946 add a memory implementation of the DropList

Maxxx873:

### DIFF
--- a/server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
+++ b/server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
@@ -24,6 +24,11 @@ import static org.apache.james.droplists.api.OwnerScope.GLOBAL;
 import java.util.Objects;
 import java.util.Optional;
 
+import jakarta.mail.internet.AddressException;
+
+import org.apache.james.core.Domain;
+import org.apache.james.core.MailAddress;
+
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 
@@ -62,7 +67,7 @@ public class DropListEntry {
             return this;
         }
 
-        public DropListEntry build() {
+        public DropListEntry build() throws AddressException {
             OwnerScope scope = ownerScope.orElse(GLOBAL);
             Preconditions.checkArgument(owner != null && !owner.trim().isBlank(), "owner must not be null, empty, or blank");
             Preconditions.checkArgument(owner.length() <= MAXIMUM_DOMAIN_LENGTH,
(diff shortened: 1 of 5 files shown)


server/data/data-api/src/main/java/org/apache/james/droplists/api/DropListEntry.java
```diff
@@ -71,6 +76,11 @@ public DropListEntry build() {
             Preconditions.checkArgument(deniedEntity != null && !deniedEntity.isBlank(), "`deniedEntity` must not be null, empty, or blank");
             Preconditions.checkArgument(deniedEntity.length() <= MAXIMUM_DOMAIN_LENGTH,
                 "deniedEntity length should not be longer than %s characters", MAXIMUM_DOMAIN_LENGTH);
+            if (deniedEntityType.equals(DeniedEntityType.DOMAIN)) {
+                deniedEntity = Domain.of(deniedEntity).asString();
+            } else {
+                deniedEntity = new MailAddress(deniedEntity).toString();
+            }
```
chibenwa, line 83: Generally speaking we can avoid variable reallocation by method extraction.
This is more elegant and less complex.

chibenwa (commented):
Hello,

Thanks to put this together!

While the datastructure for the memory structure is not optimal (`O(N)` scans)
And while we could easily do better
(one multimap for global scope, one table for domain scope, one table for user scope) this does not matter much performance wise in a testing environment.

Other remarks however likely deserve attention.

Maxxx873:
> Hello,
> 
> Thanks to put this together!
> 
> While the datastructure for the memory structure is not optimal (`O(N)` scans) And while we could easily do better (one multimap for global scope, one table for domain scope, one table for user scope) this does not matter much performance wise in a testing environment.
> 
> Other remarks however likely deserve attention.

Thanks for the excellent code review! I will try to correct everything that was pointed out to me in the remarks.

server/data/data-api/src/test/java/org/apache/james/droplists/api/DropListContract.java
```diff
@@ -0,0 +1,142 @@
...
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public interface DropListContract {
+
+    DropList dropList();
+
+    @Test
+    default void shouldAddEntry() throws AddressException {
+        DropListEntry dropListEntry = getDropListTestEntries().toList().getFirst();
```
chibenwa, line 45: That's quite complex for just getting one item.

Calling the builder here without hiding things behind a method call would seem more straight forward to me...

server/data/data-api/src/test/java/org/apache/james/droplists/api/DropListContract.java
```diff
@@ -0,0 +1,142 @@
...
+
+        Mono<Void> result = dropList().add(dropListEntry);
+
+        assertThat(Objects.requireNonNull(dropList().list(GLOBAL, dropListEntry.getOwner()).collectList().block()).size()).isEqualTo(1);
```
chibenwa, line 49: I think Objects.requireNonNull is overkill here.

 -> A NPE thrown would still be understandable for a dev debugging this
 -> Reading is harder and we do not have such a convention at all through the code base.
 
 Suggestion: relax your IDE warning or ignore those.

chibenwa (commented):
Looks good to me. Tiny cosmetic fixes.

Can you confirm the test suite establishes a complete test coverage?

Arsnael (commented):
Read it, nothing to add.

Thank you for the work, appreciated as always :)

Maxxx873:
> Read it, nothing to add.
> 
> Thank you for the work, appreciated as always :)

Thank you too, I'm really glad to be involved in this project and follow its development.

Maxxx873:
> Looks good to me. Tiny cosmetic fixes.
> 
> Can you confirm the test suite establishes a complete test coverage?

image

Arsnael (approved):

chibenwa (approved):

PR #2215, opened 2024-04-25
JAMES-2586 Remove mailbox index

Arsnael:
I don't see any sql select where we query on only username and namespace, there is always also the maibox name as well (cf `findMailboxByPath`, `findMailboxWithPathLike`, `hasChildren` methods in `PostgresMailboxDAO`)

Thus I conclude that this field was missing from the index?

### DIFF
--- a/mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/PostgresMailboxModule.java
+++ b/mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/PostgresMailboxModule.java
@@ -23,7 +23,6 @@ import static org.jooq.impl.SQLDataType.BIGINT;
 
 import java.util.UUID;
 
-import org.apache.james.backends.postgres.PostgresIndex;
 import org.apache.james.backends.postgres.PostgresModule;
 import org.apache.james.backends.postgres.PostgresTable;
 import org.jooq.Field;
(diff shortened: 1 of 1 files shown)


chibenwa (approved):

chibenwa:
Any way to use `EXPLAIN` utility for analysing `findMailboxByPath` and `findMailboxWithPathLike` and ensure the index is well used in both case.

Arsnael:
Didn't know EXPLAIN, you are right, it shows that we actually do not use the index atm.

EXPLAIN for findMailboxByPath: 

explain-mailbox-1

EXPLAIN with just username and namespace:

explain-mailbox-2

We can see that the request done by findMailboxByPath does not use the index, while I verified that the explain command really tells you that you are using the index when you do, like in the second example with only namespace and username, that the current index is only based on

When I deploy this fix I will verify again with explain that we hit this time the index for both methods

Arsnael:
Actually, wait... reading again... the first one uses the `mailbox_mailbox_name_user_name_mailbox_namespace_key` unique constraint defined in mailbox? 

=> https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS

according to the doc, an index is created for unique constraints. So technically we can deduce that wherever we declare in the code a unique constraint, we dont need an other index next to it?

Should we delete the index on mailbox then, as the unique key constraint on those 3 fields seems to do the job?

Arsnael:
Explain findMailboxWithPathLike 

explain-mailbox-3

Using unique constraint key too

vttranlina:
> Explain findMailboxWithPathLike
> 
> explain-mailbox-3
> 
> Using unique constraint key too

can you try query with `LIKE "prefix%"` ? 
(contain `%` character)

This index was created at pr: https://github.com/apache/james-project/pull/1984#issue-2121940886

Arsnael:
explain-mailbox-4

Fair enough, thanks for the explanation.

Forget this PR :)

PR #2216, opened 2024-04-25
JAMES-3946 add the IsInDropList matcher

Maxxx873:

### DIFF
--- a/server/mailet/mailets/src/main/java/org/apache/james/transport/matchers/IsInDropList.java
+++ b/server/mailet/mailets/src/main/java/org/apache/james/transport/matchers/IsInDropList.java
@@ -18,6 +18,12 @@
  ****************************************************************/
 package org.apache.james.transport.matchers;
 
+import static org.apache.james.droplists.api.DropList.Status.ALLOWED;
+import static org.apache.james.droplists.api.OwnerScope.DOMAIN;
+import static org.apache.james.droplists.api.OwnerScope.GLOBAL;
+import static org.apache.james.droplists.api.OwnerScope.USER;
+import static reactor.function.TupleUtils.function;
+
 import java.util.Collection;
 
 import jakarta.inject.Inject;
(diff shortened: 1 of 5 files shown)


server/mailet/mailets/src/main/java/org/apache/james/transport/matchers/IsInDropList.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private boolean isRecipientAllowed(Mail mail, MailAddress recipient) {
+        MailAddress sender = mail.getMaybeSender().get();
+        return isAllowed(OwnerScope.GLOBAL, recipient.asString(), sender) &&
+            isAllowed(OwnerScope.DOMAIN, recipient.getDomain().asString(), sender) &&
+            isAllowed(OwnerScope.USER, recipient.asString(), sender);
+    }
+
+    private boolean isAllowed(OwnerScope ownerScope, String owner, MailAddress sender) {
+        DropList.Status status = dropList.query(ownerScope, owner, sender).block();
```
chibenwa, line 67: Rather than calling 3 time .block we could combine those monos in order to call it olny once and paralelize db access?

server/mailet/mailets/src/test/java/org/apache/james/transport/matchers/IsInDropListTest.java
```diff
@@ -0,0 +1,134 @@
+package org.apache.james.transport.matchers;
```
chibenwa, line 1: You are missing the Apache V2 license

Maxxx873: In my opinion, such things should be checked automatically using the maven-checkstyle-plugin.
https://github.com/apache/james-project/pull/2218

Arsnael (approved):

chibenwa (approved):

PR #2220, opened 2024-04-29
[FIX] JMAP: Only log ERROR for `serverFail` errors

quantranhong1999:
IMO we should not log user input errors.

Or @chibenwa do you want to keep the `RequestTooLargeException` log for example?


chibenwa (changes requested):
Reviewing this I think all those user input related errors would deserved to be info logged, so for `RequestTooLargeException`  but also `IdentityIdNotFoundException`, `ChangeNotFoundException`, `MailboxNotFoundException`, `IllegalArgumentException`, `UnsupportedNestingException`, `UnsupportedFilterException`, `UnsupportedSortException`, `UnsupportedRequestParameterException`, `ForbiddenAccountManagementException`, `AccountNotFoundException`...

Loosing this valuable debugging exception feels like a waste while clearly ERROR logging is overkill...

quantranhong1999:
I reworked the changes.

Arsnael (approved):

vttranlina (commented):
read it

chibenwa (approved):

PR #2225, opened 2024-05-02
[METRICS] Add some useful S3 connection metrics

quantranhong1999:
Print some S3 client buillt-in metrics:
```log
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] ApiCall
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] ┌─────────────────────────────────────────┐
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] │ MarshallingDuration=PT0.000082267S      │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] │ RetryCount=0                            │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] │ ApiCallSuccessful=true                  │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] │ OperationName=DeleteBucket              │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] │ ApiCallDuration=PT0.011391168S          │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] │ CredentialsFetchDuration=PT0.000002273S │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] │ ServiceEndpoint=http://localhost:32814  │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] │ ServiceId=S3                            │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81] └─────────────────────────────────────────┘
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     ApiCallAttempt
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     ┌───────────────────────────────────────────┐
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     │ HttpStatusCode=204                        │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     │ BackoffDelayDuration=PT0S                 │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     │ SigningDuration=PT0.000143366S            │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     │ AwsRequestId=9198495c0d439b60cd0f         │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     │ ServiceCallDuration=PT0.010762143S        │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     │ AwsExtendedRequestId=9198495c0d439b60cd0f │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]     └───────────────────────────────────────────┘
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         HttpClient
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         ┌───────────────────────────────────────────┐
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         │ AvailableConcurrency=100                  │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         │ MaxConcurrency=100                        │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         │ LeasedConcurrency=0                       │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         │ ConcurrencyAcquireDuration=PT0.000391434S │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         │ PendingConcurrencyAcquires=0              │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         │ HttpClientName=NettyNio                   │
17:47:15.719 [INFO ] s.a.a.m.LoggingMetricPublisher - [94c9c81]         └───────────────────────────────────────────┘
```

[LeasedConcurrency](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/http/HttpMetric.html#LEASED_CONCURRENCY) -> likely the wanted `the count of S3 connection in use`.

TODO: extract some useful metrics into James metrics.

### DIFF
--- a/server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/JamesS3MetricPublisher.java
+++ b/server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/JamesS3MetricPublisher.java
@@ -1,0 +1,83 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.blob.objectstorage.aws;
+
+import static software.amazon.awssdk.core.metrics.CoreMetric.API_CALL_DURATION;
+import static software.amazon.awssdk.http.HttpMetric.AVAILABLE_CONCURRENCY;
+import static software.amazon.awssdk.http.HttpMetric.CONCURRENCY_ACQUIRE_DURATION;
+import static software.amazon.awssdk.http.HttpMetric.LEASED_CONCURRENCY;
+import static software.amazon.awssdk.http.HttpMetric.PENDING_CONCURRENCY_ACQUIRES;
+
+import java.time.Duration;
+
+import org.apache.james.metrics.api.GaugeRegistry;
+import org.apache.james.metrics.api.MetricFactory;
+import org.apache.james.metrics.api.TimeMetric;
+
+import software.amazon.awssdk.metrics.MetricCollection;
+import software.amazon.awssdk.metrics.MetricPublisher;
+
+public class JamesS3MetricPublisher implements MetricPublisher {
+    private final GaugeRegistry.SettableGauge<Integer> availableConcurrency; // The number of remaining concurrent requests that can be supported by the HTTP client without needing to establish another connection.
+    private final GaugeRegistry.SettableGauge<Integer> leasedConcurrency; // The number of request currently being executed by the HTTP client.
+    private final GaugeRegistry.SettableGauge<Integer> pendingConcurrencyAcquires; // The number of requests that are blocked, waiting for another TCP connection or a new stream to be available from the connection pool.
+    private final TimeMetric concurrencyAcquireDuration; // The time taken to acquire a channel from the connection pool.
+    private final TimeMetric apiCallDuration; // The total time taken to finish a request (inclusive of all retries).
+
+    public JamesS3MetricPublisher(MetricFactory metricFactory, GaugeRegistry gaugeRegistry) {
+        this.availableConcurrency = gaugeRegistry.settableGauge("s3_httpClient_availableConcurrency");
+        this.leasedConcurrency = gaugeRegistry.settableGauge("s3_httpClient_leasedConcurrency");
+        this.pendingConcurrencyAcquires = gaugeRegistry.settableGauge("s3_httpClient_pendingConcurrencyAcquires");
+        this.concurrencyAcquireDuration = metricFactory.timer("s3_httpClient_concurrencyAcquireDuration");
+        this.apiCallDuration = metricFactory.timer("s3_apiCall_apiCallDuration");
+    }
+
+    @Override
+    public void publish(MetricCollection s3ClientMetrics) {
+        extractS3ClientMetrics(s3ClientMetrics);
+    }
+
+    private void extractS3ClientMetrics(MetricCollection s3ClientMetrics) {
+        // Extract useful metrics from https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/metrics-list.html
+        s3ClientMetrics.stream().forEach(metricRecord -> {
+            if (metricRecord.metric().equals(AVAILABLE_CONCURRENCY)) {
+                availableConcurrency.setValue((Integer) metricRecord.value());
+            }
+            if (metricRecord.metric().equals(LEASED_CONCURRENCY)) {
+                leasedConcurrency.setValue((Integer) metricRecord.value());
+            }
+            if (metricRecord.metric().equals(PENDING_CONCURRENCY_ACQUIRES)) {
+                pendingConcurrencyAcquires.setValue((Integer) metricRecord.value());
+            }
+            if (metricRecord.metric().equals(CONCURRENCY_ACQUIRE_DURATION)) {
+                concurrencyAcquireDuration.record((Duration) metricRecord.value());
+            }
+            if (metricRecord.metric().equals(API_CALL_DURATION)) {
+                apiCallDuration.record((Duration) metricRecord.value());
+            }
+        });
+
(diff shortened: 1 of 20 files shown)


quantranhong1999:
Tested the metrics from sandbox env:
```
s3_httpClient_leasedConcurrency_total 4.0
s3_httpClient_pendingConcurrencyAcquires_total 0.0
s3_httpClient_maxConcurrency_total 100.0
s3_httpClient_availableConcurrency_total 7.0

s3_httpClient_concurrencyAcquireDuration{quantile="0.5",} 2.96959E-4
s3_httpClient_concurrencyAcquireDuration{quantile="0.75",} 6.26687E-4
s3_httpClient_concurrencyAcquireDuration{quantile="0.95",} 0.0033914870000000003
s3_httpClient_concurrencyAcquireDuration{quantile="0.98",} 0.010289151
s3_httpClient_concurrencyAcquireDuration{quantile="0.99",} 0.035651583
s3_httpClient_concurrencyAcquireDuration{quantile="0.999",} 0.073400319
s3_httpClient_concurrencyAcquireDuration_count 979064.0
```

metrics/metrics-dropwizard/src/main/java/org/apache/james/metrics/dropwizard/DropWizardMetric.java
```diff
@@ -57,6 +57,11 @@ public void remove(int value) {
         meter.mark(-1 * value);
     }
 
+    @Override
+    public void set(int newValue) {
+        meter.mark(- meter.getCount() + newValue);
```
chibenwa, line 62: This, is likely to be non atomic

chibenwa (commented):
Here we have a non othodox usage of metrics.

Had this been perf tested?

I would need also to see a flame graph...

chibenwa:
You are using Metric while what we truely want is a Gauge.

chibenwa:
```
package org.apache.james.metrics.api;

import java.util.concurrent.atomic.AtomicReference;

public class UpdatableGauge<T> implements Gauge<T> {
    private final AtomicReference<T> ref = new AtomicReference();
    
    public void set(T value) {
        ref.set(value);
    }
    
    @Override
    public T get() {
        return ref.get();
    }
}
```

+ record gauges ?

WDYT?

server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/S3BlobStoreDAO.java
```diff
@@ -159,6 +162,7 @@ public long size() {
             .endpointOverride(authConfiguration.getEndpoint())
             .region(configuration.getRegion().asAws())
             .serviceConfiguration(pathStyleAccess)
+            .overrideConfiguration(builder -> builder.addMetricPublisher(new JamesS3MetricPublisher(metricFactory)))
```
chibenwa, line 165: Configuration parameter (or system property) to enable/disable this just in case this is a bad idea.

quantranhong1999:
> record gauges ?
WDYT?

Let me try that.

> Here we have a non othodox usage of metrics.
Had this been perf tested?

From what I see in the SDK code, the SDK collects its metrics implicitly regardless of `MetricPublisher` usage. -> At least we don't need to care about the collection cost as it has been there the whole time.

But we may care to care about the `MetricPublisher`'s cost.

Following [SDK docs](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/metrics/MetricPublisher.html#publish(software.amazon.awssdk.metrics.MetricCollection)): `Implementations are strongly encouraged to complete any further aggregation and publishing of metrics in an asynchronous manner to avoid blocking the calling thread.`

`JamesS3MetricPublisher` records some metrics in RAM is not that much blocking IMO. Do you think it deserved to be async?

Anyway I can do some perf test to see more.

chibenwa:
`JamesS3MetricPublisher records some metrics in RAM is not that much blocking IMO. Do you think it deserved to be async?`

Pushing duration is OK, timeMetric does the aggregation with reservoirs.

Pushing counters (metrics) through gauge would also ensure one value is published every 30s.

We may want to filter metrics a bit more and remove uninteresting ones...

server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/JamesS3MetricPublisher.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public void publish(MetricCollection s3ClientMetrics) {
+        extractS3ClientMetrics(s3ClientMetrics);
+    }
+
+    private void extractS3ClientMetrics(MetricCollection s3ClientMetrics) {
+        // Extract useful metrics from https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/metrics-list.html
+        s3ClientMetrics.stream().forEach(metricRecord -> {
+            if (metricRecord.metric().equals(MAX_CONCURRENCY)) {
```
chibenwa, line 63: Not interesting as this is a conf parameter we set.

Arsnael (commented):
Read it nothing to add

quantranhong1999:
### Perf test
screencapture-file-home-hp-copySSH-s3-metrics-s3-metrics-1-root-upn-james-james-gatling-target-gatling-platformvalidationsimulation-20240506070043714-index-html-2024-05-06-15_20_02

A bit slower than TMail 0.10.0 (p99 was 800ms) (expected as `postgres-rls` is being provisioned during this performance test).
APPEND, STORE, FETCH headers, FETCH body is a bit slower too as a consequence.

### Flame graph
[s3-metrics-flame-graph.zip](https://github.com/apache/james-project/files/15218777/s3-metrics-flame-graph.zip)

`MetricCollector` -> 0.22% CPU time for the S3 client to collect the metrics.
`JamesS3MetricPublisher` -> 0.11% CPU time for James to collect the metrics from the S3 client and convert them to its metrics.

Conclusion: S3 metrics do have a bit of impact on the performance, but far from anything crazy.

-> I disabled S3 metrics by default. We could enable it when we suspect S3 issues.

chibenwa (approved):

chibenwa:
Given how useful this metric is, and given the low impact, I would actually turn this on by default.

quantranhong1999:
> Given how useful this metric is, and given the low impact, I would actually turn this on by default.

Done.

chibenwa (approved):

Arsnael (approved):

Arsnael:
Some Guice injection issues in test? https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2225/4/testReport/junit/org.apache.james/SMTPJamesServerTest/___/

quantranhong1999:
cc @Arsnael  can you merge this?

PR #2228, opened 2024-05-03
[ENHANCEMENT] Better reactify Identity methods

chibenwa:
The underlying RRT exposes iterables which are inherantly blocking.

`recipientRewriteTable.listSources(...)` actually is just a mere wrapper around a cassandra table and would be easy to reactify.

Then the overall algorithm could be reactified CF https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#expand-java.util.function.Function-

This is called upon sending email (likely to be blocking anyway) but also listing identities which is not an uncommon request.

Screenshot from 2024-05-02 10-49-14

### DIFF
--- a/server/container/util/src/main/java/org/apache/james/util/StreamUtils.java
+++ b/server/container/util/src/main/java/org/apache/james/util/StreamUtils.java
@@ -21,20 +21,11 @@ package org.apache.james.util;
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.List;
 import java.util.Optional;
-import java.util.Spliterator;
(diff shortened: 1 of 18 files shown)


Arsnael (approved):

Arsnael:
```
07:14:30,800 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.3.0:check (check-style) on project james-server-data-api: You have 2 Checkstyle violations. -> [Help 1]
```

quantranhong1999 (approved):

PR #2232, opened 2024-05-06
[ENHANCEMENT] Metrics for rabbitmq channel pool

chibenwa:
Will ease channel pool sizing as well as debugging of RabbitMQ channelPool related errors.


backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/ReactorRabbitMQChannelPool.java
```diff
@@ -697,6 +699,11 @@ public ReactorRabbitMQChannelPool(Mono<Connection> connectionMono, Configuration
             .then()
             .subscribeOn(Schedulers.boundedElastic()))
             .buildPool();
+
+        gaugeRegistry.register("rabbitmq.channels.aquired.size", () -> newPool.metrics().acquiredSize());
```
quantranhong1999, line 703:
```suggestion
        gaugeRegistry.register("rabbitmq.channels.acquired.size", () -> newPool.metrics().acquiredSize());
```

quantranhong1999 (approved):

Arsnael (approved):

PR #1789, opened 2023-11-10
JAMES-3955 Better handle and avoid RabbitMQ message processing timeouts

chibenwa:

PR #1778, opened 2023-11-03
JAMES-3955 Check for queues without consumer and resume consumption a…

chibenwa:
…s needed

PR #2236, opened 2024-05-07
JAMES-3955 RabbitMQ works Backports [3.7.x]

Arsnael:
Just https://github.com/apache/james-project/pull/1778 for now (was busy on other things in parallel)

Need as well to backport:
- https://github.com/apache/james-project/pull/1789
- https://github.com/apache/james-project/pull/1802

At least I want to see the CI and if tests are green on what I've been doing, as I had to adapt the code on some parts to fit the backport

### DIFF
--- a/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/QueueArguments.java
+++ b/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/QueueArguments.java
@@ -42,6 +42,11 @@ public class QueueArguments {
             return this;
         }
 
+        public Builder consumerTimeout(long consumerTimeoutInMillisecond) {
+            arguments.put("x-consumer-timeout", consumerTimeoutInMillisecond);
+            return this;
+        }
+
         public ImmutableMap<String, Object> build() {
             return arguments.build();
         }
--- a/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/RabbitMQConfiguration.java
+++ b/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/RabbitMQConfiguration.java
@@ -30,6 +30,8 @@ import java.io.File;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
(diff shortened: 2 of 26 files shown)


quantranhong1999 (approved):

server/apps/distributed-pop3-app/README.adoc
```diff
@@ -22,7 +22,7 @@ Third party compulsory dependencies:
 ----
 $ docker run -d --network james -p 9042:9042 --name=cassandra cassandra:3.11.10
 $ docker run -d --network james -p 9200:9200 --name=elasticsearch --env 'discovery.type=single-node' docker.elastic.co/elasticsearch/elasticsearch:7.10.2
-$ docker run -d --network james -p 5672:5672 -p 15672:15672 --name=rabbitmq rabbitmq:3.8.18-management
+$ docker run -d --network james -p 5672:5672 -p 15672:15672 --name=rabbitmq rabbitmq:3.12.1-management
```
chibenwa, line 25: That's really annoying on a maintenance branch. 

Please offer me a way to still use 3.8 and turn off via system properties whatever options was not there yet in 3.8...

(Full retro-compatibility should be achievable)

And document it fully in the upgrade instructions.

My expectations:

```
XYZ option used requires rabbit 3.12
However -DdisableXYZ=true JVM option will disable XYZ and allow you to still use 3.8
```

Arsnael: Technically it is? Just thought if we backport a feature where rabbitmq 3.12 is necessary if you want to use it, the code should be upgraded to use it as well at least. Does not force people to upgrade to it though, and the doc changes are actually quite clear on that.

It is still retro compatible (even in our current master branch). If you look the change in upgrade instructions in commit Increase consumer timeout foir TaskManagerWorkQueue, I think it's rather clear:

```
If you want the old behavior where consumer timeout defaults to 30 minutes, you can either not upgrade your RabbitMQ or set `task.queue.consumer.timeout=30minutes` in `rabbitmq.properties`.
```

If it's just a matter of reverting a few docker-compose files that nobody will look at though likely anyways then sure I can do it

chibenwa: `I want the old behaviour` is not the same as "I can still use rabbit 3.8" (which needs to be tested at the very least, and is another kind of breaking change at least to me.

Arsnael: The commit has been removed, the all code is compiling and testing back on rabbbitmq 3.8, and it's green. 

Need something else? :)

PR #2243, opened 2024-05-13
[ENHANCEMENT] Apply RabbitMQ classic queue version 2

quantranhong1999:
Claimed to be better than version 1 cf:
- https://www.rabbitmq.com/blog/2023/05/17/rabbitmq-3.12-performance-improvements#classic-queues-massively-improved-classic-queues-v2-cqv2
- https://www.rabbitmq.com/blog/2023/05/17/rabbitmq-3.12-performance-improvements#312-cqv1-vs-cqv2

### Before (all classic queues v1)
Screenshot from 2024-05-13 16-46-30

### After (all classic queues v2)
Screenshot from 2024-05-13 16-37-40

Better performance overall (p99 805ms vs 914ms).
SELECT p99 1874ms vs 2049ms
UNSELECT p99 1420ms vs 1755ms

### DIFF
--- a/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/QueueArguments.java
+++ b/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/QueueArguments.java
@@ -70,6 +70,11 @@ public class QueueArguments {
             return this;
         }
 
+        public Builder classicQueueVersion(int version) {
+            arguments.put("x-queue-version", version);
+            return this;
+        }
+
         public Builder put(String key, Object value) {
             arguments.put(key, value);
             return this;
--- a/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/RabbitMQConfiguration.java
+++ b/backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/RabbitMQConfiguration.java
@@ -601,6 +601,7 @@ public class RabbitMQConfiguration {
 
     private static final String URI_PROPERTY_NAME = "uri";
     private static final String MANAGEMENT_URI_PROPERTY_NAME = "management.uri";
+    private static final boolean STICK_TO_CLASSIC_QUEUES_VERSION_1 = Boolean.parseBoolean(System.getProperty("james.rabbitmq.stick.to.classic.queues.version.1", "false"));
 
     public static RequireAmqpUri builder() {
         return amqpUri -> managementUri -> managementCredentials -> new Builder(amqpUri, managementUri, managementCredentials);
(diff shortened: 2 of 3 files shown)


backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/RabbitMQConfiguration.java
```diff
@@ -817,6 +817,8 @@ public QueueArguments.Builder workQueueArgumentsBuilder() {
         if (useQuorumQueues) {
             builder.quorumQueue().replicationFactor(quorumQueueReplicationFactor);
             quorumQueueDeliveryLimit.ifPresent(builder::deliveryLimit);
+        } else {
+            builder.put("x-queue-version", 2);
```
quantranhong1999, line 821: Its inter-compatibility with the older James versions is good.
This works (be skipped) with old RabbitMQ version 3.8.x and 3.9.x for example.

backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/QueueArguments.java
```diff
@@ -70,6 +70,11 @@ public Builder deliveryLimit(long deliveryLimit) {
             return this;
         }
 
+        public Builder classicQueueVersion(int version) {
+            arguments.put("x-queue-version", version);
```
quantranhong1999, line 74: Its inter-compatibility with the older James versions is good.
This works (be skipped) with old RabbitMQ version 3.8.x and 3.9.x for example.

chibenwa (commented):
Could you please...

Start a james prior this cange and queue one
Stop it, remove it
And start a james after this change and with queue 2

IE ensure this is not a breaking change.

Thanks!

quantranhong1999:
> IE ensure this is not a breaking change.

Sadly this is a breaking change (queue purge is needed).
I would add a configuration to avoid this breaking change e.g. for the older James versions.

backends-common/rabbitmq/src/main/java/org/apache/james/backends/rabbitmq/RabbitMQConfiguration.java
```diff
@@ -601,6 +601,7 @@ private int resolvePort() {
 
     private static final String URI_PROPERTY_NAME = "uri";
     private static final String MANAGEMENT_URI_PROPERTY_NAME = "management.uri";
+    private static final boolean STICK_TO_CLASSIC_QUEUES_VERSION_1 = Boolean.parseBoolean(System.getProperty("james.rabbitmq.stick.to.classic.queues.version.1", "false"));
```
Arsnael, line 604: Can't we find something shorter?

Like `fallback.classic.queues.v1` or something similar?

chibenwa (approved):

PR #2246, opened 2024-05-15
JAMES-4034 SMTP submission: validate FROM header

chibenwa:
Will ease channel pool sizing as well as debugging of RabbitMQ channelPool related errors.

### DIFF
--- a/server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SenderAuthIdentifyVerificationHook.java
+++ b/server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SenderAuthIdentifyVerificationHook.java
@@ -18,7 +18,13 @@
  ****************************************************************/
 package org.apache.james.smtpserver;
 
+import java.util.stream.Stream;
+
 import jakarta.inject.Inject;
+import jakarta.mail.Address;
+import jakarta.mail.MessagingException;
+import jakarta.mail.internet.AddressException;
+import jakarta.mail.internet.InternetAddress;
 
 import org.apache.james.core.Domain;
 import org.apache.james.core.MailAddress;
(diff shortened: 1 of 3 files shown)


Arsnael (approved):

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SenderAuthIdentifyVerificationHook.java
```diff
@@ -90,4 +98,55 @@ protected boolean isSenderAllowed(Username connectedUser, Username sender) {
         }
         return allowed;
     }
+
+    @Override
+    public HookResult onMessage(SMTPSession session, Mail mail) {
+        System.out.println("mess");
```
quantranhong1999, line 104:
```suggestion
```

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SenderAuthIdentifyVerificationHook.java
```diff
@@ -90,4 +98,55 @@ protected boolean isSenderAllowed(Username connectedUser, Username sender) {
...
+            return Stream.empty();
+        }
+    }
+
+    private boolean fromDoesNotMatchAuthUser(SMTPSession session, Address from) {
+        if (from instanceof InternetAddress internetAddress) {
+            try {
+                MailAddress mailAddress = new MailAddress(internetAddress.getAddress());
+                return session.getUsername() != null &&
+                    (!fromMatchSessionUser(mailAddress, session) || !belongsToLocalDomain(mailAddress));
```
quantranhong1999, line 134: Is delegated user supported?

chibenwa: No need to be supported as

1. that's the same validation layer than JMAP
2. the authenticated user is identified as the delegated user (impersonnation)

quantranhong1999 (approved):

vttranlina (approved):

PR #2256, opened 2024-05-22
JAMES-3946 DropList Cassandra and Memory app

Maxxx873:


chibenwa:
Hello!

Please do a separate commit for massive removals. This make it easier to track!

server/mailet/droplists-cassandra/extensions.properties
```diff
@@ -0,0 +1 @@
+guice.extension.module=org.apache.james.droplists.cassandra.CassandraDropListsModule
```
chibenwa, line 1: Maybe we could move this file into a `sample-configuration` sub folder to make the intent clear?

server/mailet/droplists-cassandra/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.james</groupId>
+        <artifactId>james-server</artifactId>
+        <version>3.9.0-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>james-server-droplists-cassandra</artifactId>
```
chibenwa, line 13: To be fair I am not super fan of this extra maven module.

We could move the CassandraDropListsModule into `CassandraDropListsModule` ?

server/mailet/droplists-cassandra/src/main/java/org/apache/james/droplists/cassandra/CassandraDropListsModule.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.inject.AbstractModule;
+import com.google.inject.Scopes;
+
+public class CassandraDropListsModule extends AbstractModule {
+    @Override
+    protected void configure() {
+        bind(CassandraDropList.class).in(Scopes.SINGLETON);
+        bind(DropList.class).to(CassandraDropList.class);
+
+        bind(DropListRoutes.class).in(Scopes.SINGLETON);
```
chibenwa, line 34: Why binding the routes in Singleton?

server/mailet/droplists-memory/pom.xml
```diff
@@ -0,0 +1,72 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>org.apache.james</groupId>
+        <artifactId>james-server</artifactId>
+        <version>3.9.0-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>james-server-droplists-memory</artifactId>
```
chibenwa, line 13: Idem I would see `MemoryDropListsModule` onto server/container/guice/memory

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/WhiteListManager.java
```diff
@@ -1,793 +0,0 @@
-/****************************************************************
```
chibenwa, line 1: After a second though, to remove this class we need to be able to make the droplist work with the JPA app with a SQL backend as otherwize James JPA users would not be able to have an alternative.

Maybe we can implement a JPA backend prior to the removal? WDYT?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/WhiteListManager.java
```diff
@@ -1,793 +0,0 @@
...
- * Licensed to the Apache Software Foundation (ASF) under one   *
```
chibenwa, line 2: Also removals shall be documented as a breaking change in upgrade-instructions.md with an alternative....

server/mailet/droplists-cassandra/src/main/java/org/apache/james/droplists/cassandra/CassandraDropListsModule.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
```
chibenwa, line 36: Question here: CassandraModules are not declared. THis means that James would not manage their lifecycle.

We would need to register them for James to manage them.

Actually there is a further issue: child injectors are not able to modify the parent injector. THis means the module likely cannot register them easily...

The way I would see it would be:
 - A/ Accept an external lifecycle for the cassandra extensions. In such case we need to document the CQL code to create the table to the minimum.... 
 - B/ Allow registration of UserDefinedCassandraModule. Those would subclass cassandra modules and be explicitly registered into cassandra.properties. The guice wiring would then assemble the "core" cassandraModules and the user defined ones. Boiler plate but doable. Could then be reused for other extensions.
 - C/ Decide that the extension is core enough to get a module chooser. We would have a config toogle to turn it on/off. And we would not encounter this limitation as the guice modules would then be combined. This would be nice as separate routes for instance won't need extra registration. As everything is already on the classpath, impact is limited.
 - D/ Bind everything directly in each and every app. I do not like it as we should allow users to not have too many cassandra tables and disable those of features he do not uses...
 
 Do you see other options? Any preference / thoughts?

chibenwa:
(and thanks a lot for the contribution)

Maxxx873:
> (and thanks a lot for the contribution)

As always, thank you very much for the code review. I will need some more time to analyze the comments and make corrections.

server/mailet/droplists-cassandra/src/main/java/org/apache/james/droplists/cassandra/CassandraDropListsModule.java
Maxxx873: I like option C. I even started implementing it, but I doubted that it was core enough. I think we should go back to C.

server/mailet/droplists-cassandra/src/main/java/org/apache/james/droplists/cassandra/CassandraDropListsModule.java
Maxxx873: I should have used Multibinder...

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/WhiteListManager.java
Maxxx873: Yes I agree. Before deleting, I'll try to implement a JPA backend.

server/mailet/droplists-cassandra/src/main/java/org/apache/james/droplists/cassandra/CassandraDropListsModule.java
chibenwa: I share and support this analysis.

server/mailet/droplists-cassandra/src/main/java/org/apache/james/droplists/cassandra/CassandraDropListsModule.java
chibenwa: You can not use parent multi-binder in a child module...

This could be doable if we use module combinaison for loading extensions but this leads to a lower isolation between components. Child modules were historically choosen on the project to provide higher level isolation...

Anyway https://github.com/apache/james-project/pull/2256#discussion_r1612285222 will address this.

PR #2259, opened 2024-05-24
JAMES-2586 Fix sequential issue with updating flags in the reactive p…

Arsnael:
…ipeline

The deployments tests (that aren't being tested anymore in apache james) were not passing for TMail since Postgresql latest updates with pg-pool => https://github.com/linagora/tmail-backend/pull/1035

The store imap command on message flags seems not always returning in the right order fetch responses. The one little change flatMap to concatMap here makes sure we keep it sequential in order, that's where the UpdatedFlags were being mixed up in order in the pipeline.

### DIFF
--- a/mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/PostgresMessageMapper.java
+++ b/mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/PostgresMessageMapper.java
@@ -20,6 +20,7 @@
 package org.apache.james.mailbox.postgres.mail;
 
 import static org.apache.james.blob.api.BlobStore.StoragePolicy.LOW_COST;
+import static org.apache.james.util.ReactorUtils.DEFAULT_CONCURRENCY;
 
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


quantranhong1999 (approved):
Good catch!

hungphan227 (approved):

mailbox/postgres/src/main/java/org/apache/james/mailbox/postgres/mail/PostgresMessageMapper.java
```diff
@@ -290,7 +290,7 @@ private Flux<UpdatedFlags> updatedFlags(List<ComposedMessageIdWithMetaData> list
                                             FlagsUpdateCalculator flagsUpdateCalculator) {
         return modSeqProvider.nextModSeqReactive(mailbox.getMailboxId())
             .flatMapMany(newModSeq -> Flux.fromIterable(listMessagesMetaData)
-                .flatMap(messageMetaData -> updateFlags(messageMetaData, flagsUpdateCalculator, newModSeq)));
+                .concatMap(messageMetaData -> updateFlags(messageMetaData, flagsUpdateCalculator, newModSeq)));
```
chibenwa, line 293: Range updates are frequent in IMAP. doing this sequentially might be a show stopper.

How about using `flatMapSequential` ?

Also, we are missing the concurrency control: let's use the default concurrency defined in reactor utils.

chibenwa (commented):
Nice diagnostic!

But I miss non regression tests, TDD style.

Arsnael:
> But I miss non regression tests, TDD style.

Not sure what kind of test I can add here. As said there is imap mpt tests that do similar store flags commands but somehow they are green

chibenwa:
Unit tests maybe on the mapper layer?

chibenwa (approved):

PR #2265, opened 2024-05-30
JAMES-4041 Fix OOM upon IMAP COPY

chibenwa:
Issue arised from command lines like:

```
a0 COPY 1:4294967295 otherMailbox
```

Semantically valid but it actually caused the batching algorithm to generate with a batch size of 10 429.496.730 ranges. With a memory overhead of ~52 bytes per range we would occupy a space of ~20GB thus the OOM.

The root cause lies on the MUA using an arbitrary large range instead of:

```
a0 COPY 1:* otherMailbox
```

We believe the batching semantic also is incorrect. Browsing large amounts of metadata actuall is not an issue, but the post traitment might be.

We thus:
 - First consolidate the range against the actual messages in the store
 - Window it using the batch size.
 - And finally execute like before the copy/move logic.

This avoids generating a large amount of potentially enpty ranges.

MessageBatcher is furthermore removed as it no longer is used.


Arsnael (approved):

quantranhong1999 (approved):

Arsnael:
Can you squash the fixup @chibenwa ?

PR #2269, opened 2024-06-02
[CLEAN CODE] Annotation repetitions should not be wrapped

Maxxx873:
@NamedQueries has been removed.

[Repeating Annotations](https://docs.oracle.com/javase/tutorial/java/annotations/repeating.html)


chibenwa (approved):
Thanks ;-)

Arsnael (approved):
Good catch, thanks for the fix :)

PR #2275, opened 2024-06-04
JAMES-4015 VacationMailet should sanitize broken Reply-To fields

chibenwa:


server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/VacationMailet.java
```diff
@@ -86,6 +95,30 @@ public void service(Mail mail) {
...
+                .map(LenientAddressParser.DEFAULT::parseAddressList)
+                .flatMap(Collection::stream)
+                .filter(Mailbox.class::isInstance)
+                .map(Mailbox.class::cast)
+                .map(Mailbox::getAddress)
+                .map(Throwing.function(InternetAddress::new))
+                .toArray(InternetAddress[]::new);
+
+            if (replyTo.length > 0) {
+                LOGGER.info("Recovering from badly formatted Reply-To. Originaal value {}, deduced value {}",
```
Arsnael, line 112:
```suggestion
                LOGGER.info("Recovering from badly formatted Reply-To. Original value {}, deduced value {}",
```

Arsnael (approved):

quantranhong1999 (approved):

PR #2276, opened 2024-06-04
[IMPROVEMENT] Configure fastview projection miss threshold

chibenwa:
We observe some warning from perfectly well running systems. This threshold is hard coded, a bit of liberty can be nice.


Arsnael (approved):

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/projections/MessageFastViewProjectionHealthCheck.java
```diff
@@ -35,7 +37,9 @@
 public class MessageFastViewProjectionHealthCheck implements HealthCheck {
 
     private static final ComponentName COMPONENT_NAME = new ComponentName("MessageFastViewProjection");
-    private static final double MAXIMUM_MISS_PERCENTAGE_ACCEPTED = 10;
+    private static final double MAXIMUM_MISS_PERCENTAGE_ACCEPTED = Optional.ofNullable(System.getProperty("james.jmap.fastview.projection.max.miss.percentage.accepted", null))
```
quantranhong1999, line 40: Should we document this in `jvm.adoc`?

chibenwa: IMO this is niche...

Arsnael:
Merge or not?

PR #2289, opened 2024-06-09
JAMES-3946 Add a DropLists JPA backend

Maxxx873:


server/data/data-jpa/pom.xml
```diff
@@ -33,7 +33,7 @@
     <name>Apache James :: Server :: Data :: JPA Persistence</name>
 
     <properties>
-        <target.jdk>11</target.jdk>
+        <target.jdk>21</target.jdk>
```
chibenwa, line 36: Is this even needed to get this property defined here?

We likely should just inherit this from parent project...

chibenwa (approved):
Looks overall good to me!

server/data/data-jpa/pom.xml
Maxxx873: Yes indeed, I'll try to inherit this

quantranhong1999 (approved):
Thank you!

chibenwa (approved):

PR #2291, opened 2024-06-10
[FIX] registrations multimap made fully thread safe

romainmoreau:
It used to throw a `ConcurrentModificationException` when there was a high level of concurrency.

Here's a simple demo that reproduces the issue:

```java
import java.util.Collections;
import java.util.HashMap;
import java.util.concurrent.ConcurrentHashMap;
import java.util.function.Supplier;
import java.util.stream.IntStream;

import com.google.common.collect.HashMultimap;
import com.google.common.collect.Multimap;
import com.google.common.collect.Multimaps;

public class SimpleDemo {
    private static final String KEY = "key";

    public static void main(String[] args) {
        System.out.println("Thread safe demo");
        demo(SimpleDemo::threadSafe);
        System.out.println("Partially thread safe demo");
        demo(SimpleDemo::partiallyThreadSafe);
    }

    private static Multimap<String, Integer> threadSafe() {
        return Multimaps.synchronizedSetMultimap(
                Multimaps.newSetMultimap(new HashMap<>(), () -> Collections.newSetFromMap(new ConcurrentHashMap<>())));
    }

    private static Multimap<String, Integer> partiallyThreadSafe() {
        return Multimaps.synchronizedSetMultimap(HashMultimap.create());
    }

    private static void demo(Supplier<Multimap<String, Integer>> multimapSupplier) {
        var multimap = multimapSupplier.get();
        IntStream.range(0, 128).parallel().forEach(i -> {
            multimap.put(KEY, i);
            multimap.get(KEY).stream().mapToInt(j -> j).sum();
        });
    }
}
```

### DIFF
--- a/event-bus/api/src/test/java/org/apache/james/events/EventBusConcurrentTestContract.java
+++ b/event-bus/api/src/test/java/org/apache/james/events/EventBusConcurrentTestContract.java
@@ -28,7 +28,9 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.awaitility.Awaitility.await;
 
 import java.time.Duration;
+import java.util.List;
 import java.util.Set;
+import java.util.stream.IntStream;
 
 import org.apache.james.events.EventBusTestFixture.EventListenerCountingSuccessfulExecution;
 import org.apache.james.events.EventBusTestFixture.GroupA;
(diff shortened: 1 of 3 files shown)


quantranhong1999 (approved):
Thanks. Should we have a unit test to reproduce this?

chibenwa (approved):

chibenwa:
Good catch!

Maybe we could craft a unit test using `ConcurrentTestRunner` against th in vm event bus and try to trigger the bug? (Bonus...)

romainmoreau:
Sure I'll try to craft a unit test.

romainmoreau:
I added a test that reproduces the issue to the generic concurrent test suite and added this suite to the in VM implementation (not sure why it wasn't already before).

chibenwa (approved):

quantranhong1999 (approved):

PR #2303, opened 2024-06-14
[FIX] Allow delivering emails to a&b@linagora.com

chibenwa:
Reported by @samiulsami on gitter chat.

Delivery to this valid mail address and username failed due to quota root limitations that was not pushed to system edge.

We lift this limitation bu reworking the way
DefaultUserQuotaRootResolver is working.


quantranhong1999:
```java
emptyQuotaRootShouldBeWellDeSerialized – org.apache.james.event.json.dtos.QuotaRootTest
14s
Error
Expecting actual throwable to be an instance of:
  org.apache.james.mailbox.exception.MailboxException
but was:
  java.lang.IllegalArgumentException: username should not be null or empty after being trimmed
```

Related.

chibenwa:
(was :green_apple: , I force pushed by mistake)

quantranhong1999 (approved):

PR #2304, opened 2024-06-14
[FIX] race condition avoided

romainmoreau:
With PLAIN auth, flushing before popping `ImapLineHandler` in `AuthenticateProcessor` will result sometimes in this `ImapLineHandler` to also consume the next IMAP command.

It can be reproduced with this simple demo:

```java
for (var i = 0; i < 10_000; i++) {
	Session.getInstance(getProperties()).getStore().connect("user", "password");
}
```

Without this patch, the problem occurs on average every 500 authentications in my environment.
With the patch, the demo loop completes every time.

### DIFF
--- a/protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractAuthProcessor.java
@@ -96,7 +96,6 @@ public abstract class AbstractAuthProcessor<R extends ImapRequest> extends Abstr
                         .action("AUTH")
                         .log("IMAP Authentication succeeded.");
                     okComplete(request, responder);
-                    responder.flush();
                     session.stopDetectingCommandInjection();
                 } catch (BadCredentialsException e) {
                     authFailure = true;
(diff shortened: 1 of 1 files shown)


chibenwa (commented):
Ok looks bad indeed.

Please provide a test for this in ImapServerTest, taking a few seconds on it is OK.

BTW doesn't break something in ImapServerTest? I am likely the one inserting this "flush" and it definitly is not there by accident... Hopefully this is covered by a test case.

chibenwa:
We can do a @RepeatedTest(50) 

One run would not be enough to catch the issue but over several builds we would catch regressions.

romainmoreau:
Yes I noticed your previous PR #1432 and I tested my change with Evolution, I didn't see any error when logging in with login/password or listing; also no unit test seems to fail, so I was hoping you'd remember why it was added.

Thanks for the pointers for making the unit test.

chibenwa (approved):
Then let's go....

romainmoreau:
#2307 was opened for the unit test (I finally did like `shouldSupportManyConcurrentSSLConnections` because it was more effective to reproduce the issue).

PR #2315, opened 2024-06-26
JAMES-4046 Upgrade Lucene

quantranhong1999:
resolve https://github.com/linagora/james-project/issues/5001

status: fine at version 6

### DIFF
--- a/mailbox/lucene/pom.xml
+++ b/mailbox/lucene/pom.xml
@@ -103,17 +103,17 @@
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
-            <artifactId>lucene-analyzers</artifactId>
+            <artifactId>lucene-analyzers-common</artifactId>
+            <version>${lucene.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-analyzers-smartcn</artifactId>
             <version>${lucene.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-core</artifactId>
-            <version>${lucene.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.lucene</groupId>
-            <artifactId>lucene-smartcn</artifactId>
             <version>${lucene.version}</version>
         </dependency>
         <dependency>
--- a/mailbox/lucene/pom.xml
+++ b/mailbox/lucene/pom.xml
@@ -117,6 +117,12 @@
             <version>${lucene.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-misc</artifactId>
+            <version>${lucene.version}</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.awaitility</groupId>
             <artifactId>awaitility</artifactId>
             <scope>test</scope>
--- a/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LenientImapSearchAnalyzer.java
+++ b/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LenientImapSearchAnalyzer.java
@@ -22,7 +22,7 @@ import java.io.Reader;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.TokenStream;
-import org.apache.lucene.analysis.WhitespaceTokenizer;
+import org.apache.lucene.analysis.core.WhitespaceTokenizer;
 import org.apache.lucene.analysis.shingle.ShingleFilter;
 import org.apache.lucene.util.Version;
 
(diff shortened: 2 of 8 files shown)


Arsnael (commented):
Read it, nothing to comment so far.

Seeing that latest release of Lucene is 9.11... good luck :)

quantranhong1999:
I am starting to get lost debugging the `testSearchAtomsUS` MPT test.

I confirm that this test failed from Lucene 4.x already.

```
C: A83 STORE 1:9  +FLAGS (\FLAGGED)
S: \* 1 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: \* 2 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: \* 3 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: \* 4 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: \* 5 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: \* 6 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: \* 7 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: \* 8 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: \* 9 FETCH \(FLAGS \(\\Flagged \\Recent\)\)
S: A83 OK STORE completed\.

C: A84 STORE 1:4  +FLAGS (\ANSWERED)
S: \* 1 FETCH \(FLAGS \(\\Answered \\Flagged \\Recent\)\)
S: \* 2 FETCH \(FLAGS \(\\Answered \\Flagged \\Recent\)\)
S: \* 3 FETCH \(FLAGS \(\\Answered \\Flagged \\Recent\)\)
S: \* 4 FETCH \(FLAGS \(\\Answered \\Flagged \\Recent\)\)
S: A84 OK STORE completed\.

C: A85 STORE 10:14  +FLAGS (\ANSWERED)
S: \* 10 FETCH \(FLAGS \(\\Answered \\Recent\)\)
S: \* 11 FETCH \(FLAGS \(\\Answered \\Recent\)\)
S: \* 12 FETCH \(FLAGS \(\\Answered \\Recent\)\)
S: \* 13 FETCH \(FLAGS \(\\Answered \\Recent\)\)
S: \* 14 FETCH \(FLAGS \(\\Answered \\Recent\)\)
S: A85 OK STORE completed\.
```

After this script, only messages 10 to 14 are added the `Answered` flag, while it is not the case for messages 1 to 4. Messages 1 to 4 after this script still stuck with only the `Flagged` flag.

The weird thing is if we do not update the  `Flagged` flag for messages 1 to 4 in the first place, then setting the `Answer` flag for them is OK.

My feeling is we have an issue updating the flag for the messages that already have a flag. But I could not progress more in that direction...

chibenwa:
Have you checked for race conditions? I suspect the lucene indexes updates not to be idempotent...

quantranhong1999:
> Have you checked for race conditions? 

Synchronized keyword does not solve the issue for example.

> With the UpdatedFlags ad hand we can instead subtilly mix added flags / deletes flags?
BTW we might adopt a similar indempotent friendly way to apply updates into OpenSearch too I believe...

It is likely not impossible with Lucene or at least our case for partial update (Lucene supports partially update only NUMERIC and BINARY field type while our flags need to be STRING to easily query). `Lucene does not provide a direct API to update just a part of a document. The typical approach in Lucene is to delete the old document and add a new one with the updated fields. This is because Lucene's indexing structure does not allow for in-place updates of individual fields without reindexing the entire document.`


> My feeling is we have an issue updating the flag for the messages that already have a flag.

  So after a `.updateDocument` call:
   - If I use the old `IndexReader`, I see no change on the document but can search for the document.
   - If I use the new `IndexReader` (to try to get the latest snapshot of the index), I see no change on the document and even can not search for the document.

  I tried to make `IndexWriter` commit everywhere and open new `IndexReader` but can not see the change got reflected (right away). And somehow the update is really laggy.
e.g:
- after we set the `FLAGGED` flag on the message, we see the message has no flag.
- after we add the `ANSWER` flag to the message, we see the message has only the `FLAGGED` flag.

I do not understand how that happens....

I tried to implement the optimistic concurrent control for the update flag, but either the `.updateDocument` API of Lucene suck as it just returns void (no way to know if the update succeeds or not so we can retry) or the `IndexReader` is not reliable to get the latest status of the document.

chibenwa:
Ok I see. This is because we read the document from Lucene, patch it, then upload it again.

It is very possible for the document not to be updated internally yet (because index changes are async) and thus not seen when updating the document a second time...

chibenwa:
Nuke option: read again from the DB :smiling_imp:

Smart option: see how partial document update had been implemented in OpenSearch / friends. They likely solved the problem already...

chibenwa:
Otherwize I think adding some awaits in the tests might be ok, we thus would aknowlege there is a data race in our Lucene code...

quantranhong1999:
> Otherwize I think adding some awaits in the tests might be ok, we thus would aknowlege there is a data race in our Lucene code...

I added some AWAIT 10 seconds in the MPT test but it did not solve the issue.

> It is very possible for the document not to be updated internally yet (because index changes are async) and thus not seen when updating the document a second time...

Likely not as during debugging I stop for a long time and after that, the update is still not reflected.
Of course with a new `IndexReader` opened (snapshot of current index).

quantranhong1999:
I tried to put AWAIT after every STORE +FLAG command -> not work.

I tried to use the [Luke](https://github.com/DmitryKey/luke) tool to introspect the index during debugging, but not work as we can only check the index after the test...

Likely I would put this work in the standby mode for now. Unless anyone wants to have fun with it...

Arsnael:
How about disabling the mpt tests for now, then continue the migration towards 9.11 to see if this has been done differently or solved on lucene side after?

Arsnael:
It was just a suggestion. This task aint a priority at all, just a nice plus if we can deal with it.

Dont hesitate to take a break on something else :)

PR #2337, opened 2024-07-03
JAMES-3824 SMTP Extension for Message Transfer Priorities

Maxxx873:
[JAMES-3824](https://issues.apache.org/jira/projects/JAMES/issues/JAMES-3824?filter=allopenissues)

### DIFF
--- a/docs/modules/servers/partials/RemoteDelivery.adoc
+++ b/docs/modules/servers/partials/RemoteDelivery.adoc
@@ -74,6 +74,10 @@ within the mailet code.<br>
 Note: This facility should be used with extreme care by expert users with a thorough knowledge of the relevant RFCs and
 the ability to perform their own problem resolutions.
 * *debug* (optional) - a Boolean (true/false) indicating whether debugging is on. Default is false.
+* *usePriority* (optional) - a Boolean value (true/false) that indicates whether email priority support is enabled.
+When the property is set to false, it allows for the disabling of priorities when interacting with unknown third-party
+systems during remote delivery MX resolution. When this option is enabled, any email will be assigned the lowest priority
+during retry attempts. Default is false.
 
 ==== Security
 
--- a/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/ReceivedDataLineFilter.java
+++ b/protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/ReceivedDataLineFilter.java
@@ -53,6 +53,7 @@ public class ReceivedDataLineFilter extends SeparatingDataLineFilter {
     private static final AtomicInteger COUNTER = new AtomicInteger(0);
     private final ProtocolSession.AttachmentKey<Boolean> headersPrefixAdded = ProtocolSession.AttachmentKey.of("HEADERS_PREFIX_ADDED" + COUNTER.incrementAndGet(), Boolean.class);
     private final ProtocolSession.AttachmentKey<Boolean> headersSuffixAdded = ProtocolSession.AttachmentKey.of("HEADERS_SUFFIX_ADDED" + COUNTER.incrementAndGet(), Boolean.class);
+    private final ProtocolSession.AttachmentKey<Integer> mtPriority = ProtocolSession.AttachmentKey.of("MT-PRIORITY", Integer.class);
 
     /**
      * Return the service type which will be used in the Received headers.
@@ -111,23 +112,28 @@ public class ReceivedDataLineFilter extends SeparatingDataLineFilter {
                 sslSession.getProtocol(),
                 Optional.ofNullable(sslSession.getCipherSuite())
                     .orElse("")))
-                .ifPresent(header::add);
+            .ifPresent(header::add);
 
         headerLineBuffer.append("by ").append(session.getConfiguration().getHelloName()).append(" (").append(session.getConfiguration().getSoftwareName()).append(") with ").append(getServiceType(session, heloMode.orElse("NOT-DEFINED")));
         headerLineBuffer.append(" ID ").append(session.getSessionID());
 
         List<MailAddress> rcptList = session.getAttachment(SMTPSession.RCPT_LIST, State.Transaction).orElse(ImmutableList.of());
+
+        String priorityValue = session.getAttachment(mtPriority, State.Transaction)
+            .map(p -> " (PRIORITY " + p + ")").orElse("");
+
         if (rcptList.size() == 1) {
             // Only indicate a recipient if they're the only recipient
             // (prevents email address harvesting and large headers in
             // bulk email)
             header.add(headerLineBuffer.toString());
-            
+
             headerLineBuffer = new StringBuilder();
-            headerLineBuffer.append("for <").append(rcptList.get(0).toString()).append(">;");
+            headerLineBuffer.append("for <").append(rcptList.getFirst().toString()).append(">");
+            headerLineBuffer.append(priorityValue).append(";");
         } else {
             // Put the ; on the end of the 'by' line
-            headerLineBuffer.append(";");
+            headerLineBuffer.append(priorityValue).append(";");
         }
         header.add(headerLineBuffer.toString());
         headerLineBuffer = new StringBuilder();
@@ -165,7 +171,7 @@ public class ReceivedDataLineFilter extends SeparatingDataLineFilter {
      */
     private Response addHeaders(SMTPSession session, byte[] line, LineHandler<SMTPSession> next) {
         Response response;
-        for (Header header: headers(session)) {
+        for (Header header : headers(session)) {
             response = header.transferTo(session, next);
             if (response != null) {
                 return response;
--- a/server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/remote/delivery/DeliveryRunnableTest.java
+++ b/server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/remote/delivery/DeliveryRunnableTest.java
@@ -36,6 +36,7 @@ import java.util.function.Supplier;
 
 import org.apache.james.domainlist.api.DomainList;
 import org.apache.james.metrics.tests.RecordingMetricFactory;
+import org.apache.james.queue.api.MailPrioritySupport;
 import org.apache.james.queue.api.MailQueue;
 import org.apache.mailet.Attribute;
 import org.apache.mailet.AttributeValue;
@@ -61,6 +62,7 @@ public class DeliveryRunnableTest {
     void setUp() {
         FakeMailetConfig mailetConfig = FakeMailetConfig.builder()
             .setProperty(RemoteDeliveryConfiguration.DEBUG, "true")
+            .setProperty(RemoteDeliveryConfiguration.USE_PRIORITY, "true")
             .setProperty(RemoteDeliveryConfiguration.DELAY_TIME, "1000,2000,3000,4000,5000")
             .build();
 
@@ -140,6 +142,7 @@ public class DeliveryRunnableTest {
         verify(mailQueue).enQueue(FakeMail.builder()
                 .name("name")
                 .attribute(DeliveryRetriesHelper.makeAttribute(1))
+                .attribute(MailPrioritySupport.LOW_PRIORITY_ATTRIBUTE)
                 .attribute(new Attribute(IS_DELIVERY_PERMANENT_ERROR, AttributeValue.of(false)))
                 .state(Mail.ERROR)
                 .lastUpdated(FIXED_DATE)
@@ -163,6 +166,7 @@ public class DeliveryRunnableTest {
         verify(mailQueue).enQueue(FakeMail.builder()
                 .name("name")
                 .attribute(DeliveryRetriesHelper.makeAttribute(3))
+                .attribute(MailPrioritySupport.LOW_PRIORITY_ATTRIBUTE)
                 .attribute(new Attribute(IS_DELIVERY_PERMANENT_ERROR, AttributeValue.of(false)))
                 .state(Mail.ERROR)
                 .lastUpdated(FIXED_DATE)
@@ -186,6 +190,7 @@ public class DeliveryRunnableTest {
         verify(mailQueue).enQueue(FakeMail.builder()
                 .name("name")
                 .attribute(DeliveryRetriesHelper.makeAttribute(5))
+                .attribute(MailPrioritySupport.LOW_PRIORITY_ATTRIBUTE)
                 .attribute(new Attribute(IS_DELIVERY_PERMANENT_ERROR, AttributeValue.of(false)))
                 .state(Mail.ERROR)
                 .lastUpdated(FIXED_DATE)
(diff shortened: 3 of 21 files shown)


server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/priority/SmtpMtPriorityParameterHook.java
```diff
@@ -0,0 +1,77 @@
...
+            return HookResult.builder()
+                .smtpReturnCode(SMTPRetCode.SYNTAX_ERROR_ARGUMENTS)
+                .hookReturnCode(HookReturnCode.deny())
+                .smtpDescription("The Mail parameter cannot contain more than one MT-PRIORITY parameter at the same time")
+                .build();
+        }
+        if (paramName.equals(MT_PRIORITY_PARAMETER)) {
+            try {
+                Optional<SmtpMtPriorityParameters.MtPriority> mtPriority = SmtpMtPriorityParameters.MtPriority.from(paramValue);
+                if (mtPriority.filter(p -> session.getUsername() == null && p.getPriorityValue() > 0).isPresent()) {
```
chibenwa, line 55: Who is able to set which priory is actually quite crucial and saying "high priorities are only for connected users" seems somehow arbitrary. (We can be more severe than the spec and allow eg some high priority just to some individuals)

I would encapsulate the check into a dedicated method of smtp session - so that it could get easily overloaded if need be:

```
default boolean isAuthorized(MtPriority p) { 
     return session.getUsername() != null && p.getPriorityValue() <= 0;
}
```

And then:

```
mtPriority.ifPresent(p -> {
    if (!session.isAuthorized(p)) {
        // reject
    }
    session.setAttachment(MT_PRIORITY_ATTACHMENT_KEY, p, Transaction))
});
```

(possibly even extracting the resulting lambda as a method)

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/priority/SmtpMtPriorityParameters.java
```diff
@@ -0,0 +1,100 @@
...
+            return new SmtpMtPriorityParameters.MtPriority(priority);
+        }
+
+        public static Optional<MtPriority> from(String value) {
+            if (Integer.parseInt(value) < MIN_MT_PRIORITY_VALUE || Integer.parseInt(value) > MAX_MT_PRIORITY_VALUE) {
+                throw new IllegalArgumentException("Invalid priority: According to RFC-6710 allowed values are from -9 to 9 inclusive");
+            }
+            return Optional.of(value)
+                .map(SmtpMtPriorityParameters.MtPriority::of);
+        }
```
chibenwa, line 53: Why an optional? We actually always return a value....

Also we call `Integer.parseInt(value)` 3 times, twice in the check, once in `of`.

Suggestion:

 - [ ] Move the check (-9 < p < 9) into the constructor
 - [ ] And remove `from` method
 - [ ] Adapt callers

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/priority/SmtpMtPriorityParameters.java
```diff
@@ -0,0 +1,100 @@
...
+
+        public static MtPriority defaultValue() {
+            return new SmtpMtPriorityParameters.MtPriority(0);
+        }
```
chibenwa, line 57: This can be a constant:

```
public static final MtPriority DEFAULT = new MtPriority(0);
```

server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpMtPriorityMessageHookTest.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.mailet.Mail;
+import org.assertj.core.api.SoftAssertions;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+class SmtpMtPriorityMessageHookTest {
```
chibenwa, line 42: I do not see test that shows rejection of high value MT priority for not connected users.

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/priority/SmtpMtPriorityParameterHook.java
```diff
@@ -0,0 +1,77 @@
...
+                    LOGGER.debug("Needs to be logged to use positive priorities");
+                    session.setAttachment(MT_PRIORITY_ATTACHMENT_KEY, SmtpMtPriorityParameters.MtPriority.defaultValue(), Transaction);
+                    return HookResult.DECLINED;
+                }
+                mtPriority.ifPresent(p -> session.setAttachment(MT_PRIORITY_ATTACHMENT_KEY, p, Transaction));
```
chibenwa, line 60: We would need to have an INFO log about priority setting CF

```
   3.  The received MT-PRIORITY parameter value SHOULD be logged as part
       of any logging of message transactions.
```

server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpMtPriorityMessageHookTest.java
```diff
@@ -0,0 +1,147 @@
...
+
+        SoftAssertions.assertSoftly(softly -> {
+            softly.assertThat(smtpProtocol.getReplyCode()).isEqualTo(250);
+            softly.assertThat(smtpProtocol.getReplyString()).contains("250 MT-PRIORITY");
+        });
+    }
+
+    @ParameterizedTest
+    @CsvSource({"-9,0", "-8,0", "-7,0", "-6,0", "-5,0", "-4,0", "-3,0", "-2,0", "-1,0", "0,0", "1,1",
+        "2,2", "3,3", "4,4", "5,5", "6,6", "7,7", "8,8", "9,9"})
```
chibenwa, line 73: Yes true actually the mail queue implementations do not support negative priorities as of today yet I believe this is more of a mistake and mis-understanding of legacy implementors rather than a concious choice.

It would be great if we could modify the JMS mailque to support this and alter PriorityMailQueueContract accordingly, and thus having here:

```suggestion
    @CsvSource({"-9,-9", "-8,-8", "-7,-7", "-6,-6", "-5,-5", "-4,-4", "-3,-3", "-2,-2", "-1,-1", "0,0", "1,1",
        "2,2", "3,3", "4,4", "5,5", "6,6", "7,7", "8,8", "9,9"})
```

While we do not violate the spec (that accounts for such rounding) we could at least plan this as an improvement.

chibenwa (commented):
Happy to see you did well with Apache James SMTP layers!

Several remarks:

 - We do not add the PRIORITY clause in Received headers
 
 CF
 
 ```
2.  When inserting a Received header field as specified in [Section](https://www.rfc-editor.org/rfc/rfc5321#section-4.4)
       [4.4 of [RFC5321]](https://www.rfc-editor.org/rfc/rfc5321#section-4.4), the compliant MTA/MSA (Mail Submission Agent)
       SHOULD include the "PRIORITY" clause whose syntax is specified in
       [Section 7](https://www.rfc-editor.org/rfc/rfc6710.html#section-7).
```

That is strongly encouraged and might not be that hard to implement. Maybe we should...

 - We do not preserve the priority upon relays. Ideally RemoteDelivery should comply, and cary other to external systems that advertize support for it the mail priority.

**Comment by chibenwa** [comment]
(a rebase is needed to solve conflict with distributed server doc relocation)

**Comment by Maxxx873** [comment]
As always, I'm amazed at the level of expertise in the code review, thank you! I will be happy to continue working on all comments

**Comment by chibenwa** [comment]
Sure. If you need pointers, especially on how to test "RemoteDelivery" then do not hesitate to ask.

File: server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpMtPriorityMessageHookTest.java (hunk 6 above)
Comment (Maxxx873) [reply server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpMtPriorityMessageHookTest.java:73]: If all goes well, I plan to implement this improvement in a separate PR

**Comment by Maxxx873** [comment]
I have worked through the main points. I plan to finalize "RemoteDelivery" as part of this PR a little later.

File: server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpMtPriorityMessageHookTest.java (hunk 4 above)
Comment (Maxxx873) [reply server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpMtPriorityMessageHookTest.java:42]: in this `shouldSetDefaultPriorityWhenUnauthenticated(String priorityValue)`

File: protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/ReceivedDataLineFilter.java [hunk 7]
```diff
@@ -165,7 +171,7 @@ protected Response onHeadersLine(SMTPSession session, byte[] line, LineHandler<S
      */
     private Response addHeaders(SMTPSession session, byte[] line, LineHandler<SMTPSession> next) {
         Response response;
-        for (Header header: headers(session)) {
+        for (Header header : headers(session)) {
```
chibenwa, line 174: Checkstyle

chibenwa (commented):
Good so far to me!

docs/modules/servers/partials/RemoteDelivery.adoc
```diff
@@ -74,6 +74,9 @@ within the mailet code.<br>
 Note: This facility should be used with extreme care by expert users with a thorough knowledge of the relevant RFCs and
 the ability to perform their own problem resolutions.
 * *debug* (optional) - a Boolean (true/false) indicating whether debugging is on. Default is false.
+* *usePriority* (optional) - a Boolean value (true/false) that indicates whether email priority support is enabled.
+When this option is enabled, any email that does not have a specified priority will be assigned the highest priority upon
+delivery and the lowest priority during retry attempts. Default is false.
```
chibenwa, line 79: If enabled, I think we should not communicate priorities for mails that do not have one.

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/remote/delivery/DeliveryRunnableTest.java
```diff
@@ -240,6 +245,29 @@ void deliveryTemporaryFailureShouldResetDeliveryCountOnNonErrorState() throws Ex
...
+                .attribute(new Attribute(IS_DELIVERY_PERMANENT_ERROR, AttributeValue.of(false)))
+                .state(Mail.ERROR)
+                .lastUpdated(FIXED_DATE)
+                .build(),
+            Duration.ofSeconds(1));
+        verifyNoMoreInteractions(mailQueue);
+    }
+
+    @Test
+    void deliveryTemporaryFailureShouldRetryDeliveryWithPreservedPriority() throws Exception {
```
chibenwa, line 258: We should ensure that all outgoing components are able to set priorities based on the outgoing queue content too.

See https://github.com/apache/james-project/tree/master/server/mailet/remote-delivery-integration-testing for writing such tests.

I would expect that James transmit the mails precisely with the priority I did submit it with.

docs/modules/servers/partials/RemoteDelivery.adoc
Maxxx873: If this property `usePriority` is enabled but does not affect emails without priorities set, I don't quite understand what it is for. Based on my understanding, this property was initially designed to hardcode email priorities through configuration, with the highest priority assigned on send and the lowest on retry. To maintain the original intent while also supporting the new priority functionality, I propose using this property to assign priorities to emails that currently lack a specified priority. 

I also have a couple of alternative suggestions: 

1. Always assign the highest priority when sending and the lowest when retrying, irrespective of any priority specified in the email, if `usePriority` is enabled. . 
2. Eliminate this property entirely and only consider the priority if it is explicitly set in the email.

What are your thoughts on these options?

chibenwa (commented):
Can you please add a test with both DSNs and MT-PRIORITY ?

chibenwa:
Also I would like to see 

```
    @Test
    void remoteDeliveryShouldNotSetupDefaultValueWhenNone(DockerMockSmtp mockSmtp) throws Exception {
        AuthenticatingSMTPClient smtpClient = new AuthenticatingSMTPClient("TLS", "UTF-8");
        try {
            smtpClient.connect("localhost", jamesServer.getProbe(SmtpGuiceProbe.class).getSmtpPort().getValue());
            smtpClient.auth(PLAIN, FROM, PASSWORD);
            smtpClient.ehlo(DEFAULT_DOMAIN);
            smtpClient.mail("<" + FROM + ">");
            smtpClient.rcpt("<" + RECIPIENT + ">");
            smtpClient.sendShortMessageData("A short message...");
        } finally {
            smtpClient.disconnect();
        }

        calmlyAwait.atMost(TEN_SECONDS).untilAsserted(() -> assertThat(mockSmtp.getConfigurationClient().listMails())
            .hasSize(1)
            .extracting(Mail::getEnvelope)
            .containsExactly(Mail.Envelope.builder()
                .from(new MailAddress(FROM))
                .addRecipient(Mail.Recipient.builder()
                    .address(new MailAddress(RECIPIENT))
                    .build())
                .build()));
    }
```

As a test.

docs/modules/servers/partials/RemoteDelivery.adoc
Maxxx873: > If this property `usePriority` is enabled but does not affect emails without priorities set, I don't quite understand what it is for. Based on my understanding, this property was initially designed to hardcode email priorities through configuration, with the highest priority assigned on send and the lowest on retry. To maintain the original intent while also supporting the new priority functionality, I propose using this property to assign priorities to emails that currently lack a specified priority.
> 
> I also have a couple of alternative suggestions:
> 
> 1. Always assign the highest priority when sending and the lowest when retrying, irrespective of any priority specified in the email, if `usePriority` is enabled. .
> 2. Eliminate this property entirely and only consider the priority if it is explicitly set in the email.
> 
> What are your thoughts on these options?

I chose the third option. I left only the lowest priority for retry attempts.

chibenwa: > only consider the priority if it is explicitly set in the email.

This is what I meant.

The property is however useful when set to false as it allows tiurning off priorities when dealing with unknown third party systems when performing remote delivery mx resolution.

(Prioritiesmakes sens localy or with known 3rd party)

chibenwa:
A last obe test for remoteDelivery

GIVEN usePriority is false
WHEN i send a mail with priority 8
THEN It gets transmitted without priority

(See comlent above)

Maxxx873:
@chibenwa Thank you very much for taking the time to review the code and provide such detailed feedback. I appreciate your insights. And if anyone else has additional comments, I am happy to work on them.

chibenwa (approved):

server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpMtPriorityMessageHookTest.java
Maxxx873: > Yes true actually the mail queue implementations do not support negative priorities as of today yet I believe this is more of a mistake and mis-understanding of legacy implementors rather than a concious choice.
> 
> It would be great if we could modify the JMS mailque to support this and alter PriorityMailQueueContract accordingly, and thus having here:
> 
> While we do not violate the spec (that accounts for such rounding) we could at least plan this as an improvement.

I am currently looking for an optimal solution, considering that queues implement the The Jakarta Messaging API.
The Jakarta Messaging API defines ten levels of priority value, with 0 as the lowest priority and 9 as the highest.  Any ideas on how to use negative priorities? What might a modification or an alternative of the JMS look like?

chibenwa:
Can you squash fixups please?

Arsnael (approved):
LGTM :) Thank you for this @Maxxx873 

Good to merge.

PR #2340, opened 2024-07-07
[FIX] shouldAddSSLInformationInReceivedHeaders() test

Maxxx873:
The consumer argument of the hasOnlyOneElementSatisfying method does not include any assertions, causing the test to pass regardless of the expected string.


chibenwa (approved):

chibenwa:
(BTW could you please privately send me your mail address? Mine is `btellier[at]apache.org`.

Arsnael (approved):

PR #2341, opened 2024-07-08
Update RightManager to support setRights in reactive way

hungphan227:


mailbox/api/src/main/java/org/apache/james/mailbox/RightManager.java
```diff
@@ -170,6 +170,7 @@ public interface RightManager {
      */
     void setRights(MailboxPath mailboxPath, MailboxACL mailboxACL, MailboxSession session) throws MailboxException;
 
+    Publisher<Void> setRightsReactive(MailboxPath mailboxPath, MailboxACL mailboxACL, MailboxSession session) throws MailboxException;
```
chibenwa, line 173: We cannot use `applyRightsCommandReactive` for this job?

PR #2342, opened 2024-07-09
JAMES-4046 Upgrade Lucene

Arsnael:
Just tried to give it a shot to upgrade directly to latest lucene version from Quan's work : https://github.com/apache/james-project/pull/2315

in the lucene module I have 3 tests failing left, all returning the following error:

```
java.lang.IllegalArgumentException: cannot change field "uid" from doc values type=NUMERIC to inconsistent doc values type=NONE
```

Maybe @quantranhong1999 has an idea? :)

I have conscience that there is still some tests that have been disabled and of course didn't check mpt tests yet (one step at a time)

### DIFF
--- a/mailbox/lucene/pom.xml
+++ b/mailbox/lucene/pom.xml
@@ -103,17 +103,17 @@
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
-            <artifactId>lucene-analyzers</artifactId>
+            <artifactId>lucene-analyzers-common</artifactId>
+            <version>${lucene.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-analyzers-smartcn</artifactId>
             <version>${lucene.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
             <artifactId>lucene-core</artifactId>
-            <version>${lucene.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.lucene</groupId>
-            <artifactId>lucene-smartcn</artifactId>
             <version>${lucene.version}</version>
         </dependency>
         <dependency>
--- a/mailbox/lucene/pom.xml
+++ b/mailbox/lucene/pom.xml
@@ -103,12 +103,12 @@
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
-            <artifactId>lucene-analyzers-common</artifactId>
+            <artifactId>lucene-analysis-common</artifactId>
             <version>${lucene.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
-            <artifactId>lucene-analyzers-smartcn</artifactId>
+            <artifactId>lucene-analysis-smartcn</artifactId>
             <version>${lucene.version}</version>
         </dependency>
         <dependency>
--- a/mailbox/lucene/pom.xml
+++ b/mailbox/lucene/pom.xml
@@ -117,6 +117,12 @@
             <version>${lucene.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-misc</artifactId>
+            <version>${lucene.version}</version>
+        </dependency>
+
+        <dependency>
             <groupId>org.awaitility</groupId>
             <artifactId>awaitility</artifactId>
             <scope>test</scope>
--- a/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LenientImapSearchAnalyzer.java
+++ b/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LenientImapSearchAnalyzer.java
@@ -22,7 +22,7 @@ import java.io.Reader;
 
 import org.apache.lucene.analysis.Analyzer;
 import org.apache.lucene.analysis.TokenStream;
-import org.apache.lucene.analysis.WhitespaceTokenizer;
+import org.apache.lucene.analysis.core.WhitespaceTokenizer;
 import org.apache.lucene.analysis.shingle.ShingleFilter;
 import org.apache.lucene.util.Version;
 
@@ -47,7 +47,9 @@ public final class LenientImapSearchAnalyzer extends Analyzer {
     }
     
     @Override
-    public TokenStream tokenStream(String arg0, Reader reader) {
-        return new ShingleFilter(new UpperCaseFilter(new WhitespaceTokenizer(Version.LUCENE_31, reader)), 2, maxTokenLength);
+    protected TokenStreamComponents createComponents(String fieldName, Reader reader) {
+        WhitespaceTokenizer source = new WhitespaceTokenizer(Version.LUCENE_40, reader);
+        TokenStream filter = new ShingleFilter(new UpperCaseFilter(source), 2, maxTokenLength);
+        return new TokenStreamComponents(source, filter);
     }
 }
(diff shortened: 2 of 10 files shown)


quantranhong1999:
Hi Rene,

Thanks for putting this together!

> java.lang.IllegalArgumentException: cannot change field "uid" from doc values type=NUMERIC to inconsistent doc values type=NONE

It seems we declared the `uid` field with DocValues `NUMERIC` (to be able to search/sort) but somehow during `update` the flag (modified the FLAGS_FIELD of the document at https://github.com/apache/james-project/blob/5ffd1bd106b722dfd3cd61a38c0486637562c58b/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LuceneMessageSearchIndex.java#L1276), it seems somehow the DocValues for the `uid` field broke to `NONE`?

quantranhong1999:
I have pushed 2 commits to fix the above 3 tests and another issue: https://github.com/quantranhong1999/james-project/commits/upgrade-lucene-rene/

Now we have 4 MPT failing tests left:
image

quantranhong1999:
I am stuck at debugging the MPT test (again).

It seems in the end (at the failing assertion), searching by flag returns no UIDs leading to James just returning all possible UIDs in the mailboxes.

I tried:
  - add await in MPT test -> not work
  - TextField, Store.Yes for the flags field -> not work
  - partial update -> only work with DocValues field -> try DocValues for the flags field -> seems not suitable
  - `writer.commit` everywhere -> not work
  - apply delete / write = true when declaring the IndexReader -> not work

Do you have any other clue or idea @Arsnael?

Arsnael:
I think something is wrong with flags update in general, and maybe the search query on flags too.

On testSearch tests, I tried to modify a bit the script. I tried to add the flag flagged to just messages 3 and 4 and do a search on it, it returns 3 and 4.

But then search unflagged => returns all messages.

I tried to debug yesterday as well, I felt like maybe (aint completely sure) we were creating more documents each time with updates...

I think we might need to review the flags part of the code. In the createFlagQuery for example you can see some comments like `// lucene does not support simple NOT queries so we do some nasty hack here` . Maybe lucenesupports it now for example and the hack is creating issues now? Things like that...

 Or also I wondered why flags and messages are being in separated documents with lucene? Was it because of some limitations of the version 2? Maybe would make sense in latest version to have it all organized in just one document? It looks like to me we have it all in one doc in opensearch for example if I'm not wrong

quantranhong1999:
> I tried to debug yesterday as well, I felt like maybe (aint completely sure) we were creating more documents each time with updates...

Yes very likely, when Lucene updates the document, it would delete and create a new document.

Somehow multiple updates on the same document break the flags field's data.

> Or also I wondered why flags and messages are being in separated documents with lucene? Was it because of some limitations of the version 2? Maybe would make sense in latest version to have it all organized in just one document? It looks like to me we have it all in one doc in opensearch for example if I'm not wrong

Yes. I think we could try to refactor that. But I am not sure if we should do that in this upgrade or in another ticket.

Arsnael:
Pushing the debugging further...

I modified the search test to:
- append 4 messages
- store Flagged flag on message 3 and 4
- search messages with Flagged
- search unflagged messages

Fails on last step where it returns all messages.

By modifying the code in the search query  for flags instead of doing a not flags, I did a search on all documents with flags.

Returns 6 documents (and not 4). The original 4 plus the 2 extra that were created when adding the Flagged flag.

Where I'm confused is that according to the lucene doc, the updateDocument method is supposed to delete the current doc and then create a new one. Or here it's obvious the old one does not get deleted.

Why? Need to search more. Maybe there is an other way to do an update in recent lucene? Maybe a bug? Maybe the delete is async?... 

I might try as well just deleting and adding the document

Arsnael:
@quantranhong1999 

When I read the javadoc of IndexWriter => https://lucene.apache.org/core/9_11_1/core/org/apache/lucene/index/IndexWriter.html#close()

I understand that we might need to review our writer usage. It seems to wait for a call to close() for example before flushing out changes (however our writer is always opened)

But I tried quickly a commit() after the updateDocument(). It's costly and shouldnt be used in prod code but just for testing... Unfortunately I still got 6 documents :(

So I'm not really sure to understand anymore what we do wrong here... Could try an other time a quick refactoring where I open and close the writer after usage everywhere but... will see

Arsnael:
I think I'm gonna drop it for now...

I tried to refactor the code to open and close the IndexWriter after each use in the code, I tried to use `commit`, `forceMergeDeletes` methods, tried to explicitly delete the old doc and add the new one, tried to change the FSDirectory (MMapDirectory, NIOFSDirectory)... Does not change anything.

Tried changing the deletion policy, the merging policy...

I think there is probably something we are missing during that migration, with number of systems using lucene I hardly believe we have a bug here... but I'm a bit at a loss here.

chibenwa:
Yes those are the most relevant ones.

Arsnael:
@woj-tek if you can clear the mpt imap tests for lucene, we are likely good yes :)

Arsnael:
@woj-tek I think you got it right about what the Search.test is supposed to do yes :)

> Was there a consensus why it was done that way and if it should be changed?

No idea honestly, that Lucene implementation is very old and was done way before I join the project I believe.

But it would not change the fact that for some reason Lucene does not seem to remove the old documents when doing its updates, so it just creates new ones, and that's why the Search.test is failing.

chibenwa:
Yes for the test scenari.

I believe the current mpt test has value as it allowed catching issues regular unit tests did not. I would be extremely caitious if alterations are needed.

Btw we pass this very test with opensearch meaning this is doable on a lucene based setup...

Thanks for having a look @woj-tek !

Arsnael:
Hi @woj-tek !
Huge investigation work here, thank you very much :)

So the issue is with the StringField ID_FIELD tokenization correct? You drop in the end the ID_FIELD and update the document with the query instead. Not too sure to get what exactly happens there but looks good to me :) 

You want me to cherry-pick your commit on this PR or you want to open your own PR with the fix? :)

Thanks for the hard work, really appreciated :)

quantranhong1999:
Thanks a lot @woj-tek for your hard work and putting this together!

> PS. If someone has a formatter for IDEA that would produce sources formatted to checkstyle liking I'd be very thankfull if shared :)

There you go: https://github.com/apache/james-project/blob/master/src/site/xdoc/server/dev-build.xml#L138

Arsnael:
Hi @uschindler 

Thanks for that long explanation on how Lucene is working and what seems to be our issues. TBH I don't think people here are familiar with Lucene, it was a very old implementation made by other people and we just tried to upgrade without having much knowledge around it.

As @woj-tek said, we have different implementation already, like with OpenSearch, but seems some people like @woj-tek in our community are still using Lucene as a more lightweight approach for hosting their own mail server.

With all those information we could likely propose something better as Lucene indexing implementation then I agree. I think we could try to do this!

gautamworah96:
Hi @uschindler 

I didn't understand a few details

> Lucene does not know how the document was indexed originally, the information is not stored anywhere in index. When you then reindex the document it will aply default settings for all the fields in the Document (which is analyzed/tokenized with the default analyzer). This will transform all StringField instances to TextField.

Let's take the code [here](https://github.com/tigase/james-project/commit/e5fe4010131f085754cfadcffdb14224612bb848) as a reference for the purpose of this discussion. In the code, on L1293#[LuceneMessageSearchIndex.java](https://github.com/tigase/james-project/commit/e5fe4010131f085754cfadcffdb14224612bb848#diff-a7c2a3c5cdb7e4a2914c899409991e27df6b25ad54488f197bc533193e3a03d0), when they tried using a TermQuery on the ID field, that too failed. That atleast should've worked right? Sure Lucene does not store indexing information, but it would've stored the untokenized ID right (without info on whether or not it was tokenized)? All the TermQuery had to do was match against the terms indexed in the ID field?

Even the updates made on L1282 used the same StringField

chibenwa:
Could this get closed?

PR #2344, opened 2024-07-13
[IMPROVEMENT] results handlers also called for pushed line handlers

romainmoreau:
Some responses don't currently pass through `ProtocolHandlerResultHandler`, for instance the response to the base64 encoded identifiers after the SMTP `AUTH PLAIN` command (i.e. 235 or 535), because these responses are issued by a dynamically pushed LineHandler, so this MR adds a call to the result handlers in this case.

### DIFF
--- a/protocols/netty/src/main/java/org/apache/james/protocols/netty/BasicChannelInboundHandler.java
+++ b/protocols/netty/src/main/java/org/apache/james/protocols/netty/BasicChannelInboundHandler.java
@@ -63,7 +63,7 @@ import io.netty.util.AttributeKey;
 /**
  * {@link ChannelInboundHandlerAdapter} which is used by the SMTPServer and other line based protocols
  */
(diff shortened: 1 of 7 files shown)


chibenwa (commented):
Hello

I wion't be surprised this could have been intendeed. Line handlers are called for each line of the body of a SMTP message (a lot) and calling an iterator on each smtp line likely is non neglictible.
On the counter side, do we have use cases that relies on ech smtp line of the message to go through  resultHandlers?

romainmoreau:
My use case was getting the result of `UsersRepositoryAuthHook`; after taking a second look, I noticed there was already the `HookResultHook` mecanism but it's partially implemented on `AuthCmdHandler`, so I created #2345.

About the data lines there is already `DataLineFilter` that seem to cover this use case, so indeed this PR is not needed, I'm closing it.

PR #2345, opened 2024-07-14
[FIX] HookResultHook added to marker interfaces of AuthCmdHandler

romainmoreau:
`AuthCmdHandler` supports `HookResultHook` in `wireExtensions` and with its `rHooks` field but not in `getMarkerInterfaces` so `rHooks` will always be empty; this PR fixes this to be able to get `AuthHook` results in `HookResultHook` (e.g. getting `UsersRepositoryAuthHook` result).


chibenwa (approved):

quantranhong1999 (approved):

Arsnael (approved):
LGTM, thank you for the little fix :)

PR #2366, opened 2024-07-26
JAMES-4054 Add X-SMIME-Status in SMIMECheckSignature process

hungphan227:
Will update the ticket number later


chibenwa:
Please let's rely on https://certificate.nikhef.nl/info/smime-manual.html

I also think working from x509 certificates directly would ease configuration a lot!

chibenwa (commented):
`keystore2` please share also the associated public and private PEM files to enable working with it using openssl.

server/mailet/integration-testing/src/main/java/org/apache/james/mailets/TemporaryJamesServer.java
```diff
@@ -139,7 +139,8 @@ public static Builder builder() {
         "pop3server.xml",
         "recipientrewritetable.xml",
         "usersrepository.xml",
-        "smime.p12");
+        "smime.p12",
+        "keystore2");
```
quantranhong1999, line 143: Maybe a better name for this keystore (to reflect the SMIME use case)?

quantranhong1999 (approved):
We need ticket number

Arsnael (approved):

chibenwa:
Please let me ~30 mins for review

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/crypto/SMIMECheckSignatureIntegrationTest.java
```diff
@@ -0,0 +1,142 @@
...
+            .sendMessageWithHeaders(FROM, RECIPIENT,
+                ClassLoaderUtils.getSystemResourceAsString("eml/mail_with_no_signature.eml"));
+
+        testIMAPClient.connect(LOCALHOST_IP, jamesServer.getProbe(ImapGuiceProbe.class).getImapPort())
+            .login(RECIPIENT, PASSWORD)
+            .select(TestIMAPClient.INBOX)
+            .awaitMessage(awaitAtMostOneMinute);
+        assertThat(testIMAPClient.readFirstMessage()).containsSequence("X-SMIME-Status: Not signed");
+    }
+
```
chibenwa, line 141: Can you add a mail for an unknown CA?

chibenwa (commented):
Overall good to me so far

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/crypto/SMIMECheckSignatureIntegrationTest.java
chibenwa: In such case we shall report `X-SMIME-Status: Unknown`

hungphan227: Do you mean the signature is valid but not trusted?

chibenwa: Nevermid, after re reading RFC-9219 that served as insoiration this do not make sens.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/crypto/SMIMECheckSignatureIntegrationTest.java
```diff
@@ -0,0 +1,142 @@
...
+    public void setup(@TempDir File temporaryFolder) throws Exception {
+        MailetContainer mailetContainer = MailetContainer.builder()
+            .putProcessor(CommonProcessors.root())
+            .putProcessor(CommonProcessors.error())
+            .putProcessor(ProcessorConfiguration.transport()
+                .addMailet(MailetConfiguration.BCC_STRIPPER)
+                .addMailet(MailetConfiguration.builder()
+                    .mailet(SMIMECheckSignature.class)
+                    .matcher(All.class)
+                    .addProperty("keyStoreFileName", temporaryFolder.toPath().resolve("conf").resolve("smime_cert_keystore").toAbsolutePath().toString())
```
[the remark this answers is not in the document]
hungphan227: added .md file to show used commands

hungphan227:
I am a bit curious about this case: When I change email address domain in "From" header to make it different than the email address domain in the signature, the test still pass. I am not sure if it is the right behavior or not.

chibenwa (approved):

PR #2367, opened 2024-07-27
JAMES-4044 Add a cache for LDAP attribute matchers

chibenwa:
(help welcome to address review comments if needed)

### DIFF
--- a/server/mailet/ldap/src/main/java/org/apache/james/transport/matchers/CacheSettings.java
+++ b/server/mailet/ldap/src/main/java/org/apache/james/transport/matchers/CacheSettings.java
@@ -1,0 +1,74 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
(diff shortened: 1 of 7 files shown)


server/mailet/ldap/src/main/java/org/apache/james/transport/matchers/CacheSettings.java
```diff
@@ -0,0 +1,106 @@
...
+        int separatorPosition = condition.indexOf('=');
+        String key = condition.substring(0, separatorPosition).toLowerCase(Locale.US);
+        String value = condition.substring(separatorPosition + 1);
+        return Pair.of(key, value);
+    }
+
+    private final Duration duration;
+    private final int size;
+
+    public CacheSettings(Duration duration, int size) {
```
mbaechler, line 75: aren't we using java 21 yet? Why don't we use Records for such things?

chibenwa: Fair ^^

quantranhong1999 (approved):

Arsnael (approved):

Arsnael:
Green. Can we squash this?

chibenwa:
Feel free yes

PR #2372, opened 2024-08-08
JAMES-4055 Configure S-MIME check signature with pem files

hungphan227:


docs/modules/servers/partials/SMIMECheckSignature.adoc
```diff
@@ -38,6 +41,11 @@ compatible with standard email clients like Outlook Express and Thunderbird.
 store path.
 * keyStorePassword (default: ""): Certificate store password.
 
+If fileType is "pem". The following configuration parameters would be used:
+
+* keyStoreType: same as above.
```
chibenwa, line 46: If using pem why do i need  to specify keystoretype?

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
```diff
@@ -0,0 +1,137 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(KeyStoreHolderFactory.class);
+    private static final KeyStoreHolderFactory INSTANCE = new KeyStoreHolderFactory();
+    private static final String KEYSTORE_FILE_TYPE = "keystore";
+    private static final String PEM_FILE_TYPE = "pem";
+
+    private KeyStoreHolderFactory() {
+    }
+
+    public static KeyStoreHolderFactory getInstance() {
+        return INSTANCE;
```
chibenwa, line 50: Is singlton pattern needed? Why not just call "INSTANCE"?

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    public KeyStoreHolder createKeyStoreHolder(MailetConfig config) throws MessagingException {
+        try {
+            initJCE();
+            String fileType = config.getInitParameter("fileType");
+            if (fileType == null) {
+                fileType = KEYSTORE_FILE_TYPE;
```
chibenwa, line 58: Method extraction to avoid variable reallocation.

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
```diff
@@ -0,0 +1,137 @@
...
+            switch (fileType) {
+                case KEYSTORE_FILE_TYPE:
+                    return createFromKeyStoreFile(config);
+                case PEM_FILE_TYPE:
+                    return createFromPemFile(config);
+                default:
+                    throw new RuntimeException("Unsupported file type " + fileType);
+            }
+        } catch (Exception e) {
+            throw new MessagingException("Error loading the trusted certificate store", e);
```
chibenwa, line 69: Why not runtime like above?

server/mailet/integration-testing/src/main/java/org/apache/james/mailets/TemporaryJamesServer.java
```diff
@@ -140,7 +140,8 @@ public static Builder builder() {
         "recipientrewritetable.xml",
         "usersrepository.xml",
         "smime.p12",
-        "smime_cert_keystore");
+        "smime_cert_keystore",
+        "trusted_certificate.pem");
```
chibenwa, line 144: Please do 2 tests: one with pem, one with keystore

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
hungphan227: In the original code, everything is in this try catch block:

```
try {
    ........
} catch (Exception e) {
    throw new MessagingException("Error loading the trusted certificate store", e);
}
```
As I understand, somewhere on top use MessagingException for a specific purpose.

server/mailet/integration-testing/src/main/java/org/apache/james/mailets/TemporaryJamesServer.java
hungphan227: added two different test classes (one with keystore, one with pem) for SMIMECheckSignature

quantranhong1999 (commented):
Nothing to add. Should be a fixup commit instead of a force push IMO.

Arsnael (approved):
LGTM !

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
```diff
@@ -0,0 +1,125 @@
...
+public class KeyStoreHolderFactory {
+    private static final Logger LOGGER = LoggerFactory.getLogger(KeyStoreHolderFactory.class);
+    private static final String KEYSTORE_FILE_TYPE = "keystore";
+    private static final String PEM_FILE_TYPE = "pem";
+
+    public static KeyStoreHolder createKeyStoreHolder(MailetConfig config) throws MessagingException {
+        try {
+            initJCE();
+            String fileType = config.getInitParameter("fileType");
+            if (fileType == null) {
```
chibenwa, line 49: Use optionals fir null check please

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
```diff
@@ -0,0 +1,125 @@
...
+
+public class KeyStoreHolderFactory {
+    private static final Logger LOGGER = LoggerFactory.getLogger(KeyStoreHolderFactory.class);
+    private static final String KEYSTORE_FILE_TYPE = "keystore";
+    private static final String PEM_FILE_TYPE = "pem";
+
+    public static KeyStoreHolder createKeyStoreHolder(MailetConfig config) throws MessagingException {
+        try {
+            initJCE();
+            String fileType = config.getInitParameter("fileType");
```
chibenwa, line 48: String typing: I would use an enum to represent valid choices

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
```diff
@@ -0,0 +1,125 @@
...
+    private static KeyStoreHolder createFromKeyStoreFile(MailetConfig config) throws Exception {
+        String type = config.getInitParameter("keyStoreType");
+        String fileName = config.getInitParameter("keyStoreFileName");
+        String password = config.getInitParameter("keyStorePassword");
+
+        if (fileName != null) {
+            return createFromKeyStoreFile(fileName, password, type);
+        } else {
+            LOGGER.info("No trusted store path specified, using default store.");
+            return createFromKeyStoreFile(password);
```
chibenwa, line 74: createFromPlatformKeyStoreFile ?

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
hungphan227: just an usual way to handle default case of switch

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderFactory.java
hungphan227: Isn't it "KeyStoreFile" enough? Don't understand what "platform" mean.

chibenwa: The name you propose is confusing and duplicated.
Here platform means defaut java trust store.
A specific, discriminative name is always better.

hungphan227: oh sorry I misunderstood. Anyway, I have that method removed.

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+    public static final String KEY_STORE_PASSWORD_DEFAULT_VALUE = "";
+
+    public static class Builder {
+        private Optional<String> fileType;
+        private Optional<String> keyStoreType;
+        private Optional<String> keyStoreFileName;
+        private Optional<String> keyStorePassword;
+        private Optional<String> pemFileName;
+
+        public Builder setFileType(Optional<String> fileType) {
```
chibenwa, line 43: Why not use the strong type here?

mailet/crypto/src/main/java/org/apache/james/transport/KeyStoreHolderConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+            return this;
+        }
+
+        public Builder setPemFileName(Optional<String> pemFileName) {
+            this.pemFileName = pemFileName;
+            return this;
+        }
+
+        public KeyStoreHolderConfiguration build() {
+            return new KeyStoreHolderConfiguration(this.fileType.map(type -> KeyFileType.valueOf(type.toUpperCase())).orElse(KeyFileType.KEYSTORE),
```
chibenwa, line 69: Move `this.fileType.map(type -> KeyFileType.valueOf(type.toUpperCase())).orElse(KeyFileType.KEYSTORE)` into a `KeyFileType::parse` method...

chibenwa, line 69: We need checks on the state IMO:
 - Does specifying pem file with keystore makes sense?
 - Does specifying keystore stuff when dealing with PEM make sense?

hungphan227: fixed

quantranhong1999 (commented):
LGTM.

Do we have a test for the "second" signature format?

Arsnael:
> Do we have a test for the "second" signature format? (could be in another PR I guess?)

I would say logically any code implemented should have its tests

hungphan227:
> LGTM.
> 
> Do we have a test for the "second" signature format? (could be in another PR I guess?)

will create another PR

Arsnael:
If it's just adding tests for something implemented here add the tests here please. make more sense IMO.

hungphan227:
> If it's just adding tests for something implemented here add the tests here please. make more sense IMO.

The fix for customer issue does not really relate to this ticket. I will create another PR

PR #2373, opened 2024-08-12
JAMES-4046 Upgrade Lucene to version 9.11.1

woj-tek:
A followup work on upgrading James to use latest Lucene based on @quantranhong1999 and @Arsnael work in #2315 and #2342

JIRA: https://issues.apache.org/jira/browse/JAMES-4046


Arsnael:
I agree, let's start by getting a green build with this it seems good enough for the Lucene update for now. Then for something more correct and performant we can apply the useful comments form the Lucene community on following PRs. 

Can we mark this work Ready for Review @woj-tek ? :)

mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LuceneMessageSearchIndex.java
```diff
@@ -1266,50 +1267,58 @@ public Mono<Void> update(MailboxSession session, MailboxId mailboxId, List<Updat
     }
 
     private void update(MailboxId mailboxId, MessageUid uid, Flags f) throws IOException {
+        var flagsID = createFlagsIdField(mailboxId, uid);
+        var term = new Term(ID_FIELD, flagsID);
+
+
+
+
+        // TODO: remove! just for debugging purposes
```
Arsnael, line 1277: Unecessary line skips, and is that TODO still in place here? The log traces?

quantranhong1999:
> I agree, let's start by getting a green build with this it seems good enough for the Lucene update for now. Then for something more correct and performant we can apply the useful comments form the Lucene community on following PRs.

+1

mailbox/lucene/pom.xml
```diff
@@ -103,7 +103,12 @@
         </dependency>
         <dependency>
             <groupId>org.apache.lucene</groupId>
-            <artifactId>lucene-analyzers</artifactId>
+            <artifactId>lucene-analysis-common</artifactId>
+            <version>${lucene.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.lucene</groupId>
+            <artifactId>lucene-analysis-smartcn</artifactId>
```
uschindler, line 111: Is this still needed? Below you no longer use the Chinese code.

mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LenientImapSearchAnalyzer.java
```diff
@@ -47,7 +44,9 @@ public LenientImapSearchAnalyzer() {
     }
     
     @Override
-    public TokenStream tokenStream(String arg0, Reader reader) {
-        return new ShingleFilter(new UpperCaseFilter(new WhitespaceTokenizer(Version.LUCENE_31, reader)), 2, maxTokenLength);
+    protected TokenStreamComponents createComponents(String fieldName) {
```
uschindler, line 47: This whole class can be removed. Better use the builder pattern of CustomAnalyzer. Subclassing analyzers like that is no longer recommended.

mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/StrictImapSearchAnalyzer.java
```diff
@@ -50,12 +48,14 @@ public StrictImapSearchAnalyzer(int minTokenLength, int maxTokenLength) {
...
 
-   /**
-    * @see org.apache.lucene.analysis.Analyzer#tokenStream(java.lang.String, java.io.Reader)
-    */
-   @Override
-   public TokenStream tokenStream(String fieldName, Reader reader) {
-       return new NGramTokenFilter(new UpperCaseFilter(new SentenceTokenizer(reader)), minTokenLength, maxTokenLength);
-   }
+    /**
+     * @see org.apache.lucene.analysis.Analyzer#tokenStream(java.lang.String, java.io.Reader)
```
uschindler, line 52: Same here.

Arsnael (approved):

quantranhong1999 (approved):

mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/StrictImapSearchAnalyzer.java
```diff
@@ -18,44 +18,40 @@
...
-   }
-   
+    static Analyzer getAnalyzer(int minTokenLength, int maxTokenLength) {
+        try {
+            return CustomAnalyzer.builder()
+                    .withTokenizer(WhitespaceTokenizerFactory.NAME)
+                    .addTokenFilter(UpperCaseFilterFactory.NAME)
+                    .addTokenFilter(NGramFilterFactory.NAME,
+                            "minGramSize", String.valueOf(minTokenLength),
+                            "maxGramSize", String.valueOf(maxTokenLength),
```
chibenwa, line 50: Performance of this is likely to be terrible. Maybe we should not make this the default?

chibenwa (approved):
Thanks all for pulling this out! updating only the flags sub-document indeed is smart.

(August vacation review - sorry for not digging deeper onto the topic...)

mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LuceneMessageSearchIndex.java
```diff
@@ -1339,11 +1351,11 @@ public Mono<Void> deleteAll(MailboxSession session, MailboxId mailboxId) {
...
     public void delete(MailboxId mailboxId, MessageRange range) throws IOException {
-        BooleanQuery query = new BooleanQuery();
-        query.add(new TermQuery(new Term(MAILBOX_ID_FIELD, mailboxId.serialize())), BooleanClause.Occur.MUST);
-        query.add(createQuery(range), BooleanClause.Occur.MUST);
+        BooleanQuery.Builder queryBuilder = new BooleanQuery.Builder();
+        queryBuilder.add(new TermQuery(new Term(MAILBOX_ID_FIELD, mailboxId.serialize())), BooleanClause.Occur.MUST);
+        queryBuilder.add(createQuery(range), BooleanClause.Occur.MUST);
 
-        writer.deleteDocuments(query);
+        writer.deleteDocuments(queryBuilder.build());
```
chibenwa, line 1358: Shall we also explicitly delete flags subdocuments?


I get that by not having messages exposing them they cannot be seen from a user perspective. But wouldn't they be leaked?

uschindler:
Looks fine with the custom analyzer. Maybe the extra classes can be removed at some point. Basically a singleton in some static class would be fine, too. No need to create a new one every time.

uschindler:
P.S.: The IOException comes from the resources files. If there are no resources, the CustomAnalyzer won't  throw UOE, so can be warpped with an AssertionError.

mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/StrictImapSearchAnalyzer.java
uschindler: What do you mean with performance? Creating the analyzer? I would change the analyzer to be a singleton.

Arsnael:
@woj-tek looks good to me for merging! Just need to cleanup that commit history. You think you can help squash the fixups, add the jira ticket number on your commits as well, and removing the WIP part of commit description of upgrade lucene to 9.11.1 ? If not I could do it later as well it's ok :)

Thanks to the Lucene community for their analysis and support on this work, it's been very helpful as well!

Arsnael:
> EDIT: though, should I remove StrictImapSearchAnalyzer as suggested beforehand?

You can I think, not used anyways :)

Arsnael:
> I wonder if we should handle it somehow internally (just drop the index as such a big upgrade is not supported either way or maybe use external library like https://github.com/hakanai/luceneupgrader) or handle it externally (most likely dropping version as well)?

Good question !

There is a webadmin endpoint to reindex all messages already for that purpose... On such a big version upgrade of Lucene I would think it's probably better to drop the index and reindex all messages in a clean index using this:  https://james.apache.org/server/manage-webadmin.html#ReIndexing_all_mails

Prefer the rebuildAllNoCleanup mode to skip the cleanup phase, as you would restart on an empty index, so should be better :)

I expect the James running with Lucene to be minimal enough so that there is not that much emails to reindex, shouldn't be too long. I never run that task with Lucene, but ran it multiple times with OpenSearch for reindex :)

Would that be acceptable @woj-tek ? Maybe you could try?

uschindler:
You have to delete the index anyways, because an empty index would still cause the message.

So I'd suggest rebuildAllNoCleanup, but you need to make sure to completely delete the index directory first.

I'd not recommend to use that Index upgrader code mentioned above. This should only be used as ultimate action if you for example have no source documents anymore, which is not the case here. In addition, the resulting index would have no doc values, so would break anyways. I had customers that needed to upgrade such old indexes with not source file backup. In that case there were also on the fly migrations needed (like converting numeric fields to points, adding doc values). Those migrations are doable but it's a big issue and aren't worth to do on small indexes.

Arsnael:
> One more thing, and this was more the intention of my previous question - should we be bothered in a way with doing it automatically in the indexer (e.g. handle the exception and remove the index and call re-index automatically) - it would be more "automatic" and possibly more "user friendly" but I'm very much weary about doing things automatically and prefer to be more explicit (thus in my/our case manually dropping and re-indexing is better)

I honestly don't think we should automatically force migrate an index, there could be many reasons that an admin might not want to behind his back. Think it's always better that this kind of operations is done by an admin aware of the situation.

However you make me think that we should add an entry in upgrade-instructions.md file, explaining that for people using Lucene we bumped up to version 9.11.1, which is a breaking change on the index, and that they should drop the index and reindex manually, with the link to the wedadmin task in question :)

Do you think you could add it @woj-tek as part of this PR? :)

> And yet another thing, slightly touching on the above - should we call automatic re-indexing if we detect no-/empty-index maybe?

That's an other matter. I feel like such a decision would probably better deserve a discussion on the ML for example before being taken

uschindler:
> However you make me think that we should add an entry in upgrade-instructions.md file, explaining that for people using Lucene we bumped up to version 9.11.1, which is a breaking change on the index, and that they should drop the index and reindex manually, with the link to the wedadmin task in question :)

Maybe catch the org.apache.lucene.index.IndexFormatTooOldException and log a useful instruction for the sysadmin.

Uwe

chibenwa:
Does it crash at startup or runtime?

Runtime would be problematic, i'd prefer a clean error in the starting sequence - easier to spot. Bonus points for explaining what to do in that log of course.

+1 to rebuild the index: as explaibed james can do this easily.

This should be also documented in upgrade-instructions.md

uschindler:
Write lock files are never deleted, unless it is using SimpleFSLockFactory (only available in very old versions).

Basically the lock file is not a signal that an index is locked, the file is just an empty container to place a kernel file lock on it. 

So lock files are always visible, also on unlocked indexes.

uschindler:
If your code still uses SimpleFSLock when creating the directory, please use the modern replacement.

uschindler:
https://lucene.apache.org/core/9_11_1/core/org/apache/lucene/store/NativeFSLockFactory.html

uschindler:
> However, it's still required to call `writer.close()` when shutting down?

Yes. If you don't close the writer it won't commit. So either call close or commit. Of course close is more correct, as it will close all files.

chibenwa:
Sorry for long reaction times... Coming back from vacations, at work tomorrow. Being off the grid was definitly necessary!

>  James indeed supports @PreDestroy

We took that useful bit of infrastructure from defunct Onami project.

> docker-compose and it sends SIGKILL after 10s grace period by default

THANKS for cathing this! THis ticks the unsolved https://issues.apache.org/jira/browse/JAMES-3931 ticket off my inbox.

> Increasing grace period I found out that the shutdown takes almost 20s. Bulk of the shutdown comes from disposing services where each takes ~2s (and with 8 services it totals to 16s already)

This burns down to the Netty gracefull shutdown process, which takes 2 seconds sleep for each event loops processed sequentially.

+1 this is an issue, I regulary get annoyed with this.

Severall things here:
 - a. gracefull shutdown can be disabled via configuration.
 - b. We could allow configuration of the shutdown period, Netty allows for this.
 - c. I'm peacemealed regarding sharing the event loop across all Netty servers. One shared event loop for sure is a Netty performance recommendation (less context switch, less threads) and in this case means only one shut down. Downside is that it brings risks (blocking the event loop would be truely disastrous). Additionally to implement it we would need to find the right way to express "share the event loop" in the configuration. I would feel confident implementing this with Guice but I would not be able to conduct the spring wiring for this enhancement.
 - d. Parallelize the shutdown process. IE cheat on the guice wiring to register Netty components, register them, and do this in parallel. Here we need to be mindful that it do not mess up with the order of the shutdown - which relies on listening the order of instanciation of Guice components.
 - e. Start less TCP servers. EG ban startTLS for SMTP and IMAP, no POP3...

I saw a mail about a JIRA issue on the topic, I will answer there too. Likely d. is the simplest fix, and I can propose a fix somewhere this week.

chibenwa:
> GH seems more convenient :) )

Well, sure but the ASF also closely monitor mailing list activity...

My take on this, is that it is nice to state on the mailing list when important topics are discussed on github like you did ^^

chibenwa:
> How does that fit into whole organization principle?

I mean this is a metric reported as part of the board reports and is used to evaluate project health.

Arsnael:
Can we merge this then?

chibenwa:
Here we go!

Thanks to all the contributors for making this much needed update, and related fix happen!

PR #2376, opened 2024-08-14
JAMES-4057 Enable fuzziness search

quantranhong1999:
Fuzziness search helps improve search experience as allow typos in user input. cf: https://opensearch.org/docs/latest/query-dsl/full-text/match/#fuzziness

We support TextCriterion and SubjectCriterion but not HeaderCriterion as of now.


quantranhong1999:
Do you think we should support the fuzzy search for `HeaderCriterion` (e.g. address searching)?

docs/modules/servers/pages/distributed/configure/opensearch.adoc
```diff
@@ -97,6 +97,11 @@ turning off headers indexing result in non-strict compliance with the IMAP / JMA
 If the message is not found, it will fall back to the old behavior (The message will be indexed from the blobStore source)
 Default to false.
 
+| opensearch.text.fuzziness.search
+| Use fuzziness on text searches. This option helps to correct user typing mistakes.
+
+Default to true.
```
chibenwa, line 103: I'm peacemealed: have this been performance tested both on the indexing and search part?

Until it is I would rather be defensive and keep this disabled by default.

Arsnael (commented):
Read, nothing to add.

I would agree that this deserves a perf run though :)

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2376/1/testReport/junit/org.apache.james.mpt.imapmailbox.elasticsearch/OpenSearchUidSearchOnIndexTest/testSearchAtomsUS/

Is that related, or unstable test?

docs/modules/servers/pages/distributed/configure/opensearch.adoc
quantranhong1999: > on the indexing and search part?

This change only affects the search phase.

> have this been performance tested

Benchmark on Email/query with random text: https://github.com/linagora/james-gatling/blob/master/src/main/scala-2.13/org/apache/james/gatling/jmap/rfc8621/JmapEmail.scala#L49

Fuzzy search off:
image

Fuzzy search on:
image

p99 94ms -> p99 173ms: indeed with fuzzy on, the OpenSearch query is slower, but it is still acceptable IMO.

quantranhong1999:
> Is that related, or unstable test?

It was related. Because with fuzzy search we allow more flexible input that leads to more flexible results.

Therefore I disable fuzzy search by default, and let the admin choose himself.

docs/modules/servers/pages/distributed/configure/opensearch.adoc
chibenwa: The email query shall exercise text search for this to be relevant. Is it the case?

quantranhong1999: Yes it is.

Arsnael (approved):

mailbox/opensearch/src/test/java/org/apache/james/mailbox/opensearch/OpenSearchOptimizeMoveAndFuzzySearchIntegrationTest.java
```diff
@@ -0,0 +1,91 @@
...
+import org.apache.james.mailbox.model.ComposedMessageId;
+import org.apache.james.mailbox.model.MailboxPath;
+import org.apache.james.mailbox.model.SearchQuery;
+import org.apache.james.mime4j.dom.Message;
+import org.junit.jupiter.api.Test;
+import org.opensearch.client.opensearch._types.query_dsl.QueryBuilders;
+
+import reactor.core.publisher.Flux;
+
+class OpenSearchOptimizeMoveAndFuzzySearchIntegrationTest extends OpenSearchIntegrationTest {
```
chibenwa, line 38: In general running more test isn't much of an issue but OpenSearch tests are notoriously long.

What are the benefits of running the whole test suite with fuziness, all of the time?

I'd run only the test specific to fuziness in order to cut build time. We would save likely several minutes I believe.

chibenwa (approved):

PR #2378, opened 2024-08-18
Proposal to Extend Priority Range for Mail Queues

Maxxx873:


chibenwa (approved):

chibenwa:
```
The Jakarta Messaging API defines ten levels of priority value, with 0 as the lowest priority and 9 as the highest. Clients should consider priorities 0-4 as gradations of normal priority and priorities 5-9 as gradations of expedited priority. Priority is set to 4 by default.
```

I guess you raised the topic on the mailing list but this means the JMS implementation will not handle well 19 priority levels.

This PR is IMO great as the JMS constraints was leaking to the SMTP API (thus potentially impacting all other  implementations) yet this still means we need to adapt the JMS implementation in order to handle those lower negative priorities.

quantranhong1999 (approved):

Arsnael (approved):

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2378/1/

Related test failures :)

Arsnael (approved):

Arsnael:
Thanks for the nice contribution @Maxxx873 :)

PR #2380, opened 2024-08-19
Improving James shutdown speed

woj-tek:
(followup to: https://github.com/apache/james-project/pull/2373#issuecomment-2295396500) 

Initial work on making shutdown speed little better, main changes:
* make shutdown parallel
* decrease netty grace period (either to be reversed or made configurable)

(added "cosmetics" with information bout startup and shutdown time)

### DIFF
--- a/protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractAsyncServer.java
+++ b/protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractAsyncServer.java
@@ -22,9 +22,12 @@ import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.james.protocols.api.ProtocolServer;
 import org.apache.james.util.concurrent.NamedThreadFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.github.fge.lambdas.Throwing;
 import com.google.common.collect.ImmutableList;
@@ -51,6 +54,8 @@ import io.netty.util.concurrent.ImmediateEventExecutor;
  * Abstract base class for Servers which want to use async io
  */
 public abstract class AbstractAsyncServer implements ProtocolServer {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractAsyncServer.class);
 
     public static final int DEFAULT_IO_WORKER_COUNT = Runtime.getRuntime().availableProcessors() * 2;
     public static final int DEFAULT_BOSS_WORKER_COUNT = 2;
@@ -175,6 +180,7 @@ public abstract class AbstractAsyncServer implements ProtocolServer {
     
     @Override
     public synchronized void unbind() {
+        LOGGER.trace("Unbinding service: {}", this);
         if (!started) {
             return;
         }
@@ -184,13 +190,15 @@ public abstract class AbstractAsyncServer implements ProtocolServer {
         bossGroup.ifPresent(boss -> futures.add(boss.shutdownGracefully()));
 
         if (workerGroup != null) {
-            futures.add(workerGroup.shutdownGracefully());
+            futures.add(workerGroup.shutdownGracefully(100, 3000, TimeUnit.MILLISECONDS));
         }
 
         futures.add(channels.close());
 
         if (gracefulShutdown) {
-            futures.forEach(Throwing.<Future<?>>consumer(Future::await).sneakyThrow());
+            futures.stream()
+                    .parallel()
+                    .forEach(Throwing.<Future<?>>consumer(Future::await).sneakyThrow());
         }
 
         started = false;
--- a/protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractAsyncServer.java
+++ b/protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractAsyncServer.java
@@ -47,7 +47,8 @@ import io.netty.channel.socket.SocketChannel;
 import io.netty.channel.socket.nio.NioServerSocketChannel;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.ImmediateEventExecutor;
-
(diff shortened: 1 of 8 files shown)


server/protocols/protocols-library/src/main/java/org/apache/james/protocols/lib/ProtocolHandlerChainImpl.java
```diff
@@ -158,9 +158,6 @@ public <T> Optional<T> getFirstHandler(Class<T> type) {
      */
     @Override
     public void destroy() {
-        LinkedList<ProtocolHandler> lHandlers = getHandlers(ProtocolHandler.class);
-        for (ProtocolHandler handler : lHandlers) {
-            handler.destroy();
-        }
+        getHandlers(ProtocolHandler.class).stream().parallel().forEach(ProtocolHandler::destroy);
```
chibenwa, line 161: Is this making a difference? I would expect those handlers to not be linked to netty...

server/container/guice/common/src/main/java/org/apache/james/GuiceJamesServer.java
```diff
@@ -104,11 +106,13 @@ public void start() throws Exception {
     }
 
     public void stop() {
+        final double shutdownStart = System.currentTimeMillis();
```
chibenwa, line 109: Use Guava's Stopwatch ?

protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractAsyncServer.java
```diff
@@ -184,13 +190,15 @@ public synchronized void unbind() {
...
+            futures.add(workerGroup.shutdownGracefully(100, 3000, TimeUnit.MILLISECONDS));
         }
 
         futures.add(channels.close());
 
         if (gracefulShutdown) {
-            futures.forEach(Throwing.<Future<?>>consumer(Future::await).sneakyThrow());
+            futures.stream()
+                    .parallel()
+                    .forEach(Throwing.<Future<?>>consumer(Future::await).sneakyThrow());
```
chibenwa, line 201: I'd prefer Reactor to handle concurrency:

```
Flux.fromIterables(futures)
   .flatmap(future -> Mono.fromRunnable(() -> future.await()))
   .then()
   .block();
```

protocols/netty/src/main/java/org/apache/james/protocols/netty/AbstractAsyncServer.java
```diff
@@ -184,13 +190,15 @@ public synchronized void unbind() {
         bossGroup.ifPresent(boss -> futures.add(boss.shutdownGracefully()));
 
         if (workerGroup != null) {
-            futures.add(workerGroup.shutdownGracefully());
+            futures.add(workerGroup.shutdownGracefully(100, 3000, TimeUnit.MILLISECONDS));
```
chibenwa, line 193: Being able to differenciate the timeout from the quiet period indeed is a good finding!

I agree having a short quiet period, and longer timeouts. Though 100ms looks low, maybe 500ms - personal tastes...

Also we need to name magic constants here so that the intent can be understood without having to pro-actively read the Netty documentation.

```
futures.add(workerGroup.shutdownGracefully(SHUTDOWN_QUIET_PERIOD, SHUTDOWN_TIMEOUT, 3000, TimeUnit.MILLISECONDS));
```

chibenwa (commented):
This PR delivers partial work, as it do paralelise only within the scope of a signle protocole.

I would prefer a solution that paralelise also shutdown across protocols, IE stop SMTP+POP3+IMAP at the same time.

chibenwa:
> I would prefer a solution that paralelise also shutdown across protocols, IE stop SMTP+POP3+IMAP at the same time.

The solution I had in mind is achievable with Guice but way harder with Spring.

I agree with shutting down each of the protocols sequentially as proposed here (until we get rid of Spring!)

chibenwa:
> Could you give some pointers?

The idea I had was to have a `ServerDestroyer` (names could be improved) wich takes as an input `Set<AbstractAsyncServerFactory>` (injected) from there gets the list of servers, and iterates over it (reactor) to close things in parallel. We would then remove the @Predestroy annotation in the factory.

The issue is that unless some further Spring code is written there won't be any cleanup ran with Spring.

I would have kept this as an applicative concern and avoided to modify onami code.

> I guess you referred to this when mentioning proper order

Not just that but let's say it did lead to horrible bugs!

> Is there any plan to do so? Any ETA or depreciation of Spring elements?

Few releases in the future. This topic was discussed on the mailing list.

CF https://www.mail-archive.com/server-dev@james.apache.org/msg73550.html

server/protocols/protocols-library/src/main/java/org/apache/james/protocols/lib/ProtocolHandlerChainImpl.java
chibenwa: Yes please. Handlers shall be quick to dispose. (+ I'm opinionated on parallel streams... - sorry)

server/protocols/protocols-library/src/main/java/org/apache/james/protocols/lib/netty/AbstractServerFactory.java
```diff
@@ -52,9 +54,7 @@ public void configure(HierarchicalConfiguration<ImmutableNode> config) {
     @PostConstruct
     public void init() throws Exception {
         servers = createServers(config);
-        for (AbstractConfigurableAsyncServer server: servers) {
-            server.init();
-        }
+        servers.parallelStream().forEach(Throwing.consumer(AbstractConfigurableAsyncServer::init).sneakyThrow());
```
chibenwa, line 57: Parallel streams spotted?

chibenwa (approved):

chibenwa:
> I'm not fluent in ASF inner workings but I guess a vote on a list?

Yes

https://www.apache.org/foundation/voting.html

Arsnael (approved):

quantranhong1999 (approved):

PR #2381, opened 2024-08-19
Explicit commit after running reindex

woj-tek:
As per mailinglist:

>>    should we finish the operation with the explicit "commit" operation?
> +1
> 
> We could add a `postReindexing()` method for search indexes which commits for Lucene and noop for Opensearch + scanning.
> 
> Then calling it in reindexing task would be a no brainer.
> 
> Do you think you can contribute this?--
> 
> Best regards,
> 
> Benoit TELLIER

### DIFF
--- a/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LuceneMessageSearchIndex.java
+++ b/mailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LuceneMessageSearchIndex.java
@@ -1351,6 +1351,15 @@ public class LuceneMessageSearchIndex extends ListeningMessageSearchIndex {
     }
 
     @Override
+    public void postReindexing() {
+        try {
+            commit();
+        } catch (IOException e) {
+            throw new RuntimeException("Error while commiting to index", e);
+        }
+    }
+
+    @Override
     public Mono<Flags> retrieveIndexedFlags(Mailbox mailbox, MessageUid uid) {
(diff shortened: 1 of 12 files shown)


chibenwa (approved):

Arsnael (approved):

Arsnael:
Seen those tests failing twice in a row. related?

https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2381/4/testReport/

chibenwa:
> do I remember correctly and such improvement could be considered?

+1

Though the CI tests needs to be as extensive as possible.

chibenwa (approved):
(as always I hate mocks ^^ )

+1

chibenwa:
Yet more mocks

To be fair those are used as last resorts. Generally we favor asserting the actual outcome, which is more resilient to change.

Here it is likely that spawning a full search index was deemed unnecessary and too expensive hence mocked used.

chibenwa:
> Of course, it's better to rely on (correct) outcome but if it could catch spurious calls then I'd say it's beneficial as well :)

Agreed.

(To a similar end, instrumenting the Cassandra driver to programaticaly intercept CQL calls was valuable and contributed to limit count of mocks.)

Arsnael:
Likely unrelated this time indeed, I restarted a build

chibenwa:
> Are the builds / surefire "fail fast", i.e. they stop the run on the first module with failing tests?

Yes

PR #2383, opened 2024-08-20
JAMES-4054 Fix remove X-SMIME-Status header in mailetcontainer.xml

Arsnael:
MailetContainerModule seems to check that RemoveMimeHeader with the parameter name strictly equal to `bcc` is present. If not, it crashes james startup. 

That would create issues during next release I think and this is the fix.

Why do we do such a strict checkup btw does somebody now? sound absurd to me.

### DIFF
--- a/server/apps/cassandra-app/sample-configuration/mailetcontainer.xml
+++ b/server/apps/cassandra-app/sample-configuration/mailetcontainer.xml
@@ -70,7 +70,11 @@
             </matcher>
 
             <mailet match="All" class="RemoveMimeHeader">
-                <name>bcc,X-SMIME-Status</name>
+                <name>bcc</name>
+                <onMailetException>ignore</onMailetException>
+            </mailet>
+            <mailet match="All" class="RemoveMimeHeader">
+                <name>X-SMIME-Status</name>
                 <onMailetException>ignore</onMailetException>
             </mailet>
             <mailet match="All" class="RecipientRewriteTable">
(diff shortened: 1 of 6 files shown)


chibenwa:
> MailetContainerModule seems to check that RemoveMimeHeader with the parameter name strictly equal to bcc is present. If not, it crashes james startup.

Not possible to make that check smarter ?

Arsnael:
> BTW those mailetcontainer checks are significantly annoying me for a long, long long time! I would happily have a "I know what I am doing and you don't so please do not validate my mailetcontainer against your stupid rules" option somewhere.

Ok with this.

Also was thinking making the check just less strict? Obviously you don't need to make a strictly equals on name for checking bcc is there, relaxing it by just checking if it contains it (as there is nothing wrong by adding other names in the syntax) would be an easy win/good start IMO

chibenwa:
> Also was thinking making the check just less strict? Obviously you don't need to make a strictly equals on name for checking bcc is there, relaxing it by just checking if it contains it (as there is nothing wrong by adding other names in the syntax) would be an easy win/good start IMO

+1.

Do you want to open the ticket for this @Arsnael ?

I still think we would benefit from a JVM system property to disable Mailet container checks. We could add that in the ticket IMO...

Arsnael:
=> https://issues.apache.org/jira/browse/JAMES-4059

chibenwa (approved):

PR #2385, opened 2024-08-21
WIP reproduce FETCH invalid message set issue

chibenwa:
Cc @woj-tek

### DIFF
--- a/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
+++ b/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
@@ -32,6 +32,12 @@ S: \* 2 FETCH \(UID 3\)
 S: \* 3 FETCH \(UID 4\)
 S: a OK FETCH completed.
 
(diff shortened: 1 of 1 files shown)


chibenwa:
> (what's the best way to proceed with this? should I use gh pr checkout 2385 and then create new branch locally?

Yes

> I think I can't push to this pr/branch directly?)

image

Only ASF james commiters can edit this PR, yes.

I think it is easier to open a new branch, new PR, cherry-pick my work then close this one.

chibenwa:
I am not fan of basing a decision (skip the data) on a projection (the state of the local in-memory data structure "Selected Mailbox" may be temporarilry out-of-date especially in a distributed setup)

On the example provided the MSN id set conversion `1:*` bear the meaning *all messages* and it would make sense to me to convert it into `MessageRange.all()`. (right now we are translating `1:*` into `MessageRange.of(firstUid, MessageUid.MAX)`.

This would even make sens regarding performances as downstream code akka the mailbox is optimized to handle `MessageRange.all()`.

Thoughts?

chibenwa:
CF https://github.com/apache/james-project/pull/2386

PR #2386, opened 2024-08-21
JAMES-4060: Fix UID FETCH command failing on empty mailbox

woj-tek:
(ref: https://issues.apache.org/jira/browse/JAMES-4060, https://www.mail-archive.com/server-dev@james.apache.org/msg74362.html)
(followup to https://github.com/apache/james-project/pull/2385 #2385)

* extending testcase to include `UID FETCH`
* fix the issue by ignoring range retrieval when the inbox is empty
* improvements to debug logs (compacting DefaultImapDecoder logs to single entry; add missing `toString()` and simplify toString in `ImmutableStatusResponse`):

Log afterwards looks nicer/cleaner IMHO (it's not perfect as it's mostly parsed data but gives more information about request-response chatter):
```
17:14:05.268 [DEBUG] o.a.j.m.p.ProtocolSession - C: b1 CREATE box
17:14:05.268 [TRACE] o.a.j.i.d.m.DefaultImapDecoder - Processing b1 CREATE CreateRequest{mailboxName=box}
17:14:05.268 [DEBUG] o.a.j.m.s.StoreMailboxManager - createMailbox #private:imapuser:box
17:14:05.269 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  MAILBOXID OK b1 completed. CREATE
17:14:05.269 [DEBUG] o.a.j.m.p.ProtocolSession - S: b1 OK [MAILBOXID (4)] CREATE completed.
17:14:05.269 [DEBUG] o.a.j.m.p.ProtocolSession - C: b2 SELECT box
17:14:05.270 [TRACE] o.a.j.i.d.m.DefaultImapDecoder - Processing b2 SELECT SelectRequest{mailboxName=box, condstore=false, lastKnownUidValidity=UidValidity{UNKNOWN}, knownModSeq=null, uidSet=null, knownUidSet=null, knownSequenceSet=null}
17:14:05.270 [DEBUG] o.a.j.m.s.StoreMailboxManager - Loaded mailbox #private:imapuser:box
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  MAILBOXID OK Ok
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  UIDVALIDITY OK UIDs valid
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  PERMANENTFLAGS OK Limited
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  HIGHESTMODSEQ OK Highest
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  UIDNEXT OK Predicted next UID
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  READ-WRITE OK b2 completed. SELECT
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [MAILBOXID (4)] Ok
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * FLAGS (\Answered \Deleted \Draft \Flagged \Seen)
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * 0 EXISTS
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * 0 RECENT
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [UIDVALIDITY 277573197] UIDs valid
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [PERMANENTFLAGS (\Answered \Deleted \Draft \Flagged \Seen \*)] Limited
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [HIGHESTMODSEQ 0] Highest
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [UIDNEXT 1] Predicted next UID
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: b2 OK [READ-WRITE] SELECT completed.
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - C: b3 FETCH 1:* (UID)
17:14:05.273 [TRACE] o.a.j.i.d.m.DefaultImapDecoder - Processing b3 FETCH FetchRequest{useUids=false, idSet=[IdRange ( 1->9223372036854775807 )], fetch=FetchData{items=[UID], setSeen=false, bodyElements=[], changedSince=-1, vanished=false}}
17:14:05.273 [DEBUG] o.a.j.m.s.StoreMailboxManager - Loaded mailbox 4 #private:imapuser:box
17:14:05.273 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  OK b3 completed. FETCH
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: b3 OK FETCH completed.
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - C: b4 UID FETCH 1:* (UID)
17:14:05.274 [TRACE] o.a.j.i.d.m.DefaultImapDecoder - Processing b4 UID FetchRequest{useUids=true, idSet=[IdRange ( 1->9223372036854775807 )], fetch=FetchData{items=[UID], setSeen=false, bodyElements=[], changedSince=-1, vanished=false}}
17:14:05.274 [DEBUG] o.a.j.m.s.StoreMailboxManager - Loaded mailbox 4 #private:imapuser:box
17:14:05.274 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  OK b4 completed. FETCH
17:14:05.274 [DEBUG] o.a.j.m.p.ProtocolSession - S: b4 OK FETCH completed.
```

### DIFF
--- a/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
+++ b/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
@@ -38,6 +38,14 @@ S: \* 2 FETCH \(UID 3\)
 S: \* 3 FETCH \(UID 4\)
 S: a2 OK FETCH completed.
 
+C: a21 FETCH * (UID)
+S: \* 3 FETCH \(UID 4\)
+S: a21 OK FETCH completed.
+
+C: a22 FETCH *:* (UID)
+S: \* 3 FETCH \(UID 4\)
+S: a22 OK FETCH completed.
+
 #UID fetch
 C: a UID FETCH 2:3 (INTERNALDATE)
 S: \* 1 FETCH \(INTERNALDATE "[^"]*" UID 2\)
--- a/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
+++ b/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
@@ -32,6 +32,12 @@ S: \* 2 FETCH \(UID 3\)
 S: \* 3 FETCH \(UID 4\)
 S: a OK FETCH completed.
 
+C: a2 FETCH 1:* (UID)
+S: \* 1 FETCH \(UID 2\)
+S: \* 2 FETCH \(UID 3\)
+S: \* 3 FETCH \(UID 4\)
+S: a2 OK FETCH completed.
+
 #UID fetch
 C: a UID FETCH 2:3 (INTERNALDATE)
 S: \* 1 FETCH \(INTERNALDATE "[^"]*" UID 2\)
@@ -77,6 +83,24 @@ S: a OK SEARCH completed.
 # Cleanup
 C: a1 DELETE copied
 S: a1 OK DELETE completed.
+
+C: b1 CREATE box
+S: b1 OK \[MAILBOXID \(.*\)\] CREATE completed.
+
+C: b2 SELECT box
+S: \* OK \[MAILBOXID \(.+\)\] Ok
+S: \* FLAGS \(\\Answered \\Deleted \\Draft \\Flagged \\Seen\)
+S: \* 0 EXISTS
+S: \* 0 RECENT
+S: \* OK \[UIDVALIDITY \d+\].*
+S: \* OK \[PERMANENTFLAGS \(\\Answered \\Deleted \\Draft \\Flagged \\\Seen( \\\*)?\)\].*
+S: \* OK \[HIGHESTMODSEQ \d+\].*
+S: \* OK \[UIDNEXT 1\].*
+S: b2 OK \[READ-WRITE\] SELECT completed.
+
+C: b3 FETCH 1:* (UID)
+S: b3 OK FETCH completed.
+
 
 
 
--- a/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
+++ b/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
@@ -101,6 +101,9 @@ S: b2 OK \[READ-WRITE\] SELECT completed.
 C: b3 FETCH 1:* (UID)
 S: b3 OK FETCH completed.
 
+C: b4 UID FETCH 1:* (UID)
+S: b4 OK FETCH completed.
+
 
 
 
--- a/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
+++ b/mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
@@ -98,12 +98,11 @@ S: \* OK \[HIGHESTMODSEQ \d+\].*
 S: \* OK \[UIDNEXT 1\].*
 S: b2 OK \[READ-WRITE\] SELECT completed.
 
-C: b3 FETCH 1:* (UID)
+C: b3 UID FETCH 1:* (UID)
 S: b3 OK FETCH completed.
 
-C: b4 UID FETCH 1:* (UID)
+C: b4 UID FETCH *:* (UID)
 S: b4 OK FETCH completed.
 
-
-
-
+C: b5 UID FETCH * (UID)
+S: b5 OK FETCH completed.
--- a/protocols/imap/src/main/java/org/apache/james/imap/api/Tag.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/api/Tag.java
@@ -21,7 +21,6 @@ package org.apache.james.imap.api;
 
 import java.util.Objects;
 
-import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
 
 public class Tag {
--- a/protocols/imap/src/main/java/org/apache/james/imap/api/Tag.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/api/Tag.java
@@ -53,8 +53,6 @@ public class Tag {
 
     @Override
     public String toString() {
-        return MoreObjects.toStringHelper(this)
-            .add("value", value)
-            .toString();
+        return value;
     }
 }
--- a/protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
@@ -75,7 +75,6 @@ public class DefaultImapDecoder implements ImapDecoder {
     }
 
     private ImapMessage decodeCommandTagged(ImapRequestLineReader request, Tag tag, ImapSession session) {
-        LOGGER.debug("Got <tag>: {}", tag);
         try {
             String commandName = request.atom();
             return decodeCommandNamed(request, tag, commandName, session);
@@ -110,7 +109,6 @@ public class DefaultImapDecoder implements ImapDecoder {
     }
 
     private ImapMessage decodeCommandNamed(ImapRequestLineReader request, Tag tag, String commandName, ImapSession session) {
-        LOGGER.debug("Got <command>: {}", commandName);
         ImapCommandParser command = imapCommands.getParser(commandName);
         if (command == null) {
             LOGGER.info("Missing command implementation for commmand {}", commandName);
@@ -121,6 +119,7 @@ public class DefaultImapDecoder implements ImapDecoder {
         if (count == null || (int) count > 0) {
             session.setAttribute(INVALID_COMMAND_COUNT, 0);
         }
+        LOGGER.trace("Processing {} {} {}", tag.asString(), commandName, message.toString());
         return message;
     }
 }
--- a/protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
@@ -119,7 +119,7 @@ public class DefaultImapDecoder implements ImapDecoder {
         if (count == null || (int) count > 0) {
             session.setAttribute(INVALID_COMMAND_COUNT, 0);
         }
-        LOGGER.trace("Processing {} {} {}", tag, commandName, message);
+        LOGGER.trace("Processing {} {} {} [{}] ", tag, commandName, message, session.getUserName());
         return message;
     }
 }
--- a/protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
@@ -119,7 +119,7 @@ public class DefaultImapDecoder implements ImapDecoder {
         if (count == null || (int) count > 0) {
             session.setAttribute(INVALID_COMMAND_COUNT, 0);
         }
-        LOGGER.trace("Processing {} {} {}", tag.asString(), commandName, message.toString());
+        LOGGER.trace("Processing {} {} {}", tag, commandName, message);
         return message;
     }
 }
--- a/protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
@@ -31,6 +31,7 @@ import org.apache.james.imap.decode.ImapCommandParser;
 import org.apache.james.imap.decode.ImapCommandParserFactory;
 import org.apache.james.imap.decode.ImapDecoder;
 import org.apache.james.imap.decode.ImapRequestLineReader;
+import org.apache.james.util.MDCStructuredLogger;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -119,7 +120,12 @@ public class DefaultImapDecoder implements ImapDecoder {
         if (count == null || (int) count > 0) {
             session.setAttribute(INVALID_COMMAND_COUNT, 0);
         }
-        LOGGER.trace("Processing {} {} {} [{}] ", tag, commandName, message, session.getUserName());
+        // Avoid cost of initializing MDC locally
+        if (LOGGER.isTraceEnabled()) {
+            new MDCStructuredLogger(LOGGER)
+                    .field("username", String.valueOf(session.getUserName()))
+                    .log(logger -> logger.trace("Processing {} {} {} ", tag, commandName, message));
+        }
         return message;
     }
 }
--- a/protocols/imap/src/main/java/org/apache/james/imap/message/response/ImmutableStatusResponse.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/message/response/ImmutableStatusResponse.java
@@ -24,8 +24,6 @@ import org.apache.james.imap.api.Tag;
 import org.apache.james.imap.api.display.HumanReadableText;
 import org.apache.james.imap.api.message.response.StatusResponse;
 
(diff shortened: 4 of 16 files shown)


protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
```diff
@@ -121,6 +119,7 @@ private ImapMessage decodeCommandNamed(ImapRequestLineReader request, Tag tag, S
         if (count == null || (int) count > 0) {
             session.setAttribute(INVALID_COMMAND_COUNT, 0);
         }
+        LOGGER.trace("Processing {} {} {}", tag.asString(), commandName, message.toString());
```
chibenwa, line 122: Red flag: tostring is eagerly instanciated here

Wrap it in a if (LOGGER.isTraceEnabled() ?

chibenwa: Please put the username from the session in there.

protocols/imap/src/main/java/org/apache/james/imap/message/response/UnpooledStatusResponseFactory.java
```diff
@@ -26,12 +26,18 @@
...
 
 public class UnpooledStatusResponseFactory extends AbstractStatusResponseFactory implements StatusResponseFactory {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(UnpooledStatusResponseFactory.class);
+
     @Override
     protected StatusResponse createResponse(Type type, Tag tag, ImapCommand command, HumanReadableText displayTextKey, ResponseCode code) {
-        return new ImmutableStatusResponse(type, tag, command, displayTextKey, code);
+        var response = new ImmutableStatusResponse(type, tag, command, displayTextKey, code);
+        LOGGER.trace("Created IMAP response: {}", response);
```
chibenwa, line 39: I'm missing a username in this log...

protocols/imap/src/main/java/org/apache/james/imap/processor/StatusProcessor.java
```diff
@@ -128,6 +128,7 @@ Mono<MailboxStatusResponse> sendStatus(MessageManager mailbox, StatusDataItems s
                     if (response.getHighestModSeq() != null) {
                         condstoreEnablingCommand(session, responder, metaData, false);
                     }
+                    LOGGER.trace("Status on mailbox named {}, response: {}", mailbox, response);
```
chibenwa, line 131: Doesn;t it duplicates the log in UnpooledStatusResponseFactory ?

protocols/imap/src/main/java/org/apache/james/imap/processor/fetch/FetchProcessor.java
```diff
@@ -201,10 +201,12 @@ private Mono<Void> doFetch(SelectedMailbox selected, FetchRequest request, Respo
         List<MessageRange> ranges = new ArrayList<>();
 
         for (IdRange range : request.getIdSet()) {
-            MessageRange messageSet = messageRange(session.getSelected(), range, request.isUseUids())
-                .orElseThrow(() -> new MessageRangeException(range.getFormattedString() + " is an invalid range"));
-            if (messageSet != null) {
-                MessageRange normalizedMessageSet = normalizeMessageRange(selected, messageSet);
+            if (session.getSelected().getLastUid().isEmpty()) {
+                continue;
+            }
```
chibenwa, line 206: See https://github.com/apache/james-project/pull/2385#issuecomment-2302370425

I am not fan of basing a decision (skip the data) on a projection (the state of the local in-memory data structure "Selected Mailbox" may be temporarilry out-of-date especially in a distributed setup)

On the example provided the MSN id set conversion `1:*` bear the meaning *all messages* and it would make sense to me to convert it into `MessageRange.all()`. (right now we are translating `1:*` into `MessageRange.of(firstUid, MessageUid.MAX)`.

This would even make sens regarding performances as downstream code akka the mailbox is optimized to handle `MessageRange.all()`.

Thoughts?

protocols/imap/src/main/java/org/apache/james/imap/decode/main/DefaultImapDecoder.java
chibenwa: > .trace() do the check itself - should we check if the logger is enabled still?

Without explicit calls to 'toString' of course we can aggressively call `.trace()`

> I wonder - shouldn't this be handled by MDC (and would be more general)?

As explained on the mailing list, MDC needs to be explicitly carried other accross threds in reactor wich could be expensive. 

In 'classic' log statements not integrated with reactor syntax you can assume the log would not be here.

One other alternative could be to rely explicitly on MDCStructuredLogger:

```
// Avoid cost of initializing MDC locally
if (LOGGER.isTraceEnabled()) {
    new MDCStructuredLogger(LOGGER)
        .field("username", username)
        ...
        .log(logger -> logger.trace("My trace"));
}
```

A bit boiler plate but this may be better than seriously messing up reactor syntax to cary other the MDC in this case, and cleaner than just adding the username in the message......

protocols/imap/src/main/java/org/apache/james/imap/message/response/UnpooledStatusResponseFactory.java
chibenwa: Maybe because this is not necessarily the right place for that log. 

Few things first with this location we miss some answers:

```
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  MAILBOXID OK Ok
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  UIDVALIDITY OK UIDs valid
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  PERMANENTFLAGS OK Limited
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  HIGHESTMODSEQ OK Highest
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  UIDNEXT OK Predicted next UID
17:14:05.272 [TRACE] o.a.j.i.m.r.UnpooledStatusResponseFactory - Created IMAP response:  READ-WRITE OK b2 completed. SELECT
```

VS

```
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [MAILBOXID (4)] Ok
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * FLAGS (\Answered \Deleted \Draft \Flagged \Seen)
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * 0 EXISTS
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * 0 RECENT
17:14:05.272 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [UIDVALIDITY 277573197] UIDs valid
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [PERMANENTFLAGS (\Answered \Deleted \Draft \Flagged \Seen \*)] Limited
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [HIGHESTMODSEQ 0] Highest
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: * OK [UIDNEXT 1] Predicted next UID
17:14:05.273 [DEBUG] o.a.j.m.p.ProtocolSession - S: b2 OK [READ-WRITE] SELECT completed.
```

(it only includes responses with a status)

In practice putting the log in `ChannelImapResponseWriter::write(byte[])` should yield very good results:
 - You are guarantied to get the exact output sent
 - You would get all responses but avoid expensive litterals - litterals sending could be traced too btw but maybe without content
 - We could pass the IMAP session as a parameter easily to that class and access the username for structured logging

Though this hooks onto the TCP transport layer provided by server/protocols/protocols-imap4 and won't be seen with MPT that exercise only the underlying logic - not an issue as MPT itself allow to see/assert the exchanges.

That is definitly the way I would undertook.

protocols/imap/src/main/java/org/apache/james/imap/processor/StatusProcessor.java
chibenwa: By logging the exact IMAP chatter output isn't this limitation lifted?

protocols/imap/src/main/java/org/apache/james/imap/processor/fetch/FetchProcessor.java
chibenwa: Yes this was.

protocols/imap/src/main/java/org/apache/james/imap/message/response/ImmutableStatusResponse.java
```diff
@@ -72,12 +70,22 @@ public ImapCommand getCommand() {
...
+        if (getTag() != null) {
+            sb.append(' ').append(getTag());
+        }
+        if (getTextKey() != null) {
+            sb.append(' ').append(getTextKey());
+        }
+        if (getCommand() != null) {
+            sb.append(' ').append(getCommand());
+        }
+        return sb.toString();
```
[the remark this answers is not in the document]
chibenwa: Sorry for getting chatty it is very interesting technical discussions ^^

Given https://github.com/apache/james-project/pull/2386#discussion_r1725676546 first this change is no longer needed.

Keeping "toString" standard definitly is valuable: apply the IDE template, get the same everywhere.

This do not mean we should blindly apply toString to our log statements. We can apply alternative formats.  Here in this case something that kinda duplicates StatusResponseEncoder behaviour.

mpt/impl/imap-mailbox/core/src/main/resources/org/apache/james/imap/scripts/Uid.test
```diff
@@ -32,6 +32,12 @@ S: \* 2 FETCH \(UID 3\)
 S: \* 3 FETCH \(UID 4\)
 S: a OK FETCH completed.
 
+C: a2 FETCH 1:* (UID)
```
chibenwa, line 35: In a file name `UID` I would expect tests about `UID ...` not about Fetch used in MSN mode.

Those tests are valuable, maybe they can be move in an existing `fetch...` script?

protocols/imap/src/main/java/org/apache/james/imap/api/Tag.java
```diff
@@ -53,8 +52,6 @@ public final int hashCode() {
 
     @Override
     public String toString() {
-        return MoreObjects.toStringHelper(this)
-            .add("value", value)
-            .toString();
+        return value;
```
chibenwa, line 55: +1 here MoreObject definitly is overkill

chibenwa:
> Or this was more of a comment for future/followup improvements?

No need to bother, this was more a comment for the future.

protocols/imap/src/main/java/org/apache/james/imap/message/response/ImmutableStatusResponse.java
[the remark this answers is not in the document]
chibenwa: >  should we use asString() for custom string output?

+1

> 

That's true!

In non critical code portions or non costly operations we don't really care but for big string concatenations this might be a concern.

I'm sorry to be peacky, this kind of stuff in IMAP code tend to have a non neglictible impact on CPU (mdc context handling) and memory allocations in what I could see with flame graphs (cf async-profiler) and being careful here do not cost much.

If wrapping things in ifs is for you also not your cup of tea we apparently can use suppliers in the logging event builder (I just found this out!)

```
        LOGGER.makeLoggingEventBuilder(Level.DEBUG)
            .setMessage("My message {}")
            .addArgument(() -> "I'm costly and I know it!")
            .log();
```

And under the hood (log4j but other log provider would do the same):

```
logBuilder.log(this.message, this.arguments.toArray());
```

Could be nice to defer expensive argument evaluation, eg for custom displays?

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMailboxProcessor.java
```diff
@@ -444,14 +444,10 @@ protected Optional<MessageRange> messageRange(SelectedMailbox selected, IdRange
...
         // the mailbox. See IMAP-289
-        if (lowVal == Long.MAX_VALUE && highVal == Long.MAX_VALUE) {
-            Optional<MessageUid> last = selected.getLastUid();
-            if (!last.isPresent()) {
-                throw new MessageRangeException("Mailbox is empty");
-            }
-            return last.get().toRange();
+        if ((lowVal == Long.MAX_VALUE || lowVal == 1) && highVal == Long.MAX_VALUE) {
+            return MessageRange.all();
         }
```
chibenwa, line 451: We got a new bug here!

CF https://datatracker.ietf.org/doc/html/rfc3501 ABNF

```
seq-number      = nz-number / "*"
                    ; message sequence number (COPY, FETCH, STORE
                    ; commands) or unique identifier (UID COPY,
                    ; UID FETCH, UID STORE commands).
                    ; * represents the largest number in use.  In
                    ; the case of message sequence numbers, it is
                    ; the number of messages in a non-empty mailbox.
                    ; In the case of unique identifiers, it is the
                    ; unique identifier of the last message in the
                    ; mailbox or, if the mailbox is empty, the
                    ; mailbox's current UIDNEXT value.
                    ; The server should respond with a tagged BAD
                    ; response to a command that uses a message
                    ; sequence number greater than the number of
                    ; messages in the selected mailbox.  This
                    ; includes "*" if the selected mailbox is empty.
```

`1:*` -> Messages from first to last IE `MessageRange.all() (we are good)

But `*:*` means range from last message to last message IE `last.get().toRange()`

So we actually need two separate ifs: one handling `1:*` and one handling `*:*`

Also we MAY which:
 - [ ] To ensure MPT tests `C: A0 FETCH *:*`
 - [ ] To test the behaviour of `C: A0 UID FETCH *:*`
 - [ ] Technically speacking a lone `seq-number` is a valid `sequence-set` so we may which to ensure `C: A0 FETCH *` and `C: A0 UID FETCH *` are nicely handled - and tested.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ChannelImapResponseWriter.java
```diff
@@ -68,6 +83,15 @@ public void setFlushCallback(FlushCallback flushCallback) {
 
     @Override
     public void write(byte[] buffer) {
+        if (LOGGER.isTraceEnabled()) {
+            if (ctx != null) {
+                ImapSession imapSession = ctx.channel().attr(IMAP_SESSION_ATTRIBUTE_KEY).get();
+                String username = retrieveUsername(imapSession);
```
chibenwa, line 89: I'm sorry I am definitly picky here but this means a map read every time we send an IMAP response.

This is likely unecessary are the code that creates ChannelImapResponseWriter already have an ImapSession at hand.

I would add the ImapSession as a field, then pass it as an argument and avoid resolve it here.

(Sorry that's a critical section so I'm picky!)

Arsnael (commented):
Read it, nothing to add

chibenwa (approved):

Arsnael (approved):

PR #2387, opened 2024-08-22
JAMES-4056 Update SMIMECheckSignature for case mail with multiple certs

hungphan227:


Arsnael (approved):

chibenwa (approved):
Nice catch!

chibenwa:
Please add an explicit mention in JIRA JAMES-4056 of what the deffect was and what the fix was so that this gets recorded on the mailing list.

(More generally we need to put more efforts to describe our work, and record it on community channels)

hungphan227:
> Please add an explicit mention in JIRA JAMES-4056 of what the deffect was and what the fix was so that this gets recorded on the mailing list.
> 
> (More generally we need to put more efforts to describe our work, and record it on community channels)

Added a comment in https://issues.apache.org/jira/browse/JAMES-4056

PR #2393, opened 2024-08-28
JAMES-4059 MailetContainerModule should not do a strict check on RemoveMimeHeader with bcc name param

Arsnael:
I hesitated adding a test for the jvm property commit as it's a bit tricky for james startup

Maybe in memory-app with some alternative conf files in resources? Or just not necessary?

### DIFF
--- a/docs/modules/servers/pages/distributed/configure/jvm.adoc
+++ b/docs/modules/servers/pages/distributed/configure/jvm.adoc
@@ -105,10 +105,11 @@ Ex in `jvm.properties`
 
 == Disable mailet container check at James startup
 
-James is doing a check on startup making sure that the `bcc` header is being removed via
+James is doing checks on startup for validating mailet container configuration against a set of
+business rules, for instance making sure that the `bcc` header is being removed via
 `RemoveMimeHeader` mailet in the mail processing pipeline defined in `mailetcontainer.xml` file.
 
-It could be useful for some administrators that know what they are doing to disable such a check
+It could be useful for some administrators that know what they are doing to disable such checks
 during James startup.
 
 Optional. Boolean. Defaults to true.
--- a/docs/modules/servers/pages/distributed/configure/jvm.adoc
+++ b/docs/modules/servers/pages/distributed/configure/jvm.adoc
@@ -103,3 +103,18 @@ Ex in `jvm.properties`
 # james.reactor.inputstream.prefetch=4
 ----
 
+== Disable mailet container check at James startup
+
+James is doing a check on startup making sure that the `bcc` header is being removed via
+`RemoveMimeHeader` mailet in the mail processing pipeline defined in `mailetcontainer.xml` file.
+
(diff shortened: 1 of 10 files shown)


docs/modules/servers/pages/distributed/configure/jvm.adoc
```diff
@@ -103,3 +103,18 @@ Ex in `jvm.properties`
 # james.reactor.inputstream.prefetch=4
 ----
 
+== Disable mailet container check at James startup
+
+James is doing a check on startup making sure that the `bcc` header is being removed via
+`RemoveMimeHeader` mailet in the mail processing pipeline defined in `mailetcontainer.xml` file.
+
+It could be useful for some administrators that know what they are doing to disable such a check
```
chibenwa, line 111:
```suggestion
It could be useful for some administrators that know what they are doing to disable such checks
```

server/container/guice/protocols/jmap/src/test/java/org/apache/james/jmap/MailetPreconditionTest.java
```diff
@@ -191,5 +191,19 @@ void bccMailetCheckShouldNotThrowOnValidPair() throws Exception {
             assertThatCode(() -> MailetContainerModule.BCC_Check.check(pairs))
                 .doesNotThrowAnyException();
         }
+
+        @Test
+        void bccMailetCheckShouldNotThrowOnValidPairWithOtherNames() throws Exception {
```
chibenwa, line 196: Edge case: Please add a test ensuring that removing `x-bcc-bug` header is not considered as valid.

chibenwa (commented):
While we are at it, how about case insentivity?

(RemoveMimeHeader is case insentive, the check here is case sensitive)

quantranhong1999 (approved):

server/container/guice/mailet/src/main/java/org/apache/james/modules/server/MailetContainerModule.java
```diff
@@ -86,7 +86,7 @@ public class MailetContainerModule extends AbstractModule {
         "transport",
         All.class,
         RemoveMimeHeader.class,
-        pair -> Arrays.asList(pair.getMailet().getMailetConfig().getInitParameter("name").split(",")).contains("bcc"),
+        pair -> Arrays.asList(pair.getMailet().getMailetConfig().getInitParameter("name").toLowerCase().split(",")).contains("bcc"),
```
chibenwa, line 89: Other stupid edge case question: do we handle the case where RemoveMimeHeader do not have a "name" property decently?

IE we likely throw today an uggly non explicit NPE, it would be nice to throw a contextualized with decent error message IllegalStateException, and have a unit test for it?

chibenwa: Likely OK

chibenwa (approved):

PR #2401, opened 2024-09-09
JAMES-4069 Healthcheck for IMAP

hungphan227:


server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPHealthCheck.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public ComponentName componentName() {
+        return COMPONENT_NAME;
+    }
+
+    @Override
+    public Mono<Result> check() {
+        return Mono.fromCallable(() -> {
+            if (imapServerFactory.getImapServers().stream().map(IMAPServer::getReactiveThrottler).anyMatch(ReactiveThrottler::isQueueFull)) {
```
quantranhong1999, line 48: Can we extract method for this? and with better indent?

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
```diff
@@ -76,16 +78,19 @@ protected IMAPServer createServer(HierarchicalConfiguration<ImmutableNode> confi
...
             server.configure(serverConfig);
             servers.add(server);
         }
+        imapServers = servers.stream().map(IMAPServer.class::cast).toList();
 
         return servers;
-
     }
 
+    public List<IMAPServer> getImapServers() {
```
quantranhong1999, line 93: We can use https://github.com/apache/james-project/blob/master/server/protocols/protocols-library/src/main/java/org/apache/james/protocols/lib/netty/AbstractServerFactory.java#L66 and filter by IMAP server type.

image

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
```diff
@@ -76,16 +78,19 @@ protected IMAPServer createServer(HierarchicalConfiguration<ImmutableNode> confi
...
         List<HierarchicalConfiguration<ImmutableNode>> configs = config.configurationsAt("imapserver");
-        
+
         for (HierarchicalConfiguration<ImmutableNode> serverConfig: configs) {
             IMAPServer server = createServer(serverConfig);
             server.setFileSystem(fileSystem);
             server.configure(serverConfig);
             servers.add(server);
         }
+        imapServers = servers.stream().map(IMAPServer.class::cast).toList();
```
Arsnael, line 88: little indent here does not kill :)

chibenwa, line 88: What is this uggly cast?


Can't we `List<IMAPServer> servers` in the first place?

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
chibenwa: Yes it would prevent the uggly field allocation, +1 quan

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
```diff
@@ -66,7 +68,7 @@ public IMAPServerFactory(FileSystem fileSystem, ThrowingFunction<HierarchicalCon
         this.connectionCheckFactory = connectionCheckFactory;
     }
 
-    protected IMAPServer createServer(HierarchicalConfiguration<ImmutableNode> config) {
+    private IMAPServer createServer(HierarchicalConfiguration<ImmutableNode> config) {
```
vttranlina, line 69: Changing `protected` to `private` risks breaking forked projects that are attempting to reuse this function.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
hungphan227: Fixed. Sr I did not know that the attribute servers has already existed.

chibenwa (approved):

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
chibenwa: BTW what prevent us from doing the cast within IMAPHealthCheck ?

This would prevent a change here...

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServer.java
```diff
@@ -312,4 +312,8 @@ protected ChannelHandlerFactory createFrameHandlerFactory() {
     public Set<ConnectionCheck> getConnectionChecks() {
         return this.connectionChecks;
     }
+
+    public ReactiveThrottler getReactiveThrottler() {
```
chibenwa, line 316: Does it makes sense to expose the throttler outside?

Shouldn't we only expose if the queue is full?

chibenwa:
Sorry 2 more comments...

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServerFactory.java
hungphan227: Yes we can. However, I don't see any problems with "a change" in IMAPServerFactory.

server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/IMAPServer.java
hungphan227: It will be quite complicated for testing if throttler is not exposed. We are still allowed to use Reflection, right?

chibenwa: Then :
 - Still add a method to only expose if the queue is full
 - Keep this methods but reduce its visibility: package private and make it clear this is for tests: `@VisibleForTesting` annotation.

Arsnael (approved):

quantranhong1999 (approved):

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2401/4/testReport/junit/org.apache.james.webadmin.integration.rabbitmq/RabbitMQWebAdminServerIntegrationImmutableTest/healthCheckComponentsList/

Likely related?

quantranhong1999 (approved):

PR #2404, opened 2024-09-10
JAMES-4070 Implement query param to return 503 code if degraded

hungphan227:


docs/modules/servers/partials/operate/webadmin.adoc
```diff
@@ -94,6 +94,14 @@ Response codes:
 services can still be used.
 * 503: At least one check have answered with a Unhealthy status
 
+Additional query parameters are supported:
```
quantranhong1999, line 97: Please add doc to `manage-webadmin.md` too.

BTW please complement https://github.com/apache/james-project/pull/2399/files doc also.

Arsnael:
Read it nothing to add

Arsnael (approved):

quantranhong1999 (approved):

server/protocols/webadmin/webadmin-core/src/main/java/org/apache/james/webadmin/routes/HealthCheckRoutes.java
```diff
@@ -58,6 +58,38 @@
...
+
+        private Function<ResultStatus, Integer> converter;
+
+        StatusCodeConvertMode(Function<ResultStatus, Integer> converter) {
+            this.converter = converter;
+        }
+
+        public Function<ResultStatus, Integer> getConverter() {
+            return converter;
+        }
```
chibenwa, line 91: Exposing the function outside iss super verbose.

We have internally a function because the way enum work. But *please* do not leak this to the outer code!

chibenwa:
```suggestion
        public int getCorrespondingSatus(ResultStatus result) {
            return converter.apply(result);
        }
```

server/protocols/webadmin/webadmin-core/src/main/java/org/apache/james/webadmin/routes/HealthCheckRoutes.java
```diff
@@ -91,25 +123,25 @@ public void define(Service service) {
...
+        StatusCodeConvertMode convertMode = getStatusCodeConvertMode(request);
         HealthCheck healthCheck = healthChecks.stream()
             .filter(c -> c.componentName().getName().equals(componentName))
             .findFirst()
             .orElseThrow(() -> throw404(componentName));
 
         Result result = Mono.from(healthCheck.check()).block();
         logFailedCheck(result);
-        response.status(getCorrespondingStatusCode(result.getStatus(), isStrict));
+        response.status(getCorrespondingStatusCode(result.getStatus(), convertMode));
```
chibenwa, line 144: Inline that: we definitly do not need that level of indirection

```suggestion
        response.status(convertMode.getCorrespondingSatus(result.getStatus())));
```

Arsnael (approved):

PR #2405, opened 2024-09-10
James 3945

florentos17:
i am working on implementing sub-addressing for apache james.
for now it works but WITHOUT any right management.
this is still work in progress.

any comment is welcome :)

### DIFF
--- a/core/src/test/java/org/apache/james/core/MailAddressTest.java
+++ b/core/src/test/java/org/apache/james/core/MailAddressTest.java
@@ -47,7 +47,9 @@ class MailAddressTest {
                 GOOD_ADDRESS,
                 GOOD_QUOTED_LOCAL_PART,
                 "server-dev@james-apache.org",
+                "local-part+details@example.com",
                 "a&b@james-apache.org",
+                "+details@example.com",
                 "server-dev@[127.0.0.1]",
                 "server.dev@james.apache.org",
                 "\\.server-dev@james.apache.org",
@@ -93,6 +95,7 @@ class MailAddressTest {
                 "server-dev@#123.apache.org",
                 "server-dev@[127.0.1.1.1]",
                 "server-dev@[127.0.1.-1]",
+                "test@dom+ain.com",
                 "\"a..b\"@domain.com", // jakarta.mail is unable to handle this so we better reject it
                 "server-dev\\.@james.apache.org", // jakarta.mail is unable to handle this so we better reject it
                 "a..b@domain.com",
@@ -225,4 +228,89 @@ class MailAddressTest {
         EqualsVerifier.forClass(MailAddress.class)
             .verify();
     }
+
+    @Test
+    void getLocalPartDetailsShouldReturnDetails() throws AddressException {
+        MailAddress mailAddress = new MailAddress("address+details@example.com");
+        assertThat(mailAddress.getLocalPartDetails("+")).contains("details");
+    }
+
+    @Test
+    void getLocalPartDetailsShouldReturnEmptyWhenNone() throws AddressException {
+        MailAddress mailAddress = new MailAddress("address@example.com");
+        assertThat(mailAddress.getLocalPartDetails("+")).isEmpty();
+    }
+
+    @Test
+    void getLocalPartDetailsShouldReturnEmptyWhenEmpty() throws AddressException {
+        MailAddress mailAddress = new MailAddress("address+@example.com");
+        assertThat(mailAddress.getLocalPartDetails("+")).contains("");
+    }
+
+    @Test
+    void getLocalPartDetailsShouldReturnDetailsIncludingOtherOccurencesOfDelimiter() throws AddressException {
+        MailAddress mailAddress = new MailAddress("address+de+tails@example.com");
+        assertThat(mailAddress.getLocalPartDetails("+")).contains("de+tails");
+    }
+
+    @Test
+    void getLocalPartDetailsShouldReturnEmptyWhenAddressStartsWithDelimiter() throws AddressException {
+        MailAddress mailAddress = new MailAddress("+details@example.com");
+        assertThat(mailAddress.getLocalPartDetails("+")).isEmpty();
+    }
+
+    @Test
+    void getLocalPartDetailsShouldReturnDetailsWhenMultipleCharacterDelimiter() throws AddressException {
+        MailAddress mailAddress = new MailAddress("localpart--details@example.com");
+        assertThat(mailAddress.getLocalPartDetails("--")).contains("details");
+    }
+
+    @Test
+    void getLocalPartDetailsShouldBePreciseWithMultipleCharacterDelimiter() throws AddressException {
+        MailAddress mailAddress = new MailAddress("localpart---details@example.com");
+        assertThat(mailAddress.getLocalPartDetails("--")).contains("-details");
+    }
+
+    @Test
+    void stripDetailsShouldWork() throws AddressException {
+        MailAddress mailAddress = new MailAddress("localpart+details@example.com");
+        assertThat(mailAddress.stripDetails("+")).isEqualTo("localpart@example.com");
+    }
+
+    @Test
+    void stripDetailsShouldReturnTheSameWhenNoDetails() throws AddressException {
+        MailAddress mailAddress = new MailAddress("address@example.com");
+        assertThat(mailAddress.stripDetails("+")).isEqualTo("address@example.com");
+    }
+
+    @Test
+    void stripDetailsShouldRemoveSeparatorWhenEmptyDetails() throws AddressException {
+        MailAddress mailAddress = new MailAddress("address+@example.com");
+        assertThat(mailAddress.stripDetails("+")).isEqualTo("address@example.com");
+    }
+
+    @Test
+    void stripDetailsShouldStripAllDetailsWhenSeveralOccurencesOfSeparator() throws AddressException {
+        MailAddress mailAddress = new MailAddress("address+de+tails@example.com");
+        assertThat(mailAddress.stripDetails("+")).isEqualTo("address@example.com");
+    }
+
+    @Test
+    void stripDetailsShouldReturnTheSameWhenAddressStartsWithDelimiter() throws AddressException {
+        MailAddress mailAddress = new MailAddress("+details@example.com");
+        assertThat(mailAddress.stripDetails("+")).isEqualTo("+details@example.com");
+    }
+
+    @Test
+    void stripDetailsShouldWorkWhenMultipleCharacterDelimiter() throws AddressException {
+        MailAddress mailAddress = new MailAddress("localpart--details@example.com");
+        assertThat(mailAddress.stripDetails("--")).isEqualTo("localpart@example.com");
+    }
+
+    @Test
+    void stripDetailsShouldBePreciseWithMultipleCharacterDelimiter() throws AddressException {
+        MailAddress mailAddress = new MailAddress("localpart---details@example.com");
+        assertThat(mailAddress.stripDetails("--")).isEqualTo("localpart@example.com");
+    }
+
 }
--- a/docs/modules/servers/partials/SubAddressing.adoc
+++ b/docs/modules/servers/partials/SubAddressing.adoc
@@ -1,0 +1,22 @@
+=== SubAddressing
+
+SubAddressing positions a storage directive for the folder this email will be delivered in if the following criteria are met:
+
+* the sender has requested the storage directive by sending the mail to <strong> `recipient+folder@domain` </strong> instead of just `recipient@domain` ;
+* the folder <strong>exists</strong> and the recipient has <strong>allowed</strong> the sender to send a mail to that specific folder.
+
+These directives are used by <strong>LocalDelivery</strong> mailet when adding the email to the recipients mailboxes.
+
+The storage directive is recognized when a specific character or character sequence is present in the local part of the recipient address. <strong>By default, it is "+"</strong>.
+
+If the sender is not allowed to send a mail to the specified folder, then the mail is delivered in the recipient's inbox.
+Likewise, if the storage directive is empty or absent, the mail will simply be delivered in the recipient's inbox.
+
+Thus,
+
+ * a mail sent to `recipient+folder@domain` will be delivered to recipient's folder `folder` if allowed ;
+ * a mail sent to `recipient+my-super-folder@domain` will be delivered to recipient's folder `my-super-folder` if allowed ;
+ * a mail sent to `recipient@domain` or `recipient+@domain` will be delivered to recipient's inbox.
+
+Any user can position rights for other users and for its different folders. They may create whitelists or blacklists, for one or several folders.
+In the case where the sender is unknown, the mail will be delivered in the specified folder only if the recipient has allowed everyone for that folder.
--- a/server/mailet/integration-testing/src/test/java/org/apache/james/mailets/SubAddressingTest.java
+++ b/server/mailet/integration-testing/src/test/java/org/apache/james/mailets/SubAddressingTest.java
@@ -29,12 +29,19 @@ import static org.apache.james.mailets.configuration.Constants.RECIPIENT;
 import static org.apache.james.mailets.configuration.Constants.awaitAtMostOneMinute;
 
 import java.io.File;
+import java.io.IOException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
 
+import org.apache.james.core.Username;
+import org.apache.james.mailbox.model.MailboxACL;
 import org.apache.james.mailbox.model.MailboxConstants;
+import org.apache.james.mailbox.model.MailboxPath;
 import org.apache.james.mailets.configuration.CommonProcessors;
 import org.apache.james.mailets.configuration.MailetConfiguration;
 import org.apache.james.mailets.configuration.ProcessorConfiguration;
-import org.apache.james.modules.MailboxProbeImpl;
+import org.apache.james.modules.ACLProbeImpl;
 import org.apache.james.modules.protocols.ImapGuiceProbe;
 import org.apache.james.modules.protocols.SmtpGuiceProbe;
 import org.apache.james.probe.DataProbe;
@@ -59,41 +66,7 @@ class SubAddressingTest {
 
     private TemporaryJamesServer jamesServer;
 
-    void setup() throws Exception {
-        DataProbe dataProbe = jamesServer.getProbe(DataProbeImpl.class);
-        dataProbe.addDomain(DEFAULT_DOMAIN);
-        dataProbe.addUser(RECIPIENT, PASSWORD);
-        dataProbe.addUser(FROM, PASSWORD);
-
-        jamesServer.getProbe(MailboxProbeImpl.class)
-            .createMailbox(MailboxConstants.USER_NAMESPACE, RECIPIENT, TARGETED_MAILBOX);
-    }
-
-    @AfterEach
-    void tearDown() {
-        if (jamesServer != null) {
-            jamesServer.shutdown();
-        }
-    }
-
-    @Test
-    void subAddressedEmailShouldBeReceivedByDefault(@TempDir File temporaryFolder) throws Exception {
-        jamesServer = TemporaryJamesServer.builder().build(temporaryFolder);
-        jamesServer.start();
-        setup();
-
-        messageSender.connect(LOCALHOST_IP, jamesServer.getProbe(SmtpGuiceProbe.class).getSmtpPort())
-            .authenticate(FROM, PASSWORD)
-            .sendMessage(FROM, "user2+detail-test@" + DEFAULT_DOMAIN);
-
-        testIMAPClient.connect(LOCALHOST_IP, jamesServer.getProbe(ImapGuiceProbe.class).getImapPort())
-            .login(RECIPIENT, PASSWORD)
-            .select(MailboxConstants.INBOX)
-            .awaitMessage(awaitAtMostOneMinute);
-    }
-
-    @Test
-    void subAddressedEmailShouldBeDeliveredInSpecifiedFolder(@TempDir File temporaryFolder) throws Exception {
+    void setup(File temporaryFolder) throws Exception {
         jamesServer = TemporaryJamesServer.builder()
             .withMailetContainer(TemporaryJamesServer.defaultMailetContainerConfiguration()
                 .postmaster(POSTMASTER)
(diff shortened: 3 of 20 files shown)


examples/custom-james-assembly/custom-james-assembly.iml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
chibenwa, line 1: Please configure a local gitignore onto your laptop so that you do not commit ide file into an Apache project

(without modifying the project gitignore)

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -19,33 +19,58 @@
...
+                        boolean hasRight = mailboxManager.hasRight(
+                            MailboxPath.forUser(
+                                Username.fromMailAddress(recipient.stripDetails("+")),
+                                details
+                            ),
+                            MailboxACL.Right.Post,
+                            session
+                        );
+
+                        String targetFolder = hasRight ? details : "inbox";
```
chibenwa, line 67: Two things

 -> I am at war with `?` operator. It do not read like english and actually make reading the code more complex. Generally `if` reads better.
 
 -> `"inbox"` shall be a constant and not a crafted value. MailboxContants.INBOX is better.
 
 And finally if the guy has no right, do we need to do anything at all? After all, if we don't do anything, then it would land onto the recipient mailbox and if some prior storage directives had been positionned (spam checking?) we would not override it! 
 
 (that said maybe an INFO log saying "oups bob tried to address alice subfolder but did not had the right to do it" would be great!)

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -55,20 +63,36 @@ public void service(Mail mail) throws MessagingException {
...
+                            Username recipientUsername = Username.fromLocalPartWithoutDomain(recipient.getLocalPartWithoutDetails("+"));
+                            MailboxSession session = mailboxManager.createSystemSession(recipientUsername);
 
-                        String targetFolder = hasRight ? details : "inbox";
+                            MailboxACL aclCommand = mailboxManager.getMailbox(
+                                    MailboxPath.forUser(recipientUsername, details), session)
+                                .getMetaData(IGNORE, session, MessageManager.MailboxMetaData.FetchGroup.NO_COUNT)
+                                .getACL();
 
+                            hasRight = aclCommand.getEntries().containsKey(MailboxACL.ANYBODY_KEY) &&
```
chibenwa, line 77: Do not re-allocate variables

Generally method extraction is a good alternative.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -55,20 +63,36 @@ public void service(Mail mail) throws MessagingException {
...
+                            hasRight = false;
+                        }
+
+                        String targetFolder;
+                        if(hasRight) {
+                            targetFolder = details;
+                        } else {
+                            // TODO find LOG
+                            // LOG.info(maybesender + " tried to address " + recipient + "'s subfolder `" + details + "` but did not have the right to")
+                            targetFolder = MailboxConstants.INBOX;
```
chibenwa, line 90: Nope else we just do not do something. IE we only put the storage directive when we have rights?

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
```diff
@@ -59,37 +64,48 @@ public class SubAddressingTest {
...
+        MetricFactory metricFactory = new RecordingMetricFactory();
+
+        testee = new SubAddressing(usersRepository, mailboxManager, metricFactory);
+    }
 
-        mailboxManager.createMailbox(MailboxPath.forUser(Username.of("recipient"), "any"), session);
 
-        testee = new SubAddressing(usersRepository, mailboxManager, session);
+    @Test
+    void shouldNotAddStorageDirectiveWhenNoRight() throws Exception {
```
chibenwa, line 76: We need integration tests for this too

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
florentos17: 1. `?` operator & 2. MailboxContants.INBOX -> done

3. don't do anything if no right & 4. INFO log are still todo

florentos17: INFO log done

examples/custom-james-assembly/src/main/java/org/apache/james/examples/CustomJamesServerMain.java
```diff
@@ -60,22 +61,29 @@ public class CustomJamesServerMain implements JamesServerMain {
         new MemoryMailQueueModule(),
         new TaskManagerModule(),
         new RawPostDequeueDecoratorModule(),
-        binder -> binder.bind(MailetContainerModule.DefaultProcessorsConfigurationSupplier.class)
-            .toInstance(BaseHierarchicalConfiguration::new));
+        binder -> binder.bind(
+                MailetContainerModule
+                    .DefaultProcessorsConfigurationSupplier.class
+            ).toInstance(BaseHierarchicalConfiguration::new));
```
chibenwa, line 67:
```suggestion
        binder -> binder.bind(MailetContainerModule .DefaultProcessorsConfigurationSupplier.class)
                .toInstance(BaseHierarchicalConfiguration::new));
```

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public void service(Mail mail) throws MessagingException {
+        mail.getRecipients().forEach(recipient ->
+            recipient.getLocalPartDetails(UsersRepository.LOCALPART_DETAIL_DELIMITER)
+                .ifPresent(
+                    Throwing.consumer(targetFolder -> postIfHasRight(mail, recipient, targetFolder))
+                )
+        );
```
chibenwa, line 73:
```suggestion
        mail.getRecipients().forEach(recipient ->
            recipient.getLocalPartDetails(UsersRepository.LOCALPART_DETAIL_DELIMITER)
                .ifPresent(Throwing.consumer(targetFolder -> postIfHasRight(mail, recipient, targetFolder))));
```

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -0,0 +1,106 @@
...
+            MailboxSession session = mailboxManager.createSystemSession(recipientUsername);
+
+            MailboxACL aclCommand = mailboxManager
+                .getMailbox(MailboxPath.forUser(recipientUsername, targetFolder), session)
+                .getMetaData(IGNORE, session, MessageManager.MailboxMetaData.FetchGroup.NO_COUNT)
+                .getACL();
+
+            return aclCommand.getEntries().containsKey(MailboxACL.ANYBODY_KEY)
+                && aclCommand.getEntries()
+                .get(MailboxACL.ANYBODY_KEY)
```
chibenwa, line 99: Use `Optional` to wrap null case, transform your optional, and just call `aclCommand.getEntries().get(MailboxACL.ANYBODY_KEY)` once?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -0,0 +1,106 @@
...
+                .equals(MailboxACL.Rfc4314Rights.of(Collections.singletonList(MailboxACL.Right.Post)));
+
+        } catch (MailboxException e) {
+            return false;
```
chibenwa, line 103: Error handling is hard...

Problem 1: we are silenting potential problems here. It's bad! We likely shall log something when recovering from an exception.

Problem 2: recovering makes sense when facing MailboxNotFoundException -> OK


But mailbox exception is wider than this and includes other class of issue. Would it make sense to also recover from those? Or would ot be better to forward issues for problem we do not understand so that they get later retried?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -0,0 +1,106 @@
...
+        mail.getRecipients().forEach(recipient ->
+            recipient.getLocalPartDetails(UsersRepository.LOCALPART_DETAIL_DELIMITER)
+                .ifPresent(
+                    Throwing.consumer(targetFolder -> postIfHasRight(mail, recipient, targetFolder))
+                )
+        );
+    }
+
+    private void postIfHasRight(Mail mail, MailAddress recipient, String targetFolder) throws UsersRepositoryException {
+        if(hasPostRight(recipient, targetFolder)) {
```
chibenwa, line 77:
```suggestion
        if (hasPostRight(recipient, targetFolder)) {
```

Checkstyle...

chibenwa:
(While I have been asking for it privately, ) I feel saddened that we skip right resolution logic alltogether.

Saying that we do not handled setting 'p' rights for external  individuals is one things.

But over-simplifying the right resolution logic seems like a mistake.

For instance as off today we would not take into account negative ACLs (every body can port to my mailbox vietnam except Forent). And there would be nothingmore harmfull than thinking you denied Florent that right while actually you did not.

It turns out we can pretty simply, given the MailboxACL resolve the rights independently from the RightManager (which messes around with MailboxSession....)

We could use `MailboxACLResolver`:

```
mailboxACLResolver.resolveRights(sender, acl, recipient).contains(Right.P)
```

Or something like this....

And of course add tests for ACL restrictions...

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -0,0 +1,106 @@
...
+
+            MailboxACL aclCommand = mailboxManager
+                .getMailbox(MailboxPath.forUser(recipientUsername, targetFolder), session)
+                .getMetaData(IGNORE, session, MessageManager.MailboxMetaData.FetchGroup.NO_COUNT)
+                .getACL();
+
+            return aclCommand.getEntries().containsKey(MailboxACL.ANYBODY_KEY)
+                && aclCommand.getEntries()
+                .get(MailboxACL.ANYBODY_KEY)
+                .equals(MailboxACL.Rfc4314Rights.of(Collections.singletonList(MailboxACL.Right.Post)));
```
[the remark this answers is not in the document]
florentos17: correct
i replaced `.equals(MailboxACL.Rfc4314Rights.of(Collections.singletonList(MailboxACL.Right.Post)))`
with `.contains(MailboxACL.Right.Post)`
which is far more elegant...

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
florentos17: forwarding of course... :)

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -19,33 +19,86 @@
...
+                .forEach(mail::setAttribute);
+        } else {
+            LOG.info("{} tried to address {}'s subfolder `{}` but did not have the right to",
+                mail.getMaybeSender().toString(), recipient, targetFolder);
+        }
+    }
+
+    private Boolean hasPostRight(MailAddress recipient, String targetFolder) throws MailboxException {
+
+        try {
```
chibenwa, line 84: remove this extra blank line

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -19,33 +19,86 @@
...
         mail.getRecipients().forEach(recipient ->
-            recipient.getLocalPartDetails("+")
-                .ifPresent(
-                    Throwing.consumer(details ->
-                    StorageDirective.builder().targetFolders(ImmutableList.of(details)).build()
-                        .encodeAsAttributes(usersRepository.getUsername(recipient))
-                        .forEach(mail::setAttribute))
-                ));
+            recipient.getLocalPartDetails(UsersRepository.LOCALPART_DETAIL_DELIMITER)
+                .ifPresent(Throwing.consumer(targetFolder -> postIfHasRight(mail, recipient, targetFolder))));
```
chibenwa, line 68: postIfHasRight is a brittle method name: it embeds 2 meaning, a condition and an action.

Likely we can avoid that by leveraging functional semantic:

```
            recipient.getLocalPartDetails(UsersRepository.LOCALPART_DETAIL_DELIMITER)
                .filter(targetFolder -> hasPostRight(recipient, targetFolder))
                .ifPresent(Throwing.consumer(targetFolder -> post(mail, recipient, targetFolder))));
```

florentos17: it's not as simple:
1. for the `LOG.info`, `targetFolder` cannot be resolved as the `Else` part is a `Runnable` and not a `Consumer`. We could maybe log a message without specifying the targetFolder, or even remove the log message at all.
2. `service()` cannot throw a MailboxException, so it cannot call `hasPostRight` which does. I tried to do something like `.filter(ThrowingPredicate(targetFolder -> hasPostRight(mail, recipient, targetFolder)))` but it doesn't work, as targetFolder is no longer considered a String...


would something like this be acceptable ?
```
recipient.getLocalPartDetails(UsersRepository.LOCALPART_DETAIL_DELIMITER)
                .ifPresent(Throwing.consumer(targetFolder -> {
                    if (hasPostRight(mail, recipient, targetFolder)) {
                        post(mail, recipient, targetFolder);
                    } else {
                        LOG.info("{} tried to address {}'s subfolder `{}` but did not have the right to",
                            mail.getMaybeSender().toString(), recipient, targetFolder);
                    }
                })));
```

florentos17:
> (While I have been asking for it privately, ) I feel saddened that we skip right resolution logic alltogether.
> 
> Saying that we do not handled setting 'p' rights for external individuals is one things.
> 
> But over-simplifying the right resolution logic seems like a mistake.
> 
> For instance as off today we would not take into account negative ACLs (every body can port to my mailbox vietnam except Forent). And there would be nothingmore harmfull than thinking you denied Florent that right while actually you did not.
> 
> It turns out we can pretty simply, given the MailboxACL resolve the rights independently from the RightManager (which messes around with MailboxSession....)
> 
> We could use `MailboxACLResolver`:
> 
> ```
> mailboxACLResolver.resolveRights(sender, acl, recipient).contains(Right.P)
> ```
> 
> (It would be anyway cleaner than doing a naive ACL resolution...)
> 
> Or something like this....
> 
> And of course add tests for ACL restrictions...

---
Using `mailboxACLResolver.resolveRights(sender, acl, recipient)` works very well, except in the case where the MaybeSender is not known. Fortunately, resolveRights can also take `null` as a sender, and works the way we want: if the owner gives the right to everybody, then the right is granted even with a `null` sender.
This results in the following code:

```
Username username;
try {
    username = Username.fromMailAddress(mail.getMaybeSender().get());
} catch (NoSuchElementException e) {
    username = null;
}

MailboxACL.Rfc4314Rights rights = mailboxACLResolver.resolveRights(
    username,
    aclCommand,
    recipient.toString());
``` 
   
This works, but I know we don't like either variable re-allocations nor `null`. What do you suggest ? I could adapt `resolveRight()` to accept `Optional.empty()` instead of `null`, but is it my job to do it ?

chibenwa:
Using functionnal constructs we can get it slightly cleaner:

```
mail.getMaybeSender().map(usersrepository::getUsernameForMailAddress).orElse(null);
```

Which is slightly better (use UsersRepository to convert MailAddress to Username!)

> What do you suggest ? I could adapt resolveRight() to accept Optional.empty() instead of null, but is it my job to do it ?

I Am not fan either to be honest: null as a parameter for `resolveRights` usage being working seems mostly accidental...

 - Do the current code base currently uses "null" to denote absence of user or "anybody"?
 - Do the current code base actively test this specific value and specifies clearly the overall behaviour?

As none of these answer is negative we likely cannot rely on this 'null' parameter.

Once again 'p' implies resolving rights as part of the delivery context which is by essence different of applying rights directly linked to user actions. Hence having 'SubAddressing feature bridging the gap seems reasonable to me.

In pseudo code this is what I would write:

```
MailboxACL.Rfc4314Rights rights = mail.getMaybeSender().map(usersrepository::getUsernameForMailAddress)
   .map(resolveUserRights)
   .orElseGet(getAnyBodyRights); // read directly the ACL for null sender
```

What I also really do not like is the third parameter being an untyped String and not a stronger type.... Like `Username`.  We *could* get one commit to address this.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -19,33 +19,88 @@
...
+
+import com.github.fge.lambdas.Throwing;
+import com.google.common.collect.ImmutableList;
+
 
 public class SubAddressing extends GenericMailet {
     private final UsersRepository usersRepository;
+    private final MailboxManager mailboxManager;
+
+    private static final Logger LOG = LoggerFactory.getLogger(SubAddressing.class);
```
chibenwa:
```
    private static final Logger LOG = LoggerFactory.getLogger(SubAddressing.class);
    
    private final UsersRepository usersRepository;
    private final MailboxManager mailboxManager;
```

Static fileds go first (convention)

core/src/test/java/org/apache/james/core/MailAddressTest.java
```diff
@@ -225,4 +228,93 @@ void shouldMatchBeanContract() {
...
+    void getLocalPartDetailsShouldBePreciseWithMultipleCharacterDelimiter() throws AddressException {
+        MailAddress mailAddress = new MailAddress("localpart---details@example.com");
+        assertThat(mailAddress.getLocalPartDetails("--")).contains("-details");
+    }
+
+
+
+
+
+    @Test
```
chibenwa, line 278:
```suggestion
    }

    @Test
```

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/SubAddressingTest.java
```diff
@@ -0,0 +1,155 @@
...
+        messageSender.close();
+    }
+
+    @Test
+    void subAddressedEmailShouldBeDeliveredInINBOXWhenSpecifiedFolderDoesNotExist(@TempDir File temporaryFolder) throws Exception {
+        setup(temporaryFolder);
+
+        //do not create mailbox
+
+        sendMail();
```
chibenwa, line 102:
```suggestion
        sendMailToSubAddress();
```

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
chibenwa: I suggest looking at this together!

chibenwa (commented):
- We do not test nagative ACLs
 - We do not test null sender
 - We likely can further refactor SubAddressing mailet
 
 Let's have a little pair on this topic.

chibenwa:
- We miss documentation of this new mailet

chibenwa:
Can we stop force pushing all the time? I'd rather review minimal fixups...

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
```diff
@@ -96,40 +97,79 @@ void shouldNotAddStorageDirectiveWhenNoRight() throws Exception {
...
+    private void giveRightToPost(MailboxId mailboxId, MailboxSession session) throws MailboxException {
+        MailboxACL.ACLCommand command = MailboxACL.command()
+            .key(MailboxACL.ANYBODY_KEY)
+            .rights(MailboxACL.Right.Post)
+            .asAddition();
+
+        mailboxManager.applyRightsCommand(mailboxId, command, session);
+    }
+
+    private Mail sendSubAddressedMail(boolean knownSender) throws MessagingException, IOException {
```
chibenwa, line 161: boolean parameter is a code smell and proves things likely need a redesign

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
```diff
@@ -96,40 +97,79 @@ void shouldNotAddStorageDirectiveWhenNoRight() throws Exception {
...
+        giveRightToPost(mailboxId, session);
+        Mail mail = sendSubAddressedMail(false);
 
         AttributeName recipient = AttributeName.of("DeliveryPaths_recipient@localhost");
         assertThat(mail.attributes().map(this::unbox))
             .doesNotContain(Pair.of(recipient, "any"));
     }
 
-    private Mail sendSubAddressedMail() throws MessagingException, IOException {
+    private void giveRightToPost(MailboxId mailboxId, MailboxSession session) throws MailboxException {
```
chibenwa, line 152: If we make MailboxACLCommand a constant then giveRightToPost method can likely be removed as it would add nothing.

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
chibenwa: init of the mailet has nothing to do with sending mail. Can we move it in @BeforeEach ?

chibenwa: Protip: just return a FakeMail.Builder without the sender, and let caller decide if he wants sender or not and let him build the mail then execute the mailbet on it.

Caller code:

```
Mail mail = mailBuilder().sender(...).build();
testee.service(mail);
```

Which is more readable (tests shall be as straightforward as possible and avoid ccidental indirection that obfuscate them...

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
```diff
@@ -96,40 +97,79 @@ void shouldNotAddStorageDirectiveWhenNoRight() throws Exception {
...
+
+    @Test
+    void shouldAddStorageDirectiveWhenEveryoneHasRightAndSenderIsUnknown() throws Exception {
+
+        Username username = Username.of("recipient");
+        MailboxSession session = mailboxManager.createSystemSession(username);
+
+        // create Mailbox
+        MailboxId mailboxId = mailboxManager.createMailbox(
+                MailboxPath.forUser(username, "any"), session).get();
```
chibenwa, line 124: I'd just move it in @BeforeEach and handle the case mailbox do not exist with a mailbox called "notfound"

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
```diff
@@ -96,40 +97,79 @@ void shouldNotAddStorageDirectiveWhenNoRight() throws Exception {
...
+        AttributeName recipient = AttributeName.of("DeliveryPaths_recipient@localhost");
+        assertThat(mail.attributes().map(this::unbox))
+            .doesNotContain(Pair.of(recipient, "any"));
+    }
+
+    @Test
+    void shouldAddStorageDirectiveWhenEveryoneHasRightAndSenderIsUnknown() throws Exception {
+
+        Username username = Username.of("recipient");
+        MailboxSession session = mailboxManager.createSystemSession(username);
```
chibenwa, line 120: Relocate this boiler plate in @beforeEach, creating related class field as needed

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
florentos17: fair point ! 
the only thing is that initializing the ACLCommand throws an exception so i cannot initialize it as a static final constant directly. so i used a @BeforeAll method to initialize the different commands i will have to use, is that ok ?

chibenwa: Yes IMO ok

docs/modules/servers/partials/SubAddressing.adoc
```diff
@@ -0,0 +1,22 @@
+=== SubAddressing
```
chibenwa, line 1: https://james.staged.apache.org/james-project/3.9.0/servers/distributed/configure/mailets.html

It would be great to have an exemple that can easily get copy pasted

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
```diff
@@ -0,0 +1,213 @@
...
+
+        AttributeName recipient = AttributeName.of("DeliveryPaths_recipient@localhost");
+        assertThat(mail.attributes().map(this::unbox))
+            .doesNotContain(Pair.of(recipient, TARGET));
+    }
+
+    @Test
+    void shouldNotAddStorageDirectiveWhenSenderIsBlackListed() throws Exception {
+        // blacklist sender 1 and send from sender 1
+        givePostRightForKey(MailboxACL.EntryKey.createUserEntryKey(sender1Username, true));
```
chibenwa, line 147: true is a magic constant. Can't guess easily what it means. Extracting it as a constant could allow naming and help express the intent.

Also we could improve EntryKey api:

```
// API stability
MailboxACL.EntryKey.createUserEntryKey(sender1Username, false|false)

MailboxACL.EntryKey.createUserEntryKey(sender1Username)

MailboxACL.EntryKey.createNegativeUserEntryKey(sender1Username)
```

chibenwa, line 147: Grant rights to everybody before restricting them for sender1 otherwise you do not assert anything!

docs/modules/servers/partials/SubAddressing.adoc
florentos17: should i just add that ?

```
<mailet match="All" class="SubAddressing">
</mailet>
```

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
```diff
@@ -0,0 +1,220 @@
...
+
+        Mail mail = mailBuilder(TARGET).sender(SENDER1).build();
+        testee.service(mail);
+
+        AttributeName recipient = AttributeName.of("DeliveryPaths_recipient@localhost");
+        assertThat(mail.attributes().map(this::unbox))
+            .containsOnly(Pair.of(recipient, TARGET));
+    }
+
+    @Disabled
```
Arsnael, line 120: Why is this test disabled? Usually we leave a reason with it like, so that if one day we want to fix it, we can remember quickly what was wrong in the first place

```
@Disabled("[]REASON]")
```

Arsnael (commented):
LGTM overall, not easy for a first PR, good job !

Just a small comment

server/mailet/mailets/src/test/java/org/apache/james/transport/mailets/SubAddressingTest.java
florentos17: final fixups were required as the test was still red, but it should be good now ;)

docs/modules/servers/partials/SubAddressing.adoc
chibenwa: Yes

Arsnael (approved):

Arsnael:
Good to merge correct?

quantranhong1999 (approved):
Good job.

Tips: next time a better PR name would help.

PR #2409, opened 2024-09-12
JAMES-4054 Update SMIMECheckSignatureIntegrationTest to include IsSMIMESigned matcher

hungphan227:


mailet/crypto/src/main/java/org/apache/james/transport/matcher/IsSMIMESigned.java
```diff
@@ -45,8 +45,7 @@ public Collection<MailAddress> match(Mail mail) throws MessagingException {
         if (message == null) {
             return null;
         }
-        
-        
+
```
chibenwa, line 48: Where are the (unit) tests for this class?

Arsnael: From what I see there is no unit tests in general at all for mailets and matchers declared in the crypto module, everything seems tested on an integration test level?

Arsnael (approved):

Arsnael (approved):

mailet/crypto/src/main/java/org/apache/james/transport/matcher/IsSMIMESigned.java
chibenwa: THis one is a simple mailet only checking content type and its logic is not well exercised by the integration tests.

I do believe unit test makes full sense here and are quick and easy to write.

hungphan227: added

quantranhong1999 (approved):

mailet/crypto/src/test/java/org/apache/james/transport/matcher/IsSMIMESignedTest.java
```diff
@@ -0,0 +1,117 @@
...
+public class IsSMIMESignedTest {
+    private IsSMIMESigned isSMIMESigned;
+
+    @BeforeEach
+    void beforeEach() {
+        isSMIMESigned = new IsSMIMESigned();
+    }
+
+    @Test
+    void matchShouldReturnNonEmptyListWhenMessageContentTypeIsMultipartSigned() throws Exception {
```
chibenwa, line 40: The same test, again and again...

How about using junit 5 parameters?

hungphan227: updated

Arsnael (approved):

chibenwa (approved):

PR #2416, opened 2024-09-19
[JAMES-3696] fix broken tests

jeantil:
When I tried to start working on the flaky test I ran the `PulsarMailqueueTest` suite and one of the tests failed. Using `git bisect` I was able to trace it back to the introduction of the deadletter policy a while ago and yet all the pipelines were ok on the CI. 
I looked into the `unstable-test` stage and in develocity but couldn't find any trace of the `PulsarMailqueueTest` suite being executed. 

Doing local tests I realized that when the whole class is tagged, for some reason, the suite will not be executed, neither in the normal test stage or in the `unstable-test` stage. This does not only affect `PulsarMailQueueTest`. See my attempt at fixing the underlying problem in https://github.com/apache/james-project/pull/2417

This PR focuses on resolving the flakyness of the pulsar mailqueue tests by : 
- Fixing the currently broken test
- Changing the way we assert to allow for an asychronous distribution of removal filters.

Doing so led Matthieu and I to discover another issue in the `ManageableMailQueueContract` : several tests assertions were verifying exactly nothing. We fixed that in the 3rd commit of this PR (https://github.com/apache/james-project/pull/2416/commits/a7541825528f08484f885b15000205dd06eb9a13 as I write this). Doing so raised a failing test case for the `ActiveMQMailQueueBlobTest`

### DIFF
--- a/server/queue/queue-api/src/test/java/org/apache/james/queue/api/MailQueueContract.java
+++ b/server/queue/queue-api/src/test/java/org/apache/james/queue/api/MailQueueContract.java
@@ -35,6 +35,7 @@ import java.util.Date;
 import java.util.EnumSet;
 import java.util.Iterator;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -516,6 +517,12 @@ public interface MailQueueContract {
             .hasSize(totalDequeuedMessages);
     }
 
+    enum EnqueueDequeueSteps {
+        RETRY,
+        DEQUEUE,
+        END
+    }
+
     @Test
     default void concurrentEnqueueDequeueWithAckNackShouldNotFail() throws Exception {
         MailQueue testee = getMailQueue();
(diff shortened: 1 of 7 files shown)


server/queue/queue-pulsar/src/test/java/org/apache/james/queue/pulsar/PulsarMailQueueTest.java
```diff
@@ -316,4 +324,58 @@ public void browseShouldReturnEmptyWhenSingleDequeueMessageEvenWhenStoreIsGuaran
                 .isEmpty();
     }
 
+    @Test
+    void dequeueShouldAllowToNackAndRetrieveSeveralMails() throws Exception {
```
jeantil, line 328: ignore this test it was used to investigate the issue and should have been removed

server/queue/queue-pulsar/src/test/java/org/apache/james/queue/pulsar/PulsarMailQueueTest.java
```diff
@@ -316,4 +324,58 @@ public void browseShouldReturnEmptyWhenSingleDequeueMessageEvenWhenStoreIsGuaran
...
+                .containsOnly("name1", "name2","name3");
+    }
+
+    @Tag(Unstable.TAG)
+    @Override
+    public void removeByRecipientShouldRemoveSpecificEmail() throws Exception {
+        DelayedManageableMailQueueContract.super.removeByRecipientShouldRemoveSpecificEmail();
+    }
+
+    @ParameterizedTest
```
jeantil, line 374: this test is missing the unstable tag but we will actually propose a proper fix for flakyness

Arsnael (commented):
Read it . Looks good to me so far.

Noticed things sometimes as well breaking in tmail because it's not been really tested on James, there is clearly something better we could do here :)

chibenwa:
Read it.
To be honnest I do not grasp fully changes here but it seems legitimate.

server/queue/queue-api/src/test/java/org/apache/james/queue/api/MailQueueContract.java
```diff
@@ -526,22 +533,35 @@ default void concurrentEnqueueDequeueWithAckNackShouldNotFail() throws Exception
...
-                    MailQueue.MailQueueItem mailQueueItem = deque.takeLast();
-                    mailQueueItem.done(MailQueue.MailQueueItem.CompletionStatus.RETRY);
-                }
-                if (step % 3 == 2) {
+                    mailStates.put(name, EnqueueDequeueSteps.RETRY);
+                } else {
                     MailQueue.MailQueueItem mailQueueItem = deque.takeLast();
-                    dequeuedMails.add(mailQueueItem.getMail());
-                    mailQueueItem.done(MailQueue.MailQueueItem.CompletionStatus.SUCCESS);
+                    var name = mailQueueItem.getMail().getName();
```
jeantil, line 549: For this test what we want is to : 
enqueue->retry->success each mail
in the previous implementation since there was no tracking of the state of a specific mail it could result in a mail getting retried twice once another got success right away. 
Since the pulsar mailqueue has a dead letter policy that puts a message in dead letter after 1 retry, retrying a message more than once results in that message never being dequeable again. This in turn means the test woud block and timeout.

server/queue/queue-api/src/test/java/org/apache/james/queue/api/ManageableMailQueueContract.java
```diff
@@ -457,24 +454,25 @@ default void removeShouldNotFailWhenBrowsingIterating() throws Exception {
...
 
-        getManageableMailQueue().remove(ManageableMailQueue.Type.Name, "name2");
+        getManageableMailQueue().remove(ManageableMailQueue.Type.Name, "name99");
 
-        awaitRemove();
 
-        assertThatCode(() ->  Iterators.consumingIterator(items)).doesNotThrowAnyException();
+        assertThatCode(
+                () -> assertThat(consumeIterator(items)).isEqualTo(98) // and 998 here so 999 in total
+        ).doesNotThrowAnyException();
```
jeantil, line 475: interestingly, this test fails except for the pulsar implementation. 
Most implementations will return the removed item if the browse has been started before the item is removed. The pulsar implementation is able to filter it out even from an operating iterator. 

I am considering lowering the constraint here with this code
```
        assertThatCode(
                // the remove may or may not be applied to an already started iterator
                // as long as it doesn't make the iterator crash
                () -> assertThat(consumeIterator(items)).isGreaterThanOrEqualTo(98).isLessThan(100)
        ).doesNotThrowAnyException();
```

chibenwa: So the over implementation "browse" reads from a snapshot while here deletion are dynamically applied onto the "browsing side. Ok with me.

I am ok with the isolation level of the browse to be implementation specific.

We could likely benefit from a little write up regarding this? Interesting details could be documented...

jeantil: > We could likely benefit from a little write up regarding this? Interesting details could be documented...

I'm ok to give it a try, where would you expect that : in the API doc of the ManagedMailqueue, in a page of the website (new/old) or on the ML ?

Arsnael (approved):

jeantil:
finally a green build after many builds failed on flaky
I'll merge as is for now and do a followup for the documentation of the snapshot vs live filtered browse

PR #2417, opened 2024-09-19
[devscout] fix execution of unstable group

jeantil:
classes annotated unstable at classlevel are properly execucted with  -Punstable

this may not be true for 2 modules that override the surefire plugin definition but i'm not sure what they actually attempt to do : 
- distributed-jmap-rfc-8621-integration-tests
- rabbitmq-webadmin-integration-test

at the very least the situation will be better than it was before this change

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -644,7 +644,7 @@
         <maven-plugin-api.version>3.9.3</maven-plugin-api.version>
         <maven-reporting-impl.version>3.1.0</maven-reporting-impl.version>
         <maven-reporting-api.version>3.1.1</maven-reporting-api.version>
-        <maven-surefire-plugin.version>3.3.0</maven-surefire-plugin.version>
+        <maven-surefire-plugin.version>3.5.0</maven-surefire-plugin.version>
         <maven-enforcer-plugin.version>3.5.0</maven-enforcer-plugin.version>
         <qdox.version>2.0.3</qdox.version>
         <jaxb.version>2.3.1</jaxb.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -3326,7 +3326,7 @@
                         <reuseForks>true</reuseForks>
                         <!-- Fail tests longer than 30 minutes, prevent form random locking tests -->
                         <forkedProcessTimeoutInSeconds>1800</forkedProcessTimeoutInSeconds>
-                        <excludedGroups>unstable</excludedGroups>
```
Arsnael, line 3329: Isn't that declared in a bunch of other pom across the project? should we remove them too?

jeantil: well I'm not entirely sure. 

without this change the command 
```
mvn clean test -Dtest=org.apache.james.linshare.LinshareBlobExportMechanismTest -Dsurefire.failIfNoSpecifiedTests=false -am -pl :apache-james-linshare -Dcheckstyle.skip -Punstable-tests
```

doesn´t execute anything, with the change it does run the corresponding tests. 

but there are tests that are tagged `BasicFeature`, unless their module has a custom surefire config to include these tests, they would be excluded by the change i applied

so we probably need something more involved such as `none() |  BasicFeature`

group settings for junit5 are documented at 
https://maven.apache.org/surefire/maven-surefire-plugin/examples/junit-platform.html#filtering-by-tags
and if I understand correctly it is to be combined with https://junit.org/junit5/docs/current/user-guide/#running-tests-tag-expressions to produce the proper selection.

chibenwa (approved):

jeantil:
as @Arsnael remarked this in only a partial fix and will probably require more refinements for all the other occurences of 
```
                        <excludedGroups>unstable</excludedGroups>
```
its still better than before :D

PR #2423, opened 2024-09-27
[JAMES-4064] switch pulsarmailqueue filter topics to non durable subscriptions

jeantil:
As discussed in the issue, filter topics need to use unique subscription names to fetch all applicable filters upon restart. 

At the moment we use durable subscriptions which means that these subscriptions accumulate across restarts on the pulsar side, eventually reaching the limit of allowed subscriptions per topic. 

Alternatives
- We could try to be good citizens and delete the subscriptions but there is no guarantee that the process terminates properly and does the cleanup
- We could use another kind of UUID which is strictly incremental and or includes a timestamp, allowing to build a safe cleanup script, but it would be difficult to run automatically in a safe way. 
- We could expose the filter subscription names of the active nodes to make sure the cleanup script doesn't accidentally delete currently used subscriptions
 
In this PR I propose to use the [non-durable](https://pulsar.apache.org/docs/3.3.x/concepts-messaging/#what-is-a-subscription-mode) subscriptions of Pulsar which seem to me to be the safest and cleanest solution to this issue,

### DIFF
--- a/server/queue/queue-pulsar/src/main/scala/org/apache/james/queue/pulsar/PulsarMailQueue.scala
+++ b/server/queue/queue-pulsar/src/main/scala/org/apache/james/queue/pulsar/PulsarMailQueue.scala
@@ -46,7 +46,7 @@ import org.apache.james.queue.pulsar.EnqueueId.EnqueueId
 import org.apache.james.server.core.MailImpl
 import org.apache.mailet._
 import org.apache.pulsar.client.admin.PulsarAdminException.NotFoundException
-import org.apache.pulsar.client.api.{DeadLetterPolicy, Schema, SubscriptionInitialPosition, SubscriptionType}
+import org.apache.pulsar.client.api.{DeadLetterPolicy, Schema, SubscriptionInitialPosition, SubscriptionMode, SubscriptionType}
 import org.reactivestreams.Publisher
 import org.slf4j.LoggerFactory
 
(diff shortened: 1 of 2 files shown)


Arsnael (approved):

chibenwa:
I +1 this solution.

> We could try to be good citizens and delete the subscriptions but there is no guarantee that the process terminates properly and does the cleanup

Not mutually exclusive, right?

chibenwa (approved):

jeantil:
> Not mutually exclusive, right? Or does this work already deliver somehow promisses of a "good citizen cleanup"?

Unfortunately is seems to be mutually exclusive.

Putting this code either at the beginning or at the end of the `close()` method 
```
  admin.topics().deleteSubscription(filterTopic.name, filterSubscription.name)
  admin.topics().deleteSubscription(filterScheduledTopic.name, scheduledFilterSubscription.name)
```
 fails 
* at the begging with
```
org.apache.pulsar.client.admin.PulsarAdminException$PreconditionFailedException: Subscription has active connected consumers
```
* at the end with
```
org.apache.pulsar.client.admin.PulsarAdminException$NotFoundException: Subscription filter-subscription-cmieYQmrma-dce269ce-d092-4dc4-ae2b-e82a4d05cf83 not found for topic persistent://test/czSUkoXioi/pmq-filter-cmieYQmrma
```

PR #2434, opened 2024-10-03
JAMES-4044 LDAP matchers: support extracting attribute from DN

quantranhong1999:
Supports extracting LDAP attribute from its DN value. e.g.

`description: cn=active,ou=accountStatus,ou=whatever,o=gov,c=vn` => `description: active`

- [x] HasLDAPAttribute
- [x] SenderHasLDAPAttribute
- ...

### DIFF
--- a/server/mailet/ldap/src/main/java/org/apache/james/transport/matchers/HasLDAPAttribute.java
+++ b/server/mailet/ldap/src/main/java/org/apache/james/transport/matchers/HasLDAPAttribute.java
@@ -35,19 +35,15 @@ import org.apache.james.user.ldap.LDAPConnectionFactory;
 import org.apache.james.user.ldap.LdapRepositoryConfiguration;
 import org.apache.mailet.Mail;
 import org.apache.mailet.base.GenericMatcher;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import com.github.fge.lambdas.Throwing;
 import com.google.common.cache.Cache;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.unboundid.ldap.sdk.DN;
 import com.unboundid.ldap.sdk.Filter;
 import com.unboundid.ldap.sdk.LDAPConnectionPool;
 import com.unboundid.ldap.sdk.LDAPException;
 import com.unboundid.ldap.sdk.LDAPSearchException;
-import com.unboundid.ldap.sdk.RDNNameValuePair;
 import com.unboundid.ldap.sdk.SearchResult;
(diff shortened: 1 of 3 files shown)


server/mailet/ldap/src/main/java/org/apache/james/transport/matchers/HasLDAPAttribute.java
```diff
@@ -170,10 +172,27 @@ private boolean hasAttribute(SearchResultEntry entry) {
...
+    public static String extractLdapAttributeValue(String ldapValue) {
+        if (ldapValue.contains(",")) {
+            try {
+                return Arrays.stream(new DN(ldapValue).getRDNs())
+                    .flatMap(rdn -> rdn.getNameValuePairs().stream())
+                    .filter(pair -> pair.getAttributeName().equals("cn"))
+                    .findFirst()
+                    .map(RDNNameValuePair::getAttributeValue)
+                    .orElse(ldapValue);
+            } catch (LDAPException e) {
```
chibenwa, line 189: INFO log with the actual value in there.

server/mailet/ldap/src/main/java/org/apache/james/transport/matchers/HasLDAPAttribute.java
```diff
@@ -170,10 +176,28 @@ private boolean hasAttribute(SearchResultEntry entry) {
         return attributeValue.map(value -> Optional.ofNullable(entry.getAttribute(attributeName))
                 .map(attribute -> Arrays.stream(attribute.getValues()))
                 .orElse(Stream.empty())
+                .map(ldapValue -> extractLdapAttributeValue(ldapValue, attributeName, LOGGER))
```
chibenwa, line 179: Why passing the LOGGER as a parameter?

server/mailet/ldap/src/main/java/org/apache/james/transport/matchers/HasLDAPAttribute.java
```diff
@@ -170,10 +176,28 @@ private boolean hasAttribute(SearchResultEntry entry) {
...
                 .anyMatch(value::equals))
             .orElseGet(() -> entry.hasAttribute(attributeName));
     }
 
+    public static String extractLdapAttributeValue(String ldapValue, String attributeName, Logger logger) {
```
chibenwa, line 184: Could be better suited in a different class named AttributeUtils

chibenwa, line 184: I'd put attributeName first and would not be against of strong typing

Arsnael (approved):

PR #2435, opened 2024-10-03
JAMES-3945 Subaddressing - case sensitivity and subfolders

florentos17:
1) Currently we have a case sensitivity problem: folders are case sensitive whereas email addresses are not. We would want both `fazavant+folder@domain` and `fazavant+Folder@domain` to be delivered in my folder `Folder`. Moreover, if both `folder` and `Folder` exist, the correct one should be selected.

2) Since we agreed to give the `p` right to either noone or anyone, there is no need for unit tests about whitelists and blacklists: I removed them

3) We want to check whether a subfolder can be subaddressed. I wrote a unit test that sends an email to `fazavant+folder/subfoder@domain`, and it seems to be all good :) is that enough or should I write more advanced tests ?

### DIFF
--- a/mailbox/api/src/main/java/org/apache/james/mailbox/model/search/ExactNameCaseIncensitive.java
+++ b/mailbox/api/src/main/java/org/apache/james/mailbox/model/search/ExactNameCaseIncensitive.java
@@ -1,0 +1,70 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.mailbox.model.search;
+
+import com.google.common.base.Preconditions;
+
+import java.util.Objects;
+
+public class ExactNameCaseIncensitive implements MailboxNameExpression {
+
+    private final String name;
+
+    public ExactNameCaseIncensitive(String name) {
+        Preconditions.checkNotNull(name);
+        this.name = name;
+    }
+
+    @Override
+    public boolean isExpressionMatch(String mailboxName) {
+        Preconditions.checkNotNull(mailboxName);
+        return name.equals(mailboxName);
+    }
+
+    @Override
+    public String getCombinedName() {
+        return name;
+    }
+
+    @Override
+    public boolean isWild() {
+        return false;
+    }
+
+    @Override
+    public MailboxNameExpression includeChildren() {
+        return new PrefixedWildcard(name);
+    }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (o instanceof ExactNameCaseIncensitive) {
+            ExactNameCaseIncensitive exactName = (ExactNameCaseIncensitive) o;
+
+            return Objects.equals(this.name, exactName.name);
+        }
+        return false;
+    }
+
+    @Override
+    public final int hashCode() {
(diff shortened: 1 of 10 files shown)


server/mailet/integration-testing/src/test/java/org/apache/james/mailets/SubAddressingTest.java
```diff
@@ -100,10 +101,62 @@ void subAddressedEmailShouldBeDeliveredInINBOXWhenSpecifiedFolderDoesNotExist(@T
...
+
+    @Test
+    void subAddressedEmailShouldBeDeliveredInSpecifiedFolderWithCorrectUpperCaseWhenSeveralCasesExist(@TempDir File temporaryFolder) throws Exception {
+        setup(temporaryFolder);
+
+        // create mailboxes
+        testIMAPClient.sendCommand("CREATE " + TARGETED_MAILBOX);
+        testIMAPClient.sendCommand("SETACL " + TARGETED_MAILBOX + " " + "anyone" + " p");
+        testIMAPClient.sendCommand("CREATE " + TARGETED_MAILBOX_UPPER);
+        testIMAPClient.sendCommand("SETACL " + TARGETED_MAILBOX_UPPER + " " + "anyone" + " p");
```
chibenwa, line 154:
```suggestion
        testIMAPClient.sendCommand("CREATE " + TARGETED_MAILBOX);
        testIMAPClient.sendCommand("SETACL " + TARGETED_MAILBOX + " " + "anyone" + " p");
        testIMAPClient.sendCommand("CREATE " + TARGETED_MAILBOX_UPPER);
```

Then what happens?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/SubAddressing.java
```diff
@@ -83,33 +86,51 @@ public SubAddressing(UsersRepository usersRepository, @Named("mailboxmanager") M
...
     }
 
-    private void postIfHasRight(Mail mail, MailAddress recipient, String targetFolder) throws UsersRepositoryException, MailboxException {
-        if (hasPostRight(mail, recipient, targetFolder)) {
-            StorageDirective.builder().targetFolders(ImmutableList.of(targetFolder)).build()
+    private Optional<MailboxPath> getPathWithCorrectCase(MailAddress recipient, String targetFolder) throws UsersRepositoryException, MailboxException {
+        Username recipientUsername = usersRepository.getUsername(recipient);
+        MailboxSession session = mailboxManager.createSystemSession(recipientUsername);
+
+        return mailboxManager.list(session).stream()
```
chibenwa, line 99: search(MailboxQuery) !!!!!!

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/SubAddressingTest.java
florentos17: the desired outcome was already correctly achieved, but i added a test for that specific case to make it clear :)

Arsnael (commented):
Read it nothing to add

mailbox/api/src/main/java/org/apache/james/mailbox/model/search/ExactNameCaseIncensitive.java
```diff
@@ -0,0 +1,70 @@
...
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.mailbox.model.search;
+
+import com.google.common.base.Preconditions;
+
+import java.util.Objects;
+
+public class ExactNameCaseIncensitive implements MailboxNameExpression {
```
chibenwa, line 26: I do not see where case insentivity is?

chibenwa: Unit tests for these class?

mailbox/api/src/main/java/org/apache/james/mailbox/model/search/PrefixedWildcardCaseIncensitive.java
```diff
@@ -0,0 +1,69 @@
...
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.mailbox.model.search;
+
+import com.google.common.base.Preconditions;
+
+import java.util.Objects;
+
+public class PrefixedWildcardCaseIncensitive implements MailboxNameExpression {
```
chibenwa, line 26: Where is this used?

florentos17:
i created a new implementation of `MailboxNameExpression` to find only the mailboxes that match the name without taking the case into account - it is almost identical to the existing `ExactName` except for the `isExpressionMatch` and `includeChildren` methods.
i added unit tests which are also extremely similar to those of `ExactName`: just a couple of extra tests to test various cases.

chibenwa (approved):

Arsnael (approved):

PR #2439, opened 2024-10-07
URLEncoding for subaddressing

florentos17:
this is the very beginning of the implementation of URLEncoding for subaddressing.
right now, i can't manage to write the integration test correctly: even using the probe instead of IMAP commands, the creation of a mailbox with a name containing a space is rejected: `java.lang.RuntimeException: AAAC BAD CREATE failed. Illegal encoding.`...

side note: i was unsure of what to do with the checkstyle issue in TikaTextExtractorTest, i committed it separately.

### DIFF
--- a/server/mailet/integration-testing/src/test/java/org/apache/james/mailets/SubAddressingTest.java
+++ b/server/mailet/integration-testing/src/test/java/org/apache/james/mailets/SubAddressingTest.java
@@ -222,17 +222,15 @@ class SubAddressingTest {
                         .asAddition());
 
         sendSubAddressedMail(TARGETED_MAILBOX_ENCODED);
-        awaitAtMostOneMinute.until(() -> jamesServer.getProbe(MailboxProbeImpl.class)
-                .searchMessage(MultimailboxesSearchQuery.from(SearchQuery.builder().build()).build(), RECIPIENT, 1)
-                .size() == 1);
 
+        int LOAD_LIMIT = 1;
         awaitAtMostOneMinute.until(() -> jamesServer.getProbe(MailboxProbeImpl.class)
                 .searchMessage(MultimailboxesSearchQuery
                         .from(SearchQuery.builder().build())
                         .inMailboxes(targetMailboxId)
                         .build(),
                     RECIPIENT,
-                    1)
+                    LOAD_LIMIT)
                 .size() == 1);
     }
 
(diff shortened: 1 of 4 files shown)


quantranhong1999:
> side note: i was unsure of what to do with the checkstyle issue in TikaTextExtractorTest, i committed it separately.

It was my bad. Thank you for spotting it. I put a fix cf https://github.com/apache/james-project/pull/2440.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/SubAddressingTest.java
```diff
@@ -196,6 +203,28 @@ void subAddressedEmailShouldBeDeliveredInINBOXWhenSpecifiedFolderExistsWithCorre
         awaitSubAddressedMail(MailboxConstants.INBOX);
     }
 
+    @Disabled("can't manage to successfully create the mailbox...")
+    @Test
+    void subAddressedEmailShouldBeDeliveredInSpecifiedFolderWhenRequiringEncoding(@TempDir File temporaryFolder) throws Exception {
+        setup(temporaryFolder);
+
```
chibenwa, line 210:
```suggestion
jamesServer.probe(MailboxProbeImpl.class).createMailbox(MailboxPath.forUser(RECIPIENT, TARGETED_MAILBOX_REQUIRING_ENCODING));
```

chibenwa: (let's cheat for that one, utf -7 is tricky)

florentos17: right, thank you !

florentos17:
note on the integration test:
- to create the mailbox i used Benoit's suggestion
- to grant the `p` right i created a new `ACLProbeImpl.executeCommand` to use the `MailboxId` instead of the `MailboxPath`. but then i realized the old `executeCommand` was not used anymore so i removed it.
- to await the message, `testIMAPClient` did not work with the name requiring encoding, so i used `MailboxProbeImpl.searchMessage` but i realize now that i have to refine the searchQuery, i will work on this.

chibenwa:
> to await the message, testIMAPClient did not work with the name requiring encoding, so i used MailboxProbeImpl.searchMessage, is that ok ?

Let's say it's ok

However if we are curious regarding URL encoding we can try to run a IMAP LIST command.

Arsnael (approved):

PR #2444, opened 2024-10-09
James 1409 - Change JPARecipientRewriteTable to store separate record per target address

amichair:
Changed the implementation to use a record per row design instead of a single row with concatenated addresses, as suggested in the bug report.

Along the way I also found that listSources was broken with such a multi-address value in the old design, and the fix is built-in in the new design. Added a regression test for it in the second commit.

The one thing that's missing is that I don't know if there is any procedure/mechanism/command for performing data/schema updates when an existing system is upgraded to a new version, is there such a thing?


chibenwa (commented):
I am overall very happy with the fix proposed to this long-standing bug. THANKS!

To make it perfect it would be wonderful to writ a little entru into `upgrade-instructions.md` telling people how to update.  Minimum would be the `ALTER TABLE` to be performed. Ideally we would get the old data migrated as well.

Can we detail the migration strategy?

amichair:
The changes are:
1. Primary key changed from (DOMAIN_NAME, USER_NAME) to (DOMAIN_NAME, TARGET_ADDRESS, USER_NAME) (unfortunately, it is auto-generated in alphabetical order)
2. Added index USER_NAME_DOMAIN_NAME_TARGET_ADDRESS_INDEX (USER_NAME, DOMAIN_NAME, TARGET_ADDRESS)
3. Added index TARGET_ADDRESS_INDEX (TARGET_ADDRESS)
4. Columns have not changed.
5. Rows that had multiple semicolon delimited values in their TARGET_ADDRESS field need to be split into individual rows each with a single target address value. Rows with a single value remain unchanged.

Afaik there is no standard SQL way of doing the last data migration step. I can write a method that does it, but is there any mechanism to invoke it? or just put it in a main method of the class? How have schema/data migrations been addressed in the past?

amichair:
I added a draft of the migration instructions for now.

upgrade-instructions.md
```diff
@@ -34,6 +34,36 @@ Change list:
...
+This solves both the limitation on the number of mappings (imposed by the column maximum length), and the broken query
+by target address.
+
+For JPA users, the database schema must be changed as following:
+
+- Change Primary key from (DOMAIN_NAME, USER_NAME) to (DOMAIN_NAME, TARGET_ADDRESS, USER_NAME) (unfortunately,
+  JPA auto-generates it in alphabetical order).
+- Add index USER_NAME_DOMAIN_NAME_TARGET_ADDRESS_INDEX (USER_NAME, DOMAIN_NAME, TARGET_ADDRESS).
+- Add index TARGET_ADDRESS_INDEX (TARGET_ADDRESS).
+- The columns themselves have not changed.
```
chibenwa, line 56: Are the commands database specific or can we provide examples?

upgrade-instructions.md
```diff
@@ -34,6 +34,36 @@ Change list:
...
+
+In addition, the existing data must be migrated to the new structure:
+
+- Rows with a single address in the TARGET_ADDRESS column are unchanged. 
+- Rows that have multiple semicolon-delimited addresses in their TARGET_ADDRESS column need to be split into 
+  individual rows, each with a single target address value.
+  Unfortunately, there is no standard way to do this in SQL. In some databases it may be possible to build
+  a recursive query that can do this. Otherwise, it can be done manually using database management tools,
+  or via a short external script, or via exporting the data, manual editing it, and re-importing the migrated
+  data.
```
chibenwa, line 66: Or we could:
 - Ask the user to list all mappings through the CLI prior to the upgrade
 - Change the table name: a new table will be used, empty
 - Provide a new CLI command to import the mappings from the previously saved input
That way user do not looses data.
WDYT?

Arsnael (commented):
Hi @amichair , thank you for the fix, it's appreciated!

Read your work, nothing to add.

amichair:
I liked your idea, and it did take some work... I went for a webadmin export/import (seems more future-proof and intuitive to move jsons around instead of custom cli format parsing). Had to do some webadmin rearranging since mappings serialization wasn't fully implemented. Note that it's currently implemented as 'addMappings' rather than 'setMapping' (i.e. doesn't erase all existing mappings), I figured it's safer and more useful, though for the purposes of this migration it doesn't matter (the table is empty). Updated the instructions too. Didn't update the REST api docs with the new route, dunno when/how to do it, or maybe it can remain undocumented other than this migration ;-)

I added a few local tests, but didn't yet test the whole procedure for real - I'm assuming the table will get recreated when dropped but haven't tried it yet.

server/data/data-library/src/test/java/org/apache/james/rrt/lib/RecipientRewriteTableContract.java
```diff
@@ -411,6 +411,19 @@ default void listSourcesShouldReturnWhenMultipleSourceMapping() throws Exception
         assertThat(virtualUserTable().listSources(mapping)).contains(source, source2);
     }
 
+    @Test
+    default void listSourcesShouldReturnWhenSourceHasMultipleMappings() throws Exception {
```
chibenwa, line 415: Let's do also something for `getStoredMappings` ?

chibenwa:
> I went for a webadmin export/import (seems more future-proof and intuitive to move jsons around instead of custom cli format parsing). 

+1

server/data/data-library/src/test/java/org/apache/james/rrt/lib/RecipientRewriteTableContract.java
amichair: I'm not sure what you mean. This was a bug I stumbled upon in the previous code while working on this, so I added a test just to make sure it no longer occurs in the new design. Everything else I assume should be covered by existing tests...

chibenwa: With the export import procedure via webadmin it may not be relevant.

Let's keep as is.

chibenwa (approved):

Arsnael (approved):

amichair:
Thanks for merging this!

I do wonder however, why all the nicely organized commits are squashed into one when merging? This makes it much harder for anyone else now or in the future to understand or review the code, with the various changes all bunched together...

Arsnael:
> Thanks for merging this!

With pleasure !

> I do wonder however, why all the nicely organized commits are squashed into one when merging? This makes it much harder for anyone else now or in the future to understand or review the code, with the various changes all bunched together...

Sorry maybe went a bit fast on this. However all commits should start with the JIRA ticket number, however only the first one was. I thought maybe little changes making it ok was ok to squash. Obviously I was wrong sorry for this.

PR #2445, opened 2024-10-09
JAMES-2182 Sharing for IMAP

chibenwa:
My second shot ever at this.

First a BIG refactoring centralizes the definition of path <-> name convertion into the IMAP layer (pre-requisite) and 
provide a generic strategy that could be allowed to be overloaded (PathConverter + Namespace supplier).

A second part of this work is about exposing other users folder shared with the user into a `#user` folder.

Remaining work:
 - Polishing the few TODOs 
 - More test with sharee interacting with delgated mailboxes in IMAP
 - Test it with real life IMAP client to ensure this works as intended.
 
 I am happy, in 1 day coding it got quite far, though details needs to be addressed and more tests needs to be produced.


chibenwa:
Edit: evening coding session, most spotted bugs and a few others are out of the way but extensive testing will be required.

Arsnael (commented):
Quite a big one to read, but interesting. Huge work here, cool!

Just a minor remark but overall LGTM

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreMessageManager.java
```diff
@@ -675,12 +679,43 @@ private void trimFlags(Flags flags, MailboxSession session) {
...
 
-    @Override
-    public Map<MessageUid, Flags> setFlags(final Flags flags, final FlagsUpdateMode flagsUpdateMode, final MessageRange set, MailboxSession mailboxSession) throws MailboxException {
+    public Optional<MailboxException> ensureFlagsWrite(Flags flags, FlagsUpdateMode flagsUpdateMode, MailboxSession mailboxSession) {
+        MailboxACL.Rfc4314Rights myRights = storeRightManager.myRights(mailbox, mailboxSession);
 
-        if (!isWriteable(mailboxSession)) {
-            throw new ReadOnlyException(getMailboxPath());
+        if (flagsUpdateMode.equals(FlagsUpdateMode.REPLACE)) {
+            if (!myRights.contains(MailboxACL.Right.Write, MailboxACL.Right.WriteSeenFlag, MailboxACL.Right.DeleteMessages)) {
```
HoussemNasri, line 686: Maybe containsAll? I had to go into the contain method to verify that all input rights have to be contained.

protocols/imap/src/main/java/org/apache/james/imap/main/PathConverter.java
```diff
@@ -21,80 +21,230 @@
...
+                    .build();
+            }
+
+            MailboxPath basePath = computeBasePath(session, finalReferencename);
+            if (basePath.getNamespace().equals(MailboxConstants.USER_NAMESPACE)
+                && basePath.getUser() == null) {
+
+                int separatorPosition = decodedMailboxName.indexOf(mailboxSession.getPathDelimiter());
+                if (separatorPosition >= 0) {
+                    // interpret first part as the user
```
HoussemNasri, line 194: If we use `.split()` we won't need that comment.

protocols/imap/src/main/java/org/apache/james/imap/processor/NamespaceSupplier.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.james.imap.message.response.NamespaceResponse.Namespace;
+import org.apache.james.mailbox.MailboxSession;
+
+import com.google.common.collect.ImmutableList;
+
+public interface NamespaceSupplier {
+    class Default implements NamespaceSupplier {
+        @Override
+        public Collection<Namespace> personalNamespaces(MailboxSession session) {
+            return ImmutableList.of(new NamespaceResponse.Namespace("", session.getPathDelimiter()));
```
HoussemNasri, line 34: I know it's a micro-optimization but since we are returning an immutable list can't we cache the list? Maybe using `Suppliers.memoize()`?

protocols/imap/src/main/java/org/apache/james/imap/main/PathConverter.java
chibenwa: :-1: split underneath uses a regex which obfuscate code, expose to complications like REDOS, and behaves pourly in what is here a critical path.

protocols/imap/src/main/java/org/apache/james/imap/processor/NamespaceSupplier.java
chibenwa: It's micro optimisation on what is not a critical path, sorry I won't bother: I trust the JIT to do the job!

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreMessageManager.java
chibenwa: I think it's rather explicit, and it is unrelated with this change set.

Maybe others have opinions?

Arsnael: Looks rather clear to me, no need to change IMO

Arsnael (approved):

chibenwa:
Any more reviews for this or I can proceed with the merge?

quantranhong1999 (approved):

PR #2446, opened 2024-10-10
JAMES-4032 DKIMHook should also validate local from header

Arsnael:


server/protocols/protocols-smtp-dkim/src/main/java/org/apache/james/smtpserver/DKIMHook.java
```diff
@@ -165,7 +205,9 @@ public Config(boolean forceCRLF, boolean signatureRequired, Optional<Domain> onl
 
         DKIMCheckNeeded dkimCheckNeeded() {
             return onlyForSenderDomain
-                .map(DKIMCheckNeeded::onlyForSenderDomain)
+                .map(domain -> DKIMCheckNeeded.or(
+                    DKIMCheckNeeded.onlyForSenderDomain(domain),
+                    DKIMCheckNeeded.onlyForHeaderFromDomain(domain)))
```
chibenwa, line 210: What would be perfect would be to have it configurable.

`<validatedEntities>envelope,headers</validatedEntities>`

Defaults to both if unspecified, coma separated values.

This would enable greater flexibility when configuring this and would allow for easy rollback to previous state.

WDYT?

chibenwa (commented):
Good quality work!

server/protocols/protocols-smtp-dkim/src/main/java/org/apache/james/smtpserver/DKIMHook.java
Arsnael: Doable yes, more flexibility is good. Will give it a shot, thanks for the suggestion!

quantranhong1999 (approved):

server/protocols/protocols-smtp-dkim/src/main/java/org/apache/james/smtpserver/DKIMHook.java
Arsnael: Complexifies a bit the code but ok I think.

However I ain't sure how to test it in DKIMHookIntegrationTest. Would need a different smtpserver.xml file? Thought maybe doing like in CertificateReloadTest with the temporary server config reload... but makes the test structure more complex, maybe there is a better way? But maybe not. Wasn't sure ! But I might go with it later

docs/modules/servers/partials/configure/smtp-hooks.adoc
```diff
@@ -371,6 +372,7 @@ Example handlerchain configuration for `smtpserver.xml`:
 	<handler class="org.apache.james.smtpserver.DKIMHook">
 		<forceCLRF>true</forceCLRF>
 		<onlyForSenderDomain>apache.org</onlyForSenderDomain>
+        <validatedEntities>envelope,headers</validatedEntities>
```
chibenwa, line 375:
```suggestion
                <validatedEntities>envelope,headers</validatedEntities>
```

chibenwa (approved):

docs/modules/servers/partials/configure/smtp-hooks.adoc
Arsnael: Why the tab? It's not inside an xml tag like `onlyForSenderDomain`?

Arsnael:
Just been fixing a checkstyle

docs/modules/servers/partials/configure/smtp-hooks.adoc
quantranhong1999: Should there be a TAB inside?

quantranhong1999 (approved):

docs/modules/servers/partials/configure/smtp-hooks.adoc
Arsnael: Or do you mean:

```
	<onlyForSenderDomain>
                 <domain>apache.org</domain>
                 <validatedEntities>envelope,headers</validatedEntities>
        </onlyForSenderDomain>
```

? would require few more changes in the structure then

quantranhong1999: I am not sure about Benoit idea. But I suggest in the Antora doc that `validatedEntities` have the same indent level as `onlyForSenderDomain`, as the code currently is, no?

Arsnael: It's already on the same level? at least in code and from the format i see on github?

quantranhong1999: From what I see:

image

Arsnael: Ok seems maybe there was but the ide was showing me the same... tried to repush it. is it fixed on your end?

quantranhong1999: It is good now.

server/protocols/protocols-smtp-dkim/src/main/java/org/apache/james/smtpserver/DKIMHook.java
```diff
@@ -92,6 +104,35 @@ static DKIMCheckNeeded onlyForSenderDomain(Domain domain) {
...
+                try {
+                    return StreamUtils.ofNullable(mail.getMessage().getFrom())
+                        .distinct()
+                        .flatMap(DKIMCheckNeeded::parseMailAddress)
+                        .findFirst()
+                        .map(MailAddress::getDomain)
+                        .map(domain::equals)
+                        .orElse(false);
+                } catch (MessagingException me) {
+                    LOGGER.info("Unable to parse the \"FROM\" header; ignoring.");
```
chibenwa, line 118:
```suggestion
                    LOGGER.info("Unable to parse the \"FROM\" header {}; ignoring.", message.getHeader("From"));
```

chibenwa (approved):

server/protocols/protocols-smtp-dkim/src/main/java/org/apache/james/smtpserver/DKIMHook.java
chibenwa: Yep but getting the actual failing value is worthwhile: being the one reading prod logs I can tell you I want this. Note also I am not asking for parsed form but row form which is unlikely to fail...

quantranhong1999 (approved):

chibenwa (approved):

Arsnael:
Just squashing fixup

PR #2452, opened 2024-10-15
JAMES-2433 Implement EventStore for a JPA backend

amichair:
This is just a rough draft and not ready to merge yet, but since I have no previous experience with Scala or Reactive Streams or Guice or this event api or its usages, if anyone could help in improving and finalizing this, I think we can get this done :-)

### DIFF
--- a/event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
+++ b/event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
@@ -1,0 +1,118 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+package org.apache.james.eventsourcing.eventstore.jpa;
+
+
+import static org.apache.james.eventsourcing.eventstore.jpa.model.JPAEvent.DELETE_AGGREGATE_QUERY;
+import static org.apache.james.eventsourcing.eventstore.jpa.model.JPAEvent.SELECT_AGGREGATE_QUERY;
+
+import jakarta.inject.Inject;
+import jakarta.persistence.EntityManagerFactory;
+import jakarta.persistence.PersistenceUnit;
+
+import org.apache.james.backends.jpa.TransactionRunner;
+import org.apache.james.eventsourcing.AggregateId;
+import org.apache.james.eventsourcing.Event;
+import org.apache.james.eventsourcing.eventstore.EventStore;
+import org.apache.james.eventsourcing.eventstore.EventStoreFailedException;
+import org.apache.james.eventsourcing.eventstore.History;
+import org.apache.james.eventsourcing.eventstore.JsonEventSerializer;
+import org.apache.james.eventsourcing.eventstore.jpa.model.JPAEvent;
+import org.reactivestreams.Publisher;
+
+import com.github.fge.lambdas.Throwing;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+
+import reactor.core.publisher.Mono;
+import scala.collection.Iterable;
+import scala.collection.JavaConverters;
+
+@PersistenceUnit(unitName = "James")
+public class JPAEventStore implements EventStore {
+
+    /**
+     * The entity manager to access the database.
+     */
+    private EntityManagerFactory entityManagerFactory;
+
+    /**
+     * The JSON serializer to serialize the event data.
+     */
+    private JsonEventSerializer jsonEventSerializer;
+
+    /**
+     * Constructs a JPAEventStore.
+     */
+    @Inject
+    public JPAEventStore(EntityManagerFactory entityManagerFactory, JsonEventSerializer jsonEventSerializer) {
+        this.jsonEventSerializer = jsonEventSerializer;
+        this.entityManagerFactory = entityManagerFactory;
+    }
+
+    @Override
+    public Publisher<Void> appendAll(Iterable<Event> events) {
+        if (events.isEmpty()) {
+            return Mono.empty();
+        }
+        Preconditions.checkArgument(Event.belongsToSameAggregate(events));
+        AggregateId aggregateId = events.head().getAggregateId();
+        return Mono.fromRunnable(() -> new TransactionRunner(entityManagerFactory).runAndHandleException(
+            entityManager ->
+                JavaConverters.asJava(events).forEach(Throwing.consumer(e -> {
+                    JPAEvent jpaEvent = new JPAEvent(aggregateId, e.eventId(), jsonEventSerializer.serialize(e));
+                    entityManager.persist(jpaEvent);
+            })),
+            exception -> {
+                EventStoreFailedException esfe = new EventStoreFailedException("Unable to add events");
+                esfe.initCause(exception);
+                throw esfe;
+            }));
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Publisher<History> getEventsOfAggregate(AggregateId aggregateId) {
+        Preconditions.checkNotNull(aggregateId);
+        return Mono.fromSupplier(() -> new TransactionRunner(entityManagerFactory).runAndRetrieveResult(
+            entityManager -> History.of(
+                (Event[]) entityManager.createNamedQuery(SELECT_AGGREGATE_QUERY)
+                .setParameter("aggregateId", aggregateId.asAggregateKey())
+                .getResultStream()
+                .map(Throwing.function(e -> jsonEventSerializer.deserialize(((JPAEvent) e).getEvent())))
+                .toArray(Event[]::new))));
+    }
+
+    @Override
+    public Publisher<Void> remove(AggregateId aggregateId) {
+        return Mono.fromSupplier(() -> new TransactionRunner(entityManagerFactory).runAndRetrieveResult(
+            entityManager -> {
+                entityManager.createNamedQuery(DELETE_AGGREGATE_QUERY)
+                    .setParameter("aggregateId", aggregateId.asAggregateKey())
+                    .executeUpdate();
+                return null;
+            }));
+    }
+
+    @VisibleForTesting
+    protected void removeAll() {
+        new TransactionRunner(entityManagerFactory).runAndRetrieveResult(
+            entityManager -> entityManager.createQuery("DELETE FROM JPAEvent").executeUpdate());
+    }
+}
--- a/event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/model/JPAEvent.java
+++ b/event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/model/JPAEvent.java
@@ -1,0 +1,124 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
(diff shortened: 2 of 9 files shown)


event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
```diff
@@ -0,0 +1,176 @@
...
+                return e;
+            });
+            transaction.commit();
+        } catch (PersistenceException e) {
+            if (transaction.isActive()) {
+                transaction.rollback();
+            }
+            EventStoreFailedException esfe = new EventStoreFailedException("Unable to add events");
+            esfe.initCause(e);
+            throw esfe;
```
chibenwa, line 109:
```suggestion
            throw  new EventStoreFailedException("Unable to add events", e);
```

Mutability can be avoided

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
```diff
@@ -0,0 +1,176 @@
...
+
+    @Override
+    public Publisher<Void> appendAll(Iterable<Event> events) {
+        if (events.isEmpty()) {
+            return Mono.empty();
+        }
+        Preconditions.checkArgument(Event.belongsToSameAggregate(events));
+        AggregateId aggregateId = events.head().getAggregateId();
+        EntityManager entityManager = entityManagerFactory.createEntityManager();
+        final EntityTransaction transaction = entityManager.getTransaction();
```
chibenwa, line 90: We can handle this with TransactionRunner I believe

```
return Mono.fromRunnable(() -> TransactionRunner.runAndHandleException(entityManager -> {
             events.foreach(Throwing.runnable(e -> {
                    JPAEvent jpaEvent = new JPAEvent(aggregateId, e.eventId(), jsonEventSerializer.serialize(e));
                    entityManager.persist(jpaEvent);
            }));
}, e -> {
   throw  new EventStoreFailedException("Unable to add events", e);
});
```

Note that:
 - `e` variable name is ambiguous, we shall favor `event`
 - Wrap the whole transformation into a mono not just the final transformation

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
```diff
@@ -0,0 +1,176 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public Publisher<History> getEventsOfAggregate(AggregateId aggregateId) {
+        Preconditions.checkNotNull(aggregateId);
+        EntityManager entityManager = entityManagerFactory.createEntityManager();
+        try {
+            List<Event> events = (List<Event>) entityManager.createNamedQuery(SELECT_AGGREGATE_QUERY)
+                .setParameter("aggregateId", aggregateId.asAggregateKey())
+                .getResultStream()
+                .map(e -> {
```
chibenwa, line 125: Use THrowing.function ?

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
```diff
@@ -0,0 +1,176 @@
...
+            throw new RuntimeException("Unable to get events", e);
+        } finally {
+            EntityManagerUtils.safelyClose(entityManager);
+        }
+    }
+
+    @Override
+    public Publisher<Void> remove(AggregateId aggregateId) {
+        EntityManager entityManager = entityManagerFactory.createEntityManager();
+        final EntityTransaction transaction = entityManager.getTransaction();
```
chibenwa, line 144: Use transaction runner here too and wrap the whole operation onto a mono...

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
```diff
@@ -0,0 +1,176 @@
...
+                transaction.rollback();
+            }
+            throw new RuntimeException("Unable to remove events", e);
+        } finally {
+            EntityManagerUtils.safelyClose(entityManager);
+        }
+        return Mono.empty();
+    }
+
+    // for use in tests
```
chibenwa, line 162: Use @VisibleForTesting annotation them

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/model/JPAEvent.java
```diff
@@ -0,0 +1,122 @@
...
+            if (this == obj) {
+                return true;
+            }
+            if (obj == null || getClass() != obj.getClass()) {
+                return false;
+            }
+            final JPAEventId other = (JPAEventId) obj;
+            return Objects.equal(this.aggregateId, other.aggregateId)
+                && Objects.equal(this.eventId, other.eventId);
+        }
```
chibenwa, line 85: We shall test equals and hash code with equals verifier.

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/model/JPAEvent.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Id
+    @Column(name = "AGGREGATE_ID", nullable = false, length = 100)
+    private String aggregateId = "";
+
+    @Id
+    @Column(name = "EVENT_ID", nullable = false)
+    private int eventId;
+
+    @Column(name = "EVENT", nullable = false, length = 16 * 1024)
```
chibenwa, line 96: This may hurt: are we sure this is enough to fit all possible serialized events?

For instance RuleSetDefined may be larger than this for people holding many filters...

Other implementation are unbounded on the size of this field, likely so should do the JPA one.

I suggest looking into using an unbounded `byte[]` here if need be..

chibenwa (commented):
Thanks, this is a cool first shot at it!

I did do a few comments to polish the code here.

Next steps would be to wire the event store into existing app and expose features relying on event sourcing onto the same applications.

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
amichair: That's what I intended to do, but this exception doesn't support a cause parameter... I left the scala api part alone for now.

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
amichair: The TransactionWorker api isn't as shown above, but was easy enough to adjust. However the Throwing part I gave a few tries with several variations and types but couldn't get to work yet. If you have a working snippet that could save me some time trying to mess with the java/scala type system incompatibilities (Function1 is not a Function etc)...

btw I think it would have been nice if at least the events source api part remained in java (even if other module implementations were in scala) rather than making the interface itself use scala objects and have the scala types leak into the java code... scala was designed to try an play nice with java, but not necessarily the other way around. but perhaps that's just my own preference or lack of experience stitching them together :-)

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
amichair: ok

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/model/JPAEvent.java
amichair: Not sure what you mean here.

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/model/JPAEvent.java
amichair: I have no idea what this is used for and how big the events are, usually events and jsons (used in the other modules as well) are small-ish creatures :-)

I did start with 1mb and it errored, so just picked an arbitrary value to get through the prototype. I suppose if it needs to be a blob or whatnot we can do that. Maybe add to the contract (test) whatever size you think the implementations must support to make sure they're all compatible with the requirements?

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/model/JPAEvent.java
Maxxx873: > Not sure what you mean here.

Hello, you could use https://jqno.nl/equalsverifier/ for test

amichair: done

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * The JSON serializer to serialize the event data.
+     */
+    private JsonEventSerializer jsonEventSerializer;
+
+    /**
+     * Set the serializer to use.
+     */
+    @Inject
+    public void setJsonEventSerializer(JsonEventSerializer jsonEventSerializer) {
```
chibenwa, line 70: Can't we inject via the constructor?

chibenwa (approved):
Overall ok to me!

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
amichair: For now I got it to work with JavaConverters.asJava(events), a bit ugly but I don't know how to get the scala objects to play nice with Throwing. If you have a better way let me know.

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
amichair: done

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
amichair: done

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
amichair: ok, I moved both injections to a constructor, though other classes are inconsistent in constructor/setter injection point and I'm not sure if it matters... if you have a preference perhaps you should standardize the injection location throughout the code.

amichair:
> Next steps would be to wire the event store into existing app and expose features relying on event sourcing onto the same applications.

I don't know how/where/what is needed to finish integrating this, or if you intend on doing that separately after merging this PR?

amichair:
Anything else to be done here? (ci failure seems unrelated?)

quantranhong1999:
This should not be a WIP anymore, no?

chibenwa:
I turned the status from *Draft* into *Ready for review*

event-sourcing/event-store-jpa/src/main/java/org/apache/james/eventsourcing/eventstore/jpa/JPAEventStore.java
```diff
@@ -0,0 +1,118 @@
...
+import com.github.fge.lambdas.Throwing;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
+
+import reactor.core.publisher.Mono;
+import scala.collection.Iterable;
+import scala.collection.JavaConverters;
+
+@PersistenceUnit(unitName = "James")
+public class JPAEventStore implements EventStore {
```
quantranhong1999, line 48: No Guice module binding in this PR? e.g. `JPADropListsModule`.

quantranhong1999 (approved):

chibenwa:
> No Guice module binding in this PR? e.g. JPADropListsModule.

Guice modules will come with actual usages of this class

Arsnael (approved):

vttranlina:
The CI failed, but I cannot view the Jenkins results.
Screenshot_20241129_123137

Maxxx873:
I think it's a pretty good job overall, if there is an opportunity it would be nice to restart CI to eventually complete this step

Arsnael:
Merged this, sorry @amichair for missing this and thanks @Maxxx873 for the reminder!

PR #2460, opened 2024-10-20
JAMES-3823 SMTP Require TLS Option

Maxxx873:
[JAMES-3823](https://issues.apache.org/jira/projects/JAMES/issues/JAMES-3823?filter=allopenissues)

Hi! The first step in the direction of SMTP Require TLS Option is a basic set of hooks. In the future I plan to enhance this PR by processing of new REQUIRETLS attribute, handling headers, remote delivery, etc.


src/site/resources/server/rfclist/smtp/rfc8689.txt
```diff
@@ -0,0 +1,761 @@
+﻿
```
chibenwa, line 1: IMO we do not need to import the spec here, I +1 a simple link to the spec.

chibenwa (commented):
Hello,

Thanks a LOT for contributions on the topic!

A first remark is that this pull request only tackles incoming SMTP while it necessarily have impact on outgoing SMTP settings: we need to have RemoteDelivery setting up the REQUIRETLS mail parameter if needed... (section 4)

We also need to tackle section 5: preserving those parameters for `Bounce` and `DSNBounce` mailets.

Maxxx873:
Hi! I'm still working. In this step, I added the requireStartTls() method for SmtpConfiguration in integration testing and implemented RemoteDelivery.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/remote/delivery/MailDelivrerToHost.java
```diff
@@ -123,17 +130,23 @@ public ExecutionResult tryDeliveryToHost(Mail mail, Collection<InternetAddress>
         try {
             transport = (SMTPTransport) session.getTransport(outgoingMailServer);
             transport.setLocalHost(props.getProperty(inContext(session, "mail.smtp.localhost"), configuration.getHeloNameProvider().getHeloName()));
+            if (isTlsRequired(mail)) {
+                transport.setStartTLS(true);
```
chibenwa, line 134: 2 remarks:
 - If the transport layer is already configured with SSL there's no point setting SSL. That's the case EG relaying to a secure mail gateway other the internet.
 - If the receiver server do not support startTLS this need to throw.

chibenwa (commented):
IMO we need a way to test this:

GIVEN From REQUIRETLS
WHEN remoteDelivery speaks to a server that do NOT allow startTLS
THEN it fails

Don't we have a way to tweak MockSMTP into allowing testing this.

And yes and VERY good work, by the way.

Maxxx873:
> IMO we need a way to test this:
> 
> GIVEN From REQUIRETLS WHEN remoteDelivery speaks to a server that do NOT allow startTLS THEN it fails
> 
> Don't we have a way to tweak MockSMTP into allowing testing this.
> 
> And yes and VERY good work, by the way.

Thanks so much for the review and remarks! Unfortunately, I don't know how to implement it correctly yet. I tried setting the MockSMTP behavior, but that's not what's needed as far as I understand. It is possible to make a tweak to MockSMTP to respond not only to commands, but also to parameters passed. Is it necessary? Further, it is not very clear how and at what point I can check the startTLS allowance for remoteDelivery.  Also a question regarding Bounce and DSNBounce maillets. The REQUIRETLS email attribute will be saved as it is, what do I need to add additionally? I apologize for the large scope of questions.

server/mailet/remote-delivery-integration-testing/src/test/java/org/apache/james/smtp/tls/SmtpRequireTlsRelayTest.java
```diff
@@ -244,6 +240,33 @@ void remoteDeliveryWhenShouldRequireTlsAndDsnAndMtPriorityTogether(DockerMockSmt
...
+        smtpClient.sendShortMessageData("A short message...");
+
+        String dsnMessage = testIMAPClient.connect(LOCALHOST_IP, jamesServer.getProbe(ImapGuiceProbe.class).getImapPort())
+                .login(FROM, PASSWORD)
+                .select(TestIMAPClient.INBOX)
+                .awaitMessageCount(awaitAtMostOneMinute, 1)
+                .readFirstMessage();
+
+        Assertions.assertThat(dsnMessage).contains("504 The server has not implemented a command parameter");
+    }
```
chibenwa, line 268: No what we want is the mock server to HIDE that it implements STARTTLS (in the EHLO capability).

When that's the case the relay of the mail should fail.

This is different from failing at MAIL FROM

(What I mean is that the test is inaccurate).

Maxxx873: > No what we want is the mock server to HIDE that it implements STARTTLS (in the EHLO capability).
> 
> When that's the case the relay of the mail should fail.
> 
> This is different from failing at MAIL FROM
> 
> (What I mean is that the test is inaccurate).

I agree with you, and I propose a new version of the test

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/remote/delivery/MailDelivrerToHost.java
```diff
@@ -131,6 +133,9 @@ public ExecutionResult tryDeliveryToHost(Mail mail, Collection<InternetAddress>
             transport = (SMTPTransport) session.getTransport(outgoingMailServer);
             transport.setLocalHost(props.getProperty(inContext(session, "mail.smtp.localhost"), configuration.getHeloNameProvider().getHeloName()));
             connect(outgoingMailServer, transport);
+            if (!transport.getLastServerResponse().contains("STARTTLS")) {
```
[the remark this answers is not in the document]
chibenwa: Wasn;t reviwing latest diff, sorry

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/tls/SmtpTlsMessageHook.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.james.protocols.smtp.hook.HookReturnCode;
+import org.apache.james.smtpserver.JamesMessageHook;
+import org.apache.mailet.Attribute;
+import org.apache.mailet.AttributeName;
+import org.apache.mailet.AttributeValue;
+import org.apache.mailet.Mail;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SmtpTlsMessageHook implements JamesMessageHook {
```
chibenwa, line 41:
```suggestion
public class SmtpRequireTlsMessageHook implements JamesMessageHook {
```

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/tls/SmtpTlsEhloHook.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.apache.james.smtpserver.tls;
+
+import java.util.Set;
+
+import org.apache.james.protocols.smtp.SMTPSession;
+import org.apache.james.protocols.smtp.hook.HeloHook;
+import org.apache.james.protocols.smtp.hook.HookResult;
+
+public class SmtpTlsEhloHook implements HeloHook {
```
chibenwa, line 28:
```suggestion
public class SmtpRequireTlsEhloHook implements HeloHook {
```

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/tls/SmtpTlsParameterHook.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.james.protocols.api.ProtocolSession;
+import org.apache.james.protocols.smtp.SMTPRetCode;
+import org.apache.james.protocols.smtp.SMTPSession;
+import org.apache.james.protocols.smtp.hook.HookResult;
+import org.apache.james.protocols.smtp.hook.HookReturnCode;
+import org.apache.james.protocols.smtp.hook.MailParametersHook;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class SmtpTlsParameterHook implements MailParametersHook {
```
chibenwa, line 36:
```suggestion
public class SmtpRequireTlsParameterHook implements MailParametersHook {
```

Maxxx873:
> Very good work.
> 
> I endorse this, and just find comments onto the naming of the hooks.

Thank you! If there are any additional suggestions for improving this contribution, I am willing to follow up. It would be interesting if this could be tested on a test bench somehow. Thanks again for the code review and feedback on the work

chibenwa (approved):

chibenwa:
image

Test failures seems related.

Could you have a look?

Maxxx873:
> image
> 
> Test failures seems related.
> 
> Could you have a look?

I assume firstly  that in order for CI to work properly when running tests with MAIL FROM parameters with no value, we need to update the version of the mock smtp server in the docker hub. See:
https://github.com/apache/james-project/pull/2489

chibenwa:
Will do tomorrow!

(sorry)

chibenwa:
`linagora/mock-smtp-server:0.7` is pushed, as promissed.

I did also re-trigger a build.

Arsnael (approved):
The error looks unrelated on the CI so I restarted the build!

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2460/16/testReport/

@Maxxx873 second time it breaks on those, I think it's potentially related. Can you have a look please?

Maxxx873:
> https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2460/16/testReport/
> 
> @Maxxx873 second time it breaks on those, I think it's potentially related. Can you have a look please?

Yay, finally ci looks green )

Arsnael (approved):

chibenwa (approved):

PR #2465, opened 2024-10-22
JAMES-4080 Refactor EventDeadLetters redeliver code to work with multiple serializers

hungphan227:


event-bus/api/src/main/java/org/apache/james/events/EventDeadLetters.java
```diff
@@ -93,6 +94,10 @@ public String asString() {
 
     Mono<Event> failedEvent(Group registeredGroup, InsertionId failDeliveredInsertionId);
 
+    default Mono<Event> failedEvent(Group registeredGroup, InsertionId failDeliveredInsertionId, Map<Group, EventSerializer> mapGroupToEventSerializer) {
```
chibenwa, line 97: No we shall inject the event busses into event dead letter IMP we do not need this extra interface method.

server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/service/EventDeadLettersRedeliverService.java
```diff
@@ -109,4 +113,11 @@ private Optional<EventBus> findEventBus(Group group) {
             .filter(eventBus -> eventBus.listRegisteredGroups().contains(group))
             .findFirst();
     }
+
+    private Map<Group, EventSerializer> getMapGroupToSerializer() {
+        return eventBuses.stream().flatMap(eventBus -> eventBus.listRegisteredGroups()
+                .stream()
+                .map(group -> Pair.of(group, eventBus.getEventSerializer())))
+            .collect(ImmutableMap.toImmutableMap(Pair::getLeft, Pair::getRight, (serializer1, serializer2) -> serializer1));
+    }
```
chibenwa, line 122: Move this into DeadLetter ?

event-bus/api/src/main/java/org/apache/james/events/EventDeadLetters.java
hungphan227: Circular dependency injection error occurred when I tried to inject event busses into event dead letter

chibenwa: Bad answer.

Then we 'just need one more level of indirection...

How about a `EventBusses` interface that would abstract this away tfor us?

hungphan227: Could you give me some more detail? I cannot figure out the way to fix the circular dependency.

vttranlina:
I propose creating a new `EventSerializer` implementation that accepts a input `Set<EventSerializer>`, such as `JmapEventSerializer`, `MailboxEventSerializer`, `AnyEventSerializer`...
This approach will simplify Guice binding (using a set binder) and easy when one day we have a new `EventSerializer` implement. 

The code looks like:

```java
class EventSerializers implements EventSerializer {

    Set<EventSerializer> eventSerializers;

    @Inject
    public EventSerializers(Set<EventSerializer> eventSerializers) {
        this.eventSerializers = eventSerializers;
    }

    @Override
    public String toJson(Event event) {
        return "";
    }

    @Override
    public Event asEvent(String serialized) {
        eventSerializers.forEach(eventSerializer -> {
            // loop to find the correct event serializer
            }
        });
    }
}
```

Then we can use `EventSerializers` for deserialize event at `CassandraEventDeadLettersDAO`

hungphan227:
> I propose creating a new `EventSerializer` implementation that accepts a input `Set<EventSerializer>`, such as `JmapEventSerializer`, `MailboxEventSerializer`, `AnyEventSerializer`... This approach will simplify Guice binding (using a set binder) and easy when one day we have a new `EventSerializer` implement.
> 
> The code looks like:
> 
> ```java
> class EventSerializers implements EventSerializer {
> 
>     Set<EventSerializer> eventSerializers;
> 
>     @Inject
>     public EventSerializers(Set<EventSerializer> eventSerializers) {
>         this.eventSerializers = eventSerializers;
>     }
> 
>     @Override
>     public String toJson(Event event) {
>         return "";
>     }
> 
>     @Override
>     public Event asEvent(String serialized) {
>         eventSerializers.forEach(eventSerializer -> {
>             // loop to find the correct event serializer
>             }
>         });
>     }
> }
> ```
> 
> Then we can use `EventSerializers` for deserialize event at `CassandraEventDeadLettersDAO`

I do not totally agree with this approach, but because you all think it is good, I would follow

event-bus/api/src/main/java/org/apache/james/events/EventDeadLetters.java
chibenwa, line 97: Shall we remove this method?

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
```diff
@@ -19,23 +19,44 @@
...
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 public class CassandraEventDeadLetters implements EventDeadLetters {
+    private static final Logger LOGGER = LoggerFactory.getLogger(CassandraEventDeadLetters.class);
 
     private final CassandraEventDeadLettersDAO cassandraEventDeadLettersDAO;
     private final CassandraEventDeadLettersGroupDAO cassandraEventDeadLettersGroupDAO;
+    private final EventSerializer eventSerializer;
+    private final Set<EventSerializer> allEventSerializers;
```
chibenwa, line 43: Why we need both?

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
```diff
@@ -89,4 +111,26 @@ public Flux<Group> groupsWithFailedEvents() {
     public Mono<Boolean> containEvents() {
         return cassandraEventDeadLettersDAO.containEvents();
     }
+
+    private Event deserialize(String json) {
+        List<Event> events = allEventSerializers.stream()
+            .map(eventSerializer -> Optional.ofNullable(deserialize(json, eventSerializer)))
+            .filter(Optional::isPresent)
+            .map(Optional::get)
```
chibenwa, line 119: .flatMap(Optional::stream)

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
```diff
@@ -89,4 +111,26 @@ public Flux<Group> groupsWithFailedEvents() {
...
+            .toList();
+
+        switch (events.size()) {
+            case 0: throw new RuntimeException("Could not deserialize event: " + json);
+            case 1: return events.getFirst();
+            default: throw new RuntimeException("More than one serializers could deserialize event: " + json);
+        }
```
chibenwa, line 126: Is this a check we need to do at runtime or we could do it at startup?

Because then we could just findFirst...orElseThrow

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
```diff
@@ -89,4 +111,26 @@ public Flux<Group> groupsWithFailedEvents() {
...
+    }
+
+    private Event deserialize(String json, EventSerializer eventSerializer) {
+        try {
+            return eventSerializer.asEvent(json);
+        } catch (Exception ex) {
+            return null;
```
chibenwa, line 133: return null spotted!!!

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
```diff
@@ -19,23 +19,44 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableSet;
 
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
 public class CassandraEventDeadLetters implements EventDeadLetters {
```
chibenwa, line 37: Are we really going to duplicate the Event -> Json and json -> event in all classes like PG or we do plan to have an abstraction we can reuse?

chibenwa (commented):
As often this works lacks detailed care about code stylers and abstractions.

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
hungphan227: EventDeadLetters instance used by event bus has appropriate EventSerializer to serialize event. Meanwhile, EventDeadLetters instance used by webadmin could not get appropriate EventSerializer. Therefore, it has to use allEventSerializers to deserialize event json from database.

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
hungphan227: Really difficult to check at startup

Although there is a very small chance this case happen, I still want to check this case just for precaution.

event-bus/api/src/main/java/org/apache/james/events/EventDeserializers.java
```diff
@@ -0,0 +1,58 @@
...
+        List<Event> events = allEventSerializers.stream()
+            .map(eventSerializer -> deserialize(json, eventSerializer))
+            .flatMap(Optional::stream)
+            .toList();
+
+        switch (events.size()) {
+            case 0: throw new RuntimeException("Could not deserialize event: " + json);
+            case 1: return events.getFirst();
+            default: throw new RuntimeException("More than one serializers could deserialize event: " + json);
+        }
```
chibenwa, line 48:
```suggestion
    public Event deserialize(String json) {
        List<Event> events = allEventSerializers.stream()
            .map(eventSerializer -> deserialize(json, eventSerializer))
            .flatMap(Optional::stream)
            .findFirst()
            .orElseThrow(() -> new RuntimeException("Could not deserialize event: " + json));
```

chibenwa: If not having 2 serializers for the same event is important please check that at startup.

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
hungphan227: removed

chibenwa (approved):

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
```diff
@@ -21,21 +21,38 @@
...
+                                     EventSerializer eventSerializer) {
+        this(cassandraEventDeadLettersDAO, cassandraEventDeadLettersGroupDAO, eventSerializer, new EventDeserializers(ImmutableSet.of(eventSerializer)));
+    }
 
     @Inject
     public CassandraEventDeadLetters(CassandraEventDeadLettersDAO cassandraEventDeadLettersDAO,
-                                     CassandraEventDeadLettersGroupDAO cassandraEventDeadLettersGroupDAO) {
+                                     CassandraEventDeadLettersGroupDAO cassandraEventDeadLettersGroupDAO,
+                                     EventSerializer eventSerializer,
+                                     EventDeserializers deserializer) {
```
vttranlina, line 51: I still don't understand why we need separate for serialization and deserialization, and what the difference is between them.

quantranhong1999: How about support for multiple `EventSerializer` as well? I found weird as we support multiple `EventDeserializer` (s) but not for the serializer.

Maybe just create an `EventSerializersAggregate` or something to support multiple serializers + deserializers?

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
vttranlina: @chibenwa all issue come from
When we want to use webadmin to retrigger event dead letter (that need read and deserialize string from cassandra)

`EventDeadLettersRoutes` invoke `EventDeadLettersRedeliverService`
and `EventDeadLettersRedeliverService` invoke `EventDeadLetters` (SINGLETON instance)

The current code is:
Screenshot_20241024_202233

It is incorrect
it should be
`public EventDeadLettersRedeliverService(Set<EventBus, EventDeadLetters> eventBusesDeadLetterMap)`

The Hung's idea here is: 

Can have several instances of `EventDeadLetters` for "write" event to db (one EventDeadLetters per EventBus)
But have only single instance for "read" event from db.

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLetters.java
chibenwa, line 51: 2 things doing the same thing is a no-no for me!

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLettersDAO.java
```diff
@@ -121,11 +119,11 @@ Mono<Void> removeEvents(Group group) {
             .setString(GROUP, group.asString()));
     }
 
-    Mono<Event> retrieveFailedEvent(Group group, EventDeadLetters.InsertionId insertionId) {
+    Mono<String> retrieveFailedEvent(Group group, EventDeadLetters.InsertionId insertionId) {
```
chibenwa, line 122: We shall strive to keep a strongly type DAO see the proposal of API redesign.

event-bus/cassandra/src/main/java/org/apache/james/events/CassandraEventDeadLettersDAO.java
```diff
@@ -103,11 +101,11 @@ private PreparedStatement prepareContainEventStatement(CqlSession session) {
             .build());
     }
 
-    Mono<Void> store(Group group, Event failedEvent, EventDeadLetters.InsertionId insertionId) {
+    Mono<Void> store(Group group, String failedEvent, EventDeadLetters.InsertionId insertionId) {
```
chibenwa, line 104: We shall strive to keep a strongly type DAO see the proposal of API redesign.

chibenwa (changes requested):
What I do not like here is the lack of symetry in the design.

Symetry makes it easier to reason about things, and is commonly refered as beatiful, be it in art or in software design.

We can and MUST further enhance this work:

We shall...

 - add a `boolean isHandling(Event event)` into `EventSerializer` interface 
 - with this we could rename `EventDeserializers` into `EventSerializerAggregate` that implements `EventSerilizer` interface.
 
With that the EventDoedLetter can rely solely on `EventSerializerAggregate` and we could even keep a strongly types DAO.

chibenwa:
image

hungphan227:
Created EventSerializersAggregator that extend EventSerializer. This is the default event serializer. Although EventSerializersAggregator has both serialize and deserialize method, CassandraEventDeadLettersDAO instance used by webadmin only use deserialize method. Therefore, we do not need to care about serialize case. For CassandraEventDeadLettersDAO instances used by event bus, we inject appropriate EventSerializer.

`public CassandraEventDeadLettersDAO(CqlSession session, EventSerializer eventSerializer)`

chibenwa (approved):

quantranhong1999 (approved):

vttranlina (approved):

PR #2469, opened 2024-10-24
JAMES-4081 Implement MailToAllUsers mailet

quantranhong1999:


server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/MailToAllUsers.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    @Override
+    public void service(Mail mail) throws MessagingException {
+        Flux<Flux<MailAddress>> recipientsBatches = Flux.from(usersRepository.listReactive())
+            .map(Throwing.function(Username::asMailAddress))
+            .window(batchSize);
+
+        Flux.merge(sendMailToFirstRecipientsBatchDirectly(mail, recipientsBatches),
+                sendMailToRemainingRecipientsBatchAsynchronously(mail, recipientsBatches))
```
chibenwa, line 81: The issue is that we subscribe twice and the order might not be stable.

This means the first 100 items on the first subscriptions are not necessarily the ones of the next subscriptions.

How about:

```
Flux.from(usersRepository.listReactive())
            .map(Throwing.function(Username::asMailAddress))
            .window(batchSize);
            .index()
            .flatMap(tuple -> if (tuple.key() == 0) {
                return sendMailToFirstRecipientsBatchDirectly(mail, tuple.value());
            } else {
                return sendMailToRemainingRecipientsBatchAsynchronously(mail, tuple.value());
            }).then().block();

quantranhong1999: We may use `Flux.concat` as it would wait for the first Publisher to finish.

Or `sendMailToFirstRecipientsBatchDirectly.then(sendMailToRemainingRecipientsBatchAsynchronously)`.

quantranhong1999: BTW `recipientsBatches`  is a cold publisher so is it supposed to produce data isolated regardless of subscribers?

chibenwa: Yes but because it is subscribed twice data may defer.

chibenwa: first call: `a, b, c, d, e` => send mail synchronously to `a, b, c`

second call `d, e, b, a, c` => send mail asynchrounously to `a, c`

 - We called twice the ldap
 - and we sent `a, b, c, a, c` and not `a, b, c, d, e`

chibenwa (approved):

Arsnael (approved):

PR #2479, opened 2024-10-31
[Metrics] Add a few Websocket metrics

quantranhong1999:
- Opening connections count
- Requests throughput

### DIFF
--- a/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/WebSocketRoutes.scala
+++ b/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/WebSocketRoutes.scala
@@ -41,6 +41,7 @@ import org.apache.james.jmap.http.{Authenticator, UserProvisioning}
 import org.apache.james.jmap.json.{PushSerializer, ResponseSerializer}
 import org.apache.james.jmap.{Endpoint, JMAPRoute, JMAPRoutes, InjectionKeys => JMAPInjectionKeys}
 import org.apache.james.mailbox.MailboxSession
+import org.apache.james.metrics.api.{Metric, MetricFactory}
 import org.apache.james.user.api.DelegationStore
 import org.slf4j.{Logger, LoggerFactory}
 import play.api.libs.json.Json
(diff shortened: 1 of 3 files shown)


server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/WebSocketRoutes.scala
```diff
@@ -79,7 +80,12 @@ class WebSocketRoutes @Inject() (@Named(InjectionKeys.RFC_8621) val authenticato
                                  emailChangeRepository: EmailChangeRepository,
                                  pushSerializer: PushSerializer,
                                  typeStateFactory: TypeStateFactory,
-                                 delegationStore: DelegationStore) extends JMAPRoutes {
+                                 delegationStore: DelegationStore,
+                                 metricFactory: MetricFactory,
+                                 gaugeRegistry: GaugeRegistry) extends JMAPRoutes {
+  private val openingConnectionsMetric: Metric = metricFactory.generate("jmap_websocket_opening_connections")
```
chibenwa, line 86:
```suggestion
  private val openingConnectionsMetric: Metric = metricFactory.generate("jmap_websocket_opened_connections")
```

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/WebSocketRoutes.scala
```diff
@@ -113,6 +120,10 @@ class WebSocketRoutes @Inject() (@Named(InjectionKeys.RFC_8621) val authenticato
         frame.content().readBytes(bytes)
         new String(bytes, StandardCharsets.UTF_8)
       })
+      .doOnNext(_ => {
+        requestCountMetric.increment()
+        requestThroughputGauge.setValue(requestCountMetric.meanRate().intValue)
```
chibenwa, line 125: Kill that gauge, it won't be performant. And we can do a derivative for that in grafana.

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/WebSocketRoutes.scala
quantranhong1999: Do you mean opened connections of all time, or just opened connections at the moment?

chibenwa (approved):
We can do IMO the reactor.netty http metrics separately...

Arsnael (approved):

quantranhong1999:
Squashed.

PR #2483, opened 2024-11-01
[Metrics] Expose a few useful Reactor Netty metrics

quantranhong1999:
- jmap.reactor.netty.http.server.connections.active
- jmap.reactor.netty.http.server.connections.total
- jmap.reactor.netty.http.server.data.received
- jmap.reactor.netty.http.server.data.sent

cf https://projectreactor.io/docs/netty/1.1.19/reference/index.html#_metrics_4

### DIFF
--- a/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/SessionRoutes.scala
+++ b/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/SessionRoutes.scala
@@ -23,8 +23,8 @@ import java.nio.charset.StandardCharsets
 import java.util.stream.Stream
 
 import io.netty.handler.codec.http.HttpHeaderNames.{CONTENT_LENGTH, CONTENT_TYPE}
-import io.netty.handler.codec.http.HttpResponseStatus.{BAD_REQUEST, INTERNAL_SERVER_ERROR, OK, UNAUTHORIZED}
-import io.netty.handler.codec.http.{HttpMethod, HttpResponseStatus}
+import io.netty.handler.codec.http.HttpMethod
+import io.netty.handler.codec.http.HttpResponseStatus.{INTERNAL_SERVER_ERROR, OK, UNAUTHORIZED}
 import jakarta.inject.{Inject, Named}
 import org.apache.commons.lang3.tuple.Pair
 import org.apache.james.core.Username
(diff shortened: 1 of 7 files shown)


server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/SessionRoutes.scala
```diff
@@ -73,6 +75,7 @@ class SessionRoutes @Inject()(@Named(InjectionKeys.RFC_8621) val authenticator:
       .flatMap(session => sendRespond(session, response))
       .onErrorResume(throwable => SMono.fromPublisher(errorHandling(throwable, response)))
       .asJava()
+      .doOnSuccess(_ => httpClientMetrics.update())
```
quantranhong1999, line 78: Trigger the polling update upon successful GET `/jmap/session` seems ok IMO. While it could be a bit much doing it at POST `/jmap` (or we could add some random).

Arsnael (approved):

hungphan227 (approved):

server/protocols/jmap/src/main/java/org/apache/james/jmap/JMAPServer.java
```diff
@@ -90,10 +95,19 @@ public void start() {
                     .orElse(RANDOM_PORT))
                 .handle((request, response) -> handleVersionRoute(request).handleRequest(request, response))
                 .wiretap(wireTapEnabled())
+                .metrics(true, Function.identity())
```
chibenwa, line 98:
```suggestion
                .metrics(ENABLED, Function.identity())
```

And out of question what does `Function.identity()` allow to do?

chibenwa: Could i be configurable? Eg if perf impact is high...

Arsnael (approved):

server/protocols/jmap/src/main/java/org/apache/james/jmap/JMAPServer.java
```diff
@@ -90,10 +96,21 @@ public void start() {
...
+            if (REACTOR_NETTY_METRICS_ENABLE) {
+                configureReactorNettyMetrics();
+            }
         }
     }
 
+    private void configureReactorNettyMetrics() {
+        Metrics.globalRegistry
+            .config()
+            .meterFilter(MeterFilter.maximumAllowableTags(HTTP_CLIENT_PREFIX, URI, 100, MeterFilter.deny()));
```
chibenwa, line 111: 100 is a magic constant....

chibenwa (approved):

PR #2485, opened 2024-11-04
JAMES-4085 BlobStore: introduce APIs for multi-tenancy support

quantranhong1999:


vttranlina (commented):
read it

hungphan227 (approved):

Arsnael (approved):

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStore.java
```diff
@@ -43,45 +45,118 @@ interface BlobIdProvider {
 
     Publisher<BlobId> save(BucketName bucketName, byte[] data, StoragePolicy storagePolicy);
 
+    default Publisher<BlobId> save(Bucket bucket, byte[] data, StoragePolicy storagePolicy) {
+        return save(bucket.bucketName(), data, storagePolicy);
+    }
```
chibenwa, line 50: This makes it harder to use the class: we can not double the number of methods every time we add some parameters.

I'd be happy we make it easy to get a Bucket from a BucketName and sollely depend on bucket

chibenwa (approved):

Arsnael (approved):

jeantil:
can you wait a bit before merging ? I will try to do a thorough review tonight. I am concerned about `getDefaultBucket` and how it is used. 
I feel that this method is dangerous as it will never account for a tenant making it easy to ignore tenancy at the callsite. if we want the blobstore and james to be multitenant, the tenancy information should probably propagate throughout all the call sites until we the point where we can determine the information.

chibenwa:
Hi @jeantil 

> can you wait a bit before merging ?

of course ^^

> I am concerned about getDefaultBucket and how it is used.
I feel that this method is dangerous as it will never account for a tenant making it easy to ignore tenancy at the callsite. if we want the blobstore and james to be multitenant, the tenancy information should probably propagate throughout all the call sites until we the point where we can determine the information

Legitimate concerns indeed yet : 

 - the only component being currently multi-tenant is the mailbox
 - and the only multi-tenant mailbox  implementation is the PGSQL one.

The proposed changeset will then only make sense on top of that mailbox

> I am concerned about getDefaultBucket and how it is used.

We could easily say we kill that method, and make the notion of "default bucket" a caller concern

(test constant, etc...)

chibenwa:
@quantranhong1999 can you have a look at what it takes to remove `BloBstore::getDefaultBucket`

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -128,9 +129,14 @@ default Publisher<Void> save(BucketName bucketName, BlobId blobId, String data)
      * @return a successful Publisher if the bucket is deleted or did not exist
      *  otherwise an IOObjectStoreException in its error channel
      */
-    Publisher<Void> deleteBucket(BucketName bucketName);
+    Publisher<Void> deleteBucket(Bucket bucket);
 
     Publisher<BucketName> listBuckets();
```
jeantil, line 134: this should return Buckets to be consistent wit the changed apis

jeantil:
I realized that this PR is the implementation of the mail that proposes the change on multi-tenancy that was sent yesterday. 
Looking at the code, jira and mail, I realized that my concerns are more at an architectural level than related to this implementation ( though the challenge to the architectural choices may render this changeset obsolete)  I thus answered with my design questions on the ML instead

chibenwa:
Looks like the email did not went through

image

quantranhong1999:
I mark this PR as a draft as the design for this topic likely is complicated and deserves more discussion among the community.

PR #2489, opened 2024-11-05
[IMPROVEMENT] mock-smtp-server (parameters without values)

Maxxx873:
Hello! 
I think the mock-smtp-server needs tweaking to improve the integration testing work, he must accept a specific list of parameters without values. Like for example for a `REQUIRETLS` parameter in a task: 
[JAMES-3823](https://issues.apache.org/jira/projects/JAMES/issues/JAMES-3823?filter=allopenissues)
[#2460](https://github.com/apache/james-project/pull/2460)

### DIFF
--- a/server/mailet/mock-smtp-server/src/main/java/org/apache/james/mock/smtp/server/model/Mail.java
+++ b/server/mailet/mock-smtp-server/src/main/java/org/apache/james/mock/smtp/server/model/Mail.java
@@ -23,6 +23,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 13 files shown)


server/mailet/mock-smtp-server/src/main/java/org/apache/james/mock/smtp/server/model/Mail.java
```diff
@@ -41,6 +41,9 @@
 
 @JsonDeserialize(builder = Mail.Builder.class)
 public class Mail {
+
+    public static final List<String> smtpParametersWithoutValues = List.of("REQUIRETLS", "SMTPUTF8");
```
chibenwa, line 45: Can we avoid being hardcoding a list of values here?

If we just say 'if it do not contains a `=`' then we associate `""` then no specific whitelist is needed?

Maxxx873: Yes, we can get rid of the whitelist. But it seems to me that the mock server operation would look more plausible if it only allowed certain parameters without values. If we need to, we can decide how to pass the list as arguments or properties or whatever. But if it really doesn't make sense, I'll just leave it possible for any parameter to have an empty value.

chibenwa: My concern would be keep adding value and handling those two in a different fashion.

If needed we could go from `Map<String, String>` as a datastructure to something like `Map<String, Optional<String>>`?

Maxxx873:
I decided to follow the comments and make an implementation without hardcoding the whitelist. Thanks! I've added some more refactoring since we're making changes to this module anyway.
Can you update the image in the docker hub if the PR is accepted?

chibenwa (approved):

Arsnael (approved):
Thanks for the contribution @Maxxx873 , LGTM :)

PR #2500, opened 2024-11-12
Cherry pick commits from master to postgres

hungphan227:


chibenwa:
Why cherry picks? I t means we srew up the commit order?

Won't it leads to a mess? 

I'd rather just keep on rebasing master.

Ideally I'd love we fix the remaining concerns with the pg-app and just get it merged...

PR #2508, opened 2024-11-15
Convert ExtractedContacts class to a record

HoussemNasri:
## Why?

In https://github.com/linagora/tmail-backend/pull/1308, calling `ObjectMapper#readValue` and casting the result to ExtractedContacts results in the following error: `com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Cannot construct instance of org.apache.james.transport.mailets.ContactExtractor$ExtractedContacts`. 

Changing it to a record resolves the issue.

### DIFF
--- a/mailet/standard/src/main/java/org/apache/james/transport/mailets/ContactExtractor.java
+++ b/mailet/standard/src/main/java/org/apache/james/transport/mailets/ContactExtractor.java
@@ -144,21 +144,6 @@ public class ContactExtractor extends GenericMailet implements Mailet {
             .map(MimeUtil::unscrambleHeaderValue);
     }
 
-    public static class ExtractedContacts {
-        private final String userEmail;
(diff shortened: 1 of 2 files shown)


chibenwa (approved):
I do not understand how this fixes anything but the changeset do not look bad in itself.

HoussemNasri:
The class `ExtractedContacts` is [immutable](https://www.baeldung.com/jackson-deserialize-immutable-objects) and does not offer any `@JsonCreator` annotated method or constructor neither it has a custom-made deserializer which makes it impossible to deserialize with Jackson. Converting it to a record makes it deserializable for some reason, I couldn't find any documentation of this special behavior for records through.

Arsnael (approved):

quantranhong1999 (approved):

vttranlina (approved):

PR #2521, opened 2024-11-25
JAMES-3945 advertize that subaddressing is supported

florentos17:
[jira JAMES-3945 ticket for subaddressing](https://issues.apache.org/jira/browse/JAMES-3945)

subaddressing has now been implemented on the backend side, but not fully deployed. we need to have the frontend automatically detect if the backend supports subaddressing or not.

thus i modified the `urn:apache:james:params:jmap:mail:shares` capability to add a `subaddressingSupported` property (boolean). if this property is not defined, then the frontend should assume it is not supported.

### DIFF
--- a/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/core/Capability.scala
+++ b/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/core/Capability.scala
@@ -317,7 +317,7 @@ case object DelegationCapabilityFactory extends CapabilityFactory {
 }
 
 final case class SharesCapabilityProperties() extends CapabilityProperties {
-  override def jsonify(): JsObject = Json.obj()
+  override def jsonify(): JsObject = Json.obj("subaddressingSupported" -> true)
 }
 
 case object SharesCapabilityFactory extends CapabilityFactory {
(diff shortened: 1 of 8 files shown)


chibenwa:
Question 1: Why do we need this?

Question 2: Why isn't the doc updated? ( https://github.com/apache/james-project/blob/master/server/protocols/jmap-rfc-8621/doc/specs/spec/mail/rights.mdown#defined-capability )

Question 3: Why ISSUE-1324? THis is NOT an ASF bugtracker. Can we git inspect to find a related issue where we could explain what we do here? And use it as a PR title / commit message prefix?

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/core/Capability.scala
```diff
@@ -317,7 +317,7 @@ case object DelegationCapabilityFactory extends CapabilityFactory {
 }
 
 final case class SharesCapabilityProperties() extends CapabilityProperties {
-  override def jsonify(): JsObject = Json.obj()
+  override def jsonify(): JsObject = Json.obj("subaddressingSupported" -> true)
```
chibenwa, line 320: No impact on tests

florentos17: i just wrote one, i am checking that everything is fine and i push

chibenwa: It should impact existing session integration tests, no need to write new test IMO

chibenwa (approved):

Arsnael (approved):

chibenwa (approved):

PR #2525, opened 2024-11-27
[FIX] IDLE - prevent data race upon answers on channel writes

chibenwa:
image


server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/NettyImapSession.java
```diff
@@ -113,6 +115,21 @@ public void executeSafely(Runnable runnable) {
...
+    public ImapProcessor.Responder threadSafe(ImapProcessor.Responder responder) {
+        return new ImapProcessor.Responder() {
+            @Override
+            public void respond(ImapResponseMessage message) {
+                channel.eventLoop().execute(() -> responder.respond(message));
+            }
+
+            @Override
+            public void flush() {
+            channel.eventLoop().execute(responder::flush);
```
Arsnael, line 128:
```suggestion
                channel.eventLoop().execute(responder::flush);
```

Arsnael (approved):

PR #2526, opened 2024-11-27
[FIX] Ability to split emails with large recipient counts

chibenwa:
- I suspect James could struggle in distributed version to distribute very large count of recipients To this end, splitting email would help.
 - This is a default Postfix behaviour.

 The code snippet is extracted from MailToAllUsers mailet.

### DIFF
--- a/docs/modules/servers/partials/SplitMail.adoc
+++ b/docs/modules/servers/partials/SplitMail.adoc
@@ -1,0 +1,16 @@
+=== MailToAllUsers
+
+A mailet to split email with too much recipients. The first batch is sent directly, while the remaining batches are sent asynchronously.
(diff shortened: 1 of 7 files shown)


Arsnael (commented):
Missing doc changes for MailToAllUsers in MailToAllUsers.adoc and antora doc for the new mailet

docs/modules/servers/partials/SplitMail.adoc
```diff
@@ -0,0 +1,16 @@
+=== MailToAllUsers
```
Arsnael, line 1:
```suggestion
=== SplitMail
```

Arsnael:
Still missing doc changes for MailToAllUsers.adoc, as it does not send mails in batches anymore

Arsnael (approved):

vttranlina (approved):

PR #2491, opened 2024-11-06
refactor and test the use of the `anyone` keyword when sharing with JMAP

florentos17:
tackling [issue #5314](https://github.com/linagora/james-project/issues/5314)

PR #2527, opened 2024-11-27
support all EntryKeys when sharing with JMAP

florentos17:
following the refactoring done in [this pull request](https://github.com/apache/james-project/pull/2491), all EntryKeys are supported, no need to filter out some of them anymore.


chibenwa:
For me the benefits are unclear and the API (JMAP) impact is unclear.

IMO I would keep the JMAP API restricted to identified use cases.

(So likely aborting the work on this PR)

PR #2530, opened 2024-11-28
[ENHANCEMENT] Add a PartHasContentType

chibenwa:


mailet/standard/src/test/java/org/apache/james/transport/matchers/PartHasContentTypeTest.java
```diff
@@ -0,0 +1,105 @@
...
+                    MimeMessageBuilder.bodyPartBuilder()
+                        .type("image/png")
+                        .disposition("attachment")
+                        .filename("xxx.zip")))
+            .build();
+        Collection<MailAddress> matched = testee.match(mail);
+
+        assertThat(matched)
+            .isNotEmpty();
+    }
```
Arsnael, line 104: How about a test with multiple content types configured with the matcher?

chibenwa: That's the point of the refactoring: it leverages complexity of mime types and have that covered by AttachmentFileNameIs.

Code coverage is 100%

I do not see the point of adding more tests.

Arsnael (approved):

PR #2532, opened 2024-11-29
[ENHANCEMENT] Log size in the AuditTrail when sending an email

chibenwa:
This could be useful for audit purposes
as well as extracting trends / finding
offenders.


chibenwa:
(appouved in https://github.com/apache/james-project/pull/2531, reopened for CI)

PR #2536, opened 2024-12-01
[FIX] Support RFC-3848 service types in Received headers

chibenwa:
- Detail LMTP service type
 - Expose auth settings through service type
 - Extract message generation
 
 CF https://datatracker.ietf.org/doc/html/rfc3848

### DIFF
--- a/protocols/lmtp/src/main/java/org/apache/james/protocols/lmtp/core/ReceivedDataLineFilter.java
+++ b/protocols/lmtp/src/main/java/org/apache/james/protocols/lmtp/core/ReceivedDataLineFilter.java
@@ -53,7 +53,7 @@ public class ReceivedDataLineFilter extends org.apache.james.protocols.smtp.core
                     if (session.isTLSStarted()) {
                         return SERVICE_TYPE_SSL_AUTH;
                     } else {
-                        return SERVICE_TYPE_SSL;
+                        return SERVICE_TYPE_AUTH;
                     }
                 }
             }
--- a/protocols/lmtp/src/main/java/org/apache/james/protocols/lmtp/core/ReceivedDataLineFilter.java
+++ b/protocols/lmtp/src/main/java/org/apache/james/protocols/lmtp/core/ReceivedDataLineFilter.java
@@ -19,6 +19,7 @@
 package org.apache.james.protocols.lmtp.core;
 
 import org.apache.james.protocols.smtp.SMTPSession;
+import org.apache.james.protocols.smtp.core.ReceivedHeaderGenerator;
 
 /**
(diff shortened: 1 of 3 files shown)


quantranhong1999 (commented):
Read it

chibenwa:
Ideally we shall write some tests for this new message generator

protocols/lmtp/src/main/java/org/apache/james/protocols/lmtp/core/ReceivedDataLineFilter.java
```diff
@@ -19,19 +19,45 @@
...
+                        return SERVICE_TYPE;
+                    }
+                } else {
+                    // See RFC3848:
+                    // The new keyword "ESMTPA" indicates the use of ESMTP when the SMTP
+                    // AUTH [3] extension is also used and authentication is successfully achieved.
+                    if (session.isTLSStarted()) {
+                        return SERVICE_TYPE_SSL_AUTH;
+                    } else {
+                        return SERVICE_TYPE_SSL;
```
Arsnael, line 56: shouldn't we return here SERVICE_TYPE_AUTH ?

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/ReceivedHeaderGenerator.java
```diff
@@ -0,0 +1,126 @@
...
+
+public class ReceivedHeaderGenerator {
+    private static final DateTimeFormatter DATEFORMAT = DateTimeFormatter.ofPattern("EEE, dd MMM yyyy HH:mm:ss Z (zzz)", Locale.US);
+
+    private static final String EHLO = "EHLO";
+    private static final String SMTP = "SMTP";
+    private static final String ESMTPA = "ESMTPA";
+    private static final String ESMTPSA = "ESMTPA";
+    private static final String ESMTP = "ESMTP";
+    private static final String ESMTPS = "ESMTP";
```
Arsnael, line 42: Any reason that `ESMTPSA` constant is "ESMTPA" and not "ESMTPSA"?

Same question for `ESMTPS`

chibenwa:
Typos sorry

protocols/lmtp/src/main/java/org/apache/james/protocols/lmtp/core/ReceivedDataLineFilter.java
Arsnael: Was I wrong here?

Arsnael:
I believe the last fixup was forgotten and necessary too. Please dont hesitate to tell me if it was wrong.

Arsnael (approved):

PR #2547, opened 2024-12-05
[ENHANCEMENT] ip and x-forwarded-for im JMAP logging context

chibenwa:

PR #2564, opened 2024-12-11
[ENHANCEMENT] ip and x-forwarded-for im JMAP logging context

Arsnael:
Duplicate of https://github.com/apache/james-project/pull/2547

I noticed all builds are timeouts, with even sometimes some java heap space oom issues, but all the builds were continuously being built on the same worker (builds26). So maybe that worker has some issues.

Trying to force the build on an other worker here...

### DIFF
--- a/server/container/util/src/main/java/org/apache/james/util/ReactorUtils.java
+++ b/server/container/util/src/main/java/org/apache/james/util/ReactorUtils.java
@@ -194,7 +194,7 @@ public class ReactorUtils {
         return signal -> logWithContext(logStatement, signal.getContextView());
     }
 
-    private static void logWithContext(Runnable logStatement, ContextView contextView) {
+    public static void logWithContext(Runnable logStatement, ContextView contextView) {
         try (Closeable mdc = retrieveMDCBuilder(contextView).build()) {
             logStatement.run();
         } catch (IOException e) {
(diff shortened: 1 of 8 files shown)


quantranhong1999 (approved):

Arsnael:
So...

Locally I could still that some tests on memory jmap integration tests were indeed hanging forever.

After spending some time checking the code again and how context works with Reactor, I realized that @chibenwa likely mixed up some functions. `subscriberContext` and `contextWrite` are actually doing the same thing, used for writing context and propagate it to the pipeline. However `subscriberContext` is deprecated and bound to be replaced by `contextWrite`.

The `contextWrite` usage is correct, to record the IP and x-forwarded-for with the MDC. However usage of subscriberContext is wrong. What we want there is being able to log JMAP context. For this, should use `Mono.deferContextual`. And we have a function in ReactorUtils that wraps up the log statement in a `Mono.deferContextual` called `logAsMono` (which is calling logWithContext)

So I believe wrapping the AuditTrail in a logAsMono call is the right move here.

Locally tests are passing now. Let's see

let me know if I misunderstood something though.

vttranlina (approved):

quantranhong1999 (approved):

quantranhong1999:
The AuditTrail logs for JMAP disappeared. We need to bring them back cf https://github.com/linagora/james-project/issues/5384...

PR #2567, opened 2024-12-12
[PERF] Email move should send only one JMAP push notification

chibenwa:
Today moving emails triggers 2 JMAP websocket notification: it shall generate a signle one!

Key idea: on the EventBus, I shall be able to publish a list of events.

This will be serialized and transported by rabbit as a list of event. (We can have a configuration parameter to turn on/off this parameter to ease rolling migration). Deserialization will need to handle both list and single events.

It will be executed as a list of events by listeners, with a default implementation to handle the current single event (without breaking change)

MailboxChangeListener will handle the email move in a single pass: process both `Added` and `Expunged` in a single pass and return a single `StateChange`.

See https://github.com/linagora/tmail-flutter/issues/3331

### DIFF
--- a/docs/modules/servers/partials/configure/jvm.adoc
+++ b/docs/modules/servers/partials/configure/jvm.adoc
@@ -100,6 +100,15 @@ james.jmap.reactor.netty.metrics.enabled=true
 ----
 To enable the metrics.
 
+== Set up JMAP EMail/set range threshold
+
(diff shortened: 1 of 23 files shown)


quantranhong1999 (approved):

Arsnael (approved):

chibenwa:
(Was :green_apple: )

PR #2482, opened 2024-10-31
JAMES-4083 remove logback jackson usages from james

florentos17:
cf the [jira ticket](https://issues.apache.org/jira/browse/JAMES-4083)

i mounted the jar files as suggested, but i could not manage to point to them in every pom file. using ${project.basedir} in the parent pom does not work since it points to each pom's individual base directory... should i specify the path in each pom ?

PR #2569, opened 2024-12-17
[devscout] upgrade logback and crowdsec

jeantil:
This commit drops the unmaintained logback-classic dependecy and upgrades logback to the latest version.

Logback-classic has a native JsonEncoder (already in the original version). The latest version allows to have the formatted message and to enable/disable any top level field\ by configuration.

This allows using logback's own JSON encoder and get rid ot logback-contrib which is unmaintained to output json ( the repository is archived and the dependencies are quite outdated)

 A side effect was that the format of the Json output changes slightly, I adjusted the samplec configurations for the crowdsec module.

One of the change is a higher resolution for the timestamp, which required raising crowdsec to 1.6.x.

I missed the previous [merge request](https://github.com/apache/james-project/pull/2482) which responds to [JAMES-4083](https://issues.apache.org/jira/browse/JAMES-4083). I discovered this while rebasing on upstream :grimacing: but I feel getting rid of the whole abandoned logback contrib is a better solution so I still propose this change to you.

I adjusted the crowdsec integration tests and I use this setup in my "production" already,

### DIFF
--- a/protocols/netty/src/main/java/org/apache/james/protocols/netty/ProtocolMDCContextFactory.java
+++ b/protocols/netty/src/main/java/org/apache/james/protocols/netty/ProtocolMDCContextFactory.java
@@ -96,7 +96,6 @@ public interface ProtocolMDCContextFactory {
     static MDCBuilder forSession(ProtocolSession protocolSession) {
         return MDCBuilder.create()
             .addToContext(MDCBuilder.SESSION_ID, protocolSession.getSessionID())
-            .addToContext(MDCBuilder.CHARSET, protocolSession.getCharset().displayName())
             .addToContextIfPresent(MDCBuilder.USER, Optional.ofNullable(protocolSession.getUsername()).map(Username::asString));
     }
 
--- a/docs/modules/servers/partials/operate/logging.adoc
+++ b/docs/modules/servers/partials/operate/logging.adoc
@@ -25,6 +25,18 @@ and link:https://docs.fluentbit.io/[FluentBit] as centralize logging.
 Information about logback configuration can be found
 link:http://logback.qos.ch/manual/configuration.html[here].
 
+It is possible to configure a logback appender to expose structured logging
+as JSON using logback's  `JsonEncoder` like in the following example:
+
+```
+<appender name="CONSOLE" class="ch.qos.logback.core.ConsoleAppender">
+         <encoder class="ch.qos.logback.classic.encoder.JsonEncoder">
+             <withFormattedMessage>true</withFormattedMessage>
+         </encoder>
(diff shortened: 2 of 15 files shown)


chibenwa:
Hello @jeantil 

> Logback-classic has a native JsonEncoder

We (I) missed this!

THanks for catching this!

> 😬 but I feel getting rid of the whole abandoned logback contrib is a better solution so I still propose this change to you.

Huge +1

chibenwa (approved):

chibenwa:
image

It looks like the crowdsec stuff needs more adaptation!

jeantil:
I forgot to rerun all the integration tests i'll look into this tonight

Arsnael (approved):

jeantil:
So i did look into this qand I don't understand : it seems that my changes somehow prevent proxification through haproxy. the tests that  fail do so before logging anything that actually could be interepreted by crowdsec :exploding_head:

chibenwa:
>  it seems that my changes somehow prevent proxification through haproxy. the tests that fail do so before logging anything that actually could be interepreted by crowdsec 🤯

(!)

Weird indeed.

Have a look at HAProxyMessageHandler, it touches the MDC and issues some logs. Maybe impacted by the formatter change?

protocols/netty/src/main/java/org/apache/james/protocols/netty/ProtocolMDCContextFactory.java
```diff
@@ -96,7 +96,6 @@ static MDCBuilder from(Object o) {
     static MDCBuilder forSession(ProtocolSession protocolSession) {
         return MDCBuilder.create()
             .addToContext(MDCBuilder.SESSION_ID, protocolSession.getSessionID())
-            .addToContext(MDCBuilder.CHARSET, protocolSession.getCharset().displayName())
```
jeantil, line 99: @chibenwa @Arsnael I want to point this out and make sure you have seen it and agree with it. 

If I keep it, I end up with the charset mdc attribute being set twice in a single log, unfortunately this makes both logback-classic and logback-contrib json serializer fail to output the log. instead it outputs a message on STDERR 

I have not been able to reproduce this locally in a test, I tried various things such as changing the builder to make sure to clear the MDC and deduplicate the keys manually before writing them back: without success. 
removing this line systematically fixes the problem.

chibenwa: Seems legit to me.

The charset info seems to be kinda low value...

chibenwa (approved):

PR #2573, opened 2024-12-18
James-4097 Allow disabling same-domain requirement when assigning rights

j-hellenberg:
[James-4097](https://issues.apache.org/jira/browse/JAMES-4097)

Add a configuration switch lifting the same-domain requirement when assigning rights.

This feature is opt-in - the default configuration will still forbid cross-domain access.

### DIFF
--- a/docs/modules/servers/partials/configure/jvm.adoc
+++ b/docs/modules/servers/partials/configure/jvm.adoc
@@ -184,5 +184,5 @@ Optional. Boolean. Defaults to false.
 
 Ex in `jvm.properties`
 ----
-james.rights.users.crossdomain=false
+james.rights.crossdomain.allow=false
 ----
--- a/docs/modules/servers/partials/configure/jvm.adoc
+++ b/docs/modules/servers/partials/configure/jvm.adoc
@@ -173,3 +173,16 @@ james.deduplicating.blobstore.thread.switch.threshold=32768
 # Count of octet from which streams are buffered to files and not to memory
 james.deduplicating.blobstore.file.threshold=10240
 ----
+
+== Allow users to have rights for shares of different domain
+
+Typically, preventing users to obtain rights for shares of another domain is a useful security layer.
+However, in multi-tenancy deployments, this can be useful (for example, students might be given access to a shared mailbox
+residing under the @university.edu domain with their @student.university.edu address).
+
+Optional. Boolean. Defaults to false.
+
+Ex in `jvm.properties`
+----
+james.rights.users.crossdomain=false
+----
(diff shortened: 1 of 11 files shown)


docs/modules/servers/partials/configure/jvm.adoc
```diff
@@ -172,4 +172,17 @@ james.deduplicating.blobstore.thread.switch.threshold=32768
...
+
+Typically, preventing users to obtain rights for shares of another domain is a useful security layer.
+However, in multi-tenancy deployments, this can be useful (for example, students might be given access to a shared mailbox
+residing under the @university.edu domain with their @student.university.edu address).
+
+Optional. Boolean. Defaults to false.
+
+Ex in `jvm.properties`
+----
+james.rights.users.crossdomain=false
```
chibenwa, line 187: How about naming this

```
james.rights.crossdomain.allow=false
```

?

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreRightManager.java
```diff
@@ -74,6 +74,10 @@ public StoreRightManager(MailboxSessionMapperFactory mailboxSessionMapperFactory
         this.eventBus = eventBus;
     }
 
+    private boolean isCrossDomainAccessAllowed() {
+        return Boolean.parseBoolean(System.getProperty("james.rights.users.crossdomain", "false"));
```
chibenwa, line 78: Make it a constant: looking this up at class loading is enough

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreRightManager.java
```diff
@@ -185,7 +189,7 @@ public void applyRightsCommand(MailboxPath mailboxPath, ACLCommand mailboxACLCom
     @Override
     public Mono<Void> applyRightsCommandReactive(MailboxPath mailboxPath, ACLCommand mailboxACLCommand, MailboxSession session) {
         return Mono.just(mailboxSessionMapperFactory.getMailboxMapper(session))
-            .doOnNext(Throwing.consumer(mapper -> assertSharesBelongsToUserDomain(mailboxPath.getUser(), mailboxACLCommand)))
+            .doOnNext(Throwing.consumer(mapper -> assertUserHasAccessToShareDomains(mailboxPath.getUser(), mailboxACLCommand)))
```
chibenwa, line 192:
```suggestion
            .doOnNext(Throwing.consumer(mapper -> assertUserHasAccessToShareeDomains(mailboxPath.getUser(), mailboxACLCommand)))
```

mailbox/store/src/main/java/org/apache/james/mailbox/store/StoreRightManager.java
chibenwa: Ah ok I got it, that's needed because of the tests!

My suggestion is to allow the constant to be package local mutable annotated with `@VisibleForTesting` so that overloading that in tests become easy while keeping the property lookup at classloading time.

WDYT?

chibenwa (commented):
Looks overall very good indeed ;-)

PR #2583, opened 2025-01-06
JAMES-4100 Improve Search Snippet display

hungphan227:


mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/search/OpenSearchSearchHighlighter.java
```diff
@@ -90,8 +93,13 @@ private SearchSnippet buildSearchSnippet(Hit<ObjectNode> searchResult) {
...
-            .map(List::getFirst);
+            .map(List::getFirst)
+            .map(this::shortenGreaterThanCharacters);
 
         return new SearchSnippet(messageId, highlightedSubject, highlightedTextBody);
     }
+
+    private String shortenGreaterThanCharacters(String text) {
+        return GREATER_THAN_PATTERN.matcher(text).replaceAll(">");
+    }
```
chibenwa, line 104: Could we get a shot at a non regexp implementation of this please?

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/search/OpenSearchSearchHighlighter.java
```diff
@@ -100,6 +97,21 @@ private SearchSnippet buildSearchSnippet(Hit<ObjectNode> searchResult) {
     }
 
     private String shortenGreaterThanCharacters(String text) {
-        return GREATER_THAN_PATTERN.matcher(text).replaceAll(">");
+        StringBuilder result = new StringBuilder();
+        boolean previousIsTarget = false;
+
+        for (char c : text.toCharArray()) {
```
chibenwa, line 103: I have doubts about char iteration method.

Maybe we could bench with JMH this version and the REGEXP one?

Arsnael (commented):
Read it, nothing to add

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/search/OpenSearchSearchHighlighter.java
hungphan227: Tested both ways with JMH in an isolated test (run method shortenGreaterThanCharacters in a loop with 10000 iterations). Both have the same results (around 3ms)

quantranhong1999 (approved):

quantranhong1999:
Lucene tests failed: https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2583/1/testReport/

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/search/OpenSearchSearchHighlighter.java
```diff
@@ -90,8 +90,28 @@ private SearchSnippet buildSearchSnippet(Hit<ObjectNode> searchResult) {
         Optional<String> highlightedTextBody = Optional.ofNullable(highlightHit.get(JsonMessageConstants.TEXT_BODY))
             .or(() -> Optional.ofNullable(highlightHit.get(JsonMessageConstants.HTML_BODY)))
             .or(() -> Optional.ofNullable(highlightHit.get(ATTACHMENT_TEXT_CONTENT_FIELD)))
-            .map(List::getFirst);
+            .map(List::getFirst)
+            .map(this::shortenGreaterThanCharacters);
```
vttranlina, line 94: Rename it to `prettyHighlightedText` or `foldHighlightedText` is better?

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/search/OpenSearchSearchHighlighter.java
hungphan227: Should I use regex?

chibenwa: I am a fan of neither of both approaches to be fairly honnest..

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2583/1/testReport/ 

Related I believe

chibenwa:
Have you tried

StringUtils.remove(s, '>);

Or

CharMatcher.is('>').removeFrom(s)

?

hungphan227:
> Have you tried
> 
> StringUtils.remove(s, '>);
> 
> Or
> 
> CharMatcher.is('>').removeFrom(s)
> 
> ?

Don't you want to keep at least one character after removing?

chibenwa:
I do not know: how would it look like in the end user display?

hungphan227:
> I do not know: how would it look like in the end user display?

Example:
Input: >>>>>>>>>> append contentA to >>> inbox >>>>>>
Result: > append <mark>contentA</mark> to > inbox >

(<mark>contentA</mark> is the highlighted part)

chibenwa:
I wonder if `append contentA to INBOX` wouldn't be better...

hungphan227:
> I wonder if `append contentA to INBOX` wouldn't be better...

The solution for this is not just `StringUtils.remove(s, '>);`. We should not remove > from <mark> tag of search snippet.

Arsnael:
@hungphan227 Status of this?

Tests still not fixed I believe => https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2583/1/testReport/

hungphan227:
Actually I don't know what I should do.

Either removing all `>` or keeping just one `>`, there is no way to make it simple like `StringUtils.remove(s, '>);`. I know only two ways: either using regex or a loop to iterate through the string.

Therefore, I suggest closing this ticket and leave things as before. Anyway removing > is not really important.

Arsnael:
> Anyway removing > is not really important.

An end user might disagree with you on that I think :) 

Honestly I get that removing entirely the trailings `>` is tricky and might require a non efficient process as you need to be careful with initial `>` potentially present in the message (and you don't want to break the <mark> tags)

I would be ok with the initial solution, aka folding `>>>>>` into `>`  but need to fix current breaking related tests.

@chibenwa what are your thoughts here?

chibenwa:
Ok

mailbox/api/src/main/java/org/apache/james/mailbox/SearchUtil.java
```diff
@@ -0,0 +1,41 @@
...
+                    previousIsTarget = true;
+                }
+            } else {
+                result.append(c);
+                previousIsTarget = false;
+            }
+        }
+
+        return result.toString();
+    }
```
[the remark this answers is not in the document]
hungphan227: I have no idea. I don't see such thing in the image you posted https://github.com/linagora/james-project/issues/5340#issue-2683469911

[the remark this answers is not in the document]
chibenwa: The issue clearly leaves margin for decision to the author and suggest `>>>` could be removed.

Please let's try to follow `Maybe we should shorten if, and only if previous character is \n ? And completly remove > isf previous char is \n`. Ok?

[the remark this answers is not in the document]
hungphan227: Ok. What if there are space characters between \n and > like this?

```
abc def
        >> ghi klm
```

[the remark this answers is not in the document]
chibenwa: Generally `>>` quotes are at the begining of lines I presume.

If in doubt then take some sample email and run this locally to see how it looks.

Arsnael (commented):
LGTM but did you try to run locally related tests though? Would be surprised that everything is green :)

hungphan227:
> LGTM but did you try to run locally related tests though? Would be surprised that everything is green :)

Fixed inmailbox/lucene/src/main/java/org/apache/james/mailbox/lucene/search/LuceneSearchHighlighter.java

Sth wrong with CI. Need to restart CI

Arsnael:
> Sth wrong with CI. Need to restart CI

=> https://github.com/apache/james-project/pull/2599 is what's wrong with the CI :)

chibenwa:
Can we have a look at how it looks like with the frontend?

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/json/IndexableMessage.java
```diff
@@ -135,8 +135,8 @@ private Mono<IndexableMessage> instantiateIndexedMessage() throws IOException, M
                 .asMimePart(textExtractor)
                 .map(parsingResult -> {
 
-                    Optional<String> bodyText = parsingResult.locateFirstTextBody();
-                    Optional<String> bodyHtml = parsingResult.locateFirstHtmlBody();
+                    Optional<String> bodyText = parsingResult.locateFirstTextBody().map(SearchUtil::removeGreaterThanCharacters);
+                    Optional<String> bodyHtml = parsingResult.locateFirstHtmlBody().map(SearchUtil::removeGreaterThanCharacters);
```
chibenwa, line 139: Do this make sense on the HTML body which uses `<quote>previous message</quote>` ?

mailbox/store/src/main/java/org/apache/james/mailbox/store/search/SearchUtil.java
```diff
@@ -473,5 +473,22 @@ public boolean test(MessageId input) {
         };
     }
 
+    public static String removeGreaterThanCharacters(String text) {
```
chibenwa, line 476: Please indicate in the name that it is at the beginning of a line.

mailbox/api/src/test/java/org/apache/james/mailbox/searchhighligt/SearchHighLighterContract.java
```diff
@@ -532,4 +532,43 @@ default void shouldHighLightBodyWhenHTMLBodyMatched() throws Exception {
...
+            .build();
+
+        // Then highlightSearch should return the SearchSnippet with the highlightedBody containing the word (contentA)
+        List<SearchSnippet> searchSnippets = Flux.from(testee().highlightSearch(List.of(m1.getMessageId(), m2.getMessageId()), multiMailboxSearch, session))
+            .collectList()
+            .block();
+        assertThat(searchSnippets).hasSize(1);
+        assertSoftly(softly -> {
+            softly.assertThat(searchSnippets.getFirst().messageId()).isEqualTo(m1.getMessageId());
+            softly.assertThat(searchSnippets.getFirst().highlightedBody().get()).isEqualTo("Start \n append <mark>contentA</mark> to > inbox \n End");
```
chibenwa, line 571: shouldn't this lone `>` be html encoded?

Otherwise text plain stuff may be leveraged for HTML injections...

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/json/IndexableMessage.java
hungphan227: I still don't get the problem here

chibenwa: Take this message:

image

The corresponding HTML body is:

image

Which do not quote with `>...` but with `<blockquote>...</blockquote>` so removing `>>>` at the beginning of line for `text/html` is indeed a non-sense with well-formatted non testing data.

Can you provide us with real world data where removing `>>>` for HTML body makes sense?

mailbox/api/src/test/java/org/apache/james/mailbox/searchhighligt/SearchHighLighterContract.java
hungphan227: We can create another ticket for this issue

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/json/IndexableMessage.java
hungphan227: Ok I don't know it does not happen in html case. Will remove

mailbox/api/src/test/java/org/apache/james/mailbox/searchhighligt/SearchHighLighterContract.java
hungphan227: sr did not know it has been done. Test fixed.

hungphan227:
`03:21:52,863 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-remote-resources-plugin:1.6.0:process (default) on project apache-mailet-ai: Execution default of goal org.apache.maven.plugins:maven-remote-resources-plugin:1.6.0:process failed: Could not acquire lock(s)`

@Arsnael could you help me restart?

chibenwa (approved):

Arsnael (approved):

chibenwa (approved):

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2583/8/testReport/

Related

chibenwa (approved):

PR #2584, opened 2025-01-06
[FIX] Cleanup EmailQueryView when populating it.

chibenwa:
Current task only creates new entry and never remove existing entries which fails at solving some classes of inconsistencies.


quantranhong1999 (approved):

Arsnael (approved):

Arsnael:
CI not triggered?

PR #2586, opened 2025-01-07
Add auditrail to imap and jmap delete mailbox

hungphan227:


protocols/imap/src/main/java/org/apache/james/imap/processor/DeleteProcessor.java
```diff
@@ -68,6 +72,7 @@ protected Mono<Void> processRequestReactive(DeleteRequest request, ImapSession s
             .then(unsolicitedResponses(session, responder, false))
             .then(Mono.fromRunnable(() -> okComplete(request, responder)))
             .then()
+            .doOnSuccess(any -> auditTrail(session, selected))
```
Arsnael, line 75: Maybe should log with MDC context here too?

chibenwa (approved):

protocols/imap/src/main/java/org/apache/james/imap/processor/DeleteProcessor.java
hungphan227: what do you mean?

Arsnael: logAsMono?

Arsnael (approved):

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/MailboxSetDeletePerformer.scala
```diff
@@ -120,6 +124,16 @@ class MailboxSetDeletePerformer @Inject()(mailboxManager: MailboxManager,
                   .`then`())
             }
           }))
+      .doOnSuccess(_ => auditTrail(mailboxSession, id))
+
+  private def auditTrail(mailboxSession: MailboxSession, id: MailboxId): Unit =
+    ReactorUtils.logAsMono(() => AuditTrail.entry
+      .username(() => mailboxSession.getUser.asString())
+      .protocol("JMAP")
+      .action("Mailbox/set delete")
```
quantranhong1999, line 133: destroy, not delete

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/MailboxSetDeletePerformer.scala
```diff
@@ -120,6 +124,16 @@ class MailboxSetDeletePerformer @Inject()(mailboxManager: MailboxManager,
...
+      .parameters(() => ImmutableMap.of("loggedInUser", mailboxSession.getLoggedInUser.toScala.map(_.asString()).getOrElse(""),
+        "mailboxId", id.serialize()))
```
quantranhong1999, line 135: IMO we should audit trail by mailboxIds (which mailboxIds for Mailbox/set destroy that client really requests to the server). Doing this would make the impression that the client always requests Mailbox/set destroy one by one mailbox.

See audit trail for Email/set destroy.

quantranhong1999 (approved):

Arsnael (approved):

PR #2600, opened 2025-01-15
JAMES-4092 Update webadmin to filter content from mail repository

hungphan227:
Will update doc later

### DIFF
--- a/docs/modules/servers/partials/operate/webadmin.adoc
+++ b/docs/modules/servers/partials/operate/webadmin.adoc
@@ -3612,15 +3612,32 @@ The answer will contains all mailKey contained in that repository.
 Note that this can be used to read mail details.
 
 You can pass additional URL parameters to this call in order to limit
-the output: - A limit: no more elements than the specified limit will be
+the output:
+
+- limit: no more elements than the specified limit will be
 returned. This needs to be strictly positive. If no value is specified,
-no limit will be applied. - An offset: allow to skip elements. This
-needs to be positive. Default value is zero.
+no limit will be applied.
+- offset: allow to skip elements. This needs to be positive. Default value is zero.
 
 Example:
 
 ....
 curl -XGET 'http://ip:port/mailRepositories/var%2Fmail%2Ferror%2F/mails?limit=100&offset=500'
+....
+
+You can also pass the following additional URL parameters to filter results:
+
+- updatedBefore: filter mails by mail last updated. For example, if the value is `2d` and the current time is `21-12-2024 13:00:00`, the condition would be: last updated < 19-12-2024 13:00:00 (currentTime - `2d`). Some other value samples: `2d`, `2 days`, `2h`, `2 hours`.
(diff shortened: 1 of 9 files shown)


server/mailrepository/mailrepository-api/src/main/java/org/apache/james/mailrepository/api/MailRepository.java
```diff
@@ -59,6 +163,18 @@ default Publisher<Long> sizeReactive() {
      */
     Iterator<MailKey> list() throws MessagingException;
 
+    default Iterator<MailKey> list(List<Condition> conditions) throws MessagingException {
```
chibenwa, line 166: We would be more flexible by creating a `And` condition and passing a single condition here?

hungphan227:
> Can we please propose a Cassandra implementation that do not need to read object store content?

added

hungphan227:
doc added

chibenwa (commented):
We miss documentation

hungphan227:
> We miss documentation

added

docs/modules/servers/partials/operate/webadmin.adoc
```diff
@@ -3612,17 +3612,34 @@ The answer will contains all mailKey contained in that repository.
...
 You can pass additional URL parameters to this call in order to limit
-the output: - A limit: no more elements than the specified limit will be
+the output:
+
+- limit: no more elements than the specified limit will be
 returned. This needs to be strictly positive. If no value is specified,
-no limit will be applied. - An offset: allow to skip elements. This
-needs to be positive. Default value is zero.
+no limit will be applied.
+- offset: allow to skip elements. This needs to be positive. Default value is zero.
```
chibenwa, line 3620: Do we have tests combining limit, offset and filters?

We need to be sure the limit and offset are applied after the filters in order to provide reliable paging.

server/protocols/webadmin/webadmin-mailrepository/src/test/java/org/apache/james/webadmin/routes/MailRepositoriesRoutesTest.java
```diff
@@ -511,6 +512,290 @@ void zeroLimitShouldNotBeValid() {
...
+            .build());
+
+        given()
+            .param("sender", "*@domain.com")
+            .when()
+            .get(MY_REPO_MAILS)
+            .then()
+            .statusCode(HttpStatus.OK_200)
+            .body("", hasSize(2))
+            .body("", containsInAnyOrder("name1", "name2"));
```
chibenwa, line 620:
```suggestion
        given()
            .param("sender", "*@domain.com")
        .when()
            .get(MY_REPO_MAILS)
        .then()
            .statusCode(HttpStatus.OK_200)
            .body("", hasSize(2))
            .body("", containsInAnyOrder("name1", "name2"));
```

docs/modules/servers/partials/operate/webadmin.adoc
hungphan227: added

chibenwa (approved):

PR #2605, opened 2025-01-16
[FIX] JMAP AuditTrail logging was not triggered

quantranhong1999:


server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/EmailSetDeletePerformer.scala
```diff
@@ -119,5 +118,6 @@ class EmailSetDeletePerformer @Inject()(messageIdManager: MessageIdManager,
             .map(_.asString())
             .getOrElse("")))
         .log("Mails deleted."))
+        .subscribe()
```
chibenwa, line 121: Please let's not break the reactive chain. Return a SMono here.

hungphan227 (approved):

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/method/EmailSetDeletePerformer.scala
hungphan227: It is in doOnSuccess

quantranhong1999:
Hang on as the JMAP AuditTrail logs are back but the https://github.com/apache/james-project/pull/2564 goal seems still not accomplished.

Arsnael (approved):

Arsnael:
@chibenwa does the latest fixup from @quantranhong1999 answers your concern?

vttranlina (approved):

chibenwa (approved):
Yes

PR #2612, opened 2025-01-29
[CHANGELOG] Update the changelog

chibenwa:


chibenwa:
Happy to handle fixups if needed.

I'll merge this so that I can refer to it into voting threads.

PR #2617, opened 2025-02-02
JAMES-4108: James stuck in authentication loop after successful XOAUTH2 authentication

felixauringer:
I only added some test cases here to show that there is a problem. The commented line would fix those tests but I am not 100% sure that it does not have other side effects.

This also includes a very small fix (rename) for plain authentication which currently asks for a username although the correct message is present in the same file. I can also put this in another MR if you want.


chibenwa:
Thanks for the report!

Sorry for missing this backport.

See https://github.com/apache/james-project/pull/2618 instead.

PR #2631, opened 2025-02-06
[ENHANCEMENT] Allow plug custom indexers into OpenSearchListeningMessageSearchIndex

quantranhong1999:
During the listener process, we need to re-download the all message in order to...

- Index it in search engine
- Pre-compute the JMAP preview

That would cause x2 network traffic and x2 S3 bill.

We can download the message just once doing both the job to optimize the performance and cost.

### DIFF
--- a/docs/modules/servers/pages/distributed/configure/jvm.adoc
+++ b/docs/modules/servers/pages/distributed/configure/jvm.adoc
@@ -3,3 +3,16 @@
 
 :sample-configuration-prefix-url: https://github.com/apache/james-project/blob/master/server/apps/distributed-app/sample-configuration
 include::partial$configure/jvm.adoc[]
+
+== Disable OpenSearchListeningMessageSearchIndex fetch full message content
+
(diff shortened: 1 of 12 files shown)


chibenwa:
See https://github.com/chibenwa/james-project/tree/poc-quan

Do this in tmail : ComputePreviewMessageIndexer

We will backport if conclusive

quantranhong1999:
> Maybe this makes things significantly easier. I'll let you fix compil warnings and adapt.

Thanks to your POC, I have reworked this to only allow adding custom indexers into `OpenSearchListeningMessageSearchIndex`.

For the rest, let's do it in TMail first and backport to James later if conclusive IMO.

Arsnael (approved):

Arsnael:
https://ci-builds.apache.org/job/james/job/ApacheJames/job/PR-2631/2/ 

related?

quantranhong1999:
Another review maybe?

PR #2642, opened 2025-02-18
JAMES-3788 proxyFirst should be true by default

Arsnael:


docs/modules/servers/partials/configure/imap.adoc
```diff
@@ -121,7 +121,7 @@ using HAProxy's proxy protocol.
 | proxyFirst
 | Whether proxy frames should be handled before SSL handshakes. This allows setting either the loadbalancer in TCP mode
 (so transparent for SSL then Proxy frames needs to be handled first) or set up SSL termination between proxy and server
-(more suited for some cloud vendors). Defaults to true (TCP transparent).
+(more suited for some cloud vendors). Defaults to false.
```
quantranhong1999, line 124: Should the default configuration code be true instead?

chibenwa: Yes this would avoid a breaking change...

Fixups welcome.

Arsnael: It's false in the code. One or the other honestly. But faulty doc made me waste a morning trying to find the cause of why imap was not working anymore when deploying a new version with this change

quantranhong1999: We reviewed it too. It is a human mistake. Let's avoid a breaking change IMO.

chibenwa (approved):

quantranhong1999 (approved):

PR #2656, opened 2025-02-26
Refactor some event bus handlers so that they do not tightly bind to RabbitMQEventBus

hungphan227:
To support new event bus extension in TMail (refer to https://github.com/linagora/tmail-backend/issues/1556)

### DIFF
--- a/event-bus/api/src/main/java/org/apache/james/events/EventBus.java
+++ b/event-bus/api/src/main/java/org/apache/james/events/EventBus.java
@@ -101,7 +101,12 @@ public interface EventBus {
         return ImmutableList.of();
     }
 
-    void start();
+    default void start() {
+    }
 
-    void restart();
+    default void restart() {
+    }
+
+    default void stop() {
+    }
 }
--- a/event-bus/api/src/main/java/org/apache/james/events/EventBus.java
+++ b/event-bus/api/src/main/java/org/apache/james/events/EventBus.java
@@ -100,4 +100,8 @@ public interface EventBus {
     default Collection<Group> listRegisteredGroups() {
         return ImmutableList.of();
     }
+
+    void start();
+
+    void restart();
 }
--- a/event-bus/distributed/src/main/java/org/apache/james/events/GroupRegistrationHandlerGroup.java
+++ b/event-bus/distributed/src/main/java/org/apache/james/events/GroupRegistrationHandlerGroup.java
@@ -1,0 +1,23 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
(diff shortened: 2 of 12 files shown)


event-bus/api/src/main/java/org/apache/james/events/EventBus.java
```diff
@@ -100,4 +100,8 @@ default Registration register(EventListener.ReactiveGroupEventListener groupList
     default Collection<Group> listRegisteredGroups() {
         return ImmutableList.of();
     }
+
+    void start();
+
+    void restart();
```
chibenwa, line 106:
```suggestion
    default void start() {
    
    }

    default void restart() {
    
    }
```

And drop the other implems?

Arsnael (commented):
Read it nothing to add

Arsnael (approved):

event-bus/distributed/src/main/java/org/apache/james/events/GroupRegistrationHandlerGroup.java
```diff
@@ -0,0 +1,23 @@
...
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.events;
+
+public class GroupRegistrationHandlerGroup extends Group {
```
Arsnael, line 22: Was that really necessary?

quantranhong1999: I reviewed our RabbitMQ queues usage with the latest release, and this is a **breaking change** in the queues name...

Before:
image
`mailboxEvent-workQueue-org.apache.james.events.GroupRegistrationHandler$GroupRegistrationHandlerGroup`

After update:
image

`mailboxEvent-workQueue-org.apache.james.events.GroupRegistrationHandlerGroup`

Impact:
- It is not just a rolling update anymore as there is no consumer for the old queue, with 52 messages waiting to be consumed forever (unless you move it manually to the new queue!)
- Redeliver dead letters for the previous events could fail because the group is not matched anymore?

IMO we should have not made this breaking change, and we should revert this class path change @hungphan227.

quantranhong1999: > It will be used by another class in TMail

Adding `public` before `class GroupRegistrationHandler` was enough IMO...

chibenwa: Questions here: 

 - is this a Twake mail specific regression or a James one?
 - Of course we indeed should avoid accidental braking changes and should revert back to the previous one.

PR #2660, opened 2025-02-28
Experiment Apache Kvrocks to back our tests

quantranhong1999:
Experiment with Apache Kvrocks (as a Redis drop-in replacement) to see if it plays well with our test suite.

First step: standalone Kvrocks seems to play well with tests on my local.

We need to check Sentinel mode and likely Cluster mode also.


chibenwa:
Maybe we could raise the topic on the mailing list?

chibenwa:
<3

Arsnael:
Shall we adapt this to have kvrocks tests alongside redis in the code base?

chibenwa:
> Shall we adapt this to have kvrocks tests alongside redis in the code base?

+1 for having the tests

However I'm not sure we need to run 100% of them. Smoke tests maybe? 
Or even with a specific profile?
Or even disabled so that DEVs can run it if need be...

quantranhong1999:
I have reworked this to introduce Kvrock test extensions and simple tests that use them.

Arsnael (approved):

Arsnael (approved):

vttranlina (approved):
Looking at the code, its API seems quite similar to Redis. Hopefully, it's a good candidate for RSPAMD

PR #2664, opened 2025-03-04
Prevent FluentDataProbe#addDomain from throwing an exception in the event of an existing domain

HoussemNasri:


server/data/data-api/src/main/java/org/apache/james/probe/DataProbe.java
```diff
@@ -49,6 +52,15 @@ public FluentDataProbe addDomain(String domain) throws Exception {
             dataProbe.addDomain(domain);
             return this;
         }
+
+        public FluentDataProbe addDomainIfNotExists(String domain) throws Exception {
```
chibenwa, line 56: I would rather make this the default behaviour. WDYT?

HoussemNasri: Agree.

chibenwa (approved):

Arsnael (approved):

PR #2665, opened 2025-03-05
Extension for deleting redundant message fast view data

hungphan227:


chibenwa:
What don't you understand in "I do not want a task" and "I want this to run as an external JAR" ?

chibenwa:
KPI < 300 lines of code diff. 

More and I do not review

PR #2666, opened 2025-03-06
Update cassandra message fast view projection and add message service

hungphan227:
To support message fast view cleanup extension


chibenwa (changes requested):
Be hereby reminded we want to run this **as an extension jar**.

Thus we can't modify existing class (as we would need to change the docker image)
We can only add new classes.

Please  also let's follow the bloom filter approach.

We obviously do not need something generic. Something Cassandra specific will be enough.

PR #2667, opened 2025-03-06
[NO_MERGE] Extension for deleting redundant message fast view data

hungphan227:
require pr https://github.com/apache/james-project/pull/2666

### DIFF
--- a/examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/CassandraMessageDAO.java
+++ b/examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/CassandraMessageDAO.java
@@ -1,0 +1,60 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.messagefastview.cleanup;
+
+import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.selectFrom;
+import static org.apache.james.mailbox.cassandra.table.CassandraMessageIds.MESSAGE_ID;
+import static org.apache.james.mailbox.cassandra.table.CassandraMessageV3Table.TABLE_NAME;
+
+import jakarta.inject.Inject;
+
+import org.apache.james.backends.cassandra.utils.CassandraAsyncExecutor;
+import org.apache.james.mailbox.cassandra.ids.CassandraMessageId;
+import org.apache.james.mailbox.model.MessageId;
+
+import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.cql.PreparedStatement;
+import com.datastax.oss.driver.api.core.cql.SimpleStatement;
+import com.datastax.oss.driver.api.core.type.codec.TypeCodecs;
+
+import reactor.core.publisher.Mono;
+
+public class CassandraMessageDAO {
+    private final CassandraAsyncExecutor cassandraAsyncExecutor;
+    private final PreparedStatement select;
+
+    @Inject
+    public CassandraMessageDAO(CqlSession session) {
+        this.cassandraAsyncExecutor = new CassandraAsyncExecutor(session);
+
+        SimpleStatement simpleStatement = selectFrom(TABLE_NAME)
+            .column(MESSAGE_ID)
+            .all()
+            .build();
+        simpleStatement.setPageSize(1000);
+        this.select = session.prepare(simpleStatement);
+    }
+
+    public Mono<MessageId> getAllMessageIds() {
+        return cassandraAsyncExecutor.executeSingleRow(select.bind())
+            .map(row -> row.get(MESSAGE_ID, TypeCodecs.UUID))
+            .map(CassandraMessageId.Factory::of);
+    }
+}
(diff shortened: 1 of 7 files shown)


examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/MessageFastViewCleanupService.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Inject
+    public MessageFastViewCleanupService(MessageFastViewProjection messageFastViewProjection, MessageService messageService) {
+        this.messageFastViewProjection = messageFastViewProjection;
+        this.messageService = messageService;
+    }
+
+    public Mono<Void> cleanup() {
+        return Flux.from(messageFastViewProjection.getAllMessageIds())
+            .flatMap(messageId -> exist(messageId)
```
chibenwa, line 48: If I run this in prod I trigger 20.000.000 primary key reads at an unsustainable pace that will bring my Cassandra infrastructure to its knees.

Can't we think of an approach revolving around the use of bloom filters in order to do 2 sequential table scans and mitigate the resource cost of the operation?

examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/MessageFastViewCleanupRoute.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    @Override
+    public void define(Service service) {
+        service.post(getBasePath(), this::cleanup);
+    }
+
+    private String cleanup(Request request, Response response) {
+        messageFastViewCleanupService.cleanup().block();
+        return Responses.returnNoContent(response);
```
chibenwa, line 51: Could be nice to return the count of messages actually deleted, so we can have a RETEX on the usefulness of the feedback.

examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/MessageFastViewCleanupService.java
hungphan227: > If I run this in prod I trigger 20.000.000 primary key reads at an unsustainable pace that will bring my Cassandra infrastructure to its knees.

Cassandra drive has the default page size of pagination. Added setting page size to make sure.

> Can't we think of an approach revolving around the use of bloom filters in order to do 2 sequential table scans and mitigate the resource cost of the operation?

Previously I thought that using bloom filter could let some redundant message ids through. Sorry I forgot that we do not expect all redundant message ids removed. Updated

examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/MessageFastViewCleanupRoute.java
hungphan227: added

examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/MessageFastViewCleanupService.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.common.hash.Funnels;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+
+public class MessageFastViewCleanupService {
+    private static final Logger LOGGER = LoggerFactory.getLogger(MessageFastViewCleanupService.class);
+    private static final int MESSAGE_IDS_PER_SECOND = 1000;
+    private static final Funnel<CharSequence> BLOOM_FILTER_FUNNEL = Funnels.stringFunnel(StandardCharsets.US_ASCII);
+    public static final int EXPECTED_BLOOM_FILTER_INSERTIONS = 10000000;
```
chibenwa, line 41: image

https://hur.st/bloomfilter/?n=50000000&p=0.01&m=&k=

I think we can increase a bit the count of elements to 50 M

examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/MessageFastViewCleanupService.java
```diff
@@ -0,0 +1,86 @@
...
+    private final CassandraMessageFastViewDAO messageFastViewDAO;
+
+    @Inject
+    public MessageFastViewCleanupService(CassandraMessageDAO messageDAO, CassandraMessageFastViewDAO messageFastViewDAO) {
+        this.messageDAO = messageDAO;
+        this.messageFastViewDAO = messageFastViewDAO;
+    }
+
+    public Mono<Long> cleanup() {
+        return populatedBloomFilter()
```
chibenwa, line 54: Please generate a random string value  to use as a seed so that the hash of one run is not the hash of the next one.

examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/MessageFastViewCleanupService.java
```diff
@@ -0,0 +1,86 @@
...
+    private Mono<Long> cleanup(BloomFilter<CharSequence> bloomFilter) {
+        AtomicLong totalDeleted = new AtomicLong();
+        return Flux.from(messageFastViewDAO.getAllMessageIds())
+            .filter(messageId -> !bloomFilter.mightContain(messageId.serialize()))
+            .flatMap(messageId -> Mono.from(messageFastViewDAO.delete(messageId))
+                    .doOnSuccess(any -> {
+                        long total = totalDeleted.get();
+                        if (total % 1000 == 0) {
+                            LOGGER.info("Message fast view items deleted: {}", total);
+                        }
```
chibenwa, line 67: IMO log them all? I see little interest in sampling.

hungphan227: > IMO log them all? I see little interest in sampling.

It may be too much log if log all, plus logs might not be displayed in right order. However, I think it should be ok. Updated

chibenwa (approved):
Good to me.

IMO let's not merge this PR.

Please open another PR with "How to run this extension" in upgrade-instructions.md (so pointing to this PR)

Arsnael (commented):
LGTM too

chibenwa (approved):

examples/message-fast-view-cleanup-extension/src/main/java/org/apache/james/messagefastview/cleanup/CassandraMessageDAO.java
```diff
@@ -0,0 +1,60 @@
...
+        SimpleStatement simpleStatement = selectFrom(TABLE_NAME)
+            .column(MESSAGE_ID)
+            .all()
+            .build();
+        simpleStatement.setPageSize(1000);
+        this.select = session.prepare(simpleStatement);
+    }
+
+    public Mono<MessageId> getAllMessageIds() {
+        return cassandraAsyncExecutor.executeSingleRow(select.bind())
```
chibenwa, line 56: FUCK

WHY SINGLE ROW!!!!

hungphan227: I am very sorry. Do not know why I forget to update this.

PR #2669, opened 2025-03-07
JAMES-4072 Add SSL configuration for redis cluster topology

Arsnael:


chibenwa (approved):

hungphan227:
May need a test

Arsnael:
> May need a test

No, cf other draft PR on the refactoring. We concluded with @chibenwa that adding tests on all redis topologies was not really necessary and was just adding monstrous unnecessary complexity to the test suite. So just keeping one test for it (redis standalone) should be enough, we can accept for the rest

hungphan227:
> > May need a test
> 
> No, cf other draft PR on the refactoring. We concluded with @chibenwa that adding tests on all redis topologies was not really necessary and was just adding monstrous unnecessary complexity to the test suite. So just keeping one test for it (redis standalone) should be enough, we can accept for the rest

I think at least do manual test to make sure. Ssl for different topology may require different redis client code and setup

quantranhong1999 (approved):

PR #2672, opened 2025-03-13
[UPGRADE] JSPF and JDKIM to latest release versions

Arsnael:


quantranhong1999 (approved):

chibenwa (approved):

Arsnael:
DKIM tests failing : 

org/apache/james/jdkim/codec/binary/BinaryCodec

When looking at jdkim code, if i understand correctly we shade the commons-codec lib to bundle it in the final jar on an other path  `org.apache.james.dkim.codec` by sizing it down to base64 related code only and excluding the rest. However, after upgrading the commons-codec lib to 1.17.1, it seems it changes things a bit around and we need BinaryCodec now for base64 encoding code. But BinaryCodec is excluded from this shade:

 =>https://github.com/apache/james-jdkim/blob/master/main/pom.xml#L117

By removing that exclude line, recompiling jdkim locally and applying it to james, tests are green.

I see few different approaches here:

- patch jdkim lib by removing that exclude on BinaryCodec
- shade the all commons-codec lib without excluding things to avoid this kind of issue in the future
- remove the shade and have commons-codec aligned in projects using jdkim like james

I fear we need to go for an other release of jdkim though to nicely fix this.

Will open a PR for jspf upgrade alone at least but I'm opened to other suggestions though if any

Arsnael:
Let's try getting jspf upgrade at least first => https://github.com/apache/james-project/pull/2675

chibenwa:
> shade the all commons-codec lib without excluding things to avoid this kind of issue in the future

JDKIM is used by other projects than James and forcing everybody to align on our commons-codec version is not reasonnable. THus I am more in favor of shading as that given lib is very small.

Arsnael:
=> https://github.com/apache/james-jdkim/pull/19

Arsnael:
Jspf has been successfully upgraded. Jdkim will need a new release. 

Closing this.

PR #2679, opened 2025-03-17
[JAMES-4119] Core data JPA to PG migration tool

jeantil:
Hello Everyone, 

I propose to add this migration application which copies all the core data supported by JPA into the corresponding Postgres store. 

I built it because I intend to migrate `scaling-pulsar-smtp` from JPA to Postgres and wanted to offer a reasonably easy migration path if there are any users other than me.

The migration tool has a single integration test that : stores some data in JPA, executes the migration then demonstrates it can be read back from the PG stack. 

Doing so I encountered a few issues/misses from the PG implementation which I fixed in preliminary devscout commits.

### DIFF
--- a/server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
+++ b/server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
@@ -1,0 +1,504 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james;
+
+import static org.apache.james.modules.blobstore.BlobStoreModulesChooser.chooseBlobStoreDAOModule;
+import static org.apache.james.modules.blobstore.BlobStoreModulesChooser.chooseEncryptionModule;
+import static org.apache.james.modules.blobstore.BlobStoreModulesChooser.chooseStoragePolicyModule;
+
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+
+import jakarta.inject.Inject;
+import jakarta.persistence.EntityManagerFactory;
+
+import org.apache.james.backends.postgres.PostgresModule;
+import org.apache.james.core.Username;
+import org.apache.james.domainlist.api.DomainListException;
+import org.apache.james.domainlist.jpa.JPADomainList;
+import org.apache.james.domainlist.lib.DomainListConfiguration;
+import org.apache.james.domainlist.postgres.PostgresDomainList;
+import org.apache.james.droplists.jpa.JPADropList;
+import org.apache.james.droplists.postgres.PostgresDropList;
+import org.apache.james.filesystem.api.FileSystem;
+import org.apache.james.mailrepository.api.MailKey;
+import org.apache.james.mailrepository.api.MailRepository;
+import org.apache.james.mailrepository.api.MailRepositoryUrl;
+import org.apache.james.mailrepository.api.MailRepositoryUrlStore;
+import org.apache.james.mailrepository.jpa.JPAMailRepositoryFactory;
+import org.apache.james.mailrepository.jpa.JPAMailRepositoryUrlStore;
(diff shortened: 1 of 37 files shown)


server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
```diff
@@ -0,0 +1,455 @@
...
+                                jpaUsersDAO.getUserByName(username).ifPresent(user -> {
+                                    var jpaUser = (JPAUser) user;
+                                    var pgUser = new DefaultUser(
+                                            username,
+                                            jpaUser.getPasswordHash(),
+                                            jpaUser.getAlgorithm(),
+                                            algorithm
+                                    );
+                                    try {
+                                        postgresUsersDAO.addUser(username, UUID.randomUUID().toString());
```
chibenwa, line 206: Don't we have a way to preserve user password? By outputing the same hash into PG?

server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
```diff
@@ -0,0 +1,455 @@
...
+                        )
+                );
+            } catch (RecipientRewriteTableException e) {
+                throw new RuntimeException(e);
+            }
+            LOGGER.info("RRT migration complete");
+        }
+    }
+
+    static class MailRepositoryUrlMigration {
```
chibenwa, line 261: IMO no need to migrate this as it would be created by the app upon startup.

server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    void start() {
+        injector.getInstance(DomainMigration.class).doMigration();
+        injector.getInstance(UsersMigration.class).doMigration();
+        injector.getInstance(RRTMigration.class).doMigration();
+        injector.getInstance(MailRepositoryUrlMigration.class).doMigration();
+        injector.getInstance(DropListMigration.class).doMigration();
+        injector.getInstance(SieveMigration.class).doMigration();
+    }
```
chibenwa, line 453: :-P Maybe that's me over-engineering but I would have been writing a `MigrationAggregate implementing the `Migration` interface and using a `Migration::and` reduction. I would have had also used a set multi binder to inject the migration steps.

chibenwa (commented):
Looks great to me. Thanks for putting this together.

Not the way I would have ended up taking (I might have favoured webadmin) but definitly valid.

For mailbox content we shall IMO rely on `imapsync`.

server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
jeantil: the migration for users uses 2 steps : 
1. add the user with a random password
2. update the user which restores its password and hashing algorithm 

I did it this way to minimize the impact on the repository APIs

chibenwa (approved):

server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
jeantil: >What really is valuable however is migrating Mail Repositories which do not seems achieved here.
indeed not, I was blinded by my bias of building a tool to migrate scaling-pulsar-smtp which doesn't use the SQL database to store mails only the URLs :D I'll look into adding that, but i must say one would have to be careful with the volume as the migration will effectively copy all the data from one table to the other.

server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
jeantil: I guess that's why you wanted to have an interface.
As far as I can tell this list is exhaustive it also has the benefit of being very explicit on what the tool actually migrates I don't think additional architecture is needed here.

chibenwa: Agreed

server/apps/migration/core-data-jpa-to-pg/src/main/java/org/apache/james/JpaToPgCoreDataMigration.java
jeantil: I pushed a fixup that adds MailRepository migration. The fixup will not work with deduplication. The deduplication strategy [forcibly installs a webmin route](https://github.com/apache/james-project/blob/master/server/container/guice/blob/deduplication-gc/src/main/java/org/apache/james/modules/blobstore/BlobDeduplicationGCModule.java#L67) module which makes absolutely no sense for the migration tool and I want to make progress on Multi-DKIM storage and ARC instead of fighting that feature. 

It does work with passthrough as demonstrated by the test.

chibenwa (approved):

jeantil:
forcepushed to clean up the fixups but also added some documentation and zip packaging

chibenwa (approved):

jeantil:
sorry i messed up this morning and merged the wrong MR as it had green tests  ( the top off the stack instead of this one first ) ,  and benoit had approved it. I didn't realize he had approved it so early in the morning :D
The overall result is the same but if someone wanted to disagree with migrating scaling-pulsar-smtp to the postgres stack let me know I will revert so we can discuss it. 
If you spot issues with these changes, let me know and I will discuss or address them

PR #2682, opened 2025-03-18
[JAMES-4122] renames postgres data defnitions

jeantil:
I think I got them all


quantranhong1999 (approved):

chibenwa (approved):

chibenwa:
Nice refacto, it will lower brain load a lot I believe.

PR #2683, opened 2025-03-19
[JAMES-4123] rename cassandra data definitions

jeantil:
same as for pg, I think I got them all. i'll let CI complete

I did add a note to the CHANGELOG because the cassandra rename could be a breaking change for people who build their own assemblies. ( the PG part was never part of a release so its didn't need the warning)


quantranhong1999 (approved):

Arsnael (approved):

chibenwa:
> I did add a note to the CHANGELOG because the cassandra rename could be a breaking change for people who build their own assemblies. ( the PG part was never part of a release so its didn't need the warning)

+1

Indeed we will have to update Twake mail code accordingly ^^

chibenwa:
Maybe reopen a second PR to try trigger it?

jeantil:
https://github.com/apache/james-project/pull/2684 no luck there either :/

PR #2686, opened 2025-03-21
Revert [chore] upgrades develocity extension

jeantil:
Investigating if the upgrade caused the out of heap errors on the CI


jeantil:
I think the upgrade is the cause for the OOMs on master and the other branches. since I got a green build with the revert I'll retry the upgrade on a branch but I merged the revert for now to unblock the rest

PR #2689, opened 2025-03-27
[JAMES-4119] minor fixes

jeantil:
I ran the migration tool on my "production" this led me to detect a couple issues 
- a null pointer for the blobstoreConfiguration (not sure how I missed that in my local tests :/)
- the sample config suggested using deduplication for the blobstore which is not supported

With the fixes, my production data was migrated and I was able to use the jooq/postgres backed version of scaling-pulsar-smtp to transfer some emails


quantranhong1999 (approved):

Arsnael (approved):

chibenwa (approved):
Nice!

PR #2690, opened 2025-03-28
Redis refactoring: RedisConnectionFactory

Arsnael:
Goal here is to get directly a connection ready to serve depending on the topology. 

For now it's just keeping the original behavior of the code. This is a pre-task refactoring, the sentinel deeper investigation and potential fix will come later in a follow up work 

Margin for more refactorings but there is an other task for that...


chibenwa:
Thanks for this rationalization work.

If we do only use the redis connection for the healthcheck this means that the health check is very distinct from the rest of the source code.

I'd prefer we implement the healthcheck as:

```
PUT healthcheck/{UUID}
DELETE healthcheck/{UUID}
```

And drop alltogether the RedisConnection in favor of using the RedisClient everywhere.

Does this make sense?

vttranlina (approved):

Arsnael:
Hmmm... For now yes only the health check is using the connection. The redis rate limiter is using a lib, `ratelimitj`, that takes the redis client and not the connection. 

I was more thinking of other custom works (like the redis event bus we have on tmail, but could be the case with other actors I believe) that could rely on the connection factory to have directly a connection depending of the redis topology. 

But maybe just returning the client makes sense, as it seems that's how it goes with some libs (like ratelimitj). But it means we need to manage the connections behind then.

But as it's just a ping for the healthcheck I guess we could just have a pretty simple connection, as we do not do any operations on redis.

Well I think I'm gonna look at your proposition on an other PR and we can compare and decide. Would that be ok for you @chibenwa ?

quantranhong1999:
> I'd prefer we implement the healthcheck as:
PUT healthcheck/{UUID}
DELETE healthcheck/{UUID}

I think it is to better detect the READONLY redis issue that the healthcheck failed to detect before (as it currently only read testing).

chibenwa:
> but could be the case with other actors I believe 

1. it is not part of a LTS release
2. this is not an API but inner code
3. there's no known usages

So let's not get our code more complicated because of this.

> But as it's just a ping for the healthcheck I guess we could just have a pretty simple connection, as we do not do any operations on redis.

I prefer write + delete on a sample key, and using a client rahter than PING on top of a connection and getting all the code more complex.

Arsnael:
> I prefer write + delete on a sample key, and using a client rahter than PING on top of a connection and getting all the code more complex.

You still need to do a connection from the client though before being able to write anything on a redis key, right?

Arsnael:
@chibenwa alright after giving it a go I agree that your proposal clearly makes more sense. 
cf https://github.com/apache/james-project/pull/2693

Closing this one.

PR #2692, opened 2025-04-01
[ENHANCEMENT] JMAP MDC: make the original IP header configurable

quantranhong1999:


Arsnael (approved):

chibenwa (commented):
Can we get a commented value for this in `jvm.properties` of the distributed app?

chibenwa (approved):

guimard:
Following HTTP specs, the X-Forwarded-Header may contain [several IP addresses](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For). Thus the header maybe agreed or not depending on the proxy IP address.

Here I don't think we should validate intermediate IP addresses, but:
- read the good one
- validate that current IP address is in the list of agreed proxies
- and then if all is OK, replace source IP by the value given by header

chibenwa:
Agreed that this can be improved.

@quantranhong1999 maybe split on `,` and discard entries equal to the current IP, before reforming the field for the logger?

quantranhong1999:
> @quantranhong1999 maybe split on , and heep only the first one (the outer one) and drop intermediate http LBs IP?

Done.

Arsnael (approved):

vttranlina (approved):

chibenwa (approved):

PR #2693, opened 2025-04-01
Redis refactoring: ease code complexity with RedisClientFactory, RedisHealthCheck and RedisRateLimiter

Arsnael:


backends-common/redis/src/main/java/org/apache/james/backends/redis/RedisHealthCheck.scala
```diff
@@ -62,46 +61,43 @@ class RedisHealthCheck @Inject()(redisConfiguration: RedisConfiguration, redisCl
...
 sealed trait RedisHealthcheckPerform {
   def check(): SMono[Result]
-
-  def close(): Unit
 }
 
-class RedisStandaloneHealthCheckPerform(val redisConfiguration: StandaloneRedisConfiguration,
-                                        val redisClientFactory: RedisClientFactory,
-                                        val healthcheckTimeout: Duration) extends RedisHealthcheckPerform {
+class RedisRegularHealthCheckPerform(val redisClient: RedisClient, redisConfiguration: RedisConfiguration) extends RedisHealthcheckPerform {
```
quantranhong1999: use `regular` to reflect that it is not cluster mode healthcheck seems not clear to me.

Arsnael: Those classes literally disappear in the commit right after to just have only 1 healthcheck

chibenwa (approved):
Looks better to me indeed

quantranhong1999 (approved):

hungphan227 (approved):

PR #2694, opened 2025-04-01
[JAMES-4126] improve GuiceMailetLoader

jeantil:
- allows constructor injection of the MailetConfig for mailets
- allows pre-binding of configuration in GenericMailet

### DIFF
--- a/server/container/guice/mailet/src/main/java/org/apache/james/utils/GuiceMailetLoader.java
+++ b/server/container/guice/mailet/src/main/java/org/apache/james/utils/GuiceMailetLoader.java
@@ -56,9 +56,8 @@ public class GuiceMailetLoader implements MailetLoader {
             Class<Mailet> mailetClass = mailetLoader.locateClass(className);
             MailetConfig mailetConfig = resolveConfiguration(mailetClass, config);
             Mailet result = mailetLoader.withChildModule(
-                            binder -> binder.bind(MailetConfig.class).toInstance(mailetConfig)
-                    )
-                    .instantiate(className);
+                    binder -> binder.bind(MailetConfig.class).toInstance(mailetConfig)
+            ).instantiate(className);
             result.init(resolveConfiguration(result, config));
             return result;
         } catch (Exception e) {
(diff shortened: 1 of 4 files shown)


server/container/guice/mailet/src/main/java/org/apache/james/utils/GuiceMailetLoader.java
```diff
@@ -43,17 +43,22 @@ public class GuiceMailetLoader implements MailetLoader {
...
         try {
             ClassName className = new ClassName(config.getMailetName());
-            Mailet result = genericLoader.<Mailet>withNamingSheme(MAILET_NAMING_SCHEME)
-                .instantiate(className);
+            var mailetLoader = genericLoader.<Mailet>withNamingSheme(MAILET_NAMING_SCHEME);
+            Class<Mailet> mailetClass = mailetLoader.locateClass(className);
+            MailetConfig mailetConfig = resolveConfiguration(mailetClass, config);
+            Mailet result = mailetLoader.withChildModule(
+                            binder -> binder.bind(MailetConfig.class).toInstance(mailetConfig)
+                    )
```
chibenwa, line 60:
```suggestion
            Mailet result = mailetLoader.withChildModule(
                            binder -> binder.bind(MailetConfig.class).toInstance(mailetConfig))
```

(sorry)

jeantil:
The whole series of changes indirectly asks the question of when are we dropping spring :D

chibenwa:
> The whole series of changes indirectly asks the question of when are we dropping spring :D

Indeed.

The sooner the better.

I'd personnaly like we go for 3.9.0 with default being postgres guice, propose JPA guice and deprecate Spring

Then potentially drop it after tagging 3.9.0.

> The mailets can be made spring compatible by keeping config mutable and replicating the init behaviour there.

+1

> A probably better but far more costly option would be to have dedicated factories for every mailet but thats a lot of factory code to write :/

Please no!

Anyway a couple of modern mailets and extensions explicitly requires guice injects. I think it's ok.

server/container/guice/mailet/src/main/java/org/apache/james/utils/GuiceMailetLoader.java
jeantil: I fixed the indentation let me know if it's ok for you

jeantil:
> Maybe we could add a quick test of loading a custom mailet with constructor inject for the mailet config?

added in the latest fixup

server/container/guice/mailet/src/main/java/org/apache/james/utils/GuiceMailetLoader.java
chibenwa: My point is more that I am opinionated on lines starting with brackets but that may not matter much.

chibenwa (approved):

PR #2695, opened 2025-04-01
[JAMES-4127] add default behaviours on Mailet

jeantil:
PR stacked on https://github.com/apache/james-project/pull/2694, please review only JAMES-4127 commits

### DIFF
--- a/mailet/api/src/main/java/org/apache/mailet/Mailet.java
+++ b/mailet/api/src/main/java/org/apache/mailet/Mailet.java
@@ -80,7 +80,8 @@ public interface Mailet {
      *          and initialization parameters
      * @throws MessagingException if an error occurs
      */
-    void init(MailetConfig config) throws MessagingException;
+    default void init(MailetConfig config) throws MessagingException {
+    }
 
     /**
      * Services a mail message.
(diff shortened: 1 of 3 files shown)


mailet/api/src/main/java/org/apache/mailet/Mailet.java
```diff
@@ -127,14 +129,19 @@ public interface Mailet {
...
      * copyright.
+     * <p>
+     * By default, this method returns an empty string. Override this method
+     * to have it return a meaningful value.
      *
      * @return the Mailet information (as a plain text string)
      */
-    String getMailetInfo();
+    default String getMailetInfo() {
+        return "";
```
chibenwa, line 139: Maybe we could default to getClass().getSimpleName()?

(ok that is a behavioural change... So maybe in another PR, but providing sane defaults for this would look good to me. One less thing to care about)

chibenwa (approved):

mailet/api/src/main/java/org/apache/mailet/Mailet.java
jeantil: MailetInfo is not the same as the mailet name, the javadoc says its about 

> author, version and copyright 

but using `SimpleName` is an interesting idea for the next PR ;)

chibenwa: I may have had overlooked this sorry.

jeantil:
added a fixup that covers the mailet added for the test in the first PR

Arsnael (approved):
Needs a rebase but overall LGTM :)

quantranhong1999 (approved):

jeantil:
suqashed the fixups and merged since the build was green

PR #2697, opened 2025-04-01
JAMES-3819 Give tools to enforce RFC-8461 MTA-STS

chibenwa:
(Needed for one of my customer projects and I think this can be relevant for the community as well)

That's for the inbound part that outbound part of mta-sts is obviously missing...

### DIFF
--- a/docs/modules/servers/partials/architecture/implemented-standards.adoc
+++ b/docs/modules/servers/partials/architecture/implemented-standards.adoc
@@ -37,6 +37,7 @@ This page details standards implemented by the {server-name}.
 - link:https://datatracker.ietf.org/doc/rfc6710/[RFC-6710] SMTP Extension for Message Transfer Priorities
 - link:https://datatracker.ietf.org/doc/rfc8689/[RFC-8689] SMTP Require TLS Option
 - link:https://datatracker.ietf.org/doc/html/rfc1893[RFC-1893] Enhanced Mail System Status Codes
+- link:https://datatracker.ietf.org/doc/html/rfc8461[RFC-8461] SMTP MTA Strict Transport Security (MTA-STS)
 
 == LMTP
 
--- a/docs/modules/servers/partials/configure/mta-sts.adoc
+++ b/docs/modules/servers/partials/configure/mta-sts.adoc
@@ -31,7 +31,7 @@ Sample set up:
 
 === Advertizing MTA-STS support
 
-Configuring the MDA is only part of the story. Remains telling the world about it.
+Configuring the MTA is only part of the story. Remains telling the world about it.
 
 First is needed a DNS record for the MTA-STS policy:
 
--- a/docs/modules/servers/partials/configure/mta-sts.adoc
+++ b/docs/modules/servers/partials/configure/mta-sts.adoc
@@ -1,0 +1,67 @@
+link:https://datatracker.ietf.org/doc/html/rfc8461[MTA-STS] is a standard for a SMTP server to advertise its TLS
+set up and whether SSL is required for it to accept emails.
+
+== Securing incoming traffic with MTA-STS
+
+Apache James supports tools for MTA-STS on the incoming traffic.
+
+The *EnforceMtaSts* SMTP hook can be used to ensure *MAIL FROM* transactions only happens on encrypted channels
+and matches the MTA-STS 'enforce' policy.
+
+Sample set up:
+
+....xml
+<handlerchain>
+    <!-- ... -->
+    <handler class="org.apache.james.smtpserver.tls.EnforceMtaSts"/>
+</handlerchain>
+....
+
+The *TestMtaSts* SMTP hook can be used to audit *MAIL FROM* transactions on unencrypted channels
+and matches the MTA-STS 'testing' policy. It will log but accept emails from such connections.
+Sample set up:
+
+
+....xml
+<handlerchain>
+    <!-- ... -->
+    <handler class="org.apache.james.smtpserver.tls.TestMtaSts"/>
+</handlerchain>
+....
+
+=== Advertizing MTA-STS support
+
+Configuring the MDA is only part of the story. Remains telling the world about it.
+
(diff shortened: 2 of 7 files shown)


Arsnael (approved):

quantranhong1999 (approved):

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/tls/EnforceMtaSts.java
```diff
@@ -0,0 +1,57 @@
...
+ * SMTP MTA Strict Transport Security (MTA-STS)
+ *
+ * Aimed at enforcing mode enforce
+ */
+public class EnforceMtaSts implements MailHook {
+    private static final Logger LOGGER = LoggerFactory.getLogger(EnforceMtaSts.class);
+
+    @Override
+    public HookResult doMail(SMTPSession session, MaybeSender sender) {
+        if (!session.isTLSStarted()) {
```
jeantil, line 43: does `session.isTLSStarted()` account only for StartTLS or also for SSL sessions where the encryption is started before the session is even established ?

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/tls/TestMtaSts.java
```diff
@@ -0,0 +1,50 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * CF https://datatracker.ietf.org/doc/html/rfc8461
+ *
+ * SMTP MTA Strict Transport Security (MTA-STS)
+ *
+ * Aimed at enforcing mode testing
+ */
+public class TestMtaSts implements MailHook {
```
jeantil, line 37: I don't understand what this class does. it logs a warning if TLS is not started and does not return an error code but it returns a DECLINED

the implications of each HookResultCode actions is not very well documented and the code is not the most explicit. I think this will ignore the command without returning an error. 

Is this really the behaviour we want when testing the application of MTA STS ? I would expect the hook to warn but accept the command instead. 
this would let the administrator investigate how much traffic does not honor the published MTA STS policy without disrupting the traffic. 

what did it miss ?

docs/modules/servers/partials/architecture/implemented-standards.adoc
```diff
@@ -37,6 +37,7 @@ This page details standards implemented by the {server-name}.
 - link:https://datatracker.ietf.org/doc/rfc6710/[RFC-6710] SMTP Extension for Message Transfer Priorities
 - link:https://datatracker.ietf.org/doc/rfc8689/[RFC-8689] SMTP Require TLS Option
 - link:https://datatracker.ietf.org/doc/html/rfc1893[RFC-1893] Enhanced Mail System Status Codes
+- link:https://datatracker.ietf.org/doc/html/rfc8461[RFC-8461] SMTP MTA Strict Transport Security (MTA-STS)
```
jeantil, line 40: hmm my understanding is that this PR only covers the enforcement part of the MTA-STS rfc as a receiver of emails. 
it does not cover the sending side (RemoteDeliveryMailet) which should : 
- fetch the policy record from DNS
- try to connect to the remote host and start a TLS/SSL session 
- apply the policy (for example refuse to send to hosts that don't have a TLS/SSL capability)

isn't it a bit misleading to state that james implements the standard when the implementation is partial ? (not sure what the track record is here)

jeantil:
I agree this would be a nice addition to the public james. I feel that having best in class practices for SMTP security is a necessary feature to help with james adoption :D 

I asked a few questions in the diff to better understand the behaviour.

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/tls/TestMtaSts.java
chibenwa: The idea was to log when TLS is not active upon init of a transaction.

We want to warn the admin of non compliant connections, so that the admin can understand the impact of enforcing MTA-STS on his upcoming traffic.

I'll update the doc.

docs/modules/servers/partials/architecture/implemented-standards.adoc
chibenwa: +1 I will add a section onMTA-STS in the conf

docs/modules/servers/partials/configure/mta-sts.adoc
```diff
@@ -0,0 +1,67 @@
...
+....xml
+<handlerchain>
+    <!-- ... -->
+    <handler class="org.apache.james.smtpserver.tls.TestMtaSts"/>
+</handlerchain>
+....
+
+=== Advertizing MTA-STS support
+
+Configuring the MDA is only part of the story. Remains telling the world about it.
```
jeantil, line 34:
```suggestion
Configuring the MTA is only part of the story. Remains telling the world about it.
```
;)

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/tls/TestMtaSts.java
chibenwa: DECLINED means "skip me" in SMTP hooks.

Terrible syntax, I agree, I always get confused TBH...

But I won't fix this here...

jeantil (approved):

docs/modules/servers/partials/configure/mta-sts.adoc
chibenwa: Sorry for the lapsus

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/tls/EnforceMtaSts.java
jeantil: thank you !

Arsnael (approved):

PR #2700, opened 2025-04-09
[devscout] simplifies JDKIM api usage

jeantil:
It looks like James was using either outdated JDKIM APIs or at the very least JDKIM APIs that massively break the lib's encapsulation. 

This set a bad example for users of JDKIM looking for example usage, this PR proposes to change to a more idiomatic usage which has the added benefit of making hte DKIM related mailet's code much cleaner.  

The only downside I can think of is that it might have a performance impact, since the implementation of CRLFInputStream may not be optimal. It can be improved if it proves problematic.


quantranhong1999 (approved):

quantranhong1999:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.3.0:check (check-style) on project james-server-mailet-dkim: You have 12 Checkstyle violations. -> [Help 1]

Arsnael:
05:45:13,486 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.3.0:check (check-style) on project james-server-mailet-dkim: You have 2 Checkstyle violations. -> [Help 1]

Arsnael (commented):
Otherwise LGTM :)

chibenwa (approved):

jeantil:
I added a microbenchmark harness for `CRLFInputStream` that I committed along with the rest but disabled to prevent the bench from further slowing down the CI 

I also aligned the FIFO queue buffer size with the inpustream default buffer size used in transferTo to minimize unnecessary allocations in the common case. 

Here are the bench results I get on my computer
```
Benchmark                                                                               Mode  Cnt        Score     Error   Units
CRLFInputStreamBenchmark.measure_force_crlf_inputstream_wellformed                      avgt    5      410.651 ±  27.984   us/op
CRLFInputStreamBenchmark.measure_force_crlf_inputstream_wellformed:gc.alloc.rate        avgt    5     3342.409 ± 230.725  MB/sec
CRLFInputStreamBenchmark.measure_force_crlf_inputstream_wellformed:gc.alloc.rate.norm   avgt    5  1439000.726 ±   0.049    B/op
CRLFInputStreamBenchmark.measure_force_crlf_inputstream_wellformed:gc.count             avgt    5      273.000            counts
CRLFInputStreamBenchmark.measure_force_crlf_inputstream_wellformed:gc.time              avgt    5      185.000                ms
CRLFInputStreamBenchmark.measure_force_crlf_outputstream_wellformed                     avgt    5       66.754 ±   2.271   us/op
CRLFInputStreamBenchmark.measure_force_crlf_outputstream_wellformed:gc.alloc.rate       avgt    5     2670.461 ±  90.060  MB/sec
CRLFInputStreamBenchmark.measure_force_crlf_outputstream_wellformed:gc.alloc.rate.norm  avgt    5   186928.118 ±   0.003    B/op
CRLFInputStreamBenchmark.measure_force_crlf_outputstream_wellformed:gc.count            avgt    5      218.000            counts
CRLFInputStreamBenchmark.measure_force_crlf_outputstream_wellformed:gc.time             avgt    5      160.000                ms
CRLFInputStreamBenchmark.measure_wellformed                                             avgt    5        6.159 ±   0.042   us/op
CRLFInputStreamBenchmark.measure_wellformed:gc.alloc.rate                               avgt    5    17021.441 ± 115.950  MB/sec
CRLFInputStreamBenchmark.measure_wellformed:gc.alloc.rate.norm                          avgt    5   109944.011 ±   0.001    B/op
CRLFInputStreamBenchmark.measure_wellformed:gc.count                                    avgt    5     1396.000            counts
CRLFInputStreamBenchmark.measure_wellformed:gc.time                                     avgt    5      967.000                ms
```

If I understand this correctly, the CRLF inputstream implementation is an order of magnitude slower than the current CRLFoutputstream which itself is an order of magnitude slower than no CRLF handling. 

I am really bad at manually managing buffers I always end up mixing offsets so I'm not confident i can achieve a faster implementation but I think it should be possible to do it.

server/mailet/dkim/src/main/java/org/apache/james/jdkim/mailets/DKIMSign.java
```diff
@@ -153,38 +149,23 @@ public void init() throws MessagingException {
...
-                } catch (IOException e) {
-                    throw new MessagingException("Exception calculating bodyhash: " + e.getMessage(), e);
-                }
-            }
-            String signatureHeader = signer.sign(headers, bhj);
+        MimeMessage message = mail.getMessage();
+
+        try (MimeMessageInputStream stream = new MimeMessageInputStream(message)) {
+            String signatureHeader = forceCRLF ?
+                    signer.sign(new CRLFInputStream(stream)) :
```
jeantil, line 156: I'm having a doubt here : the original implementation was using a `HeaderSkippingOutputStream` whereas I apply the CRLF normalization to the whole stream and I'm not sure its valid.

jeantil: the same applied for the verifier implementation

server/mailet/dkim/pom.xml
```diff
@@ -131,6 +135,16 @@
...
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.37</version>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.37</version>
+        </dependency>
```
quantranhong1999, line 147: Should it be `test` scope?

PR #2703, opened 2025-04-17
JAMES-4124 Document RedisShake sample to migrate data from Redis to Kvrocks

quantranhong1999:


Arsnael (approved):

chibenwa (changes requested):
Let's take the time in this PR to say that KVrocks is a wip currently triage on a realistic setup.

As of today the Apache James PMC do not endorse it's use on production

chibenwa (approved):

PR #2705, opened 2025-04-23
JAMES-4077 Relax HTML escaping for highlight result

quantranhong1999:
Previously, we did HTML escaping for almost all possible characters. cf https://github.com/apache/james-project/pull/2593
The highlight rendering could be ugly. e.g. the slash `/` character was escaped. `subject": "<mark>Work</mark> from home request (Monday 31&#x2F;03)"`.

Following JMAP SearchSnippet specs (https://jmap.io/spec-mail.html#search-snippets): `&` (ampersand), `<` (less-than sign), and `>` (greater-than sign) characters **MUST** be escaped, while the remaining characters **MAY** depend on server choice.

Let's balance between security and visibility IMO: escape the 3 **MUST** `&` `<` `>` characters, and relax escaping for the rest.

### DIFF
--- a/mailbox/api/src/test/java/org/apache/james/mailbox/searchhighligt/SearchHighLighterContract.java
+++ b/mailbox/api/src/test/java/org/apache/james/mailbox/searchhighligt/SearchHighLighterContract.java
@@ -97,6 +97,132 @@ public interface SearchHighLighterContract {
     }
 
     @Test
+    default void shouldHtmlEscapeTheAmpersandCharacter() throws Exception {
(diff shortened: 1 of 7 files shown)


chibenwa (changes requested):
This is a frontend bug.

Screenshot from 2025-04-23 12-22-27

VS 

image


The fields of earch snippets needs to be interpreted as HTML by the front

IMO we can close this and escalate to the front

chibenwa:
CF https://github.com/linagora/tmail-flutter/issues/3691

PR #2706, opened 2025-04-28
JAMES-4072 Enable timeout option in RedisClientFactory

vttranlina:
Currently, in the `RedisHealthcheck`, we are using Reactor's timeout handling, not the timeout provided by Redis Lettuce.
With this configuration, the timeout will be correctly handled by the Lettuce library

Reference: https://github.com/redis/lettuce/discussions/2614

### DIFF
--- a/backends-common/redis/src/test/java/org/apache/james/backends/redis/RedisSentinelExtension.java
+++ b/backends-common/redis/src/test/java/org/apache/james/backends/redis/RedisSentinelExtension.java
@@ -109,7 +109,7 @@ public class RedisSentinelExtension implements GuiceModuleTestExtension {
             sb.append("redis-sentinel://123@");
             sb.append(this.stream().map(container -> container.getHost() + ":" + container.getMappedPort(SENTINEL_PORT))
                 .collect(Collectors.joining(",")));
-            sb.append("?sentinelMasterId=mymaster");
+            sb.append("?sentinelMasterId=mymaster&timeout=3s");
             return sb.toString();
         }
     }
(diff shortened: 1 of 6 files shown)


quantranhong1999 (approved):

chibenwa (approved):
We do not need to remove (or update) reactor timeout?

Arsnael (approved):

backends-common/redis/src/test/java/org/apache/james/backends/redis/RedisSentinelExtension.java
```diff
@@ -109,7 +109,7 @@ private String createRedisSentinelURI() {
             sb.append("redis-sentinel://123@");
             sb.append(this.stream().map(container -> container.getHost() + ":" + container.getMappedPort(SENTINEL_PORT))
                 .collect(Collectors.joining(",")));
-            sb.append("?sentinelMasterId=mymaster");
+            sb.append("?sentinelMasterId=mymaster&timeout=3s");
```
quantranhong1999, line 112: This requires the extra configuration change (not everyone is aware of it and applies it everywhere)

I would be defensive in keeping the Reactor timeout as the safeguard.

WDYT?

vttranlina: I'm no problem with both way
feel free to revert it

Arsnael: Change or not change? Status on this?

quantranhong1999: I think we can keep removing reactor timeout.

As we have healthcheck timeout safeguard already https://github.com/apache/james-project/blob/11646b05bbc539492fd3715230c22388174a30ad/server/container/guice/common/src/main/java/org/apache/james/PeriodicalHealthChecks.java#L59.

quantranhong1999 (approved):

PR #2707, opened 2025-04-28
JAMES-3925 deal with uploaded blobs directly with blobstoreDAO instea…

Arsnael:
…d of blobstore

We do not need dedup for example for uploaded blobs, to ease regular deletes

### DIFF
--- a/server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/upload/CassandraUploadRepository.java
+++ b/server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/upload/CassandraUploadRepository.java
@@ -25,7 +25,6 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
-import java.util.UUID;
 
 import jakarta.inject.Inject;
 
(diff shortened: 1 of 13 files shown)


quantranhong1999:
08:27:57,593 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.3.0:check (check-style) on project james-server-data-jmap: You have 1 Checkstyle violation. -> [Help 1]

quantranhong1999 (approved):

server/data/data-jmap-cassandra/src/main/java/org/apache/james/jmap/cassandra/upload/CassandraUploadRepository.java
```diff
@@ -48,32 +49,36 @@
...
-        this.blobStore = blobStore;
+        this.blobIdFactory = blobIdFactory;
+        this.blobStoreDAO = blobStoreDAO;
         this.clock = clock;
     }
 
     @Override
     public Mono<UploadMetaData> upload(InputStream data, ContentType contentType, Username user) {
         UploadId uploadId = generateId();
+        BlobId blobId = blobIdFactory.of(UUID.randomUUID().toString());
```
chibenwa, line 67: I would rather have the blobid set to the uploadId rather than an  unrelated random value as there exist a 1 <-> 1 relationship

chibenwa (commented):
I would love to see a test 

Given an upload
When I delete it
Then the blobStoreDao is empty

quantranhong1999 (approved):

chibenwa (approved):

PR #2716, opened 2025-05-12
Update james extension mecanism Fixes 

AlaeMghirbi:
This PR improves how the James extension mechanism works.

Fixes #1733


chibenwa:
Please prefix all commit messages with the JIRA number:  `JAMES-4130 My commit message...`

chibenwa:
No really `JAMES-4130 XXX` instead of `[TICKET 1733]: `

(rationals: we need to point to asf james bug tracker)

Please update.

PR #2719, opened 2025-05-14
JAMES-4131 Configure a fallback bucket for S3

Arsnael:


server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/S3BlobStoreDAO.java
```diff
@@ -154,23 +157,24 @@ public Publisher<InputStream> readReactive(BucketName bucketName, BlobId blobId)
...
-
     private static class FluxResponse {
         final CompletableFuture<FluxResponse> supportingCompletableFuture = new CompletableFuture<>();
         GetObjectResponse sdkResponse;
         Flux<ByteBuffer> flux;
     }
 
     private Mono<FluxResponse> getObject(BucketName bucketName, BlobId blobId) {
+        return getObjectFromStore(bucketName, blobId)
+            .onErrorResume(e -> {
```
chibenwa, line 168: Please do this only if it is not found IMO 

No need to retry arbitrary errors.

server/blob/blob-s3/src/main/java/org/apache/james/blob/objectstorage/aws/S3BlobStoreDAO.java
```diff
@@ -208,14 +212,29 @@ public void onStream(SdkPublisher<ByteBuffer> publisher) {
...
                 .onErrorMap(NoSuchKeyException.class, e -> new ObjectNotFoundException("Blob not found " + blobId.asString() + " in bucket " + resolvedBucketName.asString(), e))
                 .publishOn(Schedulers.parallel())
                 .map(BytesWrapper::asByteArrayUnsafe)
-                .onErrorMap(e -> e.getCause() instanceof OutOfMemoryError, Throwable::getCause));
+                .onErrorMap(e -> e.getCause() instanceof OutOfMemoryError, Throwable::getCause);
+    }
+
+    private Mono<ResponseBytes<GetObjectResponse>> getObjectBytes(BucketName bucketName, BlobId blobId) {
+        return getObjectBytesFromStore(bucketName, blobId)
+                .onErrorResume(e -> {
```
chibenwa, line 225: Idem

chibenwa (approved):
Good work thanks.

Only a tiny remark...

server/apps/distributed-app/helm-chart/james/configs/blob.properties
```diff
@@ -41,6 +41,10 @@ cache.sizeThresholdInBytes=16 KiB
 # Optional, default is bucketPrefix + `default`
 objectstorage.namespace=james-${env:JAMES_BUCKET_SUFFIX}
 
+# Fallback bucket name
+# Optional, read this bucket when default bukcket reads fails if configured
+# objectstorage.namespace.read.fallback=james-fallback
```
quantranhong1999, line 46: We need an entry in `blobstore.adoc` too.

quantranhong1999 (approved):

chibenwa (approved):

PR #2720, opened 2025-05-14
JAMES-2182 StoreMailboxManager should avoid throwing exception in `.filterWhen`

quantranhong1999:
We observed a few `MailboxExistsException` error logs since the change https://github.com/apache/james-project/commit/91f50b031ce2fd1040aa345272ae3f3a76e13b99#diff-51e67dcd44af864286be49fbe7d5474bff07b8fe6cd0536121b4f2898dbbda9e.

Actually, the errors are still caught by the main reactor stream, therefore no impact to users. However, it seems errors emitted by asynchronous operators like `.filterWhen` can be propagated through their own internal asynchronous paths and result in unhandled exceptions being logged (e.g., `onErrorDropped`).

### DIFF
--- a/protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMailboxProcessor.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMailboxProcessor.java
@@ -57,6 +57,7 @@ import org.apache.james.mailbox.MessageManager.MailboxMetaData;
 import org.apache.james.mailbox.MessageUid;
 import org.apache.james.mailbox.ModSeq;
 import org.apache.james.mailbox.NullableMessageSequenceNumber;
+import org.apache.james.mailbox.exception.InsufficientRightsException;
 import org.apache.james.mailbox.exception.MailboxException;
 import org.apache.james.mailbox.exception.MessageRangeException;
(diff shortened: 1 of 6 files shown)


Arsnael (approved):

protocols/imap/src/main/java/org/apache/james/imap/processor/AbstractMailboxProcessor.java
```diff
@@ -100,6 +101,10 @@ protected final Mono<Void> doProcess(R acceptableMessage, Responder responder, I
                             no(acceptableMessage, responder, HumanReadableText.DENIED_SHARED_MAILBOX);
                             return Mono.empty();
                         })
+                        .onErrorResume(InsufficientRightsException.class, e -> {
+                            no(acceptableMessage, responder, HumanReadableText.UNSUFFICIENT_RIGHTS);
+                            return ReactorUtils.logAsMono(() -> LOGGER.info("Processing failed due to insufficient rights", e));
```
chibenwa, line 106:
```suggestion
                            return ReactorUtils.logAsMono(() -> LOGGER.warn("Processing failed due to insufficient rights", e));
```

chibenwa (approved):
Cool thanks...

chibenwa (approved):

PR #2724, opened 2025-05-16
JAMES-4133 POC RFC-6532 Support for i8n emails

chibenwa:
CF https://issues.apache.org/jira/browse/JAMES-4133

Fails lamentably:

```
jakarta.mail.internet.AddressException: Invalid character in local-part (user account) at position 4 in 'ren��@james.org'
	at org.apache.james.core.MailAddress.parseUnquotedLocalPart(MailAddress.java:561)
	at org.apache.james.core.MailAddress.parseUnquotedLocalPartOrThrowException(MailAddress.java:272)
	at org.apache.james.core.MailAddress.<init>(MailAddress.java:193)
	at org.apache.james.protocols.smtp.core.RcptCmdHandler.doFilterChecks(RcptCmdHandler.java:157)
	at org.apache.james.protocols.smtp.core.AbstractHookableCmdHandler.onCommand(AbstractHookableCmdHandler.java:72)
	at org.apache.james.protocols.smtp.core.AbstractHookableCmdHandler.onCommand(AbstractHookableCmdHandler.java:50)
	at org.apache.james.protocols.api.handler.CommandDispatcher.dispatchCommandHandlers(CommandDispatcher.java:165)
	at org.apache.james.protocols.api.handler.CommandDispatcher.onLine(CommandDispatcher.java:142)
	at org.apache.james.protocols.netty.BasicChannelInboundHandler.channelRead(BasicChannelInboundHandler.java:196)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:346)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:318)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	at io.netty.channel.AbstractChannelHandlerContext.access$600(AbstractChannelHandlerContext.java:61)
	at io.netty.channel.AbstractChannelHandlerContext$7.run(AbstractChannelHandlerContext.java:425)
[...]

PR #2728, opened 2025-05-22
JAMES-4133 Add support for unicode adddresses as defined in RFC6532.

arnt:
This adds Domain tests and does enough to parse unicode email addresses.

However, I'm not quite happy. This needs documentation changes, I wouldn't mind another few Domain tests, and I do want to investigate what's up with Jakarta and `संपर्क@डाटामेल.भारत` but haven't time to do that this week.

Consider it a WIP.

### DIFF
--- a/protocols/imap/src/main/java/org/apache/james/imap/decode/ImapRequestLineReader.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/decode/ImapRequestLineReader.java
@@ -754,7 +754,7 @@ public abstract class ImapRequestLineReader {
      */
     protected String consumeQuoted(Charset charset) throws DecodingException {
         if (charset == null) {
-            return consumeQuoted(US_ASCII);
+            return consumeQuoted(UTF_8);
         } else {
             // The 1st character must be '"'
             consumeChar('"');
--- a/protocols/imap/src/main/java/org/apache/james/imap/decode/ImapRequestLineReader.java
+++ b/protocols/imap/src/main/java/org/apache/james/imap/decode/ImapRequestLineReader.java
@@ -20,6 +20,7 @@
 package org.apache.james.imap.decode;
 
 import static java.nio.charset.StandardCharsets.US_ASCII;
+import static java.nio.charset.StandardCharsets.UTF_8;
 
 import java.io.Closeable;
 import java.io.IOException;
@@ -291,6 +292,7 @@ public abstract class ImapRequestLineReader {
 
     protected char nextChar; // unknown
     protected boolean nextSeen = false;
+    private boolean utf8Accept;
     private final StringBuilder stringBuilder = new StringBuilder();
 
     /**
@@ -489,7 +491,25 @@ public abstract class ImapRequestLineReader {
      * 
      */
     public String mailbox() throws DecodingException {
-       return ModifiedUtf7.decodeModifiedUTF7(mailboxUTF7());
+        if (utf8Accept) {
+            String mailbox = astring(UTF_8);
+            if (mailbox.equalsIgnoreCase(ImapConstants.INBOX_NAME)) {
+                return ImapConstants.INBOX_NAME;
+            }
+            return mailbox;
+        }
+        return ModifiedUtf7.decodeModifiedUTF7(mailboxUTF7());
+    }
+
+    /**
+     * When set, {@link #mailbox()} treats the astring as UTF-8 and does not
+     * run Modified UTF-7 decoding. Callers should set this from
+     * {@code EnableProcessor.getEnabledCapabilities(session).contains(SUPPORTS_UTF8_ACCEPT)}
+     * after the session state is known.
+     */
+    public ImapRequestLineReader setUtf8Accept(boolean utf8Accept) {
+        this.utf8Accept = utf8Accept;
+        return this;
     }
 
     /**
@@ -501,7 +521,8 @@ public abstract class ImapRequestLineReader {
      * variants of ;; INBOX (e.g. "iNbOx") MUST be interpreted as INBOX ;; not
      * as an astring.
      * 
-     * Be aware that mailbox names are encoded via a modified UTF7. For more information RFC3501
+     * Be aware that mailbox names are encoded via a modified UTF7 in unextended
+     * IMAP. For more information see RFC3501. RFC9755 changes this.
      */
     public String mailboxUTF7() throws DecodingException {
         String mailbox = astring();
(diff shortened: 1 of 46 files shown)


chibenwa:
Hello @arnt 

Thanks for this major contribution.

I did update https://github.com/apache/james-project/pull/2724 with your changes and can happily say the SMTP-IN stack in James is RFC-6532 compliant :-)

However, I did face major issues as any modification (header for instance, but also the defensive copy done in the memory mail queue) causes the loss of the UTF-8 characters. This clearly see like a lack of support for RFC-6532 from `jakarta.mail.internet.MimeMessage` which James (sadly...) still heavily relies on. (quick checks tends to confirm this dependency do not support RFC-6532).

Best regards,

Benoit TELLIER

arnt:
@chibenwa could you possibly look at this now and fix the pom.xml, for example?

Most of it is gated on SMTPUTF8, UTF8=ACCEPT etc, but https://github.com/apache/james-project/pull/2728/commits/106201715b3c7f726f6f54a151571920870dddc2 changes James in the opposite case, namely when James' user does NOT enable support and a client tries to use foo@xn--bar anyway. I had a headache there. The code I wrote tries to preserve properties people generally think addresses have (e.g. string comparability) even when a fun-loving SMTP client connects to a server based on James and uses features the server author didn't know about.

Feel free to add additional comments and push it on the #2724 PR.

arnt:
@chibenwa have a look? I realise it's rather a lot of code.

protocols/imap/src/main/java/org/apache/james/imap/decode/ImapRequestLineReader.java
```diff
@@ -291,6 +292,7 @@ public static boolean isQuotedSpecial(char chr) {
 
     protected char nextChar; // unknown
     protected boolean nextSeen = false;
+    private boolean utf8Accept;
```
chibenwa, line 295:
```suggestion
    private boolean utf8Accept = false;
```

Let's make the default value explicit...

protocols/imap/src/main/java/org/apache/james/imap/encode/base/ImapResponseComposerImpl.java
```diff
@@ -64,6 +65,8 @@ public class ImapResponseComposerImpl implements ImapConstants, ImapResponseComp
 
     private boolean skipNextSpace;
 
+    private boolean utf8Accept;
```
chibenwa, line 68:
```suggestion
    private boolean utf8Accepted = false;
```

protocols/imap/src/main/java/org/apache/james/imap/processor/CapabilityProcessor.java
```diff
@@ -106,6 +113,16 @@ public void addProcessor(CapabilityImplementingProcessor implementor) {
...
     }
+
+    @Override
+    public List<Capability> getPermitEnableCapabilities(ImapSession session) {
+        return ENABLEABLE_CAPS;
+    }
+
+    @Override
+    public Mono<Void> enable(ImapMessage message, Responder responder, ImapSession session, Capability capability) {
+        return Mono.empty();
```
chibenwa, line 124: Here shouldn't we set allowUtf8 in the encoder and the decoder when SUPPORTS_UTF8_ACCEPT capability is passed ?

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/RcptCmdHandler.java
```diff
@@ -195,6 +221,15 @@ protected Response doFilterChecks(SMTPSession session, String command,
...
     }
 
+    private static boolean containsNonAscii(String s) {
+        for (int i = 0; i < s.length(); i++) {
+            if (s.charAt(i) > 0x7F) {
+                return true;
+            }
+        }
+        return false;
+    }
```
chibenwa, line 231: Can we move this dulicated piece of code in AddressNormalization ?

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/DSNBounce.java
```diff
@@ -528,6 +542,20 @@ private void appendRecipient(StringBuilder buffer, MailAddress mailAddress, Stri
             .append(LINE_BREAK);
     }
 
+    private static String addrType(MailAddress mailAddress) {
+        return containsNonAscii(mailAddress) ? "utf-8" : "rfc822";
+    }
+
+    private static boolean containsNonAscii(MailAddress mailAddress) {
```
chibenwa, line 549: Maybe we can move this into the MailAdress class itself ?

protocols/imap/src/main/java/org/apache/james/imap/processor/CapabilityProcessor.java
chibenwa: Edit I see that we handle that while instanciating composer / parser.

Maybe just a little comment here ?

Or better: make smtpUtf8 a prime citizen of ImapSession. That way we can write here an explicit `session.enableUtf8()`  and in the 2 other places we can write `session.utf8Enabled()` when instanciating composer / parser.

chibenwa (commented):
That's awesome!

To polish this work I see:
 - [ ] resolving the build conflict
 - [ ] Do non regression JMH runs for Domain and MailAddress POJO (to know what we sign in for perf wize)

I'll try to plan this in the coming week. If you want I can take care of pushing fixups for my remarks.

chibenwa:
Hello @arnt !

Thanks again for putting this awesome work together.

Restarting from it I worked on https://github.com/apache/james-project/pull/3130 :
 - Rebase it on origin/master
 - Fix MDN failed test
 - Fix a bug in IMAP parsing
 - Adress the remarks I made on this PR
 - Introduce a JMH test suite and address related performance degradation
 - Battle test (and fix) RemoteDelivery handling as per SMTPUTF8
 - Make SMTPUTF8 enabled by default without the need of further extension

I'll monitor the ongoing build there and keep doing what it takes to please the test suite.

I believe as such this change should be in acceptable state, I'd propose to target for it next James release cycle to let plenty of time to triage it on production before shipping this is a release.

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/RcptCmdHandler.java
arnt: I didn't see a natural place (other than java.lang.String). But you picked one, which is good.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/DSNBounce.java
arnt: IMO it doesn't matter all that much. I'm happy with the way you did it in #3130 but IMO it's a case of 2% better/worse.

PR #2737, opened 2025-06-04
Mailbox custom mapping refactoring

Arsnael:


quantranhong1999 (approved):

chibenwa (approved):

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/query/DefaultCriterionConverter.java
```diff
@@ -85,14 +85,14 @@ public DefaultCriterionConverter() {
         registerHeaderOperatorConverters();
     }
 
-    private String evaluateFuzzinessValue(boolean textFuzzinessSearchEnable) {
+    protected String evaluateFuzzinessValue(boolean textFuzzinessSearchEnable) {
```
Arsnael, line 88: Does that really make sense in the end? Would not implement the interface anymore but extend DefaultCriterionConverter instead then?

chibenwa: Limiting the code duplication is IMO desirable.

extend DefaultCriterionConverter in tmail is BTW not incompatible with interface in tmail

BTW it would ease injects...

quantranhong1999 (approved):

PR #2738, opened 2025-06-05
[ENHANCEMENT] Allow disable threads

chibenwa:
This changeset will not alter thread categorisation which works decen…tly today but only thread presentation
to the end user which is currently relying on search. As an operator I wishes to have a security trigger if things go blur.


mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/CassandraThreadIdGuessingAlgorithm.java
```diff
@@ -48,6 +48,8 @@
 import reactor.core.publisher.Mono;
 
 public class CassandraThreadIdGuessingAlgorithm implements ThreadIdGuessingAlgorithm {
+    private static final boolean DISABLE_THREADS = Boolean.valueOf(System.getProperty("james.mailbox.threads.disable", "false"));
```
quantranhong1999, line 51: Should we add this for `PostgresThreadIdGuessingAlgorithm` too?

chibenwa: Yes please feel free

Arsnael (approved):

PR #2739, opened 2025-06-10
JAMES-4137 Upgrade S3 SDK to 2.30.x onwards

quantranhong1999:
In S3 SDK version 2.30.0, AWS team introduced a breaking change in the integrity check in the driver:

- Checksum is always required by default by AWS SDK >= 2.30.x
- AWS SDK >= 2.30.x does not send MD5 checksum anymore (which 3rd party S3 storages rely on), but relies on CRC32 checksum.

cf: https://github.com/aws/aws-sdk-java-v2/discussions/5802

This breaks the compatibility with 3rd party object storage(s) that still rely on MD5 checksum. We failed to upgrade the SDK in the past: https://github.com/apache/james-project/pull/2638.

It seems after receiving a lot of complaints from the community (Apache Hadoop, Apache Iceberg, Apache Spark...), AWS has agreed to have an option to maintain backward compatibility with S3-compatible storages. cf https://github.com/aws/aws-sdk-java-v2/pull/6055.

Apache Iceberg is starting to adopt the backward compatibility option: https://github.com/apache/iceberg/pull/12264.

Let's try to adopt this on our side, with care.

### DIFF
--- a/docs/modules/servers/partials/configure/jvm.adoc
+++ b/docs/modules/servers/partials/configure/jvm.adoc
@@ -90,6 +90,18 @@ To enable blob hierarchy compatible with MinIO add in `jvm.properties`:
 james.s3.minio.compatibility.mode=true
 ----
 
+== Disable S3 checksum backward compatibility
+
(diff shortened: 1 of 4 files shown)


chibenwa (approved):

Arsnael (commented):
the backward compatibility wasn't added yet when i had a look at this before> Good catch!!!

Arsnael (approved):

PR #2742, opened 2025-06-12
JAMES-4124 Upgrade Rspamd to 3.12.0

quantranhong1999:
This would allow Rspamd to execute EVASHA_RO commands against Kvrocks replicas (instead of EVASHA, which Kvrocks replicas reject ATM).

cf: https://github.com/rspamd/rspamd/pull/5419

Then no workaround is needed anymore for Rspamd + Kvrocks sentinel.


chibenwa (approved):
Good catch

PR #2744, opened 2025-06-16
Improved JWT support

amichair:
- documentation on how to set up JWT for webadmin access
- added token expiration tests (it works, but was not documented/tested)
- added kid lookup support to the DefaultPublicKeyProvider (using JWK thumbprint), changed the PublicKeyProvider interface to have a separate get(kid) method, changed JwksPublicKeyProvider to have separate implementations for the get all and get kid methods (instead of the instance having a kid associated with it and either using it or not in the get all method), and updated JwtTokenVerifier (shared by both) to try the kid lookup if kid header exists or iterate through all keys if the header does not exist. This makes the design more intuitive, consistent and shared among all providers.
- upgrade jwks-rsa
- upgrade jjwt, including migration to the new api - both small naming/constant changes, and a bigger change in handling compression with our configurable flag

### DIFF
--- a/server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
+++ b/server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
@@ -18,43 +18,75 @@
  ****************************************************************/
 package org.apache.james.jwt;
 
-import java.security.PublicKey;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.security.Key;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 
 import io.jsonwebtoken.Claims;
-import io.jsonwebtoken.CompressionCodecResolver;
 import io.jsonwebtoken.Jws;
 import io.jsonwebtoken.JwtException;
 import io.jsonwebtoken.JwtParser;
 import io.jsonwebtoken.Jwts;
+import io.jsonwebtoken.Locator;
 import io.jsonwebtoken.MalformedJwtException;
-import io.jsonwebtoken.impl.compression.DefaultCompressionCodecResolver;
+import io.jsonwebtoken.io.CompressionAlgorithm;
 
 public class JwtTokenVerifier {
-    private static final CompressionCodecResolver DEFAULT_COMPRESSION_CODEC_RESOLVER = new DefaultCompressionCodecResolver();
-    private static final CompressionCodecResolver SECURE_COMPRESSION_CODEC_RESOLVER = header -> {
-        if (Optional.ofNullable(header.getCompressionAlgorithm()).isPresent()) {
-            throw new RuntimeException("Rejecting a ZIP JWT. Usage of ZIPPED JWT can result in " +
-                "excessive memory usage with malicious JWT tokens. To activate support for ZIPPed" +
-                "JWT please run James with the -Djames.jwt.zip.allow=true system property.");
-        }
-        return DEFAULT_COMPRESSION_CODEC_RESOLVER.resolveCompressionCodec(header);
-    };
-    private static final boolean allowZipJWT = Optional.ofNullable(System.getProperty("james.jwt.zip.allow"))
+    static boolean allowZipJWT = Optional.ofNullable(System.getProperty("james.jwt.zip.allow"))
         .map(Boolean::parseBoolean)
         .orElse(false);
-    @VisibleForTesting
-    static CompressionCodecResolver CONFIGURED_COMPRESSION_CODEC_RESOLVER = Optional.of(allowZipJWT)
-        .filter(b -> b)
-        .map(any -> DEFAULT_COMPRESSION_CODEC_RESOLVER)
-        .orElse(SECURE_COMPRESSION_CODEC_RESOLVER);
+
+    /**
+     * A CompressionAlgorithm that delegates to another compression algorithm
+     * if compression is allowed, or throws exceptions if not.
+     */
+    private static class SecureCompressionAlgorithm implements CompressionAlgorithm {
+        String id;
+        CompressionAlgorithm delegate;
+
+        public SecureCompressionAlgorithm(String id, CompressionAlgorithm delegate) {
+            this.id = id;
+            this.delegate = delegate;
+        }
+
+        private void validate() {
+            if (!allowZipJWT) {
+                throw new RuntimeException("Rejecting a ZIP JWT. Usage of ZIPPED JWT can result in " +
+                    "excessive memory usage with malicious JWT tokens. To activate support for ZIPPed" +
+                    "JWT please run James with the -Djames.jwt.zip.allow=true system property.");
+            }
+        }
+
+        @Override
+        public String getId() {
+            return id;
+        }
+
+        @Override
+        public OutputStream compress(OutputStream out) {
+            validate();
+            return delegate.compress(out);
+        }
+
+        @Override
+        public InputStream decompress(InputStream in) {
+            validate();
+            return delegate.decompress(in);
+        }
+    }
+
+    // wrap all supported compression algorithms with 'secure' ones that throw exception if disabled
+    private static final List<SecureCompressionAlgorithm> SECURE_COMPRESSION_ALGORITHMS =
+        Jwts.ZIP.get().values().stream().map(c -> new SecureCompressionAlgorithm(c.getId(), c)).toList();
 
     public interface Factory {
         JwtTokenVerifier create();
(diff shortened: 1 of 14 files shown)


chibenwa:
Hello,

Thanks a lot for this contribution, which is a real enhacement.

I only find the "kid" related code more complex than the rest of the change set and find it pretty low level. My JWKS knowlege in insufficiant to feel confortable about it.  Do we know why it is not built in the JWKS library? Have we considered contributing it upstream?

Arsnael (approved):
Ain't expert on jwt but LGTM to me! Thank you for the contribution @amichair !

chibenwa (changes requested):
440f1b1e78e9f4d16e47a1cb40e3da3f357548bd injects needless complexity that seems overly dangerous to me.

Too many concerns are bundled together which makes it hard to follow.

Please remove this commit from the current pull request so that we get the current changeset merged as it makes complete sense. Being so low level outside on the area of expertise of the project on security sensitive topics is a good sign there is something wrong.

Let's open a separate PR for 440f1b1e78e9f4d16e47a1cb40e3da3f357548bd and maybe let's further discuss on the mailing the problem we are trying to solve before throwing solutions at it.

server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
```diff
@@ -67,12 +99,19 @@ public static JwtTokenVerifier create(JwtConfiguration jwtConfiguration) {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(JwtTokenVerifier.class);
 
+    private final JwtParser kidJwtParser;
     private final List<JwtParser> jwtParsers;
 
     public JwtTokenVerifier(PublicKeyProvider pubKeyProvider) {
+        // one parser that performs key lookup by kid
+        this.kidJwtParser = toImmutableJwtParser(h -> {
```
chibenwa: h -> jwtHeaders

server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
```diff
@@ -67,12 +99,19 @@ public static JwtTokenVerifier create(JwtConfiguration jwtConfiguration) {
...
+            String kid = Objects.requireNonNull((String)h.get("kid"));
+            return pubKeyProvider.get(kid).orElse(null);
```
chibenwa, line 109: Do not use null

amichair:
The old code without that commit won't work with the newer jjwt releases, that's how I got to this seemingly unrelated change in the first place. Specifically, getClaimWithoutSignatureVerification doesn't work - there's no simple way to get a claim without also doing the verification itself (which will fail, because we haven't found what key to use yet... chicken and egg). In the new API there is only the KeyLocator, but that happens at a later stage in the verifier... our previous logic was a bit twisted, and straightening it out so it will be simpler and also work with the newer jjwt apis got me to this commit.

If I understand correctly, the only thing that doesn't feel right is tha actual kid calculation, which is a bit low-level, though it is currently encapsulated in one place to make it easy to swap or fix or use an external lib should it be necessary someday. We could have gone with some trivial kid, such as a certificate serial number or such where the code might look a bit shorter... but if we already need to calculate a kid, why not use what afaict is an emerging standard?

As for just trying them one by one - that's the fallback that still works if there is no kid in the jwt. I suppose we can just drop support for kid in the default provider and always use this fallback. It does remove functionality and consistency between providers, and the option of future enhancements (like key lookup per db per user or whatnot).

So I see several options:
1. Find another way to make the new api work, and remain without the kid support in the default provider.
2. Leave a nop kid lookup and use the scan-all-keys fallback, and remain without the kid support in the default provider.
3. Pick a different, possibly simpler, less standard kid calculation (remember, whomever creates the jwt also needs to run this calculation when creating the jwt - so it should be either very simple or very standard).
4. Leave it as is with the new design and full support.
5. If it's just the kid calculation we don't like, try to find some external library that provides a java PublickKey to JWK thumbprint calculation.
6. Other options?

server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
amichair: renamed

server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
amichair: This is jjwt's Locator interface, not ours - return the located object or null.

chibenwa:
Thanks for the sum up.

To be fairly honest I am not very happy with the current JWT code...

 - For JMAP I am unaware of existing user. We could just deprecate this.
 - For webadmin signing the claims prior the Webadmin calls is a show stopper to most people I have been talking to. They resort to unprotected webadmin on a private netword instead... I think a simpler way to secure webadmin with static secrets would serve us well.

Remains OIDC for IMAP/SMTP that if I am not wrong relies on Jwks.

> So I see several options:

Thanks for this explanation

> 4. Leave it as is with the new design and full support.

Acceptable. If kid computation is screwed up it will result in 401, and collision accross kid seems impossible to me.

> 1. Find another way to make the new api work, and remain without the kid support in the default provider.

I confess that this is what I would prefer...

amichair:
Just a note - I was one of those people who just used unprotected webadmin (or turned it on only when needed and then turned it off) - but only because it was too much work to figure out how to do it. That's why the first commit in this PR was just to add documentation that clearly explains what to do. Once documented, it's actually just a few minutes to set it up.

amichair:
> * For webadmin signing the claims prior the Webadmin calls is a show stopper

Just to clarify, you don't have to sign it for each call - you create a token once, or only occasionally, and then use it until it expires or the key is revoked, similar to oauth or other rest api tokens. The nice thing about it is that it also contains the user, or in the future can contain different roles, so we can add more granular authorization - a read-only token, a read-only-my-domain token, or whatever we want. I'm hoping the documentation will encourage everyone to move to tokens instead of unprotected access or static secrets or external workarounds :-)

amichair:
I extracted the thumbprint calculation into the last commit (which can be  moved to a separate PR).

> It sounds reasonable to me.

Suggestions for improvement?

server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
```diff
@@ -82,22 +121,28 @@ public Optional<String> verifyAndExtractLogin(String token) {
...
             T claim = jws
-                .getBody()
+                .getPayload()
                 .get(claimName, returnType);
             if (claim == null) {
                 throw new MalformedJwtException("'" + claimName + "' field in token is mandatory");
             }
             return Optional.of(claim);
-        } catch (JwtException e) {
+        } catch (JwtException e) { // also if kid was given but our locator didn't find the corresponding key
```
chibenwa, line 145: DO we have a way to get this handled separately so that we get a meaningful exception when "kid was given but our locator didn't find the corresponding key" ?

server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
```diff
@@ -67,12 +99,19 @@ public static JwtTokenVerifier create(JwtConfiguration jwtConfiguration) {
...
     private static final Logger LOGGER = LoggerFactory.getLogger(JwtTokenVerifier.class);
 
+    private final JwtParser kidJwtParser;
     private final List<JwtParser> jwtParsers;
 
     public JwtTokenVerifier(PublicKeyProvider pubKeyProvider) {
+        // one parser that performs key lookup by kid
+        this.kidJwtParser = toImmutableJwtParser(jwtHeaders -> {
+            String kid = Objects.requireNonNull((String)jwtHeaders.get("kid"));
+            return pubKeyProvider.get(kid).orElse(null);
```
chibenwa, line 109: Before we were loading public keys only at startup.

Now we are loading keys on each and every token we verify. Thus I would not be surprised performance significantly degrades.

Maybe we could load upfront a `Map<Kid -> PublicKey> in the DefaultPublicKeyProvider in order to keep this desirable property?

Maybe the

amichair: Regardless of the current changes:

- DefaultPublicKeyProvider loads all keys on every get() invocation. I'm not sure if this is good or not - on the one hand caching in the constructor is more efficient, on the other hand in theory keys can be added or removed and maybe that should be reflected in a provider's lifetime. Perhaps the provider's contract needs to be better defined in this regard (also thinking about future providers of keys per user or db lookup etc.)

- I think the only place that used the kid is from the OidcJwtTokenVerifier which creates a new provider and verifier per request (so loads the keys per request anyway). Before the change it was limited by accepting the kid in the provider constructor, which forced always constructing new ones, but after the current change that is no longer the case, which leaves room for future optimizations.

Perhaps optimizations/caching should be outside the scope of this PR with a separate discussion? They go beyond the changes here.

server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
amichair: I think we can throw our own runtime exception in the locator and catch it separately here. How would you want it to be handled differently? different log level? (the exceptions already have reasonable message text)

server/protocols/jwt/src/main/java/org/apache/james/jwt/JwtTokenVerifier.java
chibenwa: > DefaultPublicKeyProvider loads all keys on every get() invocation

But JwtTokenVerifier only calls it once.

> OidcJwtTokenVerifier which creates a new provider and verifier per request

But JwksPublicKeyProvider  provides caching logic.

> Perhaps optimizations/caching should be outside the scope of this PR with a separate discussion? 

Not really I believe this significantly changes the performance profile of webadmin JWT token validation.

amichair: Perhaps I'm missing something, but aiui OidcJwtTokenVerifier (the only place currently using kid) creates a new provider instance per request, so any caching within the providers would be meaningless? It seems like the providers themselves need to be cached instead of using stateless static methods, but that goes back to other components I haven't gone over here. Maybe I'm assuming something wrong about the other components using this verifier, or they are not called per request etc.?

chibenwa: ... right indeed.

Sorry I was convinced some level of caching was dome by the underlying Auth0 JWKS library.

amichair:
I opened [JAMES-4139](https://issues.apache.org/jira/browse/JAMES-4139) for key management optimizations.

PR #2750, opened 2025-06-23
JAMES-3516 Rely on Cassandra for Thread lookups

chibenwa:
- Reorganise the lookup DAO in order to include the ThreadId as part of the primary key and relegate the messageId as a clustering key. Acceptable for our usage pattern, this enables listing content of a thread.

 Migration strategy: none. Truncate the old v2 table and start with
 an empty v3 table which is acceptable.

  - Rely on this method for thread resolution instead of the search index

  - Pay an attention that messages in the tread are accessible to the end user (right resolution)
  
  Context: on a production system `Thread/get` takes 500+ ms.


mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/CassandraThreadIdGuessingAlgorithm.java
```diff
@@ -88,17 +86,17 @@ public Flux<MessageId> getMessageIdsInThread(ThreadId threadId, MailboxSession s
...
-            .build();
-
-        return Flux.from(mailboxManager.search(expression, session, Integer.MAX_VALUE))
-            .switchIfEmpty(Mono.error(() -> new ThreadNotFoundException(threadId)));
+        return threadLookupDAO.selectAll(threadId)
+            .collectList()
+            .map(ids -> {
+                // Handle umpopulated lookup dao
+                if (ids.isEmpty()) {
+                    return ImmutableList.of(threadId.getBaseMessageId());
```
quantranhong1999: I am thinking about fallback to OpenSearch for this:
- Keep the thread working well for previous messages
- Fix the `givenNonMailInAThreadThenGetThreadShouldThrowThreadNotFoundException` test: actual not found thread should throw `ThreadNotFoundException`, not naive based MessageId threadId.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/CassandraThreadIdGuessingAlgorithm.java
```diff
@@ -88,17 +86,17 @@ public Flux<MessageId> getMessageIdsInThread(ThreadId threadId, MailboxSession s
...
+                }
+                return ids;
+            })
+            .flatMapMany(messageIds -> messageIdManager.accessibleMessagesReactive(messageIds, session))
```
quantranhong1999: The test `givenThreeMailsInAThreadThenGetThreadShouldReturnAListWithThreeMessageIdsSortedByArrivalDate` is failing.

`messageIdManager.accessibleMessagesReactive` mixed the ordering (by date).

And I am not sure if `threadLookupDAO.selectAll(threadId)` keeps ordering well too in the first place.

quantranhong1999:
> Is it enough for passing mailbox/Cassandra tests?

There are 2 tests failing. I added review comments.

I would love to help on this too if you want.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/CassandraThreadIdGuessingAlgorithm.java
chibenwa: I am rather opposed to this as I know it will cause performance issues upon migrations and would like to prevent this.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/CassandraThreadIdGuessingAlgorithm.java
quantranhong1999: > And I am not sure if threadLookupDAO.selectAll(threadId) keeps ordering well too in the first place.

should be the case already as messageId is timeuuid

chibenwa: I think we do not need to retain ordering (`Thread/get` ordering is brittle)

chibenwa:
I abandonned tonight.

(sorry if I hurt anyone) but ThreadIdGuessingAlgorithmContract is absolute garbage and deserves a complete rewrite using directly and only MailboxManager and higher APIs.

quantranhong1999:
> Can you help me at writing those tests on top a mailbox manager without exercising directly the mailbox API ?

cf ticket https://github.com/linagora/james-project/issues/5444

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/CassandraThreadIdGuessingAlgorithm.java
quantranhong1999: JMAP Thread specs say we should have Thread/get ordering by date.

image

Using filter should be simple enough to retain ordering:
```java
    .flatMap(ids -> Mono.from(messageIdManager.accessibleMessagesReactive(ids, session))
        .map(accessibleSet -> {
            return ids.stream()
                .filter(accessibleSet::contains)
        })
    )
    .flatMapMany(Flux::fromIterable);
```

chibenwa:
> JMAP Thread specs say we should have Thread/get ordering by date.

Ok then we need to retrieve first level of metadata for sorting.

mailbox/cassandra/src/main/java/org/apache/james/mailbox/cassandra/CassandraThreadIdGuessingAlgorithm.java
chibenwa: (Fixed)

quantranhong1999:
> Ok then we need to retrieve first level of metadata for sorting.

I am not sure to get your point, as I see the current code removed the sorting assertion.

Is sorting by the clustering column with MESSAGE_ID as timeuuid not enough?

Arsnael (approved):

PR #2746, opened 2025-06-18
Summer lib updates

Arsnael:
It seems compiling at least. Let's see regarding the tests!

I did not update the following libs for now, I might do it in some later PR after this one is stable:

- netty: usually needs to be aligned with other libs, not sure we can here
- jjwt: need some changes and being taken care in an other PR https://github.com/apache/james-project/pull/2744
- derby
- parboiled
- doxia
- maven-reporting libs
- opensearch: upgraded to latest v2, but I saw some new v3 versions... that might need a more proper look too for a major version upgrade

PR #2751, opened 2025-06-24
Summer lib updates - bis

Arsnael:
Duplicate of https://github.com/apache/james-project/pull/2746

I don't know why github decided it was creating conflicts, it does not!


Arsnael:
Same... Is it too much commits for github? Weird cause until now it was ok. When the build is green it's not... This is stupid... I guess maybe I need to chunk it down... what a pain

PR #2762, opened 2025-07-02
[Update] Using http5 client instead of opensearch rest client and upgrading to v3

Arsnael:
Getting back to this but upgrading to v3 as well for opensearch.

With the rest client they seem to drop most http transitive libs from v3 that were used for ssl, async client, etc... so it felt better to try that way.

We could though still keep the current way by adding in the pom the missing dependencies I guess... we can test

Thinking maybe as well before perf tests were not good enough because we didnt really upgrade opensearch on test env, might need to see with an upgraded version of it


server/testing/src/main/java/org/apache/james/util/docker/Images.java
```diff
@@ -24,7 +24,7 @@ public interface Images {
     String RABBITMQ = "rabbitmq:4.1.1-management";
     String ELASTICSEARCH_2 = "elasticsearch:2.4.6";
     String ELASTICSEARCH_6 = "docker.elastic.co/elasticsearch/elasticsearch:6.3.2";
-    String OPENSEARCH = "opensearchproject/opensearch:2.19.2";
+    String OPENSEARCH = "opensearchproject/opensearch:3.1.0";
```
chibenwa, line 27: Can you please confirm we technically still run on 2.19.x after this upgrade or is this a breaking change?

Arsnael: I believe should be alright but let's see what the CI thinks: https://github.com/apache/james-project/pull/2763

Can test later as well in a test environment

PR #2773, opened 2025-07-23
feat(managesieve): add XOAUTH2 authentication mechanism

felixauringer:
As the managesieve server only supports plain authentication, here is a first implementation of XOAUTH2 as an additional authentication mechanism for managesieve.

I would be happy about feedback :)


chibenwa (commented):
This looks great.

Can we try to write a test for OIDC support in managedSieve like this is done for IMAP and SMTP?

Also, out of curiosity do we have a managedSive compatible client>

We are missing:
 - ManageSiee exemple in https://github.com/apache/james-project/tree/master/examples/oidc
 - Documentation on how to configure this feature

quantranhong1999 (commented):
Read it.

Thank you for your contribution.

This lacks some tests indeed.

felixauringer:
I have updated documentation and the oidc example. I added some CLI commands to test oidc authentication and made `test.sh` succeed. I did not test with thunderbird but my changes should not break the existing documentation on that.

mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/authenticate.test
```diff
@@ -18,20 +18,25 @@
 ################################################################
 
 C: AUTHENTICATE
-S: NO ManageSieve syntax is incorrect : You must specify a SASL mechanism as an argument of AUTHENTICATE command
+S: NO "ManageSieve syntax is incorrect: quoted SASL mechanism must be supplied"
```
chibenwa, line 21: +1 for fixing this likely non-compliant unpquoted no explanation (thanks)

mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/capability.test
```diff
@@ -38,8 +38,9 @@ S: "VERSION" "1.0"
 S: OK
 
 C: AUTHENTICATE "PLAIN"
-S: \+ ""
-C:  user password
+S: ""
+S: OK
+C: "user password"
```
chibenwa: I am not fully sure that this code is compliant with eg Thunderbird extension

From my memories of SASL we either have

```
C: AUTHENTICATE "PLAIN" "AGFsaWNlAHNlY3JldA=="\r\n
S: OK "Logged in"
```

Or

```
C: AUTHENTICATE "PLAIN"\r\n
S: +\r\n
C: AGFsaWNlAHNlY3JldA==\r\n
S: OK "Logged in"\r\n
```

(with base64 being `\0LOGIN\0password`)

So seing not base64 encoded, qutoed password seems indeed weird to me...

From the above exemple the initial server response is `+ \r\n`

-----------------

After reading the code below I think the client payload handling is retro-compatible.

So I would welcome tests that demonstrate previous behaviours are still supported.

protocols/managesieve/src/main/java/org/apache/james/managesieve/core/PlainAuthenticationProcessor.java
```diff
@@ -50,7 +50,7 @@ public PlainAuthenticationProcessor(UsersRepository usersRepository) {
 
     @Override
     public String initialServerResponse(Session session) {
-        return "+ \"\"";
+        return "\"\"\r\nOK";
```
chibenwa: Why do we need changing the initial server response here?

server/protocols/protocols-managesieve/src/test/java/org/apache/james/managesieveserver/AuthenticateTest.java
```diff
@@ -0,0 +1,220 @@
...
+        this.client.sendCommand("AUTHENTICATE \"PLAIN\" \"" + Base64.getEncoder().encodeToString(initialClientResponse.getBytes(StandardCharsets.UTF_8)) + "\"");
+        ManageSieveClient.ServerResponse authenticationResponse = this.client.readResponse();
+        Assertions.assertThat(authenticationResponse.responseType()).isEqualTo(ManageSieveClient.ResponseType.NO);
+    }
+
+    // The SASL PLAIN standard (https://datatracker.ietf.org/doc/html/rfc4616) defines the following message:
+    // message = [authzid] UTF8NUL authcid UTF8NUL passwd
+    // The current code is more lenient and accepts the message without the first null byte.
+    @Disabled
+    @Test
```
chibenwa, line 85: I we think it is a good thing to be more relaxed than the RFC, I see no issue:

 - Keeping the valuable comment
 - Testing the existing behaviour with a passing test.

protocols/managesieve/src/main/java/org/apache/james/managesieve/core/PlainAuthenticationProcessor.java
felixauringer: I do not understand where this `+ ""` comes from but I didn't find it in any of the RFCs relating to ManageSieve SASL authentication.  
There is one mention in the [SASL RFC](https://www.rfc-editor.org/rfc/rfc4422#appendix-A) but only for the `EXTERNAL` mechanism which we do not use here.  
The in my opinion relevant RFC here is [RFC 5804: A Protocol for Remotely Managing Sieve Scripts](https://datatracker.ietf.org/doc/html/rfc5804#section-4). The formal syntax there defines the response to an authenticate command as follows:

```
response-authenticate = *(string CRLF)
                            ((response-ok [response-capability]) /
                             response-nobye)
                            ;; <response-capability> is REQUIRED if a
                            ;; SASL security layer was negotiated and
                            ;; MUST be omitted otherwise.
```

If we want continuation, we want to have a `response-ok` and because we do not negotiate a SASL security layer, we must omit `response-capability`.  
It then simplifies to `response-authenticate = *(string CRLF) response-ok`. Assuming we do not want to have a response code or description, it further simplifies to `response-authenticate = *(string CRLF) "OK" CRLF`.  
The [description](https://datatracker.ietf.org/doc/html/rfc5804#section-2.1) says:

> Note that an empty challenge/response is sent as an empty string.

With all our supported authentication mechanisms, the challenge sent by the server is empty, so I would argue that the server should send the empty string. Sadly, the examples in the RFC do not show this case. Using the formal syntax from above, this leads to `response-authenticate = DQUOTE DQUOTE CRLF "OK" CRLF`. (One could probably also use the literal string variant without quotes but by supplying the exact string length. However, I would argue that it does not make much sense with an empty string.)

That's the reason why I changed it. In the RFC, I do not see any possibility how an (unquoted!) `+` sign could be the response to an authenticate command. I also didn't find why this server response shouldn't end with an `OK`, but I'm less sure about that (the examples seem to omit it in the challenge-response exchanges but I don't think the formal syntax allows it). What's the reason why you use the `+` here? Is there another RFC that I do not know?

mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/capability.test
felixauringer: Yes, I am pretty sure that having spaces as separator is not compliant. If that's okay for you, I am fine with removing them.

The [RFC](https://datatracker.ietf.org/doc/html/rfc4616) actually also does not mention base64 but I also have never seen it without it.

The current behavior implementation allows the following for `PLAIN`:

- null as delimiter, then base64 encoded, then quoted
- null as delimiter, then quoted
- space as delimiter, then quoted

(strings always need to be quoted except if the length is given beforehand)

Which of those behaviors would you like to be supported? I implemented all of them because the other protocols allow these options.

protocols/managesieve/src/main/java/org/apache/james/managesieve/core/PlainAuthenticationProcessor.java
felixauringer: Okay, I found the [PLAIN SASL RFC](https://datatracker.ietf.org/doc/html/rfc4616#section-4) from where your previous server response seems to come from. I really do think that it conflicts with the managesieve RFC (5804) but I am fine with keeping the old syntax then.

chibenwa: Hello @felixauringer 

Thanks for having a look

I think as an implementer I just sticked to RFC-4616 exemples, yes (but that's10 years back.... I'm a bit rusty on the topic)

Note that the `""` is likely none compulsory and not returned eg by the IMAP layer which also implement RFC-4616...

While what comes after is subject to discussion the fact of doing a continuation is mandatory.

mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/capability.test
felixauringer: Because it was supported before, I added an additional test ensuring that a space still works as delimiter here.

protocols/managesieve/src/main/java/org/apache/james/managesieve/core/PlainAuthenticationProcessor.java
felixauringer: I changed my code and all tests to behave like before and send `+ ""`.

server/protocols/protocols-managesieve/src/test/java/org/apache/james/managesieveserver/AuthenticateTest.java
felixauringer: I enabled the test and kept the comment.

felixauringer:
> Otherall agree with proposed changes but I would need clarifications on `AUTHENTICATE "PLAIN"` handling.

Now, the behavior is the same except that the code is more strict when it comes to quoting. SASL mechanisms and the values sent by the client must be quoted according to the RFC. (In the RFC, there is the possibility of sending unquoted data which must be preceded by the length in the form `{20+}` but that was not supported before and is not supported now.)

mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/capability.test
```diff
@@ -39,7 +39,7 @@ S: OK
 
 C: AUTHENTICATE "PLAIN"
 S: \+ ""
-C:  user password
+C: "user password"
```
chibenwa: When authentication happens as part of continuation I am rather sure that the crediential must nt be quoted


```suggestion
C: user password
```

CF https://datatracker.ietf.org/doc/html/rfc4616#section-4

chibenwa (commented):
Thanks for the extensive test suite it's pleasan to read.

mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/capability.test
felixauringer: I have added a commit that ensures continuation must be sent without quotes and adapts the tests accordingly.

However, I am not sure that your reference supports this behavior. All examples using continuation use the syntax where the unquoted message is preceded by the exact length of the client message. With that syntax, it is of course not necessary to quote the client message.

chibenwa (commented):
This work looks nice to me.
I can give a hand with the picky test suite if of any use. 
Just ask.
Best regards,
Benoit

mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/capability.test
```diff
@@ -39,7 +39,7 @@ S: OK
 
 C: AUTHENTICATE "PLAIN"
 S: \+ ""
-C:  user password
+C: user password
```
chibenwa, line 42: CF exemples in https://datatracker.ietf.org/doc/html/rfc4616

Example mentions `<NUL>tim<NUL>tanstaaftanstaaf`

We did test those inputs by using ` ` (space) and `<NUL> in an inter-changeable way hence I explexpect we are still able to handle `C:  user password`

Is it still the case?
Can we have a tet making sure this is still the case?

felixauringer: I would argue that the line before the one you referenced is also important here: `C: {21}`
The clients announces the exact length of its message. This syntax is defined in the managesieve standard but currently not implemented in James. I do not know whether there are any clients using it.

The exact line `C: user password` appears multiple times in the managesieve MPL tests that were there before my contribution.
I think in the MPL tests only use plain authentication with continuation, use spaces instead of NUL and do not quote the client message during continuation.
The Java Tests I wrote usually use NUL (including the leading NUL) and encode it with base64. However, there is also one test for accepting it without the leading NUL and one that accepts the credentials with one space between username and password (not base64 encoded). There are no Java tests that do test those variants also with continuation.

I am not exactly sure what additional tests you would like to have.

felixauringer: I think there are currently no tests for a leading space when the credentials are not base64 encoded, do you mean that?

chibenwa: Yes

chibenwa:
> I can give a hand with the picky test suite if of any use.

Good job it's :green_apple: :-)

mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/capability.test
felixauringer: I added an additional test.

chibenwa:
Recent changes regarding SASL improvement impact this PR

```
10:42:32,414 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.14.0:compile (default-compile) on project protocols-managesieve: Compilation failure: Compilation failure: 
more_vert
9765	11:42:32 AM	
10:42:32,414 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/api/Session.java:[27,38] cannot find symbol
more_vert
9766	11:42:32 AM	
10:42:32,414 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9767	11:42:32 AM	
10:42:32,414 [ERROR]   location: package org.apache.james.protocols.api
more_vert
9768	11:42:32 AM	
10:42:32,414 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/api/Session.java:[57,14] cannot find symbol
more_vert
9769	11:42:32 AM	
10:42:32,414 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9770	11:42:32 AM	
10:42:32,414 [ERROR]   location: interface org.apache.james.managesieve.api.Session
more_vert
9771	11:42:32 AM	
10:42:32,414 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/api/Session.java:[59,44] cannot find symbol
more_vert
9772	11:42:32 AM	
10:42:32,414 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9773	11:42:32 AM	
10:42:32,414 [ERROR]   location: interface org.apache.james.managesieve.api.Session
more_vert
9774	11:42:32 AM	
10:42:32,414 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/core/OAUTHAuthenticationProcessor.java:[34,38] cannot find symbol
more_vert
9775	11:42:32 AM	
10:42:32,414 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9776	11:42:32 AM	
10:42:32,414 [ERROR]   location: package org.apache.james.protocols.api
more_vert
9777	11:42:32 AM	
10:42:32,414 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/core/OAUTHAuthenticationProcessor.java:[40,19] cannot find symbol
more_vert
9778	11:42:32 AM	
10:42:32,414 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9779	11:42:32 AM	
10:42:32,414 [ERROR]   location: class org.apache.james.managesieve.core.OAUTHAuthenticationProcessor
more_vert
9780	11:42:32 AM	
10:42:32,414 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/core/OAUTHAuthenticationProcessor.java:[42,41] cannot find symbol
more_vert
9781	11:42:32 AM	
10:42:32,414 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9782	11:42:32 AM	
10:42:32,415 [ERROR]   location: class org.apache.james.managesieve.core.OAUTHAuthenticationProcessor
more_vert
9783	11:42:32 AM	
10:42:32,415 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/util/SettableSession.java:[28,38] cannot find symbol
more_vert
9784	11:42:32 AM	
10:42:32,415 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9785	11:42:32 AM	
10:42:32,415 [ERROR]   location: package org.apache.james.protocols.api
more_vert
9786	11:42:32 AM	
10:42:32,415 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/util/SettableSession.java:[36,22] cannot find symbol
more_vert
9787	11:42:32 AM	
10:42:32,415 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9788	11:42:32 AM	
10:42:32,415 [ERROR]   location: class org.apache.james.managesieve.util.SettableSession
more_vert
9789	11:42:32 AM	
10:42:32,415 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/util/SettableSession.java:[89,21] cannot find symbol
more_vert
9790	11:42:32 AM	
10:42:32,415 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9791	11:42:32 AM	
10:42:32,415 [ERROR]   location: class org.apache.james.managesieve.util.SettableSession
more_vert
9792	11:42:32 AM	
10:42:32,415 [ERROR] /home/jenkins/workspace/james_ApacheJames_PR-2773/protocols/managesieve/src/main/java/org/apache/james/managesieve/util/SettableSession.java:[94,51] cannot find symbol
more_vert
9793	11:42:32 AM	
10:42:32,415 [ERROR]   symbol:   class OidcSASLConfiguration
more_vert
9794	11:42:32 AM	
10:42:32,415 [ERROR]   location: class org.apache.james.managesieve.util.SettableSession
more_vert
9795	11:42:32 AM	
10:42:32,415 [ERROR] -> [Help 1]
```

A rebase is needed. If needed I can do it.

felixauringer:
I have rebased on the newest commit on main.
If the tests are green, I still have to adapt the documentation (but as far as I can see, this has to be done for everything using the new audience handling).

Before the new `validateToken` method was introduced, I had already written tests in the Managesieve implementation for all different cases regarding introspection, local validation, userinfo. Those tests would be better located in `OidcJwtTokenVerifierTest` now in my opinion. Do you agree? If so, I could move them there (currently there are no tests for the different cases in the config as far as I can tell) and focus more on the Managesieve related functionality in the Managesieve tests.
To keep my old tests working, I also had to adapt your new code a little bit. In `OidcSASLConfiguration`, there were two static variables relying on System Properties. As they are set once in the JVM, I did not find a good way to run tests with different values for them. They are therefore now read when the configuration is parsed. This should not happen often and therefore these additional calls of `System.getProperty` should be fine in my opinion but please correct me if you know better.

chibenwa (approved):
Sorry for the (very) long awaited review.

Cheers.

Arsnael (approved):
LGTM. Thanks for this contribution @felixauringer :)

chibenwa:
I merged this work as it had been open for a real long time.

Thanks for putting it together @felixauringer 

Other enhancements are likely to be done as follow up PRs if relevant.

PR #2778, opened 2025-07-25
JAMES-3775 ClamAVScan should be able to configure socket timeout when scanning a message

quantranhong1999:


third-party/clamav/src/main/java/org/apache/james/clamav/ClamAVScan.java
```diff
@@ -389,6 +396,12 @@ protected void initPingIntervalMilli() {
         }
     }
 
+    protected void initSocketTimeoutMilli() {
+        this.socketTimeoutMilli = Optional.ofNullable(getInitParameter("socketTimeoutMilli"))
+            .map(Integer::parseInt)
+            .orElse(DEFAULT_SOCKET_TIMEOUT_MILLI);
```
chibenwa: Let's use `DurationParser` and work only with `Duration` please

docs/modules/servers/partials/ClamAVScan.adoc
```diff
@@ -30,6 +30,9 @@ The default is *6*.
 * *<pingIntervalMilli>*: the interval (in milliseconds)
```
chibenwa: Lets use `DurationParser` here too please.

We can default to MilliSecond when no unit in order to preserve inter-operability.

chibenwa (approved):

PR #2780, opened 2025-07-29
[UPGRADE] Docker cassandra image to 5.0.4

Arsnael:
Need to perf test. There is no driver upgrade, `4.19.0` is still the latest one


quantranhong1999 (approved):

Arsnael:
dateof function has been removed in cassandra 5.0, should use toTimestamp now

Arsnael:
Culprit on our side with compression usage: https://github.com/apache/james-project/blob/master/server/blob/blob-cassandra/src/main/java/org/apache/james/blob/cassandra/cache/CassandraBlobCacheDataDefinition.java#L38

Arsnael:
@quantranhong1999 good remarks actually. But would there be? 

The tables would exist already after a migration, wouldnt it be fair to think the compression option on the table would be migrated as well? The problem would only occur if we setup a new cassandra 4 cluster with the updated java driver version... but wouldnt we want just running cass 5 on a new setup right away?

quantranhong1999:
> The tables would exist already after a migration, wouldnt it be fair to think the compression option on the table would be migrated as well?

Yes. I think it should the table should be migrated when upgrading to Cassandra 5.0.

> The problem would only occur if we setup a new cassandra 4 cluster with the updated java driver version... but wouldnt we want just running cass 5 on a new setup right away?

I am not sure the community would upgrade to Cassandra 5.0 soon. Then 4.19.1 driver would be a breaking change for existing 4.x Cassandra, for example, no? 
(Unless we remove the compaction options like your latest commit for example)

Arsnael:
The discussion would benefit from community opinion I agree. I'm waiting to see if there is other issues popping up before potentially starting a thread on the topic

chibenwa:
> The discussion would benefit from community opinion I agree.

+1

Arsnael:
@chibenwa sounds reasonable to me

chibenwa:
Rebased, upgraded to 5.0.9, removed upgrade instructions...

backends-common/cassandra/src/test/java/org/apache/james/backends/cassandra/DockerCassandra.java
```diff
@@ -137,7 +137,7 @@ private static Optional<String> getFixedBuildId() {
 
     @SuppressWarnings("resource")
     public DockerCassandra() {
-        this("cassandra_4_1_5-" + buildSpecificImageDiscriminator(),
+        this("cassandra_5_0_4-" + buildSpecificImageDiscriminator(),
```
quantranhong1999:
```suggestion
        this("cassandra_5_0_9-" + buildSpecificImageDiscriminator(),
```

quantranhong1999 (approved):

Arsnael:
+1

PR #2781, opened 2025-07-29
[IMPROVEMENT] CapabilityFactory should take Username into account + be reactive

quantranhong1999:
It is a common use case that capability could be varied per User, e.g. in a SaaS deployment where users can have different plan capabilities.

Also, allow evaluating capability to be reactive.

### DIFF
--- a/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/SessionSupplier.scala
+++ b/server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/SessionSupplier.scala
@@ -44,26 +44,21 @@ class SessionSupplier(capabilityFactories: Set[CapabilityFactory], configuration
   def generate(username: Username, delegatedUsers: Set[Username], urlPrefixes: UrlPrefixes): SMono[Session] = {
     val urlEndpointResolver: JmapUrlEndpointResolver = new JmapUrlEndpointResolver(urlPrefixes)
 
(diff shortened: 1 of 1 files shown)


server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/SessionSupplier.scala
```diff
@@ -40,28 +41,38 @@ class SessionSupplier(capabilityFactories: Set[CapabilityFactory], configuration
...
-        downloadUrl = urlEndpointResolver.downloadUrl,
-        uploadUrl = urlEndpointResolver.uploadUrl,
-        eventSourceUrl = urlEndpointResolver.eventSourceUrl)
-    }
+    evaluateCapabilities(username, urlPrefixes)
+      .flatMap { capabilities =>
+        accounts(username, capabilities) match {
+          case Left(e) => SMono.error(e)
+          case Right(account) =>
+            delegatedAccounts(delegatedUsers, capabilities) match {
```
Arsnael: two nested match a bit much? Maybe should extract that last bit?

quantranhong1999: Alright, it should look elegant now.

chibenwa (approved):

Arsnael (approved):

Arsnael:
Green! Can you squash the fixup @quantranhong1999 ?

PR #2783, opened 2025-08-01
[FIX] Old quota tables should not be recreated after migration to the…

Arsnael:
… new ones


Arsnael:
Spotted this after cleanup up old tables in an environment

quantranhong1999 (approved):

chibenwa (approved):

PR #2785, opened 2025-08-12
[INFRA-25461] Remote build cache

jeantil:
This PR configures the remote build cache for james now that infra has setup the cache node, it also cleans up a build warning because of duplicated lib declarations ( see specific commit )

### DIFF
--- a/.mvn/develocity.xml
+++ b/.mvn/develocity.xml
@@ -50,7 +50,8 @@
       </cleanup>
     </local>
     <remote>
-      <storeEnabled>false</storeEnabled>
+      <enabled>true</enabled>
+      <storeEnabled>#{isTrue(env['CI'])}</storeEnabled>
       <!--
       the next step is this
       <storeEnabled>#{isTrue(env['CI'])</storeEnabled>
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -96,7 +96,7 @@ pipeline {
         stage('Stable Tests') {
             steps {
                 echo 'Running tests'
-                sh 'mvn -B -e -fae test ${MVN_SHOW_TIMESTAMPS} -P ci-test ${MVN_LOCAL_REPO_OPT} -Dassembly.skipAssembly=true jacoco:report-aggregate@jacoco-report'
+                sh 'mvn -B -e -fae clean test ${MVN_SHOW_TIMESTAMPS} -P ci-test ${MVN_LOCAL_REPO_OPT} -Dassembly.skipAssembly=true jacoco:report-aggregate@jacoco-report'
             }
             post {
                 always {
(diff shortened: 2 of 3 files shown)


.mvn/develocity.xml
```diff
@@ -50,7 +50,8 @@
       </cleanup>
     </local>
     <remote>
-      <storeEnabled>false</storeEnabled>
+      <enabled>true</enabled>
```
jeantil, line 53: everyone can download built artifacts, but only the CI is allowed to store them

chibenwa (approved):
Nice ;-)

Jenkinsfile
```diff
@@ -96,7 +96,7 @@ pipeline {
         stage('Stable Tests') {
             steps {
                 echo 'Running tests'
-                sh 'mvn -B -e -fae test ${MVN_SHOW_TIMESTAMPS} -P ci-test ${MVN_LOCAL_REPO_OPT} -Dassembly.skipAssembly=true jacoco:report-aggregate@jacoco-report'
+                sh 'mvn -B -e -fae clean test ${MVN_SHOW_TIMESTAMPS} -P ci-test ${MVN_LOCAL_REPO_OPT} -Dassembly.skipAssembly=true jacoco:report-aggregate@jacoco-report'
```
jeantil, line 99: clean is required for storing results in the build cache 
the remote build cache will ensure we don't spend to much rebuilding already built stuff

anyway build time is negligible compared to test runtime so event with suboptimal caching we should benefit

Jenkinsfile
```diff
@@ -89,14 +89,14 @@ pipeline {
         stage('Build') {
             steps {
                 echo 'Building'
-                sh 'mvn -U -B -e clean install -DskipTests -Djib.skip -T1C ${MVN_SHOW_TIMESTAMPS} ${MVN_LOCAL_REPO_OPT}'
+                sh 'mvn -U -B -e clean install -DskipTests -Djib.skip  -DrerunGoals -T1C ${MVN_SHOW_TIMESTAMPS} ${MVN_LOCAL_REPO_OPT}'
```
jeantil: to be removed as I seem to have polluted the cache with bad outputs

jeantil:
build 12 success,  the [build scan](https://develocity.apache.org/s/7z45cnk2pse5a/performance/execution) confirms the 5hxx of tests are cacheable. Stable test lasted 5h39 which is the worst case and will only trigger when changing somehting that requires rerruning all tests
build 13 success ([build scan](https://develocity.apache.org/s/7y22vlkhwudju)), Stable test lasted 22minutes ! saving a bit over 5h for the best case scenario : everything from cache.

These results will be further improved by making 2 goals cacheable: 
- scala compile 
- jacoco aggregate test report 

I have a POC for scala compile but I don't think its mergeable as is, I had to exclude some properties from the cache key computation that I would rather to make sure we invalidate the cache if they change

pom.xml
```diff
@@ -4041,69 +4043,6 @@
                                 </execution>
                             </executions>
                         </plugin>
-                        <plugin>
```
jeantil, line 4044: we need these goals to be included since the cache requires the clean goal but including them is painless because the cache takes care of most of the cost

pom.xml
```diff
@@ -4112,25 +4051,21 @@
...
+                                    <id>strict-pom-ordering</id>
                                     <phase>none</phase>
                                 </execution>
                             </executions>
                         </plugin>
                         <plugin>
-                            <groupId>com.github.ekryd.sortpom</groupId>
-                            <artifactId>sortpom-maven-plugin</artifactId>
+                            <groupId>io.github.evis</groupId>
+                            <artifactId>scalafix-maven-plugin_2.13</artifactId>
```
jeantil, line 4065: we do not need to rerun scalafix (which is not cached) on the stable tests

PR #2789, opened 2025-08-24
JAMES-4133 Allow to announce SMTPUTF8

chibenwa:
Third party systems might require SMTPUT8 even when not needed thus affecting delivery.


chibenwa:
Indeed...

Sorry...

PR #2801, opened 2025-09-10
JAMES-4143 Jmap filter start-with

Insalien:
Issue: https://issues.apache.org/jira/browse/JAMES-4143

Changes:
- added new jmap filter start-with
- added UT tests and serialization / de-serialization tests

### DIFF
--- a/server/protocols/jmap-rfc-8621/src/test/java/org/apache/james/jmap/mailet/filter/JMAPFilteringTest.java
+++ b/server/protocols/jmap-rfc-8621/src/test/java/org/apache/james/jmap/mailet/filter/JMAPFilteringTest.java
@@ -1147,7 +1147,10 @@ class JMAPFilteringTest {
     @Test
     void startWithAddressMailNotMatched(JMAPFilteringTestSystem testSystem) throws Exception {
         testSystem.defineRulesForRecipient1(
-            Rule.Condition.of(FROM, START_WITH, "abcdef")
+            // start-with a random text
+            Rule.Condition.of(FROM, START_WITH, "abcdef"),
+            // start-with behaves differently from contains
+            Rule.Condition.of(FROM, START_WITH, USER_1_ADDRESS.substring(2, 5))
         );
 
         FakeMail mail = testSystem.asMail(mimeMessageBuilder()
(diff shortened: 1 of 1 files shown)


server/data/data-jmap-cassandra/src/test/resources/json/eventComplex-v2.json
```diff
@@ -3,6 +3,24 @@
   "eventId":0,
   "aggregateId":"FilteringRule/bart",
   "rules":[
+    {
```
quantranhong1999, line 6: IMO, it is not a good practice to modify the backward version JSON(s) to ensure backward compatibility for the deserializer remains (despite likely the change in this PR is not a breaking change).

Maybe we can create a new JSON to test the start with?

server/protocols/jmap-rfc-8621/src/test/java/org/apache/james/jmap/mailet/filter/JMAPFilteringTest.java
```diff
@@ -1129,4 +1130,56 @@ void orShouldNotMatchWhenNoConditionsAreMet(JMAPFilteringTestSystem testSystem)
 
         assertThat(mail.getAttribute(RECIPIENT_1_USERNAME_ATTRIBUTE_NAME).isEmpty()).isTrue();
     }
+
+    @Test
+    void startWithAddressMailMatched(JMAPFilteringTestSystem testSystem) throws Exception {
+        testSystem.defineRulesForRecipient1(
+            Rule.Condition.of(FROM, START_WITH, USER_1_ADDRESS.substring(0, 2))
```
chibenwa, line 1137: Add a test from 2->5 that do not match

(distinguish startWith and contains)

server/data/data-jmap-cassandra/src/test/resources/json/eventComplex-v2.json
chibenwa: The JSON structure do not change (no new field added nor removed) so it's acceptable as the "shape" of the test suite did not change and thus no risk / compatibility concerns arise.

(format do not change)

chibenwa (approved):

chibenwa (approved):

PR #2805, opened 2025-09-13
[JAMES-4145] fixes unnecessary dependencies 

jeantil:
Removes 
- james-server-postgres-common-guice -> james-server-guice-distributed (which brings cassandra, opensearch and rabbitmq to a module that's suppose to provide pg bindings)
- james-server-guice-common -> james-server-data-jmap (guice-common is pulled everywhere and should only contain bindings which are required by all apps, protocol specific bindings shouldn't be there)
 
Arguably mailbox, task and queue shouldn't be there either but I don´t have the courage right now.

This change moves 2 bindings to the james postgres app itself, I find it really weird to have a binding module for cassandra types in the postgres dependencie. I'm not sure if the problem is in having the DLP related types in a cassandra module or if it is having DLP in postgres.


server/apps/postgres-app/src/main/java/org/apache/james/modules/data/PostgresDLPConfigurationStoreModule.java
```diff
@@ -31,6 +31,7 @@
 import com.google.inject.TypeLiteral;
 import com.google.inject.multibindings.Multibinder;
 
+//FIXME - I don't see what this has to do with postgres when it pulls stuff from data-cassandra ...
```
chibenwa, line 34: I pulls from cassandra the event serialisation.

We could argue that `org.apache.james.dlp.eventsourcing.cassandra.DLPConfigurationModules` could be relocated in `data-library` where both the PG and the Cassandra implen can depend on it.

chibenwa (approved):
Thanks for the cleanup and sorry for missing the Cassandra hidden dependency upon review...

chibenwa:
```
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal com.github.ekryd.sortpom:sortpom-maven-plugin:4.0.0:verify (strict-pom-ordering) on project james-server-guice-mailbox-postgres: The file /home/jenkins/jenkins-agent/workspace/james_ApacheJames_PR-2805/server/container/guice/mailbox-postgres/pom.xml is not sorted
```

If you want I can fix it

server/apps/postgres-app/src/main/java/org/apache/james/modules/data/PostgresDLPConfigurationStoreModule.java
jeantil: >We could argue that org.apache.james.dlp.eventsourcing.cassandra.DLPConfigurationModules could be relocated in data-library where both the PG and the Cassandra implen can depend on it.

sounds good to me

jeantil:
I think my next PR will simply enable the automatic sorting of the pom files by default, I'll run it on all the project's pom and add the commit to the git blame ignore list and be done with trying to sort things by hand in xml

PR #2806, opened 2025-09-13
JAMES-3758 Expiration: specific user and possibility to use saved date

chibenwa:
Refactor the class to choose between the different message listing strategies. We did also review the reactive chain to make it more readable.


quantranhong1999:
> I restarted a build to see if it was a random failure from the builder and the build job succeeded, so I would say it was just a fluke.

Thank you. The build looks good now.

PR #2808, opened 2025-09-14
JAMES-4146: Add support for proxy protocol in managesieve

felixauringer:
I found that it is possible to configure proxy protocol for managesieve but the configuration is ignored.
This fixes that. I did not find any logic in managesieve that tries to act on the real client IP, so it is available now in the MDCContext but it is never used.

There is already a general `BasicChannelInboundHandler` which also decodes the proxy protocol, but as far as I understand, it is only used for SMTP. ManageSieve and IMAP both have their own InboundHandler.

Find the Jira issue [here](https://issues.apache.org/jira/projects/JAMES/issues/JAMES-4146).


chibenwa (approved):
Nice thanks!

Arsnael (approved):
Cheers for the fix!

PR #2809, opened 2025-09-14
JAMES-4135 OpenSearch: make query string usage lenient

chibenwa:
Because a strict interpretation of the syntax result in user facing errors


Arsnael (approved):

quantranhong1999 (approved):

jeantil:
Slightly offtopic food for thoughts:  

The build took ~1h30 instead of 4h thanks to build caching which I think is a good thing. 

The main contributors to the remaining build time can be seen in [this buildscan](https://develocity.apache.org/s/enx3owqeum664/timeline?cacheability=cacheable&hide-timeline&outcome=success,failed&sort=longest) 
 


40 minutes seem to stem from only 4 testsuites. Considering the length of execution of these test suites,I wonder if we could reclaim build time by dropping some integration tests that only test again what is already tested elsewhere.

chibenwa:
> 40 minutes seem to stem from only 4 testsuites. Considering the length of execution of these test suites,I wonder if we could reclaim build time by dropping some integration tests that only test again what is already tested elsewhere.

I'll open a ticket for linagorians to investigate the 5 min + maven projects

quantranhong1999:
> 40 minutes seem to stem from only 4 testsuites.

Hi @jeantil, I am curious here (sorry, I am not an expert in build caching). I see those "40 minutes" goal(s) are `Cacheable`. Does that mean results for those tests would be cached and would not be run in the next build?

PR #2810, opened 2025-09-15
[ENHANCEMENT] Smtp technical accounts

chibenwa:
My need:

Port 25: authentication is disabled
Port 465: submission, user auth, strict sender validation
Port 26: opened locally with k8s cluster. I want to add auth for it not to be opened. 

I do not want to create user for port 26 because:
 - I do not want to have a mailbox assiciated to it
 - I want the set up to be declarative to limit to the maximum manual actions
 
The idea is to deploy to port 26 a specific auth mechanism matching configuration. 
 
```
    <handlerchain coreHandlersPackage="org.apache.james.smtpserver.NoAuthCmdHandlerLoader" enableJmx="false">
        <handler class="org.apache.james.protocols.smtp.core.esmtp.AuthCmdHandler" />
        <handler class="org.apache.james.smtpserver.SetMailAttributeMessageHook" >
            <name>technicaluser</name>
            <value>true</value>
        </handler>
        <handler class="org.apache.james.smtpserver.ConfigurationAuthHook" >
            <accounts>
                <account>
                    <username>noreply-tdrive@domain.tld</username>
                    <passwords>
                        <password>secret123456</password>
                        <password>here_to_ease_secret_rotation</password>
                        <password>here_to_give_different_creds_to_each_app</password>
                    </passwords>
                </account>
                <account>
                    <username>noreply-tchat@domain.tld</username>
                    <passwords>
                        <password>secret234567</password>
                    </passwords>
                </account>
            </accounts>
        </handler>
    </handlerchain>
```

Env variables (and so secrets!) might be used for passwords...

I also add a way to position mail attributes. This would enable in my use case to for instance bypass rate limiting.

I decided to contribute this to ASF James as :
 - it is a fairly common use case
 - it can enable writing a James server without dependency on user repistories and potentially event without a DB, just a mail queue...

If non consensual I can move the code to LINAGORA repos.

### DIFF
--- a/docs/modules/servers/partials/configure/smtp-hooks.adoc
+++ b/docs/modules/servers/partials/configure/smtp-hooks.adoc
@@ -424,3 +424,76 @@ Would allow emails using `apache.org` as a MAIL FROM or from header domain if, a
 valid DKIM signature for the `apache.org` domain.
 
 include::partial$EnforceHeaderLimitationsMessageHook.adoc[]
+
+== SetMailAttributeMessageHook
+
+This class implements an SMTP hook to add attributes on incoming mails.
+
+It allows marking mails from a given SMTP port. Eg if I set up technical accounts for local apps on port 26
+then I can disable rate limiting for this traffic.
+
+Example XML configuration:
+
+....
+<handler class="org.apache.james.smtpserver.SetMailAttributeMessageHook">
+    <name>technicaluser</name>
+    <value>true</value>
+</handler>
+....
+
+== EnforceHeaderLimitationsMessageHook
+
+This class implements an SMTP hook to enforce limitations on the headers of incoming emails.
+
+It allows configuring and enforcing two types of restrictions:
+ - A maximum number of header lines (default: 500).
+ - A maximum total size of headers in bytes (default: 64 KB).
+If any of these thresholds are exceeded, the message is rejected with an SMTP error code:
+ -  `552 Too many header lines` if the number of lines exceeds the limit.
+ -  `552 Header size too large` if the total size exceeds the limit.
+
+Example XML configuration:
+
+....
+<handler class="org.apache.james.smtpserver.EnforceHeaderLimitationsMessageHook">
+    <maxLines>500</maxLines>
+    <maxSize>64KB</maxSize>
+</handler>
+....
+
+== ConfigurationAuthHook
+
+Declarative authentication.
+
+It is possible to open and configure on a dedicated port (eg port 26) to accept application traffic in parallel of user
+traffic. Authentication is then done on the supplied configuration.
+
+Convenient as it allows instantiating James as part of a complex application network with only only declarative configuration
+and no post-deployment actions.
+
+This is helpful for things like service accounts, used by other applications (and it is not desirable to create
+user accounts for those applications)
+
+....
+    <handlerchain coreHandlersPackage="org.apache.james.smtpserver.NoAuthCmdHandlerLoader" enableJmx="false">
+        <handler class="org.apache.james.protocols.smtp.core.esmtp.AuthCmdHandler" />
+        <handler class="org.apache.james.smtpserver.ConfigurationAuthHook" >
+            <accounts>
+                <account>
+                    <username>noreply-tdrive@domain.tld</username>
+                    <passwords>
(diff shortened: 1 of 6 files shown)


server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/NoAuthCmdHandlerLoader.java
```diff
@@ -56,7 +56,6 @@ public class NoAuthCmdHandlerLoader implements HandlersPackage {
             RsetCmdHandler.class.getName(),
             VrfyCmdHandler.class.getName(),
             MailSizeEsmtpExtension.class.getName(),
-            UsersRepositoryAuthHook.class.getName(),
```
jeantil, line 59: Why is this removed ? it doesn't seem the match the rest of the PR

jeantil:
Nice ! I find these quite interesting :D

docs/modules/servers/partials/configure/smtp-hooks.adoc
```diff
@@ -423,4 +423,77 @@ Example handlerchain configuration for `smtpserver.xml`:
...
+....
+    <handlerchain coreHandlersPackage="org.apache.james.smtpserver.NoAuthCmdHandlerLoader" enableJmx="false">
+        <handler class="org.apache.james.protocols.smtp.core.esmtp.AuthCmdHandler" />
+        <handler class="org.apache.james.smtpserver.ConfigurationAuthHook" >
+            <accounts>
+                <account>
+                    <username>noreply-tdrive@domain.tld</username>
+                    <passwords>
+                        <password>secret123456</password>
+                        <password>here_to_ease_secret_rotation</password>
```
jeantil, line 486: yes ! definitely a good idea

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/NoAuthCmdHandlerLoader.java
chibenwa: No to the contrary it is VERY related.

The idea is that because there is no `AuthCmdHandler` then `UsersRepositoryAuthHook` is unused.

Also the idea is to cleanup `NoAuthCmdHandler` so that it can be freely reused with other AuthHooks and thus easily and fully customize the authentication backend as is done here.

docs/modules/servers/partials/configure/smtp-hooks.adoc
chibenwa: Been there ... :-p

server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/ConfiguredAuthTest.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    @ParameterizedTest
+    @CsvSource({
+        "noreply-tdrive@domain.tld, bad",
+        "noreply-notfound@domain.tld, here_to_ease_secret_rotation",
+        "noreply-tdrive@other-domain.tld, secret123456",
+        "noreply-tdrive, secret123456"
+    })
+    void authShouldBeDeniedWhenMatchingConfiguredValue(String username, String password) throws Exception {
```
Arsnael:
```suggestion
    void authShouldBeDeniedWhenNotMatchingConfiguredValue(String username, String password) throws Exception {
```

Arsnael (approved):
Just a small typo but otherwise LGTM :)

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/SetMailAttributeMessageHook.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ */
+
+public class SetMailAttributeMessageHook implements JamesMessageHook {
+    private String name;
+    private String value;
+
+    @Override
+    public HookResult onMessage(SMTPSession session, Mail mail) {
+        System.out.println("Marking");
```
quantranhong1999:
```suggestion
```

PR #2811, opened 2025-09-15
JAMES-4143 Add a custom header Field for JMAP rule matching

chibenwa:


server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/Rule.java
```diff
@@ -129,33 +129,58 @@ public String toString() {
...
-            TO("to"),
-            CC("cc"),
-            SUBJECT("subject"),
-            RECIPIENT("recipient");
+        public interface Field {
+            Field FROM = new FixedField("from");
+            Field TO = new FixedField("to");
+            Field CC = new FixedField("cc");
+            Field SUBJECT = new FixedField("subject");
+            Field RECIPIENT = new FixedField("recipient");
```
Arsnael: Duplicated with declarations in FixedField? Those ain't necessary in the interface?

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/Rule.java
```diff
@@ -129,33 +129,52 @@ public String toString() {
...
             public String asString() {
                 return fieldName;
             }
         }
+
+        public record CustomHeaderField(String headerName) implements Field {
+            public static final String PREFIX = "header:";
+
+            public static Optional<Field> find(String fieldName) {
+                if (fieldName.startsWith(PREFIX)) {
```
quantranhong1999, line 168: Should the fieldName be lowercase first, to allow the input `"HEADER:custom"` can be case insensitive?

quantranhong1999 (approved):

server/data/data-jmap/src/main/java/org/apache/james/jmap/api/filtering/Rule.java
chibenwa: IMO not needed as we expect machines to supply this input. Sanitizing the input is thus not relevant to me.
Case insentivity of the header is handled by `jakarta.mail` itself.

Arsnael (approved):

PR #2812, opened 2025-09-16
JAMES-4057 Align subject search with other implementations

chibenwa:
(sorry for missing thins out)

### DIFF
--- a/mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/query/DefaultCriterionConverter.java
+++ b/mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/query/DefaultCriterionConverter.java
@@ -297,7 +297,7 @@ public class DefaultCriterionConverter implements CriterionConverter {
                         .toQuery();
                 }
             case ATTACHMENTS:
-                if (useQueryStringQuery) {
+                if (useQueryStringQuery && QUERY_STRING_CONTROL_CHAR.matchesAnyOf(textCriterion.getOperator().getValue())) {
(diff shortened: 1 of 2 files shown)


quantranhong1999 (commented):
Read it

mailbox/opensearch/src/main/java/org/apache/james/mailbox/opensearch/query/DefaultCriterionConverter.java
```diff
@@ -331,21 +330,33 @@ protected Query convertTextCriterion(SearchQuery.TextCriterion textCriterion) {
...
+                            .fields(ImmutableList.of(JsonMessageConstants.ATTACHMENTS + "." + JsonMessageConstants.Attachment.FILENAME))
+                            .query(textCriterion.getOperator().getValue())
+                            .defaultOperator(Operator.And)
+                            .lenient(true)
+                            .build().toQuery())
                         .build()
-                        .toQuery())
-                    .should(new TermQuery.Builder()
-                        .field(JsonMessageConstants.ATTACHMENTS + "." + JsonMessageConstants.Attachment.FILE_EXTENSION)
-                        .value(new FieldValue.Builder().stringValue(textCriterion.getOperator().getValue()).build())
```
quantranhong1999, line 344: Why don't we query string against the file extension anymore?

Arsnael: He does, below in the else... Github just made a bit of a mess on the diff

PR #2813, opened 2025-09-17
[BUILD] Improve test execution time - batch 2

quantranhong1999:
Effort to improve the test execution time of 5m+ modules:
image

cf https://github.com/apache/james-project/pull/2809#issuecomment-3290521909 cc @jeantil 

Let's see the CI build.


server/apps/distributed-app/pom.xml
```diff
@@ -501,6 +501,7 @@
                         -javaagent:"${settings.localRepository}"/org/jacoco/org.jacoco.agent/${jacoco-maven-plugin.version}/org.jacoco.agent-${jacoco-maven-plugin.version}-runtime.jar=destfile=${basedir}/target/jacoco.exec
                         -Xms512m -Xmx2048m</argLine>
                     <reuseForks>true</reuseForks>
+                    <forkCount>2</forkCount>
```
chibenwa: I'm unsure the CI will support launching 2 distributed stacks at once.

Lets see

server/apps/distributed-app/src/test/java/org/apache/james/CassandraRabbitMQLegacyQuotaJamesServerTest.java
```diff
@@ -22,8 +22,11 @@
 import org.apache.james.jmap.JmapJamesServerContract;
 import org.apache.james.modules.RabbitMQExtension;
 import org.apache.james.modules.TestJMAPServerModule;
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
+@Disabled("Disabled to save build time. Legacy quota was a long time ago that we assume people migrated." +
+    "Therefore we do not need to test it on every build.")
```
chibenwa: IMO let's simply remove the test

server/apps/distributed-app/pom.xml
quantranhong1999: Me too. Let's see how the CI goes.

server/apps/distributed-app/src/test/java/org/apache/james/WithScanningSearchMutableTest.java
```diff
@@ -19,8 +19,11 @@
 
 package org.apache.james;
 
+import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.extension.RegisterExtension;
 
+@Disabled("Save build time. Distributed app + scanning search is an edge use case." +
+    "WithScanningSearchImmutableTest is enough to cover this case.")
```
chibenwa: Keep a single specialized test in another class ensuring is starts with scanning search

-> Only asserts that the wiring is well done.

server/apps/cassandra-app/pom.xml
```diff
@@ -326,6 +326,13 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <skipTests>true</skipTests>
+                </configuration>
+            </plugin>
```
chibenwa, line 338: I'm reluctant having tests and not running them.

I'd propose to either:
 - 1. assume and fully remove the tests
 - 2. OR just keep run them until 3.9.0 and effective removal
 - 3. OR be smart and run them only for master built (we can have 2 CI profile, one for master, one PRs and other branches and disable those tests when not on main)
 
 If asked I'd go for 2 or 3.

server/protocols/webadmin/webadmin-mailbox/pom.xml
```diff
@@ -259,6 +259,7 @@
                         -javaagent:"${settings.localRepository}"/org/jacoco/org.jacoco.agent/${jacoco-maven-plugin.version}/org.jacoco.agent-${jacoco-maven-plugin.version}-runtime.jar=destfile=${basedir}/target/jacoco.exec
                         -Xms512m -Xmx2048m</argLine>
                     <reuseForks>true</reuseForks>
+                    <forkCount>2</forkCount>
```
chibenwa, line 251: What's expensive here is running OpenSearch for reindexing tests.

I propose an idea: how about rely on Lucene with RamStore instead of OpenSearch for such tests? We would be able to run **way faster** without harming test relevance.

Thoughts?

chibenwa (commented):
Very good work tacking a sensitive topics.

IMO we could cut it in 2 parts:
 - consensual stuf to be merged asap
 - seperated PRs for non consensual stuff eg 2 forks for distributed & disabled tests

server/apps/cassandra-app/pom.xml
jeantil: I don't quite remember what the difference between cassandra-app and distributed-app is ? 
If cassandra-app is only a subset of distributed maybe we can remove most of its tests since it's all tested by distributed already and only keep a test that checks that the app starts successfully for the bootstrapping part 

wdyt ?

jeantil:
you might be interested in this develocity view too : https://develocity.apache.org/scans/tests?search.relativeStartTime=P90D&search.rootProjectNames=*james*&search.tasks=clean,test,jacoco*&search.timeZoneId=Europe%2FParis&tests.sortField=MEAN_DURATION#
it lists the tests with the longest duration on average

server/protocols/webadmin/webadmin-mailbox/pom.xml
quantranhong1999: > What's expensive here is running OpenSearch for reindexing tests.

I will have a look at this.

chibenwa:
Yes there is a topic with BlobStoreDAO contract too ^^

server/apps/distributed-app/pom.xml
quantranhong1999: 1st build is green. But I would let it be part of the non-consensual part so we can validate more.

server/apps/distributed-app/src/test/java/org/apache/james/CassandraRabbitMQLegacyQuotaJamesServerTest.java
quantranhong1999: removed in https://github.com/apache/james-project/pull/2814

chibenwa:
image

master took 2h05

Test run: 43 147

Test run on mater: 44 509

Delta in count of test likely reflects removing cassandra-app, some parametric tests in distributed-webadmin-integration tests and having JMAP tests as smoke tests on the PG app.

2h master build is definitly hardware dependant, and comes at a real surprise to me. Thanks cloud bee and the foundation!

jeantil:
can you clarify what you mean by 
> 2h master build is definitly hardware dependant, and comes at a real surprise to me. Thanks cloud bee and the foundation! 

are you saying the CI machines are too weak compared to our workstations ? I never tried letting the full james build with tests complete on my laptop ... not sure how fast it is there. 

reproducing a specific build's conditions locally is not easy so the remote cache will not have the same level of impact

jeantil:
Yes the build cache is a remote build cache so as long as the inputs have not changed it will apply regardless of the branch being built. 
If I did this right, it will even be used to reduce local build time on  developer's machines (but non ci machines will not be able to write to the remote build cache). Note that this only applies if the maven command is the same which has a rather low chance of happening for the tests since no one would run 
``` 
mvn -B -e -fae clean test -Dorg.slf4j.simpleLogger.showDateTime=true

server/apps/cassandra-app/pom.xml
quantranhong1999: >  only keep a test that checks that the app starts successfully for the bootstrapping part

How about keeping the current Cassandra app test suite, and smoke testing for a few basic test classes (just like what we do for JMAP IT)?

Given the build time is good now, 1 -> 1.5 hour, keep running the whole Cassandra app test suite as of today is acceptable though IMO.

server/protocols/webadmin/webadmin-mailbox/pom.xml
quantranhong1999: solved as part of https://github.com/apache/james-project/pull/2815

server/apps/cassandra-app/pom.xml
jeantil: > Given the build time is good now, 1 -> 1.5 hour 

To give you some perspective with data points from my day job : An individual app CI pipeline takes about 30 minutes, that includes build (compile+ unit tests&integration tests), deployment to integration and end2end testing of critical path scenarios. larger pipelines that touch shared components and rebuild the 70 apps will take longer, lets say maybe 2h but because they hit various limits ( number of available workers, some e2e tests are slower to run and we only allow 1 e2e test to run at the same time for a given app so the pipeline may end up in a queue)  

There are technological differences (we use gradle which has a much more advanced build avoidance system than what can be done with maven resulting in even more efficient build caching) and we don´t have cassandra in the stack ( pg, mongo, ES, rabbit are there) there are probably less variants but there is also a lot more code 3-4M lines of code IIRC and far more assemblies)


So yes the build time is better on james but I wouldn't call it good, mostly difference of perspectives :D

jeantil: keeping the smoke tests is ok

server/apps/distributed-app/pom.xml
quantranhong1999: Without forkCount=2: https://develocity.apache.org/s/24fhy4ewyxcbw/timeline?sort=longest
With forkCount=2: https://develocity.apache.org/s/aanzaqjofa2bg/timeline?sort=longest
Test time for the distributed app module reduced from 12.5m -> 7.5m

However, as @jeantil raised some concerns about forkCount, I guess we can be defensive for the distributed app then.

server/apps/cassandra-app/pom.xml
quantranhong1999: > So yes the build time is better on james but I wouldn't call it good, mostly difference of perspectives :D

Yes, I understand. For me, James build went from 4-5h to more than 1 hour is already a lot better. Thanks @jeantil!

Of course, we still can optimize it better.

jeantil: we just had a build with an actual good build time on master here https://ci-builds.apache.org/job/james/job/ApacheJames/job/master/1768/cloudbees-pipeline-explorer/?filter=79 
25 minutes for all stages :D

and yes going from 4-5h to 1h is already helpful :D That was the easy step even if it took 1y9 months but mostly waiting time.

the next improvments are going to require a lot of thinking about what we are testing and how to improve the repartition of james tests on the test pyramid. I think (intuition not analysis) that there are too many integration tests. I have other topics I want to tackle first ( I have a huge wip on jdkim) and very limited time so I probably won´t  bother the james team any time soon with this but it will come... some day :rofl:

quantranhong1999:
I just force pushed the second batch. This batch focuses more on removing duplicate ITs.

chibenwa (approved):

quantranhong1999:
37 minutes of build (25m tests) now :-)

Arsnael (approved):

chibenwa (approved):

PR #2814, opened 2025-09-17
[BUILD] Improve test execution time - batch 1

quantranhong1999:
Consensual split part of https://github.com/apache/james-project/pull/2813

### DIFF
--- a/server/protocols/webadmin/webadmin-mailbox/pom.xml
+++ b/server/protocols/webadmin/webadmin-mailbox/pom.xml
@@ -259,6 +259,7 @@
                         -javaagent:"${settings.localRepository}"/org/jacoco/org.jacoco.agent/${jacoco-maven-plugin.version}/org.jacoco.agent-${jacoco-maven-plugin.version}-runtime.jar=destfile=${basedir}/target/jacoco.exec
                         -Xms512m -Xmx2048m</argLine>
                     <reuseForks>true</reuseForks>
+                    <forkCount>2</forkCount>
                     <!-- Fail tests longer than 30 minutes, prevent form random locking tests -->
                     <forkedProcessTimeoutInSeconds>1800</forkedProcessTimeoutInSeconds>
                     <excludedGroups>unstable</excludedGroups>


chibenwa (approved):

mailbox/opensearch/pom.xml
```diff
@@ -165,4 +165,16 @@
...
         </dependency>
     </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <reuseForks>true</reuseForks>
+                    <forkCount>2</forkCount>
```
jeantil, line 175: this is consensual ?

server/protocols/jmap-rfc-8621-integration-tests/postgres-jmap-rfc-8621-integration-tests/pom.xml
```diff
@@ -101,11 +101,37 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
+                    <argLine>-Djava.library.path=
+                        -javaagent:"${settings.localRepository}"/org/jacoco/org.jacoco.agent/${jacoco-maven-plugin.version}/org.jacoco.agent-${jacoco-maven-plugin.version}-runtime.jar=destfile=${basedir}/target/jacoco.exec
+                        -Xms512m -Xmx2048m</argLine>
                     <forkedProcessTimeoutInSeconds>3600</forkedProcessTimeoutInSeconds>
                     <reuseForks>true</reuseForks>
-                    <forkCount>2</forkCount>
+                    <groups combine.self="override">BasicFeature</groups>
```
jeantil, line 109: I 'm not sure I understand this and the correponding unstable change below.  these settings are alread defined in the project root  pom, why override them here ? this may need a comment to explain so we know why there is specific exception if we start tinkering with things a the root pom level

jeantil:
out of curiosity why aim for `forkCount` instead of -T2 ? 

the test suite is executed by 
` mvn -B -e -fae clean test -Dorg.slf4j.simpleLogger.showDateTime=true `  which doesn't enable parallelism

mailbox/opensearch/pom.xml
quantranhong1999: I tried locally with `mvn surefire:test` (so no parallelism) , and adding forkCount = 2 would help me gain around 25% test time for a module. E.g 5 minute -> 3.5 minute. So I think increasing forkCount does help.

BTW, `forkCount` tuning seems classic in our project.

server/protocols/jmap-rfc-8621-integration-tests/postgres-jmap-rfc-8621-integration-tests/pom.xml
quantranhong1999: Smoke JMAP integration tests for the Postgres app. We do the same for distributed JMAP tests.

quantranhong1999:
> out of curiosity why aim for forkCount instead of -T2 ?

I remember in the past I tried tuning `threadCount` for Maven Surefire and using ` -T xxx` in the maven command for our build, but that seems to bring too much parallelism. Our test suite can not take it. 

`forkCount` seems more classic and safer while still bringing value.

quantranhong1999:
Hi @jeantil is this work ok for you overall? Other improvements from your side are welcome too (maybe in another PR).

jeantil:
go ahaead as part of 
> Other improvements from your side are welcome too (maybe in another PR).

I might just try to disable all this forkcount tuning and see how it affects build times. 

> So forkCount does indeed help more or less.
Results on your machine have no guarantees to translate to results on the CI machines, parallelization is highly hardware dependant. Modern developper machines have insane amounts of resources compared to workers which most of the time work on virtualized CPUs and RAM which is shared with an unknowable number of other processes. While doing performance benchmarking before, I have seen context switching on the physical CPU use up to 50% of the CPU compute capacity. 

Performance can only be measured and tuned in and for a specific context which is usually highly hardware dependant. 
This is also why I mentioned that I would autorebase and dual build for at least a couple week : to obtain enough datapoints on a range of CI workers that a statistical analysis has some hope of giving results since we cannot control the worker composition and the shared usage on the hypoervisors they run on. 
Alternatively we would need to be able to provide our own controlled compute resources in which case a couple runs would be enough but I don't have that kind of money :D

PR #2815, opened 2025-09-17
[BUILD TIME] james-server-webadmin-mailbox: use lucene instead of Ope…

chibenwa:
…nSearch


Arsnael (approved):

jeantil:
sorry i'm a bit tired and don´t quite remember the original numbers, here is what the build scan says about slow tests in this build 
https://develocity.apache.org/s/sugjiob3gwhpu/tests/slowest-tests
2025-09-18_07-49-slow-tests-lucene-opensearch

PR #2816, opened 2025-09-19
[ENHANCEMENT] DefaultCheckTokenClient: increase pending acquire max

quantranhong1999:
The default pending acquire was low (~32, depending on machines). The OIDC client could fail under high load.

```java
"exception":"reactor.netty.internal.shaded.reactor.pool.PoolAcquirePendingLimitException: Pending acquire queue has reached its maximum size of 32
```

Increasing the default value to 10000 (around the current S3 HTTP client pending acquire) should be fine.

### DIFF
--- a/server/protocols/jwt/src/main/java/org/apache/james/jwt/DefaultCheckTokenClient.java
+++ b/server/protocols/jwt/src/main/java/org/apache/james/jwt/DefaultCheckTokenClient.java
@@ -21,6 +21,7 @@ package org.apache.james.jwt;
 
 import java.net.URL;
 import java.nio.charset.StandardCharsets;
+import java.util.Optional;
 
 import org.apache.james.jwt.introspection.IntrospectionEndpoint;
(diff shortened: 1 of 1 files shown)


server/protocols/jwt/src/main/java/org/apache/james/jwt/DefaultCheckTokenClient.java
```diff
@@ -42,11 +43,15 @@
 public class DefaultCheckTokenClient implements CheckTokenClient {
 
     public static final String TOKEN_ATTRIBUTE = "token";
+    public static final int PENDING_ACQUIRE_MAX = Optional.ofNullable(System.getProperty("james.oidc.client.pending.acquire.max"))
+        .map(Integer::valueOf)
+        .orElse(10000);
```
chibenwa, line 48:
```suggestion
        .orElse(512);
```

Maybe?

chibenwa (approved):

PR #2817, opened 2025-09-19
Webadmin route: Ability to run rules on folders for a specific user

Arsnael:
WIP, just started...

### DIFF
--- a/docs/modules/servers/partials/operate/webadmin.adoc
+++ b/docs/modules/servers/partials/operate/webadmin.adoc
@@ -1800,6 +1800,66 @@ the following `additionalInformation`:
 }
 ....
 
+=== Running a filtering rule on a mailbox
+
+....
+curl -XPOST http://ip:port/users/{usernameToBeUsed}/mailboxes/{mailboxName}/messages?action=triage \
+-d '{
+  "id": "1",
+  "name": "rule 1",
+  "action": {
+    "appendIn": {
+      "mailboxIds": ["23"]
+    }
+  },
+  "conditionGroup": {
+    "conditionCombiner": "OR",
+    "conditions": [
+      {
+        "comparator": "contains",
+        "field": "subject",
+        "value": "plop"
+      },
+      {
+        "comparator": "exactly-equals",
+        "field": "from",
+        "value": "bob@example.com"
+      }
+    ]
+  }
+}'
+....
+
+Will schedule a task for running a filtering rule passed as payload in ``mailboxName`` mailbox of ``usernameToBeUsed``.
+
+link:#_endpoints_returning_a_task[More details about endpoints returning
+a task].
+
+Resource name `usernameToBeUsed` should be an existing user.
+
+Resource name `mailboxName` should not be empty, nor contain `% *` characters, nor starting with `#`.
+
+Response codes:
+
+* 201: Success. Corresponding task id is returned.
+* 400: Invalid mailbox name
+* 400: Invalid JSON payload
+* 404: Invalid get on user mailboxes. The `username` or `mailboxName` does not exit
+
+The scheduled task will have the following type `RunRulesOnMailboxTask` and
+the following `additionalInformation`:
+
+....
+{
+    "mailboxName": "mbx1",
+    "rulesOnMessagesApplySuccessfully": 9,
+    "rulesOnMessagesApplyFailed": 3,
+    "timestamp": "2024-12-03T10:15:30Z",
+    "type": "RunRulesOnMailboxTask",
+    "username": "bob@domain.tld"
+}
+....
+
 === Subscribing a user to all of its mailboxes
 
 ....
--- a/server/protocols/jmap-rfc-8621/src/main/java/org/apache/james/jmap/mailet/filter/FilteringHeaders.java
+++ b/server/protocols/jmap-rfc-8621/src/main/java/org/apache/james/jmap/mailet/filter/FilteringHeaders.java
@@ -1,0 +1,92 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.jmap.mailet.filter;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Iterator;
+import java.util.List;
+
+import jakarta.mail.MessagingException;
+
+import org.apache.james.mailbox.exception.MailboxException;
+import org.apache.james.mailbox.model.Header;
+import org.apache.james.mailbox.model.Headers;
+import org.apache.james.mailbox.model.MessageResult;
+import org.apache.mailet.Mail;
+
+public interface FilteringHeaders {
+    class MailFilteringHeaders implements FilteringHeaders {
+        private final Mail mail;
+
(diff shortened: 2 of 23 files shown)


chibenwa:
I gave a little try to this and quite like how abstractions are put together.

I agree continuing building over this rule engine the Webadmin endpoints.

Thanks for the thoughts put into this.

server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/service/UserMailboxesService.java
```diff
@@ -213,4 +221,50 @@ private Stream<MailboxMetaData> listUserMailboxes(MailboxSession mailboxSession)
...
+    }
+
+    private Flux<Result> runRulesOnMessage(RuleMatcher ruleMatcher, MessageResult messageResult) throws MailboxException {
+        return Flux.fromStream(ruleMatcher.findApplicableRules(messageResult))
+            .map(Rule::getAction)
+            .concatMap(action -> applyActionOnMessage(messageResult, action));
+    }
+
+    private Mono<Result> applyActionOnMessage(MessageResult messageResult, Rule.Action action) {
+        return Mono.just(Result.COMPLETED); //TODO apply actions
```
Arsnael: Current actions with JMAPFiltering mailet:
- appendInMailboxes
- set flags (seen / important)
- keywords
- reject
- forwards

Setting flags and moving mails around in other mailboxes I think makes sense.

I don't think we treat reject, it does not make sense I think here? If we want to trash mails we just move them to the trash mailbox?
What about forwards? I don't think it has its place here too? (but less sure on this point so let me know wdyt)
And I guess we do keywords too?

server/protocols/webadmin/webadmin-mailbox/pom.xml
```diff
@@ -141,6 +141,10 @@
             <groupId>${james.groupId}</groupId>
             <artifactId>james-server-data-api</artifactId>
         </dependency>
+        <dependency>
+            <groupId>${james.groupId}</groupId>
+            <artifactId>james-server-data-jmap</artifactId>
```
chibenwa: IMO this is not cool to add JMAP-ish dependencies in here.

Can we relocate this in `webadmin-jmap`?

server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/service/UserMailboxesService.java
chibenwa: > I don't think we treat reject, it does not make sense I think here?

Yes it makes no functional sense so let's just throw a NotImplementedException

> What about forwards?

It *could* make sense but it's heavy to implement so maybe better just ingore it for now and throw too

> And I guess we do keywords too?

IMO yes ;-)

We need however to think about a set/unset/reset semantic. In the local delivery pipeline only the set semantic make sense.  Here we could have potentially both...

I do not have the spec in mind. Let's focus on getting the "rules running on folder" here. Let's keep the chageset minimal so let's just implement `move` for now. We will specify other actions afterward.

chibenwa:
> Also what would be the response of the triage tasks? Something like:

IMO it would need to be a task.

Yes the proposal could be a good begining.

server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/service/UserMailboxesService.java
```diff
@@ -213,4 +222,81 @@ private Stream<MailboxMetaData> listUserMailboxes(MailboxSession mailboxSession)
...
+
+        if (mailboxIds.isEmpty()) {
+            return Mono.just(Result.COMPLETED);
+        }
+
+        if (mailboxIds.contains(messageResult.getMailboxId().serialize())) {
+            return copyInMailboxes(messageResult.getUid(), messageResult.getMailboxId(), mailboxIds, mailboxSession, context);
+        } else {
+            return moveInMailboxes(messageResult, mailboxIds, mailboxSession, context);
+        }
```
chibenwa: We can rely on `MessageIdManager::setInMailboxesReactive` which better matches the "mailbox list" semantic?

Being more atomic it's likely to be less of a mess in case of failure.

Arsnael: Completely missed and forgot about that method... would defo make the logic easier, thanks

server/protocols/webadmin/webadmin-mailbox/pom.xml
Arsnael: Realizing I forgot this comment... will address sorry

Arsnael:
I will clean up the commit history later, to let the reviewers still having the follow up.

More or less ok to review I think for the webadmin route for rule on mailbox for a user. I think I will do an other PR based on this for the same but cluster wise

docs/modules/servers/partials/operate/webadmin.adoc
```diff
@@ -1800,6 +1800,66 @@ the following `additionalInformation`:
...
+....
+curl -XPOST http://ip:port/users/{usernameToBeUsed}/mailboxes/{mailboxName}/messages \
+-d '{
+  "id": "1",
+  "name": "rule 1",
+  "action": {
+    "appendIn": {
+      "mailboxIds": ["23"]
+    }
+  },
```
quantranhong1999, line 1814: Likely we need to implement the "moveTo" action which relies on `mailboxName` to ease usage.

cf ticket description `Action 'moveTo' based on a mailbox name (needed to trun mass treatment on several users as we cannot hard code mailboxid)`

IMO this can be a part of the cluster-wide PR (that runs on mailboxes of users, for which we can not use hardcoded mailboxId).

chibenwa: > Likely we need to implement the "moveTo" action which relies on mailboxName to ease usage.

Yes but as a follow up PR please.

server/protocols/jmap-rfc-8621/src/main/java/org/apache/james/jmap/mailet/filter/FilteringHeaders.java
```diff
@@ -0,0 +1,94 @@
...
+            return getMatchingHeaders(name);
+        }
+
+        @Override
+        public String getSubject() throws MailboxException {
+            return Arrays.stream(getMatchingHeaders("Subject"))
+                .findFirst()
+                .orElse(null);
+
+        }
```
chibenwa, line 71:
```suggestion
                .orElse(null);
        }
```

server/protocols/jmap-rfc-8621/src/main/java/org/apache/james/jmap/mailet/filter/FilteringHeaders.java
```diff
@@ -0,0 +1,94 @@
...
+
+public interface FilteringHeaders {
+    class MailFilteringHeaders implements FilteringHeaders {
+        private final Mail mail;
+
+        public MailFilteringHeaders(Mail mail) {
+            this.mail = mail;
+        }
+
+
```
chibenwa:
```suggestion
        }

```

server/protocols/webadmin/webadmin-mailbox/src/main/java/org/apache/james/webadmin/service/UserMailboxesService.java
```diff
@@ -213,4 +226,63 @@ private Stream<MailboxMetaData> listUserMailboxes(MailboxSession mailboxSession)
             .toStream();
     }
 
+    public Mono<Result> runRulesOnMailbox(Username username, MailboxName mailboxName, Rules rules, RunRulesOnMailboxTask.Context context) {
+        MailboxSession mailboxSession = mailboxManager.createSystemSession(username);
+        RuleMatcher ruleMatcher = new RuleMatcher(rules.getRules());
+
+        return Mono.from(mailboxManager.getMailboxReactive(MailboxPath.forUser(username, mailboxName.asString()), mailboxSession))
+            .flatMapMany(messageManager -> Flux.from(messageManager.getMessagesReactive(MessageRange.all(), FetchGroup.HEADERS, mailboxSession))
+                .concatMap(Throwing.function(messageResult -> runRulesOnMessage(ruleMatcher, messageResult, mailboxSession, context))))
```
chibenwa: IMO that's maybe a bit harsh

We could `flatmap` + `LOW_CONCURRENCY` ?

Arsnael:
Appreciate the quick fixes :)

quantranhong1999:
`org.apache.james.webadmin.integration.memory.MemoryWebAdminServerIntegrationTest.jmapUserTasksShouldBeExposed` related?

chibenwa (approved):
Wonderful, thanks for the awesome work @Arsnael !

PR #2825, opened 2025-09-29
Rule flag criteria

Arsnael:


Arsnael:
Missing custom flags and also something is wrong with this implem, didnt have time to debug properly yet

server/data/data-jmap-cassandra/src/test/resources/json/eventComplex-v4.json
```diff
@@ -33,6 +33,16 @@
...
+          {
+            "field": "flag",
+            "comparator": "isSet",
+            "value": "\\Seen"
+          },
+          {
+            "field": "flag",
+            "comparator": "isUnset",
+            "value": "\\Recent"
           }
```
chibenwa, line 51: Please also add a user flag in this test.

server/protocols/jmap-rfc-8621/src/main/java/org/apache/james/jmap/mailet/filter/ContentMatcher.java
```diff
@@ -95,7 +96,47 @@ public boolean match(Stream<String> contents, String valueToMatch) {
             return contents.map(ContentMatcher::asAddressHeader)
                 .anyMatch(addressHeaderToMatch::matchesIgnoreCase);
         }
+    }
+
+    class ParsedFlag {
+        private final Optional<Flag> flag;
```
chibenwa: Let's use `Keyword` as an abstraction ?

server/protocols/webadmin/webadmin-jmap/src/test/java/org/apache/james/webadmin/data/jmap/RunRulesOnMailboxRoutesTest.java
```diff
@@ -499,6 +503,103 @@ void runRulesOnMailboxShouldManageMixedCase() throws Exception {
...
+
+        ComposedMessageId messageId2 = messageManager.appendMessage(
+            MessageManager.AppendCommand.builder()
+                .build(Message.Builder.of()
+                    .setSubject("hello")
+                    .setFrom("alice@example.com")
+                    .setBody("body", StandardCharsets.UTF_8)),
+            systemSession).getId();
+
+        messageManager.setFlags(new Flags(Flags.Flag.ANSWERED), MessageManager.FlagsUpdateMode.ADD, MessageRange.one(messageId2.getUid()), systemSession);
```
chibenwa: `AppendCommand.builder().withFlags(...)` ?

server/data/data-jmap-cassandra/src/test/resources/json/eventComplex-v4.json
Arsnael: Ain't user flags custom flags? (if i understand correcdtly) did say I'm missing those for the moment

server/protocols/webadmin/webadmin-jmap/src/test/java/org/apache/james/webadmin/data/jmap/RunRulesOnMailboxRoutesTest.java
```diff
@@ -568,12 +563,12 @@ void runRulesOnMailboxShouldApplyFlagCriteria() throws Exception {
                   {
                     "comparator": "isSet",
                     "field": "flag",
-                    "value": "\\Seen"
+                    "value": "Seen"
```
Arsnael: There seems to be some json deserialization issue with the `\\` at the moment so for now I took it out. Will check it out later. 

Or maybe we can just omit in the rule the `\`?

Arsnael: Regarding this... In the test I need to put `"value": "\\\\Seen" ` for it to work, because while debugging I could observe that sparks does the escape when we get the request body, and then json deserializer is doing it too behind. 

Can we just do without the `\` for the sake of simplicity?

server/data/data-jmap-cassandra/src/test/resources/json/eventComplex-v4.json
```diff
@@ -33,6 +33,21 @@
             "field": "internalDate",
             "comparator": "isNewerThan",
             "value": "2d"
+          },
+          {
+            "field": "flag",
+            "comparator": "isSet",
+            "value": "\\Seen"
```
chibenwa:
```suggestion
            "value": "$seen"
```

Can we match exactly the JMAP keyword syntax ?

server/protocols/webadmin/webadmin-jmap/src/test/java/org/apache/james/webadmin/data/jmap/RunRulesOnMailboxRoutesTest.java
chibenwa: Indeed. I propose to align on JMAP keyword syntax 

CF https://github.com/apache/james-project/pull/2825/files#r2394716325

quantranhong1999 (approved):

server/data/data-jmap-cassandra/src/test/resources/json/eventComplex-v4.json
chibenwa: Can we update the test to match the doc for system flags (`$seen`) ?

server/protocols/jmap-rfc-8621/src/main/java/org/apache/james/jmap/mailet/filter/ContentMatcher.java
```diff
@@ -95,7 +98,45 @@ public boolean match(Stream<String> contents, String valueToMatch) {
...
+            }
+            return maybeFlag;
+        }
 
+        boolean matches(ParsedFlag otherFlag) {
+            return keyword.map(keyword1 ->
+                    otherFlag.keyword
+                        .map(keyword2 -> keyword1.getFlagName().equals(keyword2.getFlagName()))
+                        .orElse(false))
+                .orElse(false);
```
chibenwa:
```suggestion
            return OptionalUtils.matches(keyword, otherFlag.keyword,
                (k1, k2) -> k1.getFlagName().equals(k2.getFlagName()));
```

?

chibenwa (approved):

PR #2828, opened 2025-09-30
JAMES-4148 [FILTER] Implement moveTo action

quantranhong1999:
resolve https://github.com/linagora/james-project/issues/5469


chibenwa:
Please add a test when the mailbox do not exist (with a given name)

We would need it to be created if missing.

Can you please add such a test?

Arsnael (approved):

PR #2960, opened 2026-03-03
JAMES-4182 Proposal for changes in BlobStoreDAO interface

chibenwa:
Opened to collect initial feedback

See https://issues.apache.org/jira/browse/JAMES-4182 for context

Remaining points:
 - [x] better validation for metadata
 - [ ] Store metadata in blob stores
     - [ ] xattr for files
     - [ ] extra collumn for PG, cassandra
     - [ ] Object metadata for S3
     - [x] Direct storage for memory
     - [ ] Passing those around for other blobStore DAO
     - [x] AES blob store should support metadata

### DIFF
--- a/server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
+++ b/server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
@@ -108,15 +108,15 @@ public class AESBlobStoreDAO implements BlobStoreDAO {
     }
 
     @Override
-    public Publisher<byte[]> readBytes(BucketName bucketName, BlobId blobId) {
-        return Mono.from(underlying.readBytes(bucketName, blobId))
-            .map(Throwing.function(bytes -> {
-                InputStream inputStream = decrypt(new ByteArrayInputStream(bytes));
+    public Publisher<BytesBlob> readBytesBlob(BucketName bucketName, BlobId blobId) {
+        return Mono.from(underlying.readBytesBlob(bucketName, blobId))
+            .map(Throwing.function(bytesBlob -> {
+                InputStream inputStream = decrypt(new ByteArrayInputStream(bytesBlob.payload()));
                 try (UnsynchronizedByteArrayOutputStream outputStream = UnsynchronizedByteArrayOutputStream.builder()
-                    .setBufferSize(bytes.length + PBKDF2StreamingAeadFactory.SEGMENT_SIZE)
+                    .setBufferSize(bytesBlob.payload().length + PBKDF2StreamingAeadFactory.SEGMENT_SIZE)
                     .get()) {
                     IOUtils.copy(inputStream, outputStream);
-                    return outputStream.toByteArray();
+                    return BytesBlob.of(outputStream.toByteArray(), bytesBlob.metadata());
                 }
             }));
     }
--- a/server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
+++ b/server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
@@ -93,25 +93,25 @@ public class AESBlobStoreDAO implements BlobStoreDAO {
     }
 
     @Override
-    public InputStreamBlob readBlob(BucketName bucketName, BlobId blobId) throws ObjectStoreIOException, ObjectNotFoundException {
+    public InputStreamBlob read(BucketName bucketName, BlobId blobId) throws ObjectStoreIOException, ObjectNotFoundException {
         try {
-            return InputStreamBlob.of(decrypt(underlying.readBlob(bucketName, blobId).payload()));
+            return InputStreamBlob.of(decrypt(underlying.read(bucketName, blobId).payload()));
         } catch (IOException e) {
             throw new ObjectStoreIOException("Error reading blob " + blobId.asString(), e);
         }
     }
 
     @Override
-    public Publisher<InputStreamBlob> readBlobReactive(BucketName bucketName, BlobId blobId) {
-        return Mono.from(underlying.readBlobReactive(bucketName, blobId))
+    public Publisher<InputStreamBlob> readReactive(BucketName bucketName, BlobId blobId) {
+        return Mono.from(underlying.readReactive(bucketName, blobId))
             .map(InputStreamBlob::payload)
             .map(Throwing.function(this::decrypt))
             .map(InputStreamBlob::of);
     }
 
     @Override
-    public Publisher<BytesBlob> readBytesBlob(BucketName bucketName, BlobId blobId) {
-        return Mono.from(underlying.readBytesBlob(bucketName, blobId))
+    public Publisher<BytesBlob> readBytes(BucketName bucketName, BlobId blobId) {
+        return Mono.from(underlying.readBytes(bucketName, blobId))
             .map(Throwing.function(bytesBlob -> {
                 InputStream inputStream = decrypt(new ByteArrayInputStream(bytesBlob.payload()));
                 try (UnsynchronizedByteArrayOutputStream outputStream = UnsynchronizedByteArrayOutputStream.builder()
@@ -124,7 +124,7 @@ public class AESBlobStoreDAO implements BlobStoreDAO {
     }
 
     @Override
-    public Publisher<Void> saveBlob(BucketName bucketName, BlobId blobId, Blob blob) {
+    public Publisher<Void> save(BucketName bucketName, BlobId blobId, Blob blob) {
         return switch (blob) {
             case BytesBlob bytesBlob -> save(bucketName, blobId, bytesBlob.payload());
             case InputStreamBlob inputStreamBlob -> save(bucketName, blobId, inputStreamBlob.payload());
@@ -147,7 +147,7 @@ public class AESBlobStoreDAO implements BlobStoreDAO {
 
         return Mono.usingWhen(
                 Mono.fromCallable(() -> encrypt(inputStream)),
-                pair -> Mono.from(underlying.saveBlob(bucketName, blobId, byteSourceWithSize(pair.getLeft().asByteSource(), pair.getRight()))),
+                pair -> Mono.from(underlying.save(bucketName, blobId, byteSourceWithSize(pair.getLeft().asByteSource(), pair.getRight()))),
                 Throwing.function(pair -> Mono.fromRunnable(Throwing.runnable(pair.getLeft()::reset)).subscribeOn(Schedulers.boundedElastic())))
             .subscribeOn(Schedulers.boundedElastic())
             .onErrorMap(e -> new ObjectStoreIOException("Exception occurred while saving bytearray", e));
--- a/server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
+++ b/server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
@@ -93,9 +93,9 @@ public class AESBlobStoreDAO implements BlobStoreDAO {
     }
 
     @Override
-    public InputStream read(BucketName bucketName, BlobId blobId) throws ObjectStoreIOException, ObjectNotFoundException {
+    public InputStreamBlob readBlob(BucketName bucketName, BlobId blobId) throws ObjectStoreIOException, ObjectNotFoundException {
         try {
-            return decrypt(underlying.read(bucketName, blobId));
+            return InputStreamBlob.of(decrypt(underlying.readBlob(bucketName, blobId).payload()));
         } catch (IOException e) {
             throw new ObjectStoreIOException("Error reading blob " + blobId.asString(), e);
         }
--- a/server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
+++ b/server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
@@ -122,6 +122,14 @@ public class AESBlobStoreDAO implements BlobStoreDAO {
     }
 
     @Override
+    public Publisher<Void> saveBlob(BucketName bucketName, BlobId blobId, Blob blob) {
+        return switch (blob) {
+            case BytesBlob bytesBlob -> save(bucketName, blobId, bytesBlob.payload());
+            case InputStreamBlob inputStreamBlob -> save(bucketName, blobId, inputStreamBlob.payload());
+            case ByteSourceBlob byteSourceBlob -> save(bucketName, blobId, byteSourceBlob.payload());
+        };
+    }
+
     public Publisher<Void> save(BucketName bucketName, BlobId blobId, byte[] data) {
         Preconditions.checkNotNull(bucketName);
         Preconditions.checkNotNull(blobId);
@@ -130,7 +138,6 @@ public class AESBlobStoreDAO implements BlobStoreDAO {
         return save(bucketName, blobId, new ByteArrayInputStream(data));
     }
 
-    @Override
     public Publisher<Void> save(BucketName bucketName, BlobId blobId, InputStream inputStream) {
         Preconditions.checkNotNull(bucketName);
         Preconditions.checkNotNull(blobId);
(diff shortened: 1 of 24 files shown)


server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -19,86 +19,128 @@
...
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.io.ByteSource;
+import com.google.common.io.FileBackedOutputStream;
 
 public interface BlobStoreDAO {
-    class ReactiveByteSource {
-        private final long size;
-        private final Publisher<ByteBuffer> content;
+    record BlobMetadataName(String name) {
```
jeantil, line 42: how about using an enum instead of a record ? 
it would mean we support less metadata fields but we would have stronger semantics,

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -19,86 +19,128 @@
...
+        // TODO validation (a-z,A-Z,0-9 -_) &length 128 char & non empty
+    }
+    record BlobMetadataValue(String value) {
+
+    }
+
+    sealed interface Blob {
+        Map<BlobMetadataName, BlobMetadataValue> metadata();
```
jeantil: As mentionned above it would be better for keys to be known,
alternatively this could be an object with kmown fields which james cares about and a map for extra fields which are not interpreted by james

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -19,86 +19,128 @@
...
+
+        // Have the POJOs encode some conversions ?
```
jeantil, line 108: :+1:

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -19,86 +19,128 @@
...
-     */
-    Publisher<byte[]> readBytes(BucketName bucketName, BlobId blobId);
+        @Override
+        public BytesBlob asBytes() throws IOException {
+            return new BytesBlob(payload.readAllBytes(), metadata);
+        }
 
+        @Override
+        public ByteSourceBlob asByteSource() throws IOException {
+            try (FileBackedOutputStream fileBackedOutputStream = new FileBackedOutputStream(100 * 1024)) {
```
jeantil: `100*1024` magic number
if I understand correctly this will write the stream to a file if the length is about 100kbits

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -143,14 +143,7 @@ public ByteSourceBlob asByteSource() {
...
 
-    /**
-     * Reads a Blob based on its BucketName and its BlobId
-     *
-     * @return a Mono containing the content of the blob or
-     *  an ObjectNotFoundException in its error channel when the blobId or the bucket is not found
-     *  or an IOObjectStoreException when an unexpected IO error occurs
-     */
-    Publisher<byte[]> readBytes(BucketName bucketName, BlobId blobId);
+    Publisher<BytesBlob> readBytesBlob(BucketName bucketName, BlobId blobId);
```
jeantil: while the description was useless the details in the return had some value 
the whole blob library probably deserves more documentation instead of less 😬

server/blob/blob-memory/src/main/java/org/apache/james/blob/memory/MemoryBlobStoreDAO.java
```diff
@@ -62,12 +62,18 @@ public Publisher<InputStream> readReactive(BucketName bucketName, BlobId blobId)
...
     public Mono<byte[]> readBytes(BucketName bucketName, BlobId blobId) {
         return Mono.fromCallable(() -> blobs.get(bucketName, blobId))
             .switchIfEmpty(Mono.error(() -> new ObjectNotFoundException(String.format("blob '%s' not found in bucket '%s'", blobId.asString(), bucketName.asString()))));
     }
 
+    @Override
+    public Publisher<BytesBlob> readBytesBlob(BucketName bucketName, BlobId blobId) {
+        return Mono.fromCallable(() -> blobs.get(bucketName, blobId))
+            .switchIfEmpty(Mono.error(() -> new ObjectNotFoundException(String.format("blob '%s' not found in bucket '%s'", blobId.asString(), bucketName.asString()))))
+            .map(BytesBlob::of);
```
jeantil: 🤔 shouldn't the in memory be adapted to actually store the metadata ? the MemoryBlobStoreDao will always loose all metadata.

I know we can derive the size from the content but why have a Map<X,Y> for metadatas if all the stores canx extract is the size>?

same comment applied to all leafs implementations

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
```diff
@@ -102,9 +102,11 @@ public InputStream read(BucketName bucketName, BlobId blobId) throws ObjectStore
...
 
     @Override
-    public Publisher<InputStream> readReactive(BucketName bucketName, BlobId blobId) {
-        return Mono.from(underlying.readReactive(bucketName, blobId))
-            .map(Throwing.function(this::decrypt));
+    public Publisher<InputStreamBlob> readBlobReactive(BucketName bucketName, BlobId blobId) {
+        return Mono.from(underlying.readBlobReactive(bucketName, blobId))
+            .map(InputStreamBlob::payload)
+            .map(Throwing.function(this::decrypt))
+            .map(InputStreamBlob::of);
```
jeantil: this is stripping any metadata returned by the underlying store

jeantil (commented):
Looks like a good direction to me, I left some comments, I'm sure there are which you are aware of and will addres, feel free to close on sight

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -19,86 +19,128 @@
...
+import com.google.common.collect.ImmutableMap;
 import com.google.common.io.ByteSource;
+import com.google.common.io.FileBackedOutputStream;
 
 public interface BlobStoreDAO {
-    class ReactiveByteSource {
-        private final long size;
-        private final Publisher<ByteBuffer> content;
+    record BlobMetadataName(String name) {
+        // TODO validation (a-z,A-Z,0-9 -_) &length 128 char & non empty
```
jeantil: which store's constraint is that ?

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
chibenwa: I'm not fan of closing the data model and limiting the ability for people to do what they like in extensions.

Saying "we gonna get all use cases so we can close this is presomptuous. Exemple I did not get until yesterday that this mechanism could be used for AES key rotation.

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
chibenwa: Yes I can add JVM args in order to easily control this if need be.

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
chibenwa: I'd rather be strict, none precisely but I believe with  this we shall be good with almost all...

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
chibenwa: Yes indeed. I focussed so far on adding the metadata to the interface, not on making it work correctly.

Given an agreement on the design, I'd be happy to spend the 1 week to complete the work along with decent amount of testing.

server/blob/blob-memory/src/main/java/org/apache/james/blob/memory/MemoryBlobStoreDAO.java
chibenwa: > 🤔 shouldn't the in memory be adapted to actually store the metadata ? the MemoryBlobStoreDao will always loose all metadata.

Short answer: later. I don't like spending time writing non consensual code that do not get integrated.

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
jeantil: what about the mixed strategy : known/actively used by james metadata is materialized as such, the rest can be added as unstructured data ?

chibenwa: ~~OR we could document known values as constants in the interface and keep the underlying map.~~

Edited. Please see a better proposal below that allign both visions.

jeantil: at the very least do you mind hiding the Map type in a Metadata type ? and exposing a readonly api

chibenwa: IMO the map shall be immutable

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -25,23 +25,77 @@
...
+            Preconditions.checkArgument(value.length() < 128, "ContentTransferEncoding value is too long. Size exceed 128 chars");
+        }
+
+        public BlobMetadataValue asValue() {
+            return new BlobMetadataValue(value);
+        }
 
     }
 
+    record BlobMetadata(Map<BlobMetadataName, BlobMetadataValue> metadata) {
```
jeantil:
```
record BlobMetadata(ImmutableMap<BlobMetadataName, BlobMetadataValue> value) {
```
- java.util.Map allows for mutable implementations and doesn't inform consumers that the property is immutable
-  naming it metadata results in weird looking assertions calling `.metadata().metadata()`, I think `.metadata().value()` would be clearer. (ideally the containsAllOf would be applicablle to type BlobMetadata :D

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
jeantil: I meant more naming the constant `IN_MEMORY_PROCESSING_SIZE_TRESHOLD`  or something

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
chibenwa: I am not using a mutable map here but explicitly immutable one which is a common pattern for ASF james.

As for naming, do you have suggestions? I fall a bit sort here... Happy to update if we can come up with something relevant.

jeantil: sorry if I was unclear, for some reason githib's UI didn't want to handle diff mode yesterday somehow

for naming  my suggestion is
```suggestion
    record BlobMetadata(Map<BlobMetadataName, BlobMetadataValue> value) {
```

The tests are in the next commit but that's where I noticed the `.metadata().metadata()` changing that to value ( turns it in .metadata().value())` which is a pattern I use ahdn have seen used for ValueObjectś or objects with a similar intent.
I'm quite sure there are already cases of such a pattern in james 

would you mind adding a `get(BlobMetadataName name)` directly on the `BlobMetadata` which delegates to the map obviously ? 
This would reduce the need for consumers to dereference the full underlying map in order to get an attribute other than content encoding?  and improve encapsulation as well as promote the immutability of the api.

chibenwa: > `record BlobMetadata(Map<BlobMetadataName, BlobMetadataValue> value) {` 

Ok to me

Or even `record BlobMetadata(Map<BlobMetadataName, BlobMetadataValue> underlyingMap) ...`  ?

> would you mind adding a get(BlobMetadataName name) directly on the BlobMetadata which delegates to the map obviously ?

+1 very good idea it will avoid most access to the underlying map. Will be cleaner indeed.

server/blob/blob-api/src/test/java/org/apache/james/blob/api/MetadataAwareBlobStoreDAOContract.java
```diff
@@ -0,0 +1,76 @@
...
+    default void readBytesShouldPreserveMetadata() {
+        BlobStoreDAO testee = testee();
+
+        BlobStoreDAO.BytesBlob bytesBlob = BlobStoreDAO.BytesBlob.of("payload".getBytes(),
+            BlobStoreDAO.BlobMetadata.empty()
+                .withMetadata(new BlobStoreDAO.BlobMetadataName("name"), new BlobStoreDAO.BlobMetadataValue("value")));
+
+        Mono.from(testee.save(TEST_BUCKET_NAME, TEST_BLOB_ID, bytesBlob)).block();
+
+        assertThat(Mono.from(testee.readBytes(TEST_BUCKET_NAME, TEST_BLOB_ID)).block().metadata().metadata())
```
jeantil: I'll use this one as an example to let you pick the variant you prefer, I'm fine with both :D

```suggestion
         assertThat(Mono.from(testee.readBytes(TEST_BUCKET_NAME, TEST_BLOB_ID)).block().metadata().values())
          assertThat(Mono.from(testee.readBytes(TEST_BUCKET_NAME, TEST_BLOB_ID)).block().metadata().underlyingMap())
```

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
quantranhong1999: added `FILE_THRESHHOLD` constant.

server/blob/blob-memory/src/main/java/org/apache/james/blob/memory/MemoryBlobStoreDAO.java
quantranhong1999: I believe memory does record the metadata now.

server/blob/blob-aes/src/main/java/org/apache/james/blob/aes/AESBlobStoreDAO.java
quantranhong1999: AESBlobStoreDAO should respect metadata from the underlying blobstoredao now, see https://github.com/apache/james-project/pull/2960/changes/40e6383ba31a66606caba7d0b375a74ea4ab6ca0#diff-00b64d51d531f421f48ea475526730cde70476d0d192bca63088a713f86721fa

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
quantranhong1999: added back the Javadoc

quantranhong1999:
> I propose to split the blob metadata storage for other implementations (S3, file, Cassandra, Postgres) into other PRs.

See WIP: https://github.com/apache/james-project/pull/3001

chibenwa:
Thanks a lot for continuing this @quantranhong1999 

@jeantil I propose to merge this and move forward the implementation of metadata across the DAOs. If there is complementary feedback, we would handle the comments ;-)

Cheers,

Benoit

jeantil:
Hello

A couple  comments 
- i don't think there is a test documenting the metadata name format
- I'm aware this is not currently the case but adding comments or some kind of documentation explaining the design of BlobStoreDAO would be helpful. It always takes me quite some time to recontextualise this code. 

Something to explain that this is james virtual blobstore abstraction. That the bucketname is a james specific logical bucket which may be represented in various ways in actual storage connectors. In particular it should not be conflated with an S3 bucket name. 
Then an explanation over why metadata is exposed (use case which justified the implementation and potential use cases ) possibly explain the distinction between  metadata actively used by james which has direct helpers (and thus new helpers should be added if james starts to use more metadata) vs the rest which is exposed as an extension point for users of james libraries.
I know that technically this information is available in the JIRA and in this MR's discussion but this requires extra effort to discover or rediscover over having the why directly exposed in the contract.

The class documentation should be kept rather high level to limit it's tendency to rot but still explain the major design choices.

cheers

quantranhong1999:
Hi Jean,

> i don't think there is a test documenting the metadata name format

Likely we can add it in `BlobMetadataTest`.

> I'm aware this is not currently the case but adding comments or some kind of documentation explaining the design of BlobStoreDAO would be helpful. It always takes me quite some time to recontextualise this code.

+1. We would be happy to plan to add this documentation.

Thank you for your review.

Quan

quantranhong1999:
Hello @jeantil, please let me know if https://github.com/apache/james-project/pull/3034 resolve your documentation concern.

PR #2999, opened 2026-04-10
JAMES-4200 ActiveMQ: add a configuration option for adjusting usage l…

Arsnael:
…imits

If hardcoded to false, it can lead to test failing depending on local limits set by the user's system.

### DIFF
--- a/src/site/xdoc/server/config-activemq.xml
+++ b/src/site/xdoc/server/config-activemq.xml
@@ -40,7 +40,7 @@
       </p>
       <dl>
         <dt><strong>adjust.usage.limits</strong></dt>
-        <dd>Enable adjusting usage limits (defaults to true). You might want to turn it to false
+        <dd>Enable adjusting usage limits (defaults to false). You might want to turn it to false
           for limiting memory usage for example.</dd>
       </dl>
     </section>
--- a/src/site/xdoc/server/config-activemq.xml
+++ b/src/site/xdoc/server/config-activemq.xml
@@ -34,6 +34,16 @@
 
     </section>
 
+    <section name="ActiveMQ Configuration">
+      <p>
+        ActiveMQ offers some configuration options:
+      </p>
+      <dl>
+        <dt><strong>adjust.usage.limits</strong></dt>
+        <dd>Enable adjusting usage limits (defaults to true). You might want to turn it to false
+          for limiting memory usage for example.</dd>
+      </dl>
+    </section>
 
     <section name="ActiveMQ Metrics Configuration">
       <p>
--- a/src/site/xdoc/server/config-activemq.xml
+++ b/src/site/xdoc/server/config-activemq.xml
@@ -40,7 +40,7 @@
       </p>
       <dl>
         <dt><strong>adjust.usage.limits</strong></dt>
-        <dd>Enable adjusting usage limits (defaults to false). You might want to turn it to false
+        <dd>Enable adjusting usage limits (defaults to false). You might want to turn it to true
           for limiting memory usage for example.</dd>
       </dl>
     </section>
--- a/server/queue/queue-activemq/src/main/java/org/apache/james/queue/activemq/ActiveMQConfiguration.java
+++ b/server/queue/queue-activemq/src/main/java/org/apache/james/queue/activemq/ActiveMQConfiguration.java
@@ -23,22 +23,32 @@ import org.apache.commons.configuration2.Configuration;
 import org.apache.james.queue.activemq.metric.ActiveMQMetricConfiguration;
 
 public class ActiveMQConfiguration {
+    private static final String ADJUST_USAGE_LIMITS = "adjust.usage.limits";
+    private static final boolean ADJUST_USAGE_LIMITS_DEFAULT = false;
 
     private final ActiveMQMetricConfiguration metricConfiguration;
+    private final boolean adjustUsageLimits;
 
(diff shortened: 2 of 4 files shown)


Arsnael:
@ouvtam Hi can you have a look possibly please?

Following your contribution here: https://github.com/apache/james-project/pull/2996 I can't run the tests using the embedded activemq anymore, it breaks for me locally. I get the following:

```
[INFO] Running com.linagora.tmail.james.app.EncryptedPostgresTmailServerTest
18:26:22.992 [ERROR] o.a.a.b.BrokerService - Store limit is 102400 mb (current store usage is 0 mb). The data directory: /tmp/junit644828954926870502/junit10265229264622891642/var/store/activemq/brokers/KahaDB only has 82111 mb of usable space.
```

Maybe docker locally doesnt allow me that much space. I think this should be an option. I put it to true by default, for not breaking existing running envs. So you have to set it up to false to not auto adjust usage limits.

Is it ok for you? Up to discussion :)

quantranhong1999:
```
org.apache.james.app.spring.JamesAppSpringMainTest.testServer

Error Message
Error creating bean with name 'mailqueuefactory' defined in class path resource [META-INF/spring/activemq-queue-context.xml]: Cannot resolve reference to bean 'embeddedActiveMQ' while setting constructor argument
```

seems related.

quantranhong1999:
> I put it to true by default, for not breaking existing running envs.

+1 to revert it back to the existing behavior. It seems using `adjustUsageLimits=false`  couples the memory fix to disk-limit startup semantics

> So you have to set it up to false to not auto adjust usage limits.

It seems the best way to resolve `prevent the embedded broker to claim up to 70% of JVM memory that might cause memory drain of James`  (from PR https://github.com/apache/james-project/pull/2996) should have been to configure ActiveMQ's memory usage instead:

```
brokerService.getSystemUsage().getMemoryUsage().setLimit(...);

// and optionally
brokerService.getSystemUsage().getStoreUsage().setLimit(...);
brokerService.getSystemUsage().getTempUsage().setLimit(...);
```

quantranhong1999 (approved):

ouvtam:
I reproduced the bug in the tests which use the BrokerExtension and could fix it by setting a low fixed amount of resources.

```
broker.setAdjustUsageLimits(false);
broker.getSystemUsage().getMemoryUsage().setLimit(1024L * 1024 * 128); // 128 MB
broker.getSystemUsage().getTempUsage().setLimit(1024L * 1024 * 128); // 128 MB
broker.getSystemUsage().getStoreUsage().setLimit(1024L * 1024 * 128); // 128 MB
broker.getSystemUsage().getJobSchedulerUsage().setLimit(1024L * 1024 * 128); // 128 MB
```

The defaults are quite high, though:
```
systemUsage.getMemoryUsage().setLimit(1024L * 1024 * 1024 * 1); // 1 GB
systemUsage.getTempUsage().setLimit(1024L * 1024 * 1024 * 50); // 50 GB
systemUsage.getStoreUsage().setLimit(1024L * 1024 * 1024 * 100); // 100 GB
systemUsage.getJobSchedulerUsage().setLimit(1024L * 1024 * 1024 * 50); // 50 GB
```

In your case you hit the 100GB mark. What are sane defaults?

Arsnael:
> In your case you hit the 100GB mark. What are sane defaults?

I think @chibenwa answered it here, with 10GB => https://github.com/apache/james-project/pull/3008

Fine with me too though

chibenwa:
State of this work @Arsnael ?

Arsnael:
> State of this work @Arsnael ?

Been unclear to me if we wanted to go forward with it or if the activemq disk size configuration was considered enough.

I am happy to rebase it on latest master though and apply potential other remarks.

Would probably change that default value to false as well, as it is the default hardcoded value now.

EDIT: rebased and turned the default value to false. Let me know if it's enough

src/site/xdoc/server/config-activemq.xml
```diff
@@ -34,6 +34,16 @@
...
     </section>
 
+    <section name="ActiveMQ Configuration">
+      <p>
+        ActiveMQ offers some configuration options:
+      </p>
+      <dl>
+        <dt><strong>adjust.usage.limits</strong></dt>
+        <dd>Enable adjusting usage limits (defaults to true). You might want to turn it to false
+          for limiting memory usage for example.</dd>
```
chibenwa, line 44: Make it clear here to that default value is false?

Arsnael: Forgot to change that, good catch thanks

quantranhong1999 (approved):

src/site/xdoc/server/config-activemq.xml
```diff
@@ -40,7 +40,7 @@
       </p>
       <dl>
         <dt><strong>adjust.usage.limits</strong></dt>
-        <dd>Enable adjusting usage limits (defaults to true). You might want to turn it to false
+        <dd>Enable adjusting usage limits (defaults to false). You might want to turn it to false
           for limiting memory usage for example.</dd>
```
chibenwa, line 44: Not quite yet

```
        <dd>Enable adjusting usage limits (defaults to false). You might want to turn it to true
          for...</dd>
```

Arsnael: A quick fix could have been appreciated

chibenwa (approved):

PR #3015, opened 2026-04-15
JAMES-4182 Implement blob metadata storage for Postgres

quantranhong1999:


Arsnael (approved):

chibenwa (approved):
Cool! That's actually not *that* big.

Be sure that null value are handled when reading hstore for historical data...

quantranhong1999:
> Be sure that null value are handled when reading hstore for historical data...

Yes, it should already be the case.

PR #3016, opened 2026-04-16
JAMES-4182 Implement ZstdBlobStoreDAO

quantranhong1999:
TODO:
- [x] configuration pojo
- [x] ZstdBlobStoreDAO + unit tests


Arsnael (commented):
Read it

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,377 @@
...
+import com.google.common.io.ByteSource;
+import com.google.common.io.CountingOutputStream;
+import com.google.common.io.FileBackedOutputStream;
+
+import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
+
+public class ZstdBlobStoreDAO implements BlobStoreDAO {
+    private record CompressionDecision(long originalSize, long compressedSize) {
+        boolean satisfyCompressionMinRatio(float minRatio) {
```
chibenwa: I'd have a shortcut with if (minCompressionRatio == 0 => return false

And an explicit 1 line comment with the intent...

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,377 @@
...
+    private Mono<InputStreamBlob> decompressReactive(InputStreamBlob blob) {
+        return Mono.fromCallable(() -> InputStreamBlob.of(new ZstdInputStream(blob.payload()), blob.metadata()))
+            .doOnNext(ignored -> decompressCount.increment())
+            .onErrorMap(IOException.class, e -> new ObjectStoreIOException("Failed to initialize zstd decompression", e));
+    }
+
+    private Mono<BytesBlob> decompressBytesReactive(BytesBlob blob) {
+        return Mono.fromCallable(() -> decompress(blob))
+            .subscribeOn(Schedulers.parallel())
+            .doOnNext(ignored -> decompressCount.increment())
```
chibenwa, line 178: Decompress metric can be obtained throught he count property of BLOB_ZSTD_DECOMPRESS_LATENCY_METRIC_NAME

I suggest we remove.

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,377 @@
...
+
+    private Mono<Void> saveOriginal(BucketName bucketName, BlobId blobId, byte[] data, BlobMetadata sanitizedMetadata) {
+        return Mono.from(underlying.save(bucketName, blobId, BytesBlob.of(data, sanitizedMetadata)));
+    }
+
+    private Publisher<Void> save(BucketName bucketName, BlobId blobId, InputStream inputStream, BlobMetadata metadata) {
+        BlobMetadata sanitizedMetadata = sanitizeMetadata(metadata);
+
+        return Mono.usingWhen(Mono.fromCallable(() -> new FileBackedOutputStream(FILE_THRESHOLD)),
+                originalContent -> Mono.fromCallable(() -> inputStream.transferTo(originalContent))
```
chibenwa, line 246: 1. We should likely buffer here if and only if we compress, no?
2. is buffering relly needed?

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,377 @@
...
+                        return Mono.from(underlying.save(bucketName, blobId, byteSourceBlobWithSize(originalContent.asByteSource(), originalSize, sanitizedMetadata)));
+                    }),
+                originalContent -> Mono.fromRunnable(Throwing.runnable(originalContent::reset)).subscribeOn(Schedulers.boundedElastic()))
+            .subscribeOn(Schedulers.boundedElastic())
+            .onErrorMap(IOException.class, e -> new ObjectStoreIOException("Error saving blob " + blobId.asString(), e));
+    }
+
+    private Publisher<Void> save(BucketName bucketName, BlobId blobId, ByteSource byteSource, BlobMetadata metadata) {
+        return Mono.using(byteSource::openStream,
+                inputStream -> Mono.from(save(bucketName, blobId, inputStream, metadata)),
```
chibenwa: We have a bytesource which is a higher level abstraction than the InputStream. 
We should likely cary it over in order not to be forced to do sub-optimal work downstream...

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+
+    private Publisher<Void> save(BucketName bucketName, BlobId blobId, byte[] data, BlobMetadata metadata) {
+        BlobMetadata sanitizedMetadata = sanitizeMetadata(metadata);
+
+        if (shouldAttemptCompression(data.length)) {
+            return Mono.fromCallable(() -> compress(data))
+                .subscribeOn(Schedulers.parallel())
+                .publishOn(Schedulers.boundedElastic())
```
chibenwa: We are working with `byte[]`  I do not get why we shall publish on boundedElastic

chibenwa (commented):
Good test suite overall.

This is some production critical code and we MUST ensure this is efficient. We need to ponder every data copy and context switch. I left comments where I do think there is still an issue: IMO it still needs refinements...

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
quantranhong1999: Likely not. It is a bit more complicated than that.

`blobZstdDecompressLatency` only measures eager decompression in `readBytes()`. It does not cover:

- `read()`
- `readReactive()`

because those two return a `ZstdInputStream`, and the actual decompression happens later when the caller consumes the stream.

So for the current implementation:

- `BLOB_ZSTD_DECOMPRESS_COUNT_METRIC_NAME` counts all decompression paths
  - `read()`
  - `readReactive()`
  - `readBytes()`
- `BLOB_ZSTD_DECOMPRESS_LATENCY_METRIC_NAME` only measures `readBytes()`

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
quantranhong1999: Indeed. I will remove `.publishOn(Schedulers.boundedElastic())`.

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
quantranhong1999: I improved the `ByteSource` path: it no longer falls back to the generic `InputStream` flow, and we now resolve the size first (`sizeIfKnown()` / `size()`). This means we only allocate the compression buffer when compression is actually worth attempting. If the blob is below threshold, we keep the original `ByteSource` directly.

For plain `InputStream`, it is trickier: we cannot know the size unless we read the stream, and once read it is no longer replayable. The buffer is therefore still needed there, not because of `ZstdOutputStream` itself, but because we may need to:
1. know the original size,
2. try compression,
3. then still save the original payload if the ratio is not good enough.

I did try best effort for the InputStream saving path to avoid creating buffer when obviously no compression is needed.

Let me know your thoughts.

quantranhong1999:
I propose to do the remaining parts (guice binding + simple IT + docs) in another PR to split the work.

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,450 @@
...
+                })
+                .onErrorMap(IOException.class, e -> new ObjectStoreIOException("Error saving blob " + blobId.asString(), e));
+        }
+
+        recordThresholdSkipIfNeeded(data.length);
+        return Mono.from(underlying.save(bucketName, blobId, BytesBlob.of(data, sanitizedMetadata)))
+            .subscribeOn(Schedulers.boundedElastic());
+    }
+
+    private Mono<Void> saveCompressed(BucketName bucketName, BlobId blobId, byte[] data, byte[] compressed, BlobMetadata sanitizedMetadata) {
```
chibenwa, line 226: This method signature is aweful...

```suggestion
    private Mono<Void> saveCompressed(BucketName bucketName, BlobId blobId, BytesBlob compressedBlob, BytesBlob uncompressedBlob) {
```

And the caller should precompute the correct metadata. Better no?

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,450 @@
...
+        return Mono.from(underlying.save(bucketName, blobId, BytesBlob.of(compressed, withCompressionMetadata(sanitizedMetadata, data.length))))
+            .doOnSuccess(ignored -> {
+                compressSaveCount.increment();
+                savedBytes.add(data.length - compressed.length);
+            });
+    }
+
+    private Mono<Void> saveOriginal(BucketName bucketName, BlobId blobId, byte[] data, BlobMetadata sanitizedMetadata) {
```
chibenwa, line 234:
```suggestion
    private Mono<Void> saveOriginal(BucketName bucketName, BlobId blobId, BytesBlob) {
```

?

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,450 @@
...
+        return Mono.from(underlying.save(bucketName, blobId, BytesBlob.of(data, sanitizedMetadata)));
+    }
+
+    private Publisher<Void> save(BucketName bucketName, BlobId blobId, InputStream inputStream, BlobMetadata metadata) {
```
chibenwa, line 238:
```suggestion
    private Publisher<Void> save(BucketName bucketName, BlobId blobId, InputStreamBlob inputStream) {
```

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
chibenwa: Fair. For small stream we won't reach disk anyway, and for larger one we would do the copy for downstream DAO that (S3...) would need to buffer it anyway to know the size.

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,450 @@
...
+
+                        return Mono.from(underlying.save(bucketName, blobId, byteSourceBlobWithSize(originalContent.asByteSource(), originalSize, sanitizedMetadata)))
+                            .doOnSuccess(ignored -> thresholdSkipCount.increment());
+                    }),
+                originalContent -> Mono.fromRunnable(Throwing.runnable(originalContent::reset)).subscribeOn(Schedulers.boundedElastic()))
+            .subscribeOn(Schedulers.boundedElastic())
+            .onErrorMap(IOException.class, e -> new ObjectStoreIOException("Error saving blob " + blobId.asString(), e));
+    }
+
+    private Publisher<Void> save(BucketName bucketName, BlobId blobId, ByteSource byteSource, BlobMetadata metadata) {
```
chibenwa, line 260:
```suggestion
    private Publisher<Void> save(BucketName bucketName, BlobId blobId, ByteSourceBlob blob) {
```

chibenwa (commented):
IMO a bit more time would need to be invested trying to polish methods and their arguments within ZstdBlobStoreDAO

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,452 @@
...
+
+        if (shouldAttemptCompression(data.length)) {
+            return Mono.fromCallable(() -> compress(data))
+                .subscribeOn(Schedulers.parallel())
+                .flatMap(compressed -> {
+                    CompressionDecision compressionDecision = compressionDecision(data.length, compressed.length);
+                    if (compressionDecision.satisfyCompressionMinRatio(compressionConfiguration.minRatio())) {
+                        return saveCompressed(bucketName, blobId,
+                            BytesBlob.of(compressed, withCompressionMetadata(sanitizedMetadata, data.length)),
+                            uncompressedBlob);
```
chibenwa: I do not like this saveCompressed stuff

My gut feedling:
 - saveOriginal -> saveUnderlyig (can this wrapping even be stripped away ?)
 - ease metric recording with a helper - maybe even a specialized inner class for recording metrics for the ZstdBlobStoreDAO that would centralize this concern and related computation ?
 - Inline `saveCompressed` 
 - Extract the flatMap CompressionDecision -> saveXXX

In the end it could look like

```
private Mono<Void> saveCompressedIfWorthWhile(BucketName bucketName, BlobId blobId, BytesBlob compressedBlob, BytesBlob uncompressedBlob) {
    CompressionDecision compressionDecision = compressionDecision(uncompressed.length(), compressed.length());
    if (compressionDecision.satisfyCompressionMinRatio(compressionConfiguration.minRatio())) {
        return  Mono.from(underlying.save(bucketName, blobId, compressedBlob))
            .doOnSuccess(ignored -> metricRecorder.recordCompressedSaved(compressed, uncompressed));
    } else { 
        return Mono.from(underlying.save(bucketName, blobId, uncompressed)
            .doOnSuccess(ignored -> metricRecorder.recordUncompressedSave())
    }
}
```

Opinion?

If you like it we should likely try to adopt similar logic / naming with input stream and byte source...

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,452 @@
...
+                    }
+
+                    return saveOriginal(bucketName, blobId, uncompressedBlob)
+                        .doOnSuccess(ignored -> ratioSkipCount.increment());
+                })
+                .onErrorMap(IOException.class, e -> new ObjectStoreIOException("Error saving blob " + blobId.asString(), e));
+        }
+
+        recordThresholdSkipIfNeeded(data.length);
+        return Mono.from(underlying.save(bucketName, blobId, uncompressedBlob));
```
chibenwa:
```suggestion
        recordThresholdSkipIfNeeded(data.length);
        return Mono.from(underlying.save(bucketName, blobId, uncompressedBlob))
            .doOnSuccess(ignored -> metricRecorder.recordSkip());
```

Use do on success for consistency
Get rid of this "if needed". If compression is enabled and we get in here it means size criteria is not met otherwize we would be handled in saveCompressedIfWorthWhile

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,452 @@
...
+        try {
+            long originalSize = Long.parseLong(sizeMetadata.value());
+            return Math.toIntExact(originalSize);
+        } catch (NumberFormatException | ArithmeticException e) {
+            throw new IOException("Invalid " + CONTENT_ORIGINAL_SIZE.name() + " metadata value: " + sizeMetadata.value(), e);
+        }
+    }
+
+    private Publisher<Void> save(BucketName bucketName, BlobId blobId, BytesBlob bytesBlob) {
+        byte[] data = bytesBlob.payload();
```
chibenwa: Why storing data separatly?

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
```diff
@@ -0,0 +1,452 @@
...
+        BlobMetadata sanitizedMetadata = sanitizeMetadata(bytesBlob.metadata());
```
chibenwa: IMO silently filtering "reserved names is dangerous
I'd rather keep the explicit "throw" behaviour from Guava immutablemap. 
At least bugs will be catched easier.
If you really want to be state of the art have in a precondition that the name set intersection of the metadata with the reserved name is empty.

It will also futher simplify this code block by allowing renaming bytesBlob into uncompressedBlob

chibenwa (commented):
I think this can be further enhanced...

chibenwa:
Don't forget documentation ;-)

quantranhong1999:
I would continue another improvement round on this PR for ZstdBlobStoreDAO.

> Don't forget documentation ;-)

See guice binding + ITs + docs: https://github.com/apache/james-project/pull/3019

Arsnael (commented):
Read it, nothing to add

server/blob/blob-zstd/src/main/java/org/apache/james/blob/zstd/ZstdBlobStoreDAO.java
quantranhong1999: I adapted this suggestion. Should looks better now.

chibenwa (approved):

chibenwa:
Thanks for the nice improvments

PR #3034, opened 2026-05-06
JAMES-4182 Add documentation explains blob store design

quantranhong1999:
resolve https://github.com/apache/james-project/pull/2960#issuecomment-4241584375 concern


chibenwa (approved):

Arsnael (approved):

docs/modules/servers/partials/architecture/blobstore.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+Server components usually depend on the higher-level `BlobStore`. `BlobStoreDAO`
+is the lower-level virtual storage abstraction implemented by concrete storage
+connectors such as memory, file, Cassandra, Postgres, and S3 compatible object
+stores. It allows James to compose storage features such as encryption or
+compression independently of the storage connector.
+
+== Abstraction layers
+
+Most James components use `BlobStore`, which is responsible for saving content
```
jeantil, line 15: So a single Blobstore is bound to a single logical bucket name and component builders who want/need several "buckets" will need several blobstores

Allowing to create "organization" within the blobstore bucket can be done but requires custom blobstore implementation

server/blob/blob-api/src/main/java/org/apache/james/blob/api/BlobStoreDAO.java
```diff
@@ -38,6 +38,19 @@
...
+ *
+ * <p>A {@link BucketName} is a James-specific logical bucket. Each storage connector decides how this logical
+ * bucket is represented in its backend. It should not be conflated with an S3 bucket name and does not have to map one-to-one
+ * to a physical bucket.</p>
+ *
+ * <p>{@link BlobMetadata} is part of the contract so wrapper DAOs and storage implementations can keep side information
+ * needed to interpret a payload, such as compression markers. Metadata actively used by James should expose typed
+ * helpers, while the underlying metadata map remains an extension point for James library users and custom implementations.</p>
+ *
+ * <p>See {@code docs/modules/servers/partials/architecture/blobstore.adoc} for more details.</p>
```
jeantil, line 52: ideally can you add a similar short description and a link to the documentation on the BlobStore interface ?

quantranhong1999: Done. Hopefully it resolves your suggestions :-).

quantranhong1999:
Hi @jeantil, I propose to merge this as it is. If you have further review, I would be happy to resolve them in another PR.

PR #3040, opened 2026-05-15
Update documentation for IMAP, SMTP, LMTP, POP3, ManageSieve, TLS, OIDC

felixauringer:
Originally, I only wanted to update some documentation for ManageSieve after I did not include proper documentation in https://github.com/apache/james-project/pull/2773 and https://github.com/apache/james-project/pull/2808.

I noticed that every protocol had its own documentation for OIDC and PROXY protocol and introduced a new partial for OIDC and the common server configuration options.
As I found some inconsistencies between code and documented options, I also rewrote the documentation for the protocol-specific configuration options.

I know that this is a lot but I hope that you have a look at it at some time. I really think that these changes could make the configuration of James a little easier :slightly_smiling_face: 

A note regarding the extendable server: I noticed that building the docs throws some errors on master. All of them originate from the extendable server that includes partials incorrectly. As it is not included in the navigation and all included pages are also included in the distributed server, I just removed the docs for the extendable server altogether. It's a single commit, so I can drop it quickly if you have future plans to fix that part of the documentation.

### DIFF
--- a/docs/modules/servers/partials/configure/ssl.adoc
+++ b/docs/modules/servers/partials/configure/ssl.adoc
@@ -1,21 +1,159 @@
-This document explains how to enable James 3.0 servers to use Transport Layer Security (TLS)
-for encrypted client-server communication.
+The servers
+xref:{pages-path}/configure/smtp.adoc[SMTP],
+xref:{pages-path}/configure/pop3.adoc[POP3],
+xref:{pages-path}/configure/imap.adoc[IMAP], and
+xref:{pages-path}/configure/sieve.adoc[Managesieve]
+support TLS.
+The configuration is the same for all of them and this document explains how to do it.
 
-== Configure a Server to Use SSL/TLS
+== Configure a Server to Use TLS
 
-Each of the servers xref:{pages-path}/configure/smtp.adoc[SMTP - LMTP],
-xref:{pages-path}/configure/pop3.adoc[POP3] and xref:{pages-path}/configure/imap.adoc[IMAP]
-supports use of SSL/TLS.
+As described in xref:{pages-path}/configure/server.adoc[SMTP],
+there is a `tls` element inside the XML server element.
 
-TLS (Transport Layer Security) and SSL (Secure Sockets Layer) are protocols that provide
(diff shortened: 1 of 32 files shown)


docs/modules/servers/partials/configure/ssl.adoc
```diff
@@ -1,55 +1,194 @@
...
+
+| POP3
+| MUA
+| insecure: 110, TLS: 995
+
+| LMTP
+| Local MX
+| insecure: 24, no TLS port defined
+
+TLS for LMTP is not supported by James.
```
chibenwa, line 127: :-/ was not aware of this.

It seems weird not to offer the configuration option.

IMO we can plan follow up work in order to enable it.

chibenwa (approved):
Great!

chibenwa:
> A note regarding the extendable server

That's some legacy from the original contribution, removing it is the right call

docs/modules/servers/partials/configure/ssl.adoc
felixauringer: Should be relatively easy because the LMTP server extends an abstract server that supports it and just does not call `getEncryption`.
However, we have no use for this feature and I personally do not really know the LMTP codebase and have no idea what implications this change would have. I am not even sure whether STARTTLS is a thing with LMTP.

Arsnael (approved):

quantranhong1999 (approved):

Arsnael:
Hi @felixauringer 

> I know that this is a lot but I hope that you have a look at it at some time. I really think that these changes could make the configuration of James a little easier 

No it's not a lot to ask on the contrary. We all know we need to make the documentation of James easier and more user-friendly with a better structure for a better adoption in general. Thank you very much for spending time on this, it adds great value!

docs/modules/servers/partials/configure/ssl.adoc
felixauringer: I created https://issues.apache.org/jira/projects/JAMES/issues/JAMES-4208 to track this.

PR #3051, opened 2026-05-27
[ENHANCEMENT] BlobResolver::validateAccess allow to validate user access to a blobId without fetch the full content

quantranhong1999:


server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/DownloadRoutes.scala
```diff
@@ -244,6 +278,13 @@ class MessagePartBlobResolver @Inject()(val messageIdFactory: MessageId.Factory,
           .switchIfEmpty(SMono.just(NonApplicable))
     }
   }
+
+  override def validateAccess(blobId: BlobId, mailboxSession: MailboxSession): SMono[java.lang.Boolean] =
+    resolve(blobId, mailboxSession)
```
quantranhong1999: I am not confident with self-crafted code regarding part IDs, therefore relying on `resolve` for `MessagePartBlobResolver` now.

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/DownloadRoutes.scala
```diff
@@ -186,6 +197,17 @@ class UploadResolver @Inject()(val uploadService: UploadService) extends BlobRes
       }
     }
   }
+
+  override def validateAccess(blobId: BlobId, mailboxSession: MailboxSession): SMono[java.lang.Boolean] =
+    if (!blobId.value.value.startsWith(prefix)) {
+      SMono.just(FALSE)
+    } else {
+      val uploadIdAsString = blobId.value.value.substring(prefix.length)
```
chibenwa: `def asUploadId(blobId: BlobId): Option[UploadId]` for not to duplicate that code ?

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/DownloadRoutes.scala
```diff
@@ -201,6 +223,18 @@ class AttachmentBlobResolver @Inject()(val attachmentManager: AttachmentManager,
           }
       case _ => SMono.just(NonApplicable)
     }
+
+  override def validateAccess(blobId: BlobId, mailboxSession: MailboxSession): SMono[java.lang.Boolean] =
+    Try(attachmentIdFactory.from(blobId.value.value)) match {
```
chibenwa: `def asAttachmentId(blobId: BlobId): Option[AttachmentId]` ?

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/DownloadRoutes.scala
```diff
@@ -165,6 +168,14 @@ class MessageBlobResolver @Inject()(val messageIdFactory: MessageId.Factory,
         .switchIfEmpty(SMono.just(NonApplicable))
     }
   }
+
+  override def validateAccess(blobId: BlobId, mailboxSession: MailboxSession): SMono[java.lang.Boolean] =
+    Try(messageIdFactory.fromString(blobId.value.value)) match {
```
chibenwa: To be consistent with other comments `def asMessageId(blobId: BlobId): Option[MessageId]` ?

chibenwa (approved):
Minor stylish comments but that's OK with me.

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/DownloadRoutes.scala
chibenwa: I think this is fair as this is used only as a fallback.

chibenwa: Another approach is to validate access rights with the base message only and do not care about mime-path in the access check. Download sure would then fail but we are guaranteed to vsalidate correctly the access.

Arsnael (approved):

server/protocols/jmap-rfc-8621/src/main/scala/org/apache/james/jmap/routes/DownloadRoutes.scala
quantranhong1999: > Another approach is to validate access rights with the base message only and do not care about mime-path in the access check

Adapted.

chibenwa (approved):
Cool !

PR #3059, opened 2026-06-03
JAMES-4210 Generic SASL mechanism extension

quantranhong1999:
TODO:
- [x] SPI
- [x] IMAP/SMTP encode/decode bridge
- [x] SaslMechanismRegistry + SaslMechanismLoader
- [x] Introduce the SaslMechanism(s) implementations
- [x] Adapt SASL mechanisms for IMAP

### DIFF
--- a/examples/custom-imap/sample-configuration/imapserver.xml
+++ b/examples/custom-imap/sample-configuration/imapserver.xml
@@ -33,6 +33,13 @@ under the License.
         <connectionLimitPerIP>0</connectionLimitPerIP>
         <plainAuthDisallowed>false</plainAuthDisallowed>
         <gracefulShutdown>false</gracefulShutdown>
+        <auth>
+            <saslMechanisms>PlainSaslMechanismFactory,org.apache.james.examples.imap.sasl.ExampleTokenSaslMechanismFactory</saslMechanisms>
+            <exampleToken>
+                <expectedToken>secret-token</expectedToken>
+                <authorizedUser>bob@domain.tld</authorizedUser>
+            </exampleToken>
+        </auth>
         <imapPackages>org.apache.james.modules.protocols.DefaultImapPackage</imapPackages>
         <imapPackages>org.apache.james.examples.imap.PingImapPackages</imapPackages>
     </imapserver>
--- a/examples/custom-imap/src/main/java/org/apache/james/examples/imap/sasl/ExampleTokenSaslMechanism.java
+++ b/examples/custom-imap/src/main/java/org/apache/james/examples/imap/sasl/ExampleTokenSaslMechanism.java
@@ -1,0 +1,99 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.examples.imap.sasl;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+
+import org.apache.james.protocols.api.sasl.SaslAuthenticator;
+import org.apache.james.protocols.api.sasl.SaslExchange;
+import org.apache.james.protocols.api.sasl.SaslFailure;
+import org.apache.james.protocols.api.sasl.SaslIdentity;
+import org.apache.james.protocols.api.sasl.SaslInitialRequest;
+import org.apache.james.protocols.api.sasl.SaslMechanism;
+import org.apache.james.protocols.api.sasl.SaslStep;
+
+public class ExampleTokenSaslMechanism implements SaslMechanism {
+    public static final String NAME = "EXAMPLE-TOKEN";
+    public static final String CONTINUATION_PROMPT = "Go ahead";
+    public static final String SUCCESS_DATA_TOKEN_SUFFIX = ":server-data";
+    public static final String SUCCESS_DATA = "Token accepted";
+
+    private final ExampleTokenSaslConfiguration configuration;
+
+    public ExampleTokenSaslMechanism(ExampleTokenSaslConfiguration configuration) {
+        this.configuration = configuration;
+    }
+
+    @Override
+    public String name() {
+        return NAME;
+    }
+
+    @Override
+    public SaslExchange start(SaslInitialRequest request, SaslAuthenticator authenticator) {
+        Optional<byte[]> initialResponse = request.initialResponse();
+        return new ExampleTokenSaslExchange(initialResponse, configuration);
+    }
+
+    private static class ExampleTokenSaslExchange implements SaslExchange {
+        private final Optional<byte[]> initialResponse;
+        private final ExampleTokenSaslConfiguration configuration;
+
+        private ExampleTokenSaslExchange(Optional<byte[]> initialResponse, ExampleTokenSaslConfiguration configuration) {
+            this.initialResponse = initialResponse;
+            this.configuration = configuration;
+        }
+
+        @Override
+        public SaslStep firstStep() {
+            return initialResponse
+                .map(this::authenticate)
+                .orElseGet(() -> new SaslStep.Challenge(Optional.of(CONTINUATION_PROMPT
+                    .getBytes(StandardCharsets.UTF_8))));
+        }
+
+        @Override
+        public SaslStep onResponse(byte[] clientResponse) {
+            return authenticate(clientResponse);
+        }
+
+        @Override
+        public void close() {
+        }
+
+        private SaslStep authenticate(byte[] clientResponse) {
+            String token = new String(clientResponse, StandardCharsets.UTF_8);
+            if (configuration.expectedToken().equals(token)) {
+                return success(Optional.empty());
+            }
+            // allow client to request server to return data on success message, which may be used by Kerberos auth
+            if ((configuration.expectedToken() + SUCCESS_DATA_TOKEN_SUFFIX).equals(token)) {
+                return success(Optional.of(SUCCESS_DATA.getBytes(StandardCharsets.UTF_8)));
+            }
+            return new SaslStep.Failure(SaslFailure.authenticationFailed(Optional.empty(), Optional.of(configuration.authorizedUser()),
+                "EXAMPLE-TOKEN authentication failed."));
+        }
+
+        private SaslStep success(Optional<byte[]> serverData) {
+            return new SaslStep.Success(new SaslIdentity(configuration.authorizedUser(), configuration.authorizedUser()), serverData);
+        }
+    }
+}
--- a/examples/custom-imap/src/main/resources/imapserver.xml
+++ b/examples/custom-imap/src/main/resources/imapserver.xml
@@ -36,6 +36,13 @@ under the License.
         <customProperties>pong.response=customImapParameter</customProperties>
         <customProperties>prop.b=anotherValue</customProperties>
         <gracefulShutdown>false</gracefulShutdown>
+        <auth>
+            <saslMechanisms>PlainSaslMechanismFactory,org.apache.james.examples.imap.sasl.ExampleTokenSaslMechanismFactory</saslMechanisms>
+            <exampleToken>
+                <expectedToken>secret-token</expectedToken>
+                <authorizedUser>bob@domain.tld</authorizedUser>
+            </exampleToken>
+        </auth>
         <imapPackages>org.apache.james.modules.protocols.DefaultImapPackage</imapPackages>
         <imapPackages>org.apache.james.examples.imap.PingImapPackages</imapPackages>
     </imapserver>
@@ -55,6 +62,13 @@ under the License.
         <customProperties>pong.response=bad</customProperties>
         <customProperties>prop.b=baad</customProperties>
         <gracefulShutdown>false</gracefulShutdown>
+        <auth>
+            <saslMechanisms>PlainSaslMechanismFactory,org.apache.james.examples.imap.sasl.ExampleTokenSaslMechanismFactory</saslMechanisms>
+            <exampleToken>
+                <expectedToken>secret-token</expectedToken>
+                <authorizedUser>bob@domain.tld</authorizedUser>
+            </exampleToken>
+        </auth>
         <imapPackages>org.apache.james.modules.protocols.DefaultImapPackage</imapPackages>
         <imapPackages>org.apache.james.examples.imap.PingImapPackages</imapPackages>
     </imapserver>
--- a/examples/custom-imap/src/test/java/org/apache/james/examples/imap/CustomSaslMechanismTest.java
+++ b/examples/custom-imap/src/test/java/org/apache/james/examples/imap/CustomSaslMechanismTest.java
@@ -1,0 +1,209 @@
+/****************************************************************
+ * Licensed to the Apache Software Foundation (ASF) under one   *
+ * or more contributor license agreements.  See the NOTICE file *
+ * distributed with this work for additional information        *
+ * regarding copyright ownership.  The ASF licenses this file   *
+ * to you under the Apache License, Version 2.0 (the            *
+ * "License"); you may not use this file except in compliance   *
+ * with the License.  You may obtain a copy of the License at   *
+ *                                                              *
+ *   http://www.apache.org/licenses/LICENSE-2.0                 *
+ *                                                              *
+ * Unless required by applicable law or agreed to in writing,   *
+ * software distributed under the License is distributed on an  *
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY       *
+ * KIND, either express or implied.  See the License for the    *
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.examples.imap;
+
+import static org.apache.james.data.UsersRepositoryModuleChooser.Implementation.DEFAULT;
+import static org.apache.james.jmap.JMAPTestingConstants.BOB;
+import static org.apache.james.jmap.JMAPTestingConstants.BOB_PASSWORD;
+import static org.apache.james.jmap.JMAPTestingConstants.DOMAIN;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.OutputStreamWriter;
+import java.net.InetSocketAddress;
+import java.net.Socket;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.function.Predicate;
+
+import org.apache.james.GuiceJamesServer;
+import org.apache.james.JamesServerBuilder;
+import org.apache.james.JamesServerExtension;
+import org.apache.james.MemoryJamesConfiguration;
+import org.apache.james.MemoryJamesServerMain;
+import org.apache.james.examples.imap.sasl.ExampleTokenSaslMechanism;
+import org.apache.james.modules.protocols.ImapGuiceProbe;
+import org.apache.james.utils.DataProbeImpl;
+import org.apache.james.utils.TestIMAPClient;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
+
+class CustomSaslMechanismTest {
+    private static class ClientConnection implements AutoCloseable {
+        private final Socket socket;
+        private final BufferedReader reader;
+        private final BufferedWriter writer;
+
+        private ClientConnection(String host, int port) throws IOException {
+            socket = new Socket();
+            socket.connect(new InetSocketAddress(host, port));
+            socket.setSoTimeout(5_000);
+            reader = new BufferedReader(new InputStreamReader(socket.getInputStream(), StandardCharsets.US_ASCII));
+            writer = new BufferedWriter(new OutputStreamWriter(socket.getOutputStream(), StandardCharsets.US_ASCII));
+        }
+
+        private void writeLine(String line) throws IOException {
+            writer.write(line);
+            writer.write("\r\n");
+            writer.flush();
+        }
+
+        private String readUntil(Predicate<String> condition) throws IOException {
+            StringBuilder response = new StringBuilder();
+            while (true) {
+                String line = reader.readLine();
+                if (line == null) {
+                    throw new EOFException("Connection closed while waiting for IMAP response");
+                }
+                response.append(line).append("\n");
+                if (condition.test(line)) {
+                    return response.toString();
+                }
+            }
+        }
+
+        @Override
+        public void close() throws IOException {
+            socket.close();
+        }
+    }
+
+    private static final String EXPECTED_TOKEN = "secret-token";
+    private static final String LOCALHOST_IP = "127.0.0.1";
+
+    @RegisterExtension
+    static JamesServerExtension jamesServerExtension = new JamesServerBuilder<MemoryJamesConfiguration>(tmpDir ->
+        MemoryJamesConfiguration.builder()
+            .workingDirectory(tmpDir)
+            .configurationFromClasspath()
+            .usersRepository(DEFAULT)
+            .build())
+        .server(MemoryJamesServerMain::createServer)
+        .build();
+
+    @BeforeEach
+    void setup(GuiceJamesServer server) throws Exception {
+        server.getProbe(DataProbeImpl.class).fluent()
+            .addDomain(DOMAIN)
+            .addUser(BOB.asString(), BOB_PASSWORD);
+    }
+
+    @Test
+    void imapServerShouldAdvertiseCustomSaslMechanism(GuiceJamesServer server) throws IOException {
+        assertThat(new TestIMAPClient().connect("127.0.0.1", imapPort(server))
+            .sendCommand("CAPABILITY"))
+            .contains("AUTH=PLAIN", "AUTH=EXAMPLE-TOKEN");
+    }
+
+    @Test
+    void imapServerShouldAuthenticateCustomSaslMechanismUsingOwnConfiguration(GuiceJamesServer server) throws IOException {
+        TestIMAPClient client = new TestIMAPClient().connect("127.0.0.1", imapPort(server));
+
+        assertThat(client.sendCommand("AUTHENTICATE EXAMPLE-TOKEN " + encode(EXPECTED_TOKEN)))
+            .contains("OK AUTHENTICATE completed.");
+        assertThat(client.sendCommand("PING"))
+            .contains("PONG");
+    }
+
+    @Test
+    void imapServerShouldAuthenticateCustomSaslMechanismUsingContinuation(GuiceJamesServer server) throws IOException {
+        try (ClientConnection client = clientConnection(server)) {
+            client.readUntil(line -> line.startsWith("* OK"));
+
+            client.writeLine("A01 AUTHENTICATE EXAMPLE-TOKEN");
+            assertThat(client.readUntil(line -> line.startsWith("+")))
+                .contains("+ " + encode(ExampleTokenSaslMechanism.CONTINUATION_PROMPT));
+
+            client.writeLine(encode(EXPECTED_TOKEN));
+            String authenticationResponse = client.readUntil(line -> line.startsWith("A01"));
+            assertThat(authenticationResponse)
+                .contains("OK AUTHENTICATE completed.")
+                .doesNotContain("+ " + encode(ExampleTokenSaslMechanism.CONTINUATION_PROMPT));
+
+            client.writeLine("A02 PING");
+            assertThat(client.readUntil(line -> line.startsWith("A02")))
+                .contains("PONG");
+        }
+    }
+
+    @Test
+    void imapServerShouldAuthenticateCustomSaslMechanismReturningServerDataOnSuccess(GuiceJamesServer server) throws IOException {
+        try (ClientConnection client = clientConnection(server)) {
+            // GIVEN a custom SASL exchange started without SASL-IR
+            client.readUntil(line -> line.startsWith("* OK"));
+            client.writeLine("A01 AUTHENTICATE EXAMPLE-TOKEN");
+            assertThat(client.readUntil(line -> line.startsWith("+")))
+                .contains("+ " + encode(ExampleTokenSaslMechanism.CONTINUATION_PROMPT));
+
+            // WHEN the mechanism succeeds with final server data, as GSSAPI/Kerberos-like SASL mechanisms may require
+            client.writeLine(encode(EXPECTED_TOKEN + ExampleTokenSaslMechanism.SUCCESS_DATA_TOKEN_SUFFIX));
+            assertThat(client.readUntil(line -> line.startsWith("+") || line.startsWith("A01")))
+                .contains("+ " + encode(ExampleTokenSaslMechanism.SUCCESS_DATA));
+
+            // THEN the client acknowledges the final server data before IMAP completes authentication
+            client.writeLine("");
+            assertThat(client.readUntil(line -> line.startsWith("A01")))
+                .contains("OK AUTHENTICATE completed.");
+
+            // THEN the authenticated IMAP session remains usable
+            client.writeLine("A02 PING");
+            assertThat(client.readUntil(line -> line.startsWith("A02")))
+                .contains("PONG");
+        }
+    }
+
+    @Test
+    void plainSaslAuthenticationShouldStillWork(GuiceJamesServer server) throws IOException {
+        TestIMAPClient client = new TestIMAPClient().connect("127.0.0.1", imapPort(server));
+
+        assertThat(client.sendCommand("AUTHENTICATE PLAIN " + encodePlainInitialResponse()))
+            .contains("OK AUTHENTICATE completed.");
+        assertThat(client.sendCommand("PING"))
+            .contains("PONG");
+    }
+
+    @Test
+    void imapServerShouldRejectInvalidCustomSaslToken(GuiceJamesServer server) throws IOException {
+        assertThat(new TestIMAPClient().connect("127.0.0.1", imapPort(server))
+            .sendCommand("AUTHENTICATE EXAMPLE-TOKEN " + encode("invalid-token")))
+            .contains("NO AUTHENTICATE failed.");
+    }
+
+    private int imapPort(GuiceJamesServer server) {
+        return server.getProbe(ImapGuiceProbe.class).getImapPort();
+    }
+
+    private ClientConnection clientConnection(GuiceJamesServer server) throws IOException {
+        return new ClientConnection(LOCALHOST_IP, imapPort(server));
+    }
+
+    private String encode(String token) {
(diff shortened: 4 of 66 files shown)


protocols/api/src/main/java/org/apache/james/protocols/api/sasl/PasswordSaslAuthenticationService.java
```diff
@@ -0,0 +1,34 @@
...
+import org.apache.james.core.Username;
+
+/**
+ * Protocol-provided service used by password based SASL mechanisms.
+ */
+public interface PasswordSaslAuthenticationService {
+    /**
+     * Authenticates the supplied credentials and returns the authenticated SASL identity.
+     */
+    SaslAuthenticationResult authenticate(Username authenticationId, String password, Optional<Username> authorizationId);
```
chibenwa:
```suggestion
    SaslAuthenticationResult authenticate(Username authenticationId, Optional<Username> authorizationId, String password);
```

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslAuthenticationResult.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Successful authentication result.
+     */
+    record Success(SaslIdentity identity, String log) implements SaslAuthenticationResult {
+    }
+
+    /**
+     * Failed authentication result.
+     */
+    record Failure(String log) implements SaslAuthenticationResult {
```
chibenwa:
```suggestion
    record Failure(String reason) implements SaslAuthenticationResult {
```

Also how about a specialized Exception ?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
```diff
@@ -0,0 +1,45 @@
...
+public interface SaslMechanism {
+    /**
+     * Returns the SASL mechanism name advertised to clients.
+     */
+    String name();
+
+    /**
+     * Whether this mechanism can be used by the supplied protocol.
+     */
+    boolean supports(SaslProtocol protocol);
```
chibenwa: Can we supply rationals about when a SASL protocol is implemented but not supported by a given protocol?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Whether this mechanism is currently usable for the supplied session context.
+     */
+    boolean isAvailable(SaslSessionContext context);
```
chibenwa: Mmm I see where this come from: disable auth on unencrypted channels.

But how is this specific to a given SASL mechanism ? In this case we just do not want to allow ANY kind of auth no?

Also if that's about: disable XYZ auth then really the new way of doing it is so change the loaded list of SASLMechanisms. (we 'd be able to infer the sasl machanism to be loaded for retro compatibility).

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
```diff
@@ -0,0 +1,45 @@
...
+ * specific language governing permissions and limitations      *
+ * under the License.                                           *
+ ****************************************************************/
+
+package org.apache.james.protocols.api.sasl;
+
+/**
+ * Protocol-neutral SASL mechanism.
+ */
+public interface SaslMechanism {
```
chibenwa, line 25: I'm surprised that we do not provide a way (likely through a factory) to generate a SASLMechanism from commos-configuration2 Configuration

I have XOauth2 + OAuthBearer in mind here...

Edit: I saw below that to fit current code the conf is session scoped thus dynamic. Let's just be sure that it is not a regression for XOAUTH2 in IMAP

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslSessionContext.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Optional;
+
+/**
+ * Protocol-provided context exposed to SASL mechanisms.
+ */
+public interface SaslSessionContext {
+    /**
+     * Protocol currently running the SASL exchange.
+     */
+    SaslProtocol protocol();
```
chibenwa: Why does the Sasl mechanism needs to know this ?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslSessionContext.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Whether TLS is active for the current session.
+     */
+    boolean isTlsStarted();
```
chibenwa: Idem cf earlier comments

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslSessionContext.java
```diff
@@ -0,0 +1,47 @@
...
+
+    /**
+     * Looks up optional protocol or server configuration required by a mechanism.
+     */
+    <T> Optional<T> configuration(Class<T> configurationType);
+
+    /**
+     * Looks up protocol-provided services, such as password or bearer-token authentication.
+     */
+    <T> Optional<T> service(Class<T> serviceType);
```
chibenwa: This one looks too generic to be honnest

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslStep.java
```diff
@@ -0,0 +1,73 @@
...
+
+package org.apache.james.protocols.api.sasl;
+
+import java.util.Objects;
+import java.util.Optional;
+
+/**
+ * Server step produced by a SASL exchange.
+ */
+public interface SaslStep {
```
chibenwa: sealed ?

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/esmtp/SmtpSaslBridge.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.james.protocols.api.sasl.SaslStep;
+import org.apache.james.protocols.smtp.SMTPResponse;
+import org.apache.james.protocols.smtp.SMTPRetCode;
+
+public class SmtpSaslBridge {
+    /**
+     * Converts an SMTP AUTH request into a protocol-neutral SASL initial request.
+     */
+    public SaslInitialRequest initialRequest(String mechanismName, Optional<String> initialClientResponse) {
+        return new SaslInitialRequest(SaslProtocol.SMTP, mechanismName,
```
chibenwa: Can this be reused for LMTP ?
Likely just by overriding protected SaslProtocol field...

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslAuthenticationResult.java
```diff
@@ -0,0 +1,37 @@
...
+package org.apache.james.protocols.api.sasl;
+
+/**
+ * Result returned by protocol-provided authentication services.
+ */
+public interface SaslAuthenticationResult {
+    /**
+     * Successful authentication result.
+     */
+    record Success(SaslIdentity identity, String log) implements SaslAuthenticationResult {
```
chibenwa: Why is `log` needed here.

What is SASL specific?

Can't we just state a generic `XXX successfully connected (as UUU) using ZZZ protocol and YYY SASL mechanism` ? What do we need more, not encoded in context / SaslIdentity ?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
quantranhong1999: > I'm surprised that we do not provide a way (likely through a factory) to generate a SASLMechanism from commos-configuration2 Configuration

I plan to have a Guice loader for SaslMechanism(s) in a later step.

> Let's just be sure that it is not a regression for XOAUTH2 in IMAP

Yes. Always an important point in my head that we **MUST** not have a breaking change.

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
quantranhong1999: I think we should keep `isAvailable()` because we load the default mechanisms implicitly to prevent breaking changes. We still need a way to enforce 'disable auth on unencrypted channels' per session without forcing users to update their config. Also, connection state is dynamic—if a client uses STARTTLS, a mechanism that was previously unavailable on that connection suddenly needs to become available.

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslSessionContext.java
quantranhong1999: removed

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslSessionContext.java
quantranhong1999: removed

quantranhong1999:
Hello,

I did try a lot to make the POC ready so far, for IMAP.

I did rebase on the latest master to have the `IMAPServerTest` split. You can continue reviewing from the [JAMES-4210 Introduce SASL mechanism registry](https://github.com/apache/james-project/pull/3059/commits/9b05af70334d178647a4e086106417da3a290ebb) commit.

You can just jump to the last commit [JAMES-4210 Add custom IMAP SASL extension example](https://github.com/apache/james-project/pull/3059/commits/cb204cf4ccb60e205d9125453e913db85ddfbec8) to see how we could implement an SASL mechanism extension for IMAP.

Note that I did rush to make this POC alive for IMAP, so of course, review + polish and double-check would be needed later :)

examples/custom-imap/src/main/resources/imapserver.xml
```diff
@@ -55,6 +63,14 @@ under the License.
         <customProperties>pong.response=bad</customProperties>
         <customProperties>prop.b=baad</customProperties>
         <gracefulShutdown>false</gracefulShutdown>
+        <auth>
+            <saslMechanisms>PlainSaslMechanism,OauthBearerSaslMechanism,XOauth2SaslMechanism,org.apache.james.examples.imap.sasl.ExampleTokenSaslMechanism</saslMechanisms>
+            <saslAuthenticationServiceFactoryProviderExtensions>org.apache.james.examples.imap.sasl.ExampleTokenSaslAuthenticationServiceFactoryProvider</saslAuthenticationServiceFactoryProviderExtensions>
```
chibenwa: Leverage extensions.properties extra modules for this. 
And avoid to set up yet-an-other axis of modularity.
That's Guice job!

examples/custom-imap/src/main/java/org/apache/james/examples/imap/sasl/ExampleTokenSaslMechanism.java
```diff
@@ -0,0 +1,108 @@
...
+        return protocol == SaslProtocol.IMAP;
+    }
+
+    @Override
+    public Set<Class<?>> requiredServices(SaslProtocol protocol) {
+        if (supports(protocol)) {
+            return Set.of(ExampleTokenSaslAuthenticationService.class);
+        }
+        return Set.of();
+    }
```
chibenwa: If we need a ExampleTokenSaslAuthenticationService why don't we inject it (or its factory) via Guice ?

examples/custom-imap/src/test/java/org/apache/james/examples/imap/CustomSaslMechanismTest.java
```diff
@@ -0,0 +1,109 @@
...
+        assertThat(client.sendCommand("PING"))
+            .contains("PONG");
+    }
+
+    @Test
+    void imapServerShouldRejectInvalidCustomSaslToken(GuiceJamesServer server) throws IOException {
+        assertThat(new TestIMAPClient().connect("127.0.0.1", imapPort(server))
+            .sendCommand("AUTHENTICATE EXAMPLE-TOKEN " + encode("invalid-token")))
+            .contains("NO AUTHENTICATE failed.");
+    }
```
chibenwa, line 192: Can we add a second optional behaviour to this

I want to also be able to write

```
C: A0 AUTHENTICATE EXAMPLE-TOKEN 
S: + Go ahead
C: encode(EXPECTED_TOKEN)
S: A0 OK AUTHENTICATE completed.
```

An exemple of conditionnal continuation would actually be gold !

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/OauthBearerSaslMechanism.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.james.protocols.api.sasl;
+
+public class OauthBearerSaslMechanism extends AbstractOidcSaslMechanism {
+    public static final String NAME = "OAUTHBEARER";
+
+    @Override
+    public String name() {
+        return NAME;
+    }
```
chibenwa, line 28: I'd prefer we do not override the `name` method but rather pass stuff through the constructor

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Whether this mechanism can be used by the supplied protocol.
+     *
+     * <p>A mechanism may intentionally support only a subset of protocols when its
+     * wire payload, authorization semantics, or surrounding protocol state is only
+     * valid for those protocols. For example, a custom PLAIN variant may support
+     * only IMAP when it relies on IMAP-specific delegation semantics.
+     */
+    boolean supports(SaslProtocol protocol);
```
chibenwa: Do we REALLY need this ?

examples/custom-imap/src/main/java/org/apache/james/examples/imap/sasl/ExampleTokenSaslMechanism.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public boolean isAvailable(SaslSessionContext context) {
+        return context.service(ExampleTokenSaslAuthenticationService.class).isPresent();
+    }
```
chibenwa, line 48: Forgive me ut this is the only place the isAvailable method is actually non default - can we remove it?

protocols/api/src/test/java/org/apache/james/protocols/api/sasl/SaslMechanismContractTest.java
```diff
@@ -0,0 +1,371 @@
...
+
+        @Override
+        public boolean supports(SaslProtocol protocol) {
+            return protocol == SaslProtocol.IMAP;
+        }
+
+        @Override
+        public boolean isAvailable(SaslSessionContext context) {
+            return context.service(PasswordSaslAuthenticationService.class).isPresent();
+        }
```
chibenwa: If I do not want password why wouldn't I simply remove the associated sasl mechanism ?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanismRegistry.java
```diff
@@ -0,0 +1,121 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+/**
+ * Registry exposing configured SASL mechanisms per protocol.
+ */
+public final class SaslMechanismRegistry {
```
chibenwa: Why is this needed ?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslProtocol.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Protocols that can host SASL authentication.
+ */
+public enum SaslProtocol {
+    IMAP,
+    SMTP,
+    LMTP,
+    MANAGESIEVE,
+    POP3
+}
```
chibenwa: So this means that if I implement my own protocol then I cannot leverage existing SASL mechanism?
Why Such a restriction?
IMO -> switch to a plain string
-> also try to leverage a pre-existing type of the protocols

protocols/imap/src/main/java/org/apache/james/imap/processor/sasl/ImapBearerTokenSaslAuthenticationService.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.james.mailbox.exception.ForbiddenDelegationException;
+import org.apache.james.mailbox.exception.MailboxException;
+import org.apache.james.mailbox.exception.UserDoesNotExistException;
+import org.apache.james.protocols.api.sasl.BearerTokenSaslAuthenticationService;
+import org.apache.james.protocols.api.sasl.SaslAuthenticationResult;
+import org.apache.james.protocols.api.sasl.SaslIdentity;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ImapBearerTokenSaslAuthenticationService implements BearerTokenSaslAuthenticationService {
```
chibenwa: IMO we should not have per protocol SASL sub services.

Each protocols shall be able to apply sasl steps in it's AUTHENTICATE method...

IE the bridge should be enough!

protocols/imap/src/main/java/org/apache/james/imap/processor/sasl/ImapPasswordSaslAuthenticationServiceFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public Class<PasswordSaslAuthenticationService> serviceType() {
+        return PasswordSaslAuthenticationService.class;
+    }
+
+    @Override
+    public Optional<PasswordSaslAuthenticationService> create(SaslSessionContext context) {
+        if (context instanceof ImapSaslSessionContext imapContext && !imapContext.isPlainAuthDisallowed()) {
```
chibenwa: No IMO we should simply remove SASL mechanisms from the conf and infer the list of sasl mechanisms from the lagacy options.

chibenwa (commented):
I'm glad that we get something more or less working, that's an achievement in itself. Koodo.

Now remain the next step: SIMPLIFY ! ANd rationalize... I'd expect this proposal to be refined so that we drop the extra aaccidental complexity. 

Expect the process to take several working day (and to be painful) but we will be happy I believe about the end result!

I'd suggest we rationnalize before extending to other protocols. I'd expect from this protocol an actual code reduction by switching from O(SxP) implementation complexity to O(S+P) complexity where S is the count of SASL mechanism and P the count of protocols - what this implementation do not deliver yet !

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslInitialRequest.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param protocol protocol receiving the SASL exchange
+ * @param mechanismName requested SASL mechanism name
+ * @param initialResponse decoded initial client response, when supplied by the client
+ */
+public record SaslInitialRequest(SaslProtocol protocol, String mechanismName, Optional<byte[]> initialResponse) {
+    public SaslInitialRequest(SaslProtocol protocol, String mechanismName, Optional<byte[]> initialResponse) {
+        Objects.requireNonNull(protocol);
+        Objects.requireNonNull(mechanismName);
+        Objects.requireNonNull(initialResponse);
```
Arsnael: Is that necessary for an optional?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslStep.java
```diff
@@ -0,0 +1,73 @@
...
+    /**
+     * Server challenge to send back to the client.
+     */
+    record Challenge(Optional<byte[]> payload) implements SaslStep {
+        public Challenge {
+            payload = Objects.requireNonNull(payload)
```
Arsnael: No need to check null, it's an optional?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslStep.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /**
+     * Successful SASL exchange result.
+     */
+    record Success(SaslIdentity identity, Optional<byte[]> serverData, String log) implements SaslStep {
+        public Success {
+            identity = Objects.requireNonNull(identity);
+            serverData = Objects.requireNonNull(serverData)
+                .map(byte[]::clone);
```
Arsnael: idem?

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/esmtp/SmtpSaslBridge.java
```diff
@@ -0,0 +1,91 @@
...
+            Objects.requireNonNull(initialClientResponse).map(this::decodeInitialClientResponse));
```
Arsnael: idem?

Arsnael (commented):
Read it, not much to add for now, just minor comments

examples/custom-imap/src/test/java/org/apache/james/examples/imap/CustomSaslMechanismTest.java
quantranhong1999: I just added this continuation example. 
(still working on simplifying this work)

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
quantranhong1999: removed

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslSessionContext.java
quantranhong1999: removed

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
quantranhong1999: removed

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanismRegistry.java
quantranhong1999: removed

protocols/imap/src/main/java/org/apache/james/imap/processor/sasl/ImapBearerTokenSaslAuthenticationService.java
quantranhong1999: removed

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslInitialRequest.java
quantranhong1999: removed

quantranhong1999:
Hello,

I did simplify the design a lot. I think it should be more or less OK now.

Happy to receive any further remarks.

examples/custom-imap/src/main/java/org/apache/james/examples/imap/sasl/ExampleTokenSaslMechanism.java
```diff
@@ -0,0 +1,100 @@
...
+            return authenticate(clientResponse);
+        }
+
+        @Override
+        public void abort() {
+        }
+
+        @Override
+        public void close() {
+        }
```
chibenwa: What is the distinction between abort and close?

What are we supposed to do?

examples/custom-imap/sample-configuration/imapserver.xml
```diff
@@ -33,6 +33,13 @@ under the License.
         <connectionLimitPerIP>0</connectionLimitPerIP>
         <plainAuthDisallowed>false</plainAuthDisallowed>
         <gracefulShutdown>false</gracefulShutdown>
+        <auth>
+            <saslMechanisms>PlainSaslMechanism,org.apache.james.examples.imap.sasl.ExampleTokenSaslMechanism</saslMechanisms>
```
chibenwa: Wouldn't it be more straightforward to have the SaslMechanismFactory defined here?

The protocol layer would use guice directly to instanciate the factory then instanciate the mechanism: it would be a more straightforward design no?

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/OidcSaslMechanisms.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        @Override
+        public void close() {
+        }
+
+        private SaslStep authenticate(byte[] clientResponse) {
+            return OIDCSASLParser.parseDecoded(new String(clientResponse, StandardCharsets.US_ASCII))
+                .map(response -> (SaslStep) new SaslStep.Credentials(new SaslCredentials.BearerToken(
+                    response.getToken(), Username.of(response.getAssociatedUser()))))
```
chibenwa: So if I understand this correctly the Sasl layer do not perform the auth but provides the token handed over to the server.

Each calling layer would need to handled each kind of credentials.

That is not acceptable to me imo the athentication layer need to be embedded in the SASL layer so that we mutualize it and leverage a code / complexity reduction which is not the case here.

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslStep.java
```diff
@@ -0,0 +1,72 @@
...
+        public Optional<byte[]> payload() {
+            return payload.map(byte[]::clone);
+        }
+    }
+
+    /**
+     * Parsed credentials to be applied by the protocol handler.
+     */
+    record Credentials(SaslCredentials credentials) implements SaslStep {
+    }
```
chibenwa: Let's drop this to make authz a SASL concern

server/container/guice/common/src/test/java/org/apache/james/utils/GuiceSaslMechanismResolverTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+        private FactoryBackedSaslMechanism(String realm) {
+            super("FACTORY");
+            this.realm = realm;
+        }
+
+        private String realm() {
+            return realm;
+        }
+    }
```
chibenwa: By declaring factories we actually avoid this wizardry

chibenwa (commented):
Sorry to be picky, but this work shall be simplified further.

Let's invest more time on it.

examples/custom-imap/sample-configuration/imapserver.xml
quantranhong1999: > Wouldn't it be more straightforward to have the SaslMechanismFactory defined here?

Applied.

examples/custom-imap/src/main/java/org/apache/james/examples/imap/sasl/ExampleTokenSaslMechanism.java
quantranhong1999: Stateless mechanisms make both no-op today, but GSSAPI/Kerberos may need cancellation semantics and resource cleanup separately.

I provided a default method for `abort()` to reduce the implementation complexity a bit.

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/OidcSaslMechanisms.java
quantranhong1999: The authentication part should be in the SASL layer now.

quantranhong1999:
> Sorry to be picky, but this work shall be simplified further.

Hello, I did attempt to:
- simplify SASL injection
- move authentication logic into the SASL layer so we can reuse it for other protocols

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
```diff
@@ -31,5 +31,14 @@ public interface SaslMechanism {
...
      */
-    SaslExchange start(SaslInitialRequest request);
+    SaslExchange start(SaslInitialRequest request, SaslAuthenticator authenticator);
+
+    /**
+     * Whether this mechanism may be used over the current transport.
+     *
+     * @param channelEncrypted whether the underlying transport is encrypted, for example with TLS.
+     */
+    default boolean isAvailableOnTransport(boolean channelEncrypted) {
```
chibenwa, line 41: We can concider passin a raw ProtocolSession here I think if we want to be more generic

protocols/imap/src/main/java/org/apache/james/imap/processor/DefaultProcessor.java
```diff
@@ -136,6 +185,27 @@ public static ImapProcessor createDefaultProcessor(ImapProcessor chainEndProcess
...
+                mailboxManager.authenticate(username).as(otherUsername);
+                return Authorizator.AuthorizationState.ALLOWED;
+            } catch (UserDoesNotExistException e) {
+                return Authorizator.AuthorizationState.UNKNOWN_USER;
+            } catch (ForbiddenDelegationException | BadCredentialsException e) {
+                return Authorizator.AuthorizationState.FORBIDDEN;
+            }
+        };
+        return new JamesSaslAuthenticator(authenticator, authorizator);
+    }
```
chibenwa: Can't we figure out a better code location for this?

protocols/sasl/src/main/java/org/apache/james/protocols/sasl/oidc/OauthBearerSaslMechanism.java
```diff
@@ -0,0 +1,47 @@
...
+package org.apache.james.protocols.sasl.oidc;
+
+import org.apache.james.jwt.OidcJwtTokenVerifier;
+import org.apache.james.protocols.api.sasl.SaslAuthenticator;
+import org.apache.james.protocols.api.sasl.SaslExchange;
+import org.apache.james.protocols.api.sasl.SaslInitialRequest;
+import org.apache.james.protocols.api.sasl.SaslMechanism;
+import org.apache.james.protocols.api.sasl.SaslMechanismNames;
+
+public class OauthBearerSaslMechanism implements SaslMechanism {
```
chibenwa: Please mutualize this with XOAUTH2

CF https://github.com/chibenwa/james-project/commit/e99fee74bede164372ae6f060e9265b05bcfc21c

Which is quite beneficial

protocols/sasl/src/main/java/org/apache/james/protocols/sasl/JamesSaslAuthenticator.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.james.protocols.api.sasl.SaslAuthenticator;
+import org.apache.james.protocols.api.sasl.SaslFailure;
+import org.apache.james.protocols.api.sasl.SaslIdentity;
+
+public class JamesSaslAuthenticator implements SaslAuthenticator {
+    private final Authenticator authenticator;
+    private final Authorizator authorizator;
+
+    @Inject
+    public JamesSaslAuthenticator(Authenticator authenticator, Authorizator authorizator) {
```
chibenwa, line 65: Or can't we rather user SessionProvider ?

How about letting the Sasl mechanism directly play with SessionProvider ? And we just inject `SessionProvider` within SaslMechanismFacoties ?

server/container/guice/protocols/imap/src/main/java/org/apache/james/modules/protocols/ImapDefaultSaslMechanismFactories.java
```diff
@@ -17,12 +17,15 @@
...
-public interface SaslMechanismInstantiator {
-    Class<? extends SaslMechanism> locate(ClassName className) throws ClassNotFoundException;
+import java.lang.annotation.Retention;
 
-    SaslMechanism instantiate(ClassName className) throws ClassNotFoundException;
+import com.google.inject.BindingAnnotation;
+
+@BindingAnnotation
+@Retention(RUNTIME)
+public @interface ImapDefaultSaslMechanismFactories {
```
chibenwa, line 30: Please explain to me why this is needed.

Is this an alternative to `@Named` ?

quantranhong1999: Yes, it is an alternative to `@Named`, but more strongly typed.

The idea is to allow a community extended IMAP server that relies on custom IMAP packages to define their own default SASL(s). 

Example for TMail adoption: https://github.com/linagora/tmail-backend/pull/2438/changes/70442e1814882ca0286f2c8bdc06dae6fb9db1db#diff-eaea009076d7697f8f5a888b5793fb33e5b2fefc08d2773bfd88f7c5ddd8eef5

If you prefer `@Named`, I can switch it.

chibenwa: This means no longer specific  IMAP configuration on Twake mail side, right?

Nice!

> If you prefer @Named, I can switch it.

I actually quite like it. We can ctrl + click to explore which is not possible with bare @Named ...

Arsnael (commented):
Read it

protocols/sasl/src/main/java/org/apache/james/protocols/sasl/JamesSaslAuthenticator.java
quantranhong1999: IMO, we should avoid injecting `SessionProvider` directly into SASL mechanisms because `SessionProvider` returns `MailboxSession`, which is an IMAP/mailbox-specific session object.

For example, if SMTP later reuses OAUTHBEARER, the mechanism should not create a `MailboxSession`. SMTP only needs the authenticated/authorized identity to update its own SMTP session state and relay/submission policy.

That is why I prefer keeping a small `SaslAuthenticator` wrapper: mechanisms perform credential validation and authorization, then return a protocol-neutral `SaslAuthenticationResult.Success(SaslIdentity)`. Each protocol remains responsible for initializing its own session model from that identity.

protocols/api/src/main/java/org/apache/james/protocols/api/sasl/SaslMechanism.java
quantranhong1999: `ImapSession` does not extend `ProtocolSession`, therefore `ProtocolSession` is not generic enough to be used here...

Unless we refactor `ImapSession` to extend `ProtocolSession`, which feels like another core refactoring.

I guess we can live with `boolean channelEncrypted` for now, and refactor it further later if needed.

protocols/sasl/src/main/java/org/apache/james/protocols/sasl/oidc/OauthBearerSaslMechanism.java
quantranhong1999: done

protocols/imap/src/main/java/org/apache/james/imap/processor/DefaultProcessor.java
quantranhong1999: moved to `JamesSaslAuthenticator`

quantranhong1999:
I will plan SMTP adoption in another PR.

We can wait to see whether SMTP adoption is well aligned and viable following the SPI introduced, before merging this IMAP work, if needed. But I presume SMTP adoption would be good.

chibenwa:
> We can wait to see whether SMTP adoption is well aligned and viable following the SPI introduced, before merging this IMAP work, if needed. 

Yes please

protocols/sasl/src/main/java/org/apache/james/protocols/sasl/JamesSaslAuthenticator.java
chibenwa: I do not understand why we cannot unify this behaviour as well TBH

Arsnael (approved):

quantranhong1999:
Hello,

I just rebased and reworked the commit history. The changes are the same as before.

quantranhong1999:
Green build.

Objections to merging this?

chibenwa:
No objection on my side given that we work on a timely manner on Pop3 and manageSieve

chibenwa (approved):

PR #3067, opened 2026-06-19
[FIX] Improvments regarding mailbox consistency (backport 3.9.x)

chibenwa:

PR #3069, opened 2026-06-22
[FIX] StoreMailboxManager:renameSubMailboxes should change namespace

quantranhong1999:
When renaming a mailbox across namespaces, `StoreMailboxManager` preloads the mailbox and its children then renames them through `renameSubMailboxes`.

That helper updated the mailbox name and user but not the namespace, causing id-based renames across namespaces to keep the old namespace in storage.

Update the namespace during rename and add a regression test covering both the renamed mailbox and one child mailbox.


Arsnael (approved):

chibenwa (approved):
Good catch!

chibenwa:
Can we backport this is 3.9.x onto https://github.com/apache/james-project/pull/3067

Thanks ;-)

PR #3072, opened 2026-06-24
JAMES-4210 SASL modurlarization adoption for SMTP

quantranhong1999:
Follow previous IMAP work https://github.com/apache/james-project/pull/3059


protocols/sasl/src/main/java/org/apache/james/protocols/sasl/oidc/OAuthSaslMechanism.java
```diff
@@ -41,17 +41,30 @@
 public class OAuthSaslMechanism implements SaslMechanism {
     private final String name;
     private final OidcJwtTokenVerifier verifier;
+    private final boolean requiresSsl;
+    private final byte[] invalidTokenResponse;
```
chibenwa, line 45: Can't we supply a default / static value ?

I'd try if possible to get a similar invalidToken response in SMTP and IMAP in order to allign IMAP with SMTP

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/esmtp/AuthCmdHandler.java
```diff
@@ -165,356 +167,273 @@ private Response doAUTH(SMTPSession session, String argument) {
...
+        if (maybeMechanism.isEmpty()) {
+            return doUnknownAuth(authType);
+        }
+
+        SaslMechanism mechanism = maybeMechanism.get();
+        SaslExchange exchange;
+        try {
+            SaslInitialRequest request = SASL_BRIDGE.initialRequest(authType, initialResponse);
+            exchange = startExchange(session, mechanism, request);
+            return handleFirstSaslStep(session, authType, exchange);
```
chibenwa, line 140:
```
        try {
            SaslInitialRequest request = ;
188	+
            SaslExchange exchange = startExchange(session, maybeMechanism.get(), SASL_BRIDGE.initialRequest(authType, initialResponse));
189	+
            return handleFirstSaslStep(session, authType, exchange)
```

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/esmtp/AuthCmdHandler.java
```diff
@@ -165,356 +167,273 @@ private Response doAUTH(SMTPSession session, String argument) {
...
+                case SaslStep.Challenge challenge -> SASL_BRIDGE.challenge(challenge);
+                case SaslStep.Success success -> {
+                    session.popLineHandler();
+                    yield handleSaslSuccess(session, authType, exchange, success);
+                }
+                case SaslStep.Failure failure -> {
+                    session.popLineHandler();
+                    yield handleSaslFailure(session, authType, exchange, failure.failure());
+                }
+            };
```
chibenwa: Extract this block, mutualize it use a Runnable to either `() -> {}`  or `session::popLineHandler`

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/esmtp/AuthCmdHandler.java
```diff
@@ -165,356 +167,273 @@ private Response doAUTH(SMTPSession session, String argument) {
...
+        } finally {
+            SASL_BRIDGE.close(exchange);
+        }
     }
 
-    private String decodeBase64(String line) {
-        if (line != null) {
-            String lineWithoutTrailingCrLf = StringUtils.replace(line, "\r\n", "");
-            return new String(Base64.getDecoder().decode(lineWithoutTrailingCrLf), StandardCharsets.UTF_8);
+    private Response handleLoginFraming(SMTPSession session, Optional<String> initialResponse) {
```
chibenwa: Maybe it would be easier to handle login as a dedicated Sasl mechanism registered only for SMTP

Then we won't have this specific handling here.

protocols/smtp/src/main/java/org/apache/james/protocols/smtp/core/esmtp/AuthCmdHandler.java
```diff
@@ -165,356 +167,273 @@ private Response doAUTH(SMTPSession session, String argument) {
...
-                            .protocol("SMTP")
-                            .action("AUTH")
-                            .parameters(() -> ImmutableMap.of("authType", authType))
-                            .log("SMTP Authentication succeeded.");
-                    }
-                    return res;
-                }
-            }
+    private SaslExchange startExchange(SMTPSession session, SaslMechanism mechanism, SaslInitialRequest request) {
+        if (mechanism instanceof AuthHookSaslMechanism authHookSaslMechanism) {
```
chibenwa: Why sasl mechanism specific internals leaks in here ?

server/apps/scaling-pulsar-smtp/src/main/java/org/apache/james/UsersRepositoryBackedAuthenticator.java
```diff
@@ -0,0 +1,48 @@
...
+
+import jakarta.inject.Inject;
+
+import org.apache.james.core.Username;
+import org.apache.james.mailbox.Authenticator;
+import org.apache.james.mailbox.exception.MailboxException;
+import org.apache.james.user.api.UsersRepository;
+import org.apache.james.user.api.UsersRepositoryException;
+
+class UsersRepositoryBackedAuthenticator implements Authenticator {
```
chibenwa, line 32: Why is this needed ?

server/container/guice/protocols/smtp/src/main/java/org/apache/james/modules/protocols/SmtpGuiceProbe.java
```diff
@@ -56,6 +57,12 @@ private SmtpGuiceProbe(SMTPServerFactory smtpServerFactory) {
         this.smtpServerFactory = smtpServerFactory;
     }
 
+    @PreDestroy
+    void destroy() {
+        // SMTPServerFactory is provided through a factory method; dispose it explicitly on Guice shutdown.
+        smtpServerFactory.destroy();
+    }
```
chibenwa, line 64: Why is this needed ?

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/netty/SMTPServerFactory.java
```diff
@@ -35,26 +37,73 @@
...
+        static SmtpSaslMechanismLoader defaultLoader() {
+            ImmutableList<SaslMechanismFactory> defaultFactories = ImmutableList.of(
+                new PlainSaslMechanismFactory(AuthAnnouncementConfiguration.REQUIRE_SSL_DEFAULT),
+                new OauthBearerSaslMechanismFactory(),
+                new XOauth2SaslMechanismFactory());
+            return configuration -> loadBuiltInMechanisms(defaultFactories, configuration);
+        }
+
+        ImmutableList<SaslMechanism> load(HierarchicalConfiguration<ImmutableNode> configuration) throws ConfigurationException;
+    }
```
chibenwa, line 74: The interface is the same as for IMAP stack, no need to duplicate it IMO, likely idem for default mechanisms

chibenwa (commented):
Minor remarks, this looks overall good to me.

server/apps/scaling-pulsar-smtp/src/main/java/org/apache/james/UsersRepositoryBackedAuthenticator.java
quantranhong1999: Apps (including SMTP only apps) now need `Authenticator` binding. Meanwhile, `UserRepositoryAuthenticator` is placed in the `james-server-mailbox-adapter` module. 

I was not sure if it is relavant to add a "mailbox" depedency into a SMTP only app's `pom.xml`, therefore introducing this class...

server/container/guice/protocols/smtp/src/main/java/org/apache/james/modules/protocols/SmtpGuiceProbe.java
quantranhong1999: Fix JMX issue (instance already existed) when running some SMTP tests e.g. `MemoryTeamMailboxSmtpTest`.

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/netty/SMTPServerFactory.java
quantranhong1999: SMTP has a legacy behavior documented in `docs/modules/servers/partials/configure/smtp.adoc`: `auth.requireSSL=true` only prevents advertising AUTH before STARTTLS. Even when AUTH is not advertised, clients may still explicitly use `AUTH PLAIN` / `AUTH LOGIN`.

So for SMTP, `auth.requireSSL` controls EHLO capability announcement, not command execution. For IMAP, the equivalent setting really disables clear-text PLAIN authentication.

For example, the `SMIMECheckSignatureWithKeyStoreFileIntegrationTest` is failing because I am applying strictly the same IMAP behavior (`requireSSL=true` <=> no clear text PLAIN auth) for SMTP.

I may need to let SMTP use a slightly different default factory configuration to preserve backward compatibility.

Or do you think we can drop that legacy behavior of SMTP, and just align `requireSSL=true` <=> no authentication via clear text channel?

chibenwa: No we need to keep the announcement part.

Arsnael (commented):
Read it, nothing to add

server/apps/scaling-pulsar-smtp/src/main/java/org/apache/james/UsersRepositoryBackedAuthenticator.java
chibenwa: I'm peacemealed because smtp only server is meant not to depend on mailbox (and I 'd like to keep that)

However always-duplicating that stack seems wrong too.

Maybe we should just move the Authenticator override where we really need it IE smtp-only implems (maybe in a new dedicated Guice module if we need to.

Do this is a separate commit so we can easily see if it looks better or not.

quantranhong1999:
FYI, I just fixed some failing tests on the CI. I would continue to work on the remarks.

server/apps/scaling-pulsar-smtp/src/main/java/org/apache/james/UsersRepositoryBackedAuthenticator.java
quantranhong1999: > However always-duplicating that stack seems wrong too.

It seems there is only 1 duplication in `server/apps/scaling-pulsar-smtp` app needed. It is not too bad, I think.

> Maybe we should just move the Authenticator override where we really need it IE smtp-only implems (maybe in a new dedicated Guice module if we need to.
Do this is a separate commit so we can easily see if it looks better or not.

I tried to estimate on my local. It seems pretty invasive. I suggest we can live with this for now.

protocols/sasl/src/main/java/org/apache/james/protocols/sasl/oidc/OAuthSaslMechanism.java
quantranhong1999: This should already be the case: IMAP and SMTP both use the same OIDC SASL factories, and these factories build the invalid-token payload through the shared `OidcSaslMechanismFactory.invalidTokenResponse(...)` helper. The only difference is protocol framing: IMAP sends `+ <base64>`, SMTP sends `334 <base64>`.

`OAuthSaslMechanism` still keeps a `byte[] invalidTokenResponse` because the mechanism only needs to emit the prepared SASL error payload. It should not know about `OidcSASLConfiguration`; the factory owns config parsing and builds the payload from `scope` / `oidcConfigurationURL`, then injects the ready-to-use bytes into the mechanism.

server/apps/scaling-pulsar-smtp/src/main/java/org/apache/james/UsersRepositoryBackedAuthenticator.java
chibenwa: Oh I misread that chngeset is in scaling pulsar. That's thus VERY fine with me.

Sorry for confusion

chibenwa (commented):
No objections ;-)

Thanks for the awesome work!

quantranhong1999:
> Thanks for the awesome work!

There is still the Login mechanism dedicated to SMTP. Please wait for it ^^.

quantranhong1999:
> There is still the Login mechanism dedicated to SMTP. Please wait for it ^^.

Here we go. Please continue reviewing from the [JAMES-4210 Add SMTP LOGIN SASL mechanism](https://github.com/apache/james-project/pull/3072/changes/bc08a48da50585767655265e133bfe3d00ea5127) commit.

chibenwa (commented):
This work looks awesome 😎

chibenwa (approved):

quantranhong1999:
Reminder: we should merge the SPI + IMAP PR first: https://github.com/apache/james-project/pull/3059

Because in this SMTP PR, I squashed the SPI + IMAP work into a no-review commit...

chibenwa:
I suggest we adapt pop3 and managed sieve while we are at it.

Also mailing list status update of this work can be cool!

Arsnael (approved):

quantranhong1999:
Rebased this on top of the merged IMAP SASL work

chibenwa (approved):

Arsnael (approved):

PR #3073, opened 2026-06-24
JAMES-4213 Check whether alias target is valid when receiving email for alias

felixauringer:
Currently, it is only checked if the recipient is a user or if there is a mapping with the recipient as source. It is not checked that this mapping actually points to a local mailbox.
We do not want to accept emails that go directly to an error repository, so this should be catched after the SMTP RCPT command.

I opted to accept a recipient if all resolved mailboxes are present. This may break some poorly maintained groups, I guess. Because this check is optional, I think this is acceptable.

**Note:** This also prepares for https://github.com/linagora/tmail-backend/issues/2446 by separating the check whether a mailbox exists from the step that applies the RRT. See also the MR in that repository: https://github.com/linagora/tmail-backend/pull/2448.

Jira issue: https://issues.apache.org/jira/projects/JAMES/issues/JAMES-4213

### DIFF
--- a/server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
+++ b/server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
@@ -24,7 +24,6 @@ import org.apache.commons.configuration2.Configuration;
 import org.apache.commons.configuration2.ex.ConfigurationException;
 import org.apache.james.core.Domain;
 import org.apache.james.core.MailAddress;
-import org.apache.james.core.Username;
 import org.apache.james.domainlist.api.DomainList;
 import org.apache.james.domainlist.api.DomainListException;
 import org.apache.james.protocols.api.handler.ProtocolHandler;
@@ -64,17 +63,21 @@ public class ValidRcptHandler extends AbstractValidRcptHandler implements Protoc
 
     @Override
     protected boolean isValidRecipient(SMTPSession session, MailAddress recipient) throws UsersRepositoryException, RecipientRewriteTableException {
-        Username username = users.getUsername(recipient);
-
-        if (users.contains(username)) {
+        // Check existence of mailbox first to use RRT less often.
+        if (mailboxExists(recipient)) {
             return true;
         } else {
-            return supportsRecipientRewriteTable && isRedirected(recipient, username.asString());
+            // Check whether there is a valid RRT entry for the recipient.
+            return supportsRecipientRewriteTable && hasValidRRTEntry(recipient);
         }
     }
 
-    private boolean isRedirected(MailAddress recipient, String username) throws RecipientRewriteTableException {
-        LOGGER.debug("Unknown user {} check if it's an alias", username);
+    protected boolean mailboxExists(MailAddress recipient) throws UsersRepositoryException {
+        return users.contains(users.getUsername(recipient));
+    }
+
+    protected boolean hasValidRRTEntry(MailAddress recipient) throws RecipientRewriteTableException {
+        LOGGER.debug("Unknown recipient {}, resolving it via RRT", recipient);
 
         try {
             Mappings targetString = recipientRewriteTable.getResolvedMappings(recipient.getLocalPart(), recipient.getDomain());
--- a/server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
+++ b/server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
@@ -18,6 +18,9 @@
  ****************************************************************/
 package org.apache.james.smtpserver.fastfail;
 
+import java.util.EnumSet;
+import java.util.Optional;
+
 import jakarta.inject.Inject;
 
 import org.apache.commons.configuration2.Configuration;
@@ -32,6 +35,7 @@ import org.apache.james.protocols.smtp.core.fastfail.AbstractValidRcptHandler;
 import org.apache.james.rrt.api.RecipientRewriteTable;
 import org.apache.james.rrt.api.RecipientRewriteTable.ErrorMappingException;
 import org.apache.james.rrt.api.RecipientRewriteTableException;
+import org.apache.james.rrt.lib.Mapping;
 import org.apache.james.rrt.lib.Mappings;
 import org.apache.james.user.api.UsersRepository;
 import org.apache.james.user.api.UsersRepositoryException;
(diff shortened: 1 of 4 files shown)


felixauringer:
I have not adapted the tests yet, so they will probably fail.

chibenwa:
My point of view is that if there's an alias or somehting alike we manages this and if we manages this it should come in.

Some people actually have RRT mappings without local mailbox which is valid. And used.

I'd favour an approach that prevent invalid RRT creation in the first place and/or removes/correct RRT as needed,

felixauringer:
> Some people actually have RRT mappings without local mailbox which is valid. And used.

Just out of curiosity: What does happen after such RRT mappings have been applied? Is the email forwarded? I thought those emails just land in the error repository in all cases.

> I'd favour an approach that prevent invalid RRT creation in the first place and/or removes/correct RRT as needed,

As this (what I thought of as a bug) does not really bother us, I sadly have not the capacity to do a refactoring like that.

chibenwa:
>  Is the email forwarded? 

Yes

> I thought those emails just land in the error repository in all cases.

It depends on the mailetcontainer structure

> As this (what I thought of as a bug) does not really bother us, I sadly have not the capacity to do a refactoring like that.

Then we can close ?

But thanks a lot for the report and feedback: challenging the software that way is definitly very beneficial ;-)

felixauringer:
> Then we can close ?

Depends. In the Jira Issue you said that this stricter check could be behind a setting. I would be willing to implement that.

Additionally, I think having one method that checks the RRT table and one that checks the existence of the mailbox still makes sense for extensibility. What do you think?

chibenwa:
> stricter check could be behind a setting. I would be willing to implement that.

Ok

> Additionally, I think having one method that checks the RRT table and one that checks the existence of the mailbox still makes sense for extensibility. What do you think?

Not very opinionated I'd need to see how this looks like TBH

felixauringer:
> Not very opinionated I'd need to see how this looks like TBH

See the newest commit in this PR (7819a46a415cb9f5f9a92996eaa0ea2728f3fe8c).

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
```diff
@@ -62,30 +65,57 @@ public void setSupportsRecipientRewriteTable(boolean supportsRecipientRewriteTab
...
+            return false;
         }
-        return false;
+    }
+
+    private boolean allResolvedMailboxesExist(Mappings mappings) {
+        return mappings
+            .asStream()
+            .flatMap(mapping -> mapping.asMailAddress().stream())
+            .allMatch(address -> {
```
chibenwa: Question: shouldn't this be `any` ?

Eg with the forward feature with a local copy one of the resolved address would not be local.

quantranhong1999:
FYI:

```
11:11:52,561 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.6.0:check (check-style) on project james-server-protocols-smtp: You have 1 Checkstyle violation. -> [Help 1]
--
  |   |   | org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.6.0:check (check-style) on project james-server-protocols-smtp: You have 1 Checkstyle violation.

```

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
felixauringer: As this is now behind a config switch, I thought I could as well make it strict.
In my opinion, it doesn't make sense to enable it when using aliases for forwarding because one would have to create an additional local alias for every forward then.

felixauringer:
@chibenwa Do you think these checks are too strict? (or are they even benefitial?)

And what do you think about splitting the checks for the existence of the mailbox and the existence of a valid RRT entry into two methods? Maybe https://github.com/linagora/tmail-backend/pull/2448 helps to see how that makes extensions easier.

chibenwa:
> @chibenwa Do you think these checks are too strict?

It's IMO good to have a fixed header ordering. I'm happy that we don't have any longer to push "reorder header" and `add missing licenses" commits.

FYI intellij has a checkstyle integration that has the potential to make your life easier.

> And what do you think about splitting the checks for the existence of the mailbox and the existence of a valid RRT entry into two methods?

Agree

felixauringer:
> > And what do you think about splitting the checks for the existence of the mailbox and the existence of a valid RRT entry into two methods?
> 
> Agree

This is now isolated in a single commit (6b00aca2dfead12491dc29fd2aef9609fc3dcc7c).

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
felixauringer: I implemented both options now, with `any` and with `all` (although in a foreach loop because I couldn't handle the exceptions well in a streamed way). For me it's fine to have just the very lax and the very strict option. But you can decide which options you would like to add to James.

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
```diff
@@ -62,32 +81,92 @@ public void setSupportsRecipientRewriteTable(boolean supportsRecipientRewriteTab
...
+                    yield allResolvedMailboxesExist(mappings);
+                }
+            };
+        } catch (ErrorMappingException e) {
+            // Either an error mapping was encountered (ErrorMappingException) or the limit for recursively
+            // resolving a mapping was reached (TooManyMappingException).
+            return switch (this.recipientRewriteTableCheck) {
+                case MAPPING_EXISTS -> {
+                    // An error during mapping means that a mapping exists.
+                    LOGGER.debug("Error while resolving recipient via RRT, allowing recipient {}: ", recipient, e);
```
chibenwa:
```suggestion
                    LOGGER.info("Error while resolving recipient via RRT, allowing recipient {}: ", recipient, e);
```

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
```diff
@@ -62,32 +81,92 @@ public void setSupportsRecipientRewriteTable(boolean supportsRecipientRewriteTab
...
+                    yield true;
+                }
+                case ANY_TARGET_HAS_LOCAL_MAILBOX -> {
+                    // It is unclear whether at least one mapping is valid.
+                    LOGGER.debug("Error while resolving recipient via RRT, refusing recipient {}: ", recipient, e);
```
chibenwa:
```suggestion
                    LOGGER.info("Error while resolving recipient via RRT, refusing recipient {}: ", recipient, e);
```

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
```diff
@@ -62,32 +81,92 @@ public void setSupportsRecipientRewriteTable(boolean supportsRecipientRewriteTab
...
+                    yield false;
+                }
+                case ALL_TARGETS_HAVE_LOCAL_MAILBOX -> {
+                    // It is unclear whether all mappings are valid.
+                    LOGGER.debug("Error while resolving recipient via RRT, refusing recipient {}: ", recipient, e);
```
chibenwa:
```suggestion
                    LOGGER.info("Error while resolving recipient via RRT, refusing recipient {}: ", recipient, e);
```

chibenwa (approved):
Looks great to me!

server/protocols/protocols-smtp/src/main/java/org/apache/james/smtpserver/fastfail/ValidRcptHandler.java
felixauringer: Done (also the other comments).

chibenwa (approved):

Arsnael (approved):

PR #3089, opened 2026-07-13
JAMES-4210 [POP3] Pass command shall rely on PlainSaslMechanism logic

quantranhong1999:
Similar to what we did with IMAP's LOGIN command: rely on `PlainSaslMechanism`.

### DIFF
--- a/server/protocols/protocols-pop3/src/main/java/org/apache/james/pop3server/core/PassCmdHandler.java
+++ b/server/protocols/protocols-pop3/src/main/java/org/apache/james/pop3server/core/PassCmdHandler.java
@@ -18,6 +18,9 @@
  ****************************************************************/
 package org.apache.james.pop3server.core;
 
+import static org.apache.james.protocols.sasl.plain.PlainSaslMechanism.ENABLED;
+import static org.apache.james.protocols.sasl.plain.PlainSaslMechanism.REQUIRE_SSL;
+
 import java.io.IOException;
 import java.util.Optional;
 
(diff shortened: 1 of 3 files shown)


Arsnael (approved):

server/protocols/protocols-pop3/src/main/java/org/apache/james/pop3server/core/PassCmdHandler.java
```diff
@@ -55,15 +59,18 @@
  */
 public class PassCmdHandler extends AbstractPassCmdHandler  {
     private static final Logger LOGGER = LoggerFactory.getLogger(PassCmdHandler.class);
+    private static final PlainSaslMechanism PLAIN_SASL_MECHANISM = new PlainSaslMechanism(true, false);
```
chibenwa: What does `(true, false)` means?

server/protocols/protocols-pop3/src/main/java/org/apache/james/pop3server/core/PassCmdHandler.java
```diff
@@ -59,7 +62,7 @@
  */
 public class PassCmdHandler extends AbstractPassCmdHandler  {
     private static final Logger LOGGER = LoggerFactory.getLogger(PassCmdHandler.class);
-    private static final PlainSaslMechanism PLAIN_SASL_MECHANISM = new PlainSaslMechanism(true, false);
+    private static final PlainSaslMechanism PLAIN_SASL_MECHANISM = new PlainSaslMechanism(ENABLED, !REQUIRE_SSL);
```
chibenwa, line 65: While we are at it

This `!REQUIRE_SSL` is uggly

I'd prefer we:
 - Make it `true` (secure) by default
 - Use the same config path than SMTP to make it insecure

quantranhong1999: > Make it true (secure) by default

Currently, POP3 PASS command always accepts clear-text authentication. POP3 does not have the `auth.requireSSL` option yet. `!REQUIRE_SSL` therefore keeps the existing behavior.

I plan to add `auth.requireSSL` (and other related authentication configs like `auth.saslMechanisms`) in the upcoming [Implement RFC 5034 SASL for POP3](https://github.com/linagora/james-project/issues/5574) work.

And then we can make the POP3 PASS command respect the `auth.requireSSL` option (default to `true`).

Is that OK to you?

chibenwa: That's OK ;-)

chibenwa (approved):

Arsnael (approved):

PR #3092, opened 2026-07-17
JAMES-4210 Implement RFC 5034 AUTH command for POP3

quantranhong1999:
So POP3 can leverage SASL mechanisms via the AUTH command, just like IMAP/SMTP.


protocols/pop3/src/main/java/org/apache/james/protocols/pop3/sasl/Pop3SaslBridge.java
```diff
@@ -0,0 +1,107 @@
...
+            return new byte[0];
+        }
+        return decodeBase64(value);
+    }
+
+    private byte[] decodeBase64(String value) {
+        if (!CANONICAL_BASE64.matcher(value).matches()) {
+            throw new IllegalArgumentException("Invalid Base64 value");
+        }
+        return Base64.getDecoder().decode(value);
```
chibenwa: I'd rather try to decode then handle possible exception

protocols/pop3/src/main/java/org/apache/james/protocols/pop3/sasl/Pop3SaslBridge.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private byte[] decodeInitialClientResponse(String value) {
+        if (value.equals("=")) {
+            return new byte[0];
+        }
+        return decodeBase64(value);
+    }
+
+    private byte[] decodeBase64(String value) {
```
chibenwa: Why is that proptocol specific?
Won't all SASL bridges do the same>

chibenwa (commented):
I'm happy to see comprehensive tests on thePOP3 layer covering real HTTP exchanges

However the sasl bridge design seems very boiler plate to me. That's not a hard blocker but I believe it holds room for simplification

quantranhong1999:
Rebased and reused the SaslCodec (drop POP3Bridge): https://github.com/apache/james-project/pull/3092/commits/0e58e91079ed1b0c92e8a5c816f342547c8e0bde

chibenwa (approved):

Arsnael (approved):

PR #3094, opened 2026-07-20
[ENHANCEMENT] Prevent FETCH fragmentation

chibenwa:
```
* 1 FETCH (UID 1 BODY[] {255}\r\n 
* BODY LITTERAL
)\r\na6 OK FETCH completed.\r\n
```

Would be transmitted as 3 frames (3 flush)

By better aggregating things before writting litteral we let Netty decide when to best flush and save those 2 useless flush.


chibenwa:
@Arsnael @quantranhong1999 maybe ?

quantranhong1999 (approved):

Arsnael (approved):

PR #3099, opened 2026-07-24
JAMES-4215 Implement SASL GSSAPI Kerberos mechanism

quantranhong1999:


chibenwa:
Can you provide a demo video with a potential client ?

protocols/sasl-kerberos/src/main/java/org/apache/james/protocols/sasl/kerberos/SubjectSaslServer.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionException;
+
+import javax.security.auth.Subject;
+import javax.security.auth.callback.CallbackHandler;
+import javax.security.sasl.SaslException;
+import javax.security.sasl.SaslServer;
+
+final class SubjectSaslServer {
```
chibenwa: class SubjectSaslServer {

protocols/sasl-kerberos/pom.xml
```diff
@@ -0,0 +1,70 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-configuration2</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.kerby</groupId>
+            <artifactId>kerb-simplekdc</artifactId>
+            <scope>test</scope>
+        </dependency>
```
chibenwa, line 53: Why is kerby used only for test ?
Shouldn't we also rely more on it on the implem side ?

chibenwa (commented):
This looks like solid work!

Thanks for putting this together.

Arsnael (approved):

protocols/sasl-kerberos/pom.xml
quantranhong1999: Currently, this implementation uses built in JDK-provided Kerberos GSSAPI API(s), and it is already enough so far.

That is more standard and less prone than 3rd dependency like Kerby, IMO.

chibenwa:
No video in TB?

quantranhong1999:
> No video in TB?

I am planning this. Please wait.

quantranhong1999:
Thunderbird + James + Kerberos:

1. Start James memory app (leverage `SaslGssapiIntegrationTest` and Apache Kerby's Key Distribution Center server setup)
2. Obtain Kerberos ticket-granting ticket for ALICE
3. Start Thunderbird, and enjoy Kerberos as ALICE

https://github.com/user-attachments/assets/f36138a6-887f-460b-8ef1-2feaa34f1e7b

More detail on how to demo: https://github.com/quantranhong1999/james-project/commit/26e63bcc73499d5a1f20f250d1c9f8b31055dcf3

quantranhong1999:
Follow-up ticket to improve Kerberos's  resource cleanup upon client disconnection: https://github.com/linagora/james-project/issues/5577

chibenwa (approved):

PR #3106, opened 2026-07-31
JAMES-4210 Improve SaslExchange cleanup

quantranhong1999:
- Close SaslExchange upon client disconnection
- Drop SaslExchange.abort(), just rely on `.close()` instead.


chibenwa (approved):
No objections

PR #3108, opened 2026-08-13
JAMES-4210 SASL support for ManageSieve

quantranhong1999:
Allow ManageSieve to leverage modular SASL mechanisms, like SMTP/IMAP/POP3.


mpt/impl/managesieve/core/src/main/resources/org/apache/james/managesieve/scripts/listscripts.test
```diff
@@ -21,7 +21,7 @@ C: LISTSCRIPTS
 S: NO
 
 C: AUTHENTICATE "PLAIN"
-S: \+ ""
+S: ""
```
chibenwa, line 24: Is this change causing issues?

protocols/managesieve/src/main/java/org/apache/james/managesieve/sasl/ManageSieveSaslCodec.java
```diff
@@ -0,0 +1,222 @@
...
+            throw new SyntaxException("invalid authentication literal size");
+        }
+        int size;
+        try {
+            size = Integer.parseInt(sizeValue);
+        } catch (NumberFormatException e) {
+            throw new SyntaxException("invalid authentication literal size");
+        }
+        int contentStart = markerEnd + 3;
+        // RFC 5804 section 4 defines the literal size as an octet count, not a Java character count.
```
chibenwa, line 191: Ok as long as it gets explicitly tested with UTF8 string.

protocols/managesieve/src/test/java/org/apache/james/managesieve/sasl/ManageSieveSaslCodecTest.java
```diff
@@ -0,0 +1,160 @@
...
+    void parseInitialRequestShouldMeasureLiteralSizeInUtf8Octets() throws Exception {
+        ManageSieveSaslCodec.InitialRequest request = ManageSieveSaslCodec.parseInitialRequest("\"PLAIN\" {4+}\r\né p");
+
+        assertThat(request.saslInitialRequest().initialResponse()).hasValueSatisfying(response ->
+            assertThat(response).containsExactly(bytes("é\0p")));
+    }
+
+    @Test
+    void parseInitialRequestShouldRejectLiteralSizeSplittingUtf8Character() {
+        assertThatThrownBy(() -> ManageSieveSaslCodec.parseInitialRequest("\"PLAIN\" {1+}\r\né"))
```
chibenwa, line 79: IMO we should have a working exemple with utf-8 data too ?

upgrade-instructions.md
```diff
@@ -37,6 +38,25 @@ intentionally allow clear-text USER/PASS authentication must add the following t
...
 
+### JAMES-4210 ManageSieve SASL adoption
+
+Date: 13/08/2026
+
+Concerned products: ManageSieve servers and custom ManageSieve authentication extensions
+
+ManageSieve authentication now uses the shared SASL mechanism stack. The non-standard empty challenge `+ ""` is
+now emitted as the RFC 5804 string `""`. Clients should send quoted or literal strings for SASL continuations;
+legacy bare continuations remain accepted for compatibility.
```
chibenwa, line 49: Ok this is intentional.

Please check (eg with TB manage sieve extension) that clients supports this well...

chibenwa (commented):
Sounds like very good work to me.

Thanks for putting this together.

Mostly minor remarks, (or questions) that may not be that relevant...

Arsnael (approved):

protocols/managesieve/src/main/java/org/apache/james/managesieve/sasl/ManageSieveSaslCodec.java
quantranhong1999: Yes, this has been tested.

protocols/managesieve/src/test/java/org/apache/james/managesieve/sasl/ManageSieveSaslCodecTest.java
quantranhong1999: It is the `parseInitialRequestShouldMeasureLiteralSizeInUtf8Octets` test :-).

chibenwa (approved):

upgrade-instructions.md
quantranhong1999: https://github.com/user-attachments/assets/21b41847-37a6-4a13-8182-2c38b545e898

It works well. I can connect, create, modify some scripts, and activate the script.

PR #3130, opened 2026-08-23
Parse rfc6532 addresses

chibenwa:


chibenwa:
(
Simple master rebase, no changes. 
The ci failed with build compilation. 
I suppose this is due to recent master changes causing compilation issue atop this branch: conflicts but without merge conflict.
)

mailbox/opensearch/src/test/resources/eml/cve-2024-23184.eml
```diff
@@ -0,0 +1,45 @@
+MIME-Version: 1.0
```
Arsnael: This file does not seem to be used at all in the code. Is it necessary to keep it?

core/src/test/java/org/apache/james/core/DomainTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+
+class DomainTest {
+    @Test
+    void testPlainDomain() {
+        Domain d1 = Domain.of("example.com");
+        assertThat(d1.name().equals(d1.asString()));
```
Arsnael:
```suggestion
        assertThat(d1.name().equals(d1.asString())).isTrue();
```
?

chibenwa:
Please address your comments @Arsnael

Arsnael (approved):

quantranhong1999 (approved):

chibenwa:
@arnt no objection with that one?

Also may I ask for a ICLA ? That's a major contribution, as per ASF rule we would need IP clearance.

https://www.apache.org/licenses/contributor-agreements.html

Thanks by advance

arnt:
I started the day with [opening tabs for my open PRs](https://rant.gulbrandsen.priv.no/work/task-tracker), which are many and the James PR is big ;) The only reason I could write the james patches was that several colleagues were travelling so I had unexpected peace.

Not done with the PR catching-upping, today's my first pass since June. Still have a few tabs to read and close, and James is one.

I'll send the corporate CLA to the legal team at work; it should not be a problem (apart from the GPG invocations, which I am afraid the legal team will leave to me).

I need lunch now, will review this after lunch.

server/mailet/integration-testing/src/test/java/org/apache/james/mailets/AddDeliveredToHeaderTest.java
```diff
@@ -84,24 +101,59 @@ void receivedMessagesShouldContainDeliveredToHeaders() throws Exception {
 
     @Test
     void receivedMessagesShouldContainDeliveredToHeadersI8N() throws Exception {
```
arnt: Is that i8n instead of i18n?

mdn/src/main/java/org/apache/james/mdn/MDN.java
```diff
@@ -193,6 +193,19 @@ public static Optional<MDNReport> extractMDNReport(List<Entity> entities) {
             });
     }
 
+    /**
+     * RFC 6533 defines {@code message/global-disposition-notification} as UTF-8;
+     * no charset parameter is defined for it, so senders legitimately omit one and
+     * mime4j then falls back to us-ascii. Only the RFC 3798 form honours the
+     * declared charset.
+     */
+    private static String reportCharset(Entity entity) {
```
arnt, line 202: Just a comment. The code is correct, I think.

There a chance that an RFC will appear sometime in 2027/8 that deprecates message/global and says to use message/rfc822 for everything. Ditto text/globalheaders or whatever it was.

I think *generating* message/global is good, for now, but on the *parse* side I might lean towards accepting UTF8 in message/rfc822 already. Not 100% sure. I might be strict there too, it depends on the code.

server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpUtf8AnnounceTest.java
```diff
@@ -49,30 +58,87 @@ void tearDown() {
...
+        smtpProtocol.sendCommand("RCPT TO:<" + UTF8_RECIPIENT + ">");
+
+        SoftAssertions.assertSoftly(softly -> {
+            softly.assertThat(smtpProtocol.getReplyCode()).isEqualTo(553);
+            softly.assertThat(smtpProtocol.getReplyString()).contains(NON_ASCII_WITHOUT_SMTPUTF8);
+        });
+    }
+
+    @Test
+    void asciiEmailWithSmtpUtf8ShouldBeReceived() throws Exception {
```
arnt, line 128: What does this test?

mdn/src/main/java/org/apache/james/mdn/MDN.java
chibenwa: Thanks for the notice. We will follow upcoming RFCs when published I presume.

server/protocols/protocols-smtp/src/test/java/org/apache/james/smtpserver/SmtpUtf8AnnounceTest.java
chibenwa: Simply that we can use ascii addresses with SMTPUTF8

Maybe not the most useful of a test but defnitly harmless.

server/mailet/mailets/src/main/java/org/apache/james/transport/mailets/remote/delivery/MailDelivrerToHost.java
```diff
@@ -127,6 +127,20 @@ public ExecutionResult tryDeliveryToHost(Mail mail, Collection<InternetAddress>
...
+        // Angus reads mail.mime.allowutf8 in the SMTPTransport constructor, so it
+        // has to be decided here -- before the transport exists, hence before we
+        // can know whether the remote advertises SMTPUTF8. The envelope alone
+        // tells us whether UTF-8 could ever be needed, and that is enough: Angus
+        // only emits the SMTPUTF8 keyword on MAIL FROM when the remote also
+        // advertises it, and SmtpUtf8Strategy below still decides the downgrade.
+        // Left alone for the ASCII envelopes that make up the bulk of the
+        // traffic, so they neither change behaviour nor trip Angus' "allowutf8
+        // set but server doesn't advertise SMTPUTF8" log line. The pool clears
+        // the property again when the session is passivated.
```
arnt, line 139: Good point.

Looking at all this, I wonder why the tests passed.

server/testing/src/main/java/org/apache/james/utils/TestIMAPClient.java
```diff
@@ -50,38 +50,76 @@ public class TestIMAPClient extends ExternalResource implements Closeable, After
...
+     * {@link String#length()} of the lines it has decoded to know when a literal is over. Its
+     * streams therefore have to stay octet transparent - one char per octet, which is what its
+     * own ISO-8859-1 default gives. Decoding the socket as UTF-8 makes every multi-byte
+     * character count for one octet less than the server announced, so the client keeps reading
+     * past the literal, swallows the tagged reply as if it were message content and then blocks
+     * forever waiting for a completion line that has already gone by.
+     *
+     * UTF-8 is handled at {@link TestIMAPClient}'s own boundary instead: see
+     * {@link TestIMAPClient#asOctets(String)} and {@link TestIMAPClient#asText(String)}.
+     */
```
arnt, line 64: I suppose it has to be like that. IMAP must read octets and decode to chars in a context-sensitive manner. While I find it difficult to name the octets "ISO-8859-1", given the legacy I suppose it makes sense.

chibenwa: Well, working around our own test dependencies o_o

mdn/src/main/java/org/apache/james/mdn/MDN.java
arnt: Yes.

I think my brain has found an opinion: at this point it's best to always follow RFC 6533 when generating, and mostly when parsing (unless that's too complicated and requires too much code).

arnt:
Lovely work. I really am puzzled that the test suite was green after my work, before yours @chibenwa

Thanks!

chibenwa:
You are very welcome!
Thanks for embarking us in this journey, I'm pretty sure we would not have been able to do it by our own.

Arsnael (approved):

arnt:
https://github.com/jakartaee/mail-api/pull/833 was merged today, eliminating one buglet in this PR.

chibenwa:
(Just a rebase to solve the tiny conflict with master)

chibenwa:
Hello @arnt 

May I ask if CCLA and ICLA topics are moving forward?

I fear keeping his work open for long would create quite a bunch of conflicts...

(Sorry for not better anticipating ccla and icla topics...)

Regards,

Benoit

arnt:
I see, that makes sense. I pinged the legal team again tonight. That team is friendly, helpful and overworked, so the odds are 2:1 in our favour ;)

arnt:
I submitted an ICLA. That's the nicest CLA contribution process I've been through so far. Still waiting on the CCLA review by the legal team. (I want to submit both, so that there can never be any doubt whether a particular PR is covered or not.)

chibenwa:
Hello

I received a notice of ICLA.

I presume I need to wait for CCLA before merging this.

Have a good day!

Cheers,

Benoit

PR #3144, opened 2026-08-31
[FIX] ImapRequestFrameDecoder drops a pipelined command after a LITERAL+ literal

gkm2164:
> **Updated after review.** The originally submitted fix was wrong and has been replaced (force-push). The bug and the repro below are unchanged; the *fix* is different. Details in "What changed since the first version" at the bottom. Sorry for the churn, @chibenwa — the approval came in while I was still validating against the full `protocols-imap4` suite, and that suite is what caught it.

## Problem

When a client pipelines two commands using LITERAL+ (RFC 7888), the second command's bytes are silently dropped.

Example: two `APPEND`s back to back, no wait for a response in between, both delivered in a single TCP read.

- No error.
- No timeout signal to the client.
- No server-side log trace.
- The tag is simply never answered, and the message is never persisted.

Minimal repro (raw IMAP, after `LOGIN`), sent as one write:

```
a1 APPEND INBOX {46+}
From: user@example.com
Subject: msg1

body1

a2 APPEND INBOX {46+}
From: user@example.com
Subject: msg2

body2

```

Result: `a1 OK ... APPEND completed.` arrives. `a2` never does. `INBOX` only gains one new message, not two.

## How this was found

Tool: Dovecot's `imaptest` (official `dovecot/imaptest` Docker image).

Target: a live DynamoDB-backed James 3.9.0 deployment (contrabass-mailserver, an unrelated third-party project), under concurrent load with LITERAL+ `APPEND`s carrying custom keyword flags.

Command:
```
docker run --rm dovecot/imaptest host=host.docker.internal port=1143 \
  user=user@localhost pass=password123 clients=4 secs=30 \
  random_msg_size=2000 msgs=20 no_tracking
```

`imaptest` reported specific commands stuck indefinitely:
```
 - 63 stalled for 16 secs in command: 10 APPEND "INBOX" (\Flagged \Seen \Draft) "31-Aug-2026 01:56:40 +0000" {468+}
 - 118 stalled for 16 secs in command: 10 APPEND "INBOX" (\Flagged \Draft $Label2) {821+}
 - 63 stalled for 51 secs in command: 10 APPEND "INBOX" (\Flagged \Seen \Draft) "31-Aug-2026 01:56:40 +0000" {468+}
```

A follow-up run with `rawlog` captured the exact wire traffic (`O:` = client to server, `I:` = server to client, binary literal payload elided):
```
O: 12.10 APPEND "INBOX" (\Flagged \Seen \Draft $Label1 $Label5) "31-Aug-2026 02:01:10 +0000" {1041+}
O: <1041 bytes of literal content>
O: 12.11 APPEND "INBOX" "31-Aug-2026 02:01:10 +0000" {419+}
O: <419 bytes of literal content>
...
I: * 18 EXISTS
I: * 2 RECENT
I: 12.10 OK [APPENDUID 328955991 339] APPEND completed.
```

`12.11` never gets a line again. Not delayed — gone.

A minimal raw-socket reproduction (just two `LITERAL+` `APPEND`s, no other traffic) isolated this to a pure decoder issue. Checking the mailbox afterward confirmed `EXISTS` only advanced by one: the message itself, not just the response, never made it in.

## Root cause

`ImapRequestFrameDecoder` accumulates bytes into `pending` and parses over that snapshot. Copying everything Netty currently holds is deliberate and necessary — see below — but `pending` has no notion of how much the `ImapRequestLineReader` actually consumed from it. On a successful parse, `parseImapMessage()` throws the whole thing away:

```java
if (!pending.isEmpty()) {
    pending.clear();  // discards whatever of the *next* command was copied in
}
```

So when one read carries a literal-bearing command *plus* the start of another command, the trailing command is copied into `pending`, never looked at, and dropped.

Note what is **not** the bug: the greedy copy itself. `NotEnoughDataException#getNeededSize()` returns `size + read + crlf`, where `read` only counts characters pulled through `nextChar()` — bytes already consumed by an earlier `read(int, boolean)` on the same line are not included. For a command carrying several literals that value under-reports, so it cannot be used as an exact bound. Copying greedily and comparing loosely is precisely what makes that inaccuracy harmless today.

## Fix

Stop consuming before we know how much was used.

`obtainReader()` now parses against a *view* — `pending` followed by a slice of Netty's cumulation — instead of copying `in` and advancing it up front. On a successful parse, the cumulation is advanced by exactly what the command consumed:

```java
private void commitConsumedBytes(ByteBuf in, ParseAttempt attempt) {
    if (in != null && attempt.buffer() != null) {
        in.skipBytes(Math.max(0, attempt.buffer().readerIndex() - attempt.pendingSize()));
    }
}
```

Anything the command did not touch stays in `in` and is decoded as the next command by the existing `ByteToMessageDecoder` loop.

The failure path is deliberately left as it is today: if `pending` plus everything currently buffered still is not enough, the rest genuinely has to come from the network, so all of it is kept and the parse is retried when more arrives. That is what keeps multi-literal commands working.

One piece of size arithmetic remains: `NEEDED_DATA` keeps its historical meaning — how many bytes must still arrive beyond what `pending` holds — because `uploadToAFile()`'s completion check depends on it. Since the reader now counts from the start of `pending`, `requestMoreData()` rebases the reported size to that delta, leaving the `UNKNOWN_SIZE` sentinel untouched.

Because the parser now decides for itself when a retry is worthwhile, the `should_buffer` bookkeeping and the `totalSize >= size` comparison in `obtainReader()` are both gone. Net diff is negative (**+76 / −77**), and `uploadToAFile()` is left untouched.

## Testing

Added `IMAPServerAppendNonSynchronizedLitteralsTest#secondPipelinedAppendAfterLiteralShouldNotBeLost`.

- Placed right next to the existing `partialCommandAfterNonSynchronizedLiteralShouldNotFail` test.
- That existing test already pipelines a second command behind a literal, but only asserts the first `APPEND`'s own completion. It never checks whether the trailing command got a response at all.
- That's exactly how this went unnoticed.
- The new test pipelines two LITERAL+ `APPEND`s in one write and asserts both get a tagged `OK ... APPEND completed` response (time-bounded, since without the fix the second response never arrives).

Full module suite (`mvn clean test -pl server/protocols/protocols-imap4`, 36 classes, 630 tests) was run repeatedly in a Linux container, side by side against an unmodified `master` in the same environment. The only failures ever observed are pre-existing flaky ones that fail at the same rate on both trees (a response-desync signature occasionally hitting `IMAPServerPartialFetchTest`, `IMAPServerConnectionCheckTest` and `largeAppendsShouldWork` — e.g. `allowConnectWithUnbannedIp`: 5/200 repetitions on `master` vs 0/100 with this patch). The new regression test times out on an unmodified tree and passes with the patch, in both directions.

Also manually verified against a live James deployment with the original `imaptest` repro: both the missing response and the missing mailbox data (`EXISTS` count) are resolved.

## What changed since the first version

The first version bounded each read to the literal size from `getNeededSize()`, on the theory that the greedy copy was itself the bug. Running the full `protocols-imap4` suite showed that is not viable:

- `IMAPServerAppendNoLimitTest#smallAppendsShouldWork`, `IMAPServerPartialFetchTest` — `NegativeArraySizeException`, because the reported size excludes literal bytes already consumed.
- `IMAPServerSearchTest#passing2literalOnSameNetworkPackageWhenMoreThan16Chars` — deadlock, for the same reason: a command with two literals never gets enough data.

The greedy copy turned out to be load-bearing. The current version leaves it alone and fixes only the success path, where the bug actually is.

### DIFF
--- a/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapRequestFrameDecoder.java
+++ b/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapRequestFrameDecoder.java
@@ -74,7 +74,6 @@ public class ImapRequestFrameDecoder extends ByteToMessageDecoder implements Net
     public static final int UNAUTHENTICATE_LITERAL_MAX_SIZE = Optional.ofNullable(System.getProperty("james.imap.unauthenticated.literal.max.size"))
         .map(Integer::parseInt)
         .orElse(8192);
-    private static final String SHOULD_BUFFER = "should_buffer";
 
     private final ImapDecoder decoder;
     private final int inMemorySizeLimit;
@@ -130,16 +129,24 @@ public class ImapRequestFrameDecoder extends ByteToMessageDecoder implements Net
 
         Map<String, Object> attachment = ctx.channel().attr(FRAME_DECODE_ATTACHMENT_ATTRIBUTE_KEY).get();
 
-        Pair<ImapRequestLineReader, Integer> readerAndSize = obtainReader(ctx, in, attachment, readerIndex);
-        if (readerAndSize == null) {
+        ParseAttempt attempt = obtainReader(ctx, in, attachment, readerIndex);
+        if (attempt == null) {
             return;
         }
 
-        parseImapMessage(ctx, in, attachment, readerAndSize, readerIndex)
+        parseImapMessage(ctx, in, attachment, attempt, readerIndex)
             .ifPresent(out::add);
     }
 
-    private Optional<ImapMessage> parseImapMessage(ChannelHandlerContext ctx, ByteBuf in, Map<String, Object> attachment, Pair<ImapRequestLineReader, Integer> readerAndSize, int readerIndex) throws DecodingException {
+    /**
+     * One parse round: {@code reader} reads from {@code buffer}, whose first {@code pendingSize}
+     * bytes come from {@link #pending} and whose tail is a view of Netty's cumulation.
+     * {@code buffer} is null when the reader is backed by a file rather than by memory.
+     */
+    private record ParseAttempt(ImapRequestLineReader reader, int size, ByteBuf buffer, int pendingSize) {
+    }
+
+    private Optional<ImapMessage> parseImapMessage(ChannelHandlerContext ctx, ByteBuf in, Map<String, Object> attachment, ParseAttempt attempt, int readerIndex) throws DecodingException {
         ImapSession session = ctx.channel().attr(IMAP_SESSION_ATTRIBUTE_KEY).get();
 
         // check if the session was removed before to prevent a harmless NPE. See JAMES-1312
@@ -147,19 +154,20 @@ public class ImapRequestFrameDecoder extends ByteToMessageDecoder implements Net
         if (session != null && session.getState() != ImapSessionState.LOGOUT) {
             try {
 
-                ImapMessage message = decoder.decode(readerAndSize.getLeft(), session);
+                ImapMessage message = decoder.decode(attempt.reader(), session);
 
                 // if size is != -1 the case was a literal. if thats the case we
                 // should not consume the line
                 // See JAMES-1199
-                if (readerAndSize.getRight() == -1) {
+                if (attempt.size() == -1) {
                     try {
-                        readerAndSize.getLeft().consumeLine();
+                        attempt.reader().consumeLine();
                     } catch (DecodingException | NettyImapRequestLineReader.NotEnoughDataException e) {
                         // Silent
                     }
                 }
 
+                commitConsumedBytes(in, attempt);
                 enableFraming(ctx);
 
                 attachment.clear();
@@ -170,8 +178,6 @@ public class ImapRequestFrameDecoder extends ByteToMessageDecoder implements Net
             } catch (NettyImapRequestLineReader.NotEnoughDataException e) {
                 // this exception was thrown because we don't have enough data yet
                 requestMoreData(ctx, in, attachment, e.getNeededSize(), readerIndex);
-            } finally {
-                attachment.remove(SHOULD_BUFFER);
             }
         } else {
             // The session was null so may be the case because the channel was already closed but there were still bytes in the buffer.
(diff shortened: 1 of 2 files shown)


chibenwa (approved):
Looks awesome 😎 
Thanks for the nice contribution.
I will triage it a bit, but it seems definitely like something we want in!

gkm2164:
Closing this one: validating it against the full `protocols-imap4` suite showed the approach here is wrong, and the reworked fix is different enough that a fresh PR will be easier to review than a rewrite of this branch.

For the record, what this patch got wrong: it bounded each read to the literal size reported by `NotEnoughDataException#getNeededSize()`. That value is `size + read + crlf`, where `read` only counts characters pulled through `nextChar()` — bytes already consumed by an earlier `read(int, boolean)` on the same line are not included. So for a command carrying several literals it under-reports, and anything that trusts it precisely breaks (`IMAPServerSearchTest#passing2literalOnSameNetworkPackageWhenMoreThan16Chars` deadlocks). The existing "read everything, compare loosely" code is not sloppy — it is what makes that inaccuracy harmless, which I did not appreciate when I opened this.

The rework keeps that failure path exactly as it is today and instead fixes only the success path, which is where the actual bug lives: parse against a non-consuming view of Netty's cumulation, then advance the reader index by what the command actually used, so a pipelined follow-up command is left in the buffer instead of being dropped by `pending.clear()`. That needs no size arithmetic at all and removes the `should_buffer` bookkeeping.

I will open a new PR once it is green locally. Apologies for the CI cycles spent on this branch.

gkm2164:
Thanks for showing attention!  currently, I'm running jenkins, and I found several issues with current fix..

chibenwa:
Ok

The existing test suite is ... Quite exhaustive ! 

Do not hesitate to wave a hand if you want to get some help on the topic, and thanks a lot for the contribution.

gkm2164:
First of all, I'm a big fan of JAMES project. Lol

I found this while I was to build AWS infra version..(DynamoDB, S3, SNS/SQS usage version), and honestly, when testing the feature parity with Claude Code.

As far as I know, we use APPEND for doing Sent, or Draft update with minimal operations without full use of SMTP and this is a part of some small issue.

Has this been already a problem previously?

chibenwa:
> First of all, I'm a big fan of JAMES project. Lol

THANKS!

> Has this been already a problem previously?

We did not encounter this bug of piped APPEND command yet to be honnest. But (as you might have seen) the parsing of potentially long IMAP littleral is a quangmire which do not have any easy solution.

Regarding the changeset I do agree with what is done (better cumulation managenebt rather than dropping the full BytBuff + the code extractions.

For the context that's some code that had been written in the light of `CVE-2024-37358: Denial of service through the use of IMAP literals`. I wrote that code by hand and did not account for piped litterals so it's very possible that I ended up doing a mistake.

I'll wait for a :green_apple: build, let this open for 3 day to let a chance for people in the community to react. Then we will merge this timely.

If you want the fix to be backported to `3.9.x` then fill free to open a PR against it: that way you might not need to do a big work to benefit from it.

chibenwa (approved):
It's 🍏 so that is OK with me.

gkm2164:
Thanks for the information. I'm expecting this is only happening towards the bad clients, generally, some short-tempered clients where it just sends everything before each commands processed or not, like dovecot testing...

It's honor to make such a contribution!

I'll create another PR over the 3.9.x, as well, once this gets merged

Arsnael (approved):

PR #3145, opened 2026-08-31
[FIX] ImapRequestFrameDecoder drops a pipelined command after a literal (3.9.x backport)

gkm2164:
Backport of #3144 to `3.9.x`, as discussed there.

## What this is

`ImapRequestFrameDecoder` silently drops a pipelined command that arrives in the same read as a preceding literal-bearing command (two LITERAL+ `APPEND`s in one write: the second gets no response and is never persisted). Full analysis, wire captures and root cause are in #3144 — this PR carries the exact same decoder change (+76/−77, byte-identical to the master commit — I'll update this reference with the merged SHA once #3144 lands; opening as a draft until then).

Worth noting: the bug was originally *discovered on a 3.9.0 deployment* (the `imaptest` stalls in #3144 were recorded against James 3.9.0), so `3.9.x` is confirmed affected, not just theoretically.

## Differences from the master PR

- **Decoder**: none. The cherry-pick applied cleanly; the only drift between `3.9.x` and master in this file is the `LeakAware` constructor signature on a line the fix does not touch.
- **Regression test placement**: master splits IMAP server tests into per-topic classes on top of `AbstractIMAPServerTest`, which does not exist on `3.9.x`. The same test method (`secondPipelinedAppendAfterLiteralShouldNotBeLost`, unchanged) therefore lives in `IMAPServerTest`'s existing `AppendNonSynchronizedLitterals` nested class, right next to `partialCommandAfterNonSynchronizedLiteralShouldNotFail` — the same neighborhood it occupies on master.

## Testing on 3.9.x

- Bidirectional check in a Linux container (maven:3.9-eclipse-temurin-21): with the pristine `3.9.x` decoder plus only the new test, `secondPipelinedAppendAfterLiteralShouldNotBeLost` fails with `execution timed out after 5000 ms` (bug reproduced); with the patch it passes.
- Full `server/protocols/protocols-imap4` module suite with the patch: **612 tests run, 0 failures** (1 skipped), `BUILD SUCCESS` — including all 28 nested classes of `IMAPServerTest`.

One unrelated observation from validating this on a macOS host: `IMAPServerTest` contains nested classes `Ssl` and `SSL`, whose class files collide on case-insensitive filesystems. The surviving file then fails class loading (`NoClassDefFoundError ... wrong name`), and JUnit silently drops the whole class from discovery — a macOS developer running this module locally gets a green build with 568 of the tests never executed. Linux CI is unaffected. Happy to file that separately if useful.

### DIFF
--- a/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapRequestFrameDecoder.java
+++ b/server/protocols/protocols-imap4/src/main/java/org/apache/james/imapserver/netty/ImapRequestFrameDecoder.java
@@ -74,7 +74,6 @@ public class ImapRequestFrameDecoder extends ByteToMessageDecoder implements Net
     public static final int UNAUTHENTICATE_LITERAL_MAX_SIZE = Optional.ofNullable(System.getProperty("james.imap.unauthenticated.literal.max.size"))
         .map(Integer::parseInt)
         .orElse(8192);
-    private static final String SHOULD_BUFFER = "should_buffer";
 
     private final ImapDecoder decoder;
     private final int inMemorySizeLimit;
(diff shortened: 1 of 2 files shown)


chibenwa (approved):
Pretty much a cherry-pick right?

Ok with me too.

chibenwa (approved):
Pretty much a cherry-pick right?

Ok with me too.

chibenwa (approved):
Pretty much a cherry-pick right?

Ok with me too.

chibenwa (approved):
Pretty much a cherry-pick right?

Ok with me too.

gkm2164:
Yeap. And also verified this also passes the Dovecot imap test

Arsnael (approved):

quantranhong1999 (approved):

chibenwa:
Merged.
Thanks again for the nice contribution.

gkm2164:
It was a honor for me! See you on next PR😎

PR #3153, opened 2026-09-04
JAMES-4209 Improve S3 data recovery handling + JAMES-4224 entropy 256 -> 128 bits

chibenwa:
Key ideas:
 - don't pay for the 256 bit of entropy that are mathematically not needed. Let it for a grab for paranoids.
 - don't use an extra side car object for S3 recovery. Use directly a metadata of the header blob
 - better centralize blobId generation logic and encode it in BlobId.Factory
 
 What this achieves:
  - ~8% size reduction of amortized tiered per message cost in Cassandra ( ~ 560 byte per message)
  - dramatic reduction of Rados omap metadata (~40% for the side car and ~18% for shorter names)


Arsnael (approved):

Arsnael (approved):

quantranhong1999 (approved):
LGTM.

I rebased to solve conflicts.

PR #3159, opened 2026-09-08
JAMES-4222 Add reusable Ceph S3 test container

quantranhong1999:
In case we need to triage compatibility with Ceph in the future.


Arsnael (approved):

chibenwa (commented):
Nice. Can we triage this against blibstoreDaoContracts even if it do not get merged?

chibenwa (commented):
How do we complete this? 
Make it pass the BlobStoreDAOContats ?

quantranhong1999:
I missed this. I will have a look.

PR #3175, opened 2026-09-11
[UPGRADE] ActiveMQ + Artemis (numerous CVEs)

chibenwa:


server/blob/blob-s3/src/test/java/org/apache/james/blob/objectstorage/aws/S3MinioDocker.java
```diff
@@ -34,7 +34,7 @@
 
 public class S3MinioDocker {
 
-    public static final DockerImageName DOCKER_IMAGE_NAME = DockerImageName.parse("minio/minio")
+    public static final DockerImageName DOCKER_IMAGE_NAME = DockerImageName.parse("chibenwa/minio")
```
Arsnael, line 37: Sad but I could not find better alternatives unfortunately for now either... Fair

Arsnael (approved):

quantranhong1999 (approved):

PR #3185, opened 2026-09-15
[FIX] Cassandra 5 support dateof(now()) -> toTimestamp(now())

chibenwa:


quantranhong1999 (approved):

Arsnael (approved):

Arsnael:
Technically yes, on v4 it's deprecated so it should still be valid to make those moves.

Thank you @chibenwa

PR #3186, opened 2026-09-15
JAMES-4229 fix(webapi): Create domain mappings instead of domain alias mappings

felixauringer:
Previously, only the PUT endpoint of /domainMappings used domain alias mapppings. The GET and DELETE endpoints use domain mappings.


chibenwa (approved):
Thanks for spotting this...

chibenwa (commented):
Maybe we need dedicated routes for domain alias maping too. 

Do we have those ?

felixauringer:
> Maybe we need dedicated routes for domain alias maping too.
> 
> Do we have those ?

There is no dedicated endpoint for this. However, most mappings can also be created with this general endpoint: https://james.apache.org/server/manage-webadmin.html#Address_Mappings (the docs do not list domain alias mappings but I haven't tried whether it's still possible.

chibenwa:
Ok https://james.staged.apache.org/james-project/3.10.0/servers/distributed/operate/webadmin.html#_create_an_alias_for_a_domain does domainAliasMapping as far as I am aware of.

domain mappping shall be implemented via /domainMappings so that's the reverse of what this PR does 

> Previously, only the PUT endpoint of /domainMappings used domain alias mapppings. 

That's a mistake PUT shall target domain mapping

chibenwa (approved):
Yes ;-)

Arsnael (approved):

chibenwa:
@felixauringer I believe the honor of merging this shall be yours ;-)

PR #3195, opened 2026-09-22
JAMES-4232 format of SMTP AUTH ehlo response line

felixauringer:
RFC 5321 (https://www.rfc-editor.org/info/rfc5321/#section-4.1.1.1) specifies that a space must follow after the keyword in an EHLO response.
Hence, the correct format is `250-AUTH PLAIN LOGIN`. James currently additionally sends `250-AUTH=PLAIN LOGIN`. This seems to be for backward compatibility with clients that are not RFC compliant. According to the postfix docs (https://www.postfix.org/postconf.5.html#broken_sasl_auth_clients), this applies to very old clients, so this should not be the default behavior.


Arsnael (approved):
Nice catch, thanks @felixauringer

quantranhong1999 (approved):
Thank you!

chibenwa (approved):
Nice catch

